PR #87, opened 2015-04-28
Set grunt to break on all TypeScript compilation errors

ErjanGavalji:
When there are errors that are not breaking the typescript compilation, the grunt build passes making these errors hard to find in the output.


dtopuzov:
:+1:

enchev:
:+1:

Issue #51, opened 2015-04-24
Feature request: Pull-to-refresh ListView


enchev:
_From @truongsinh on April 9, 2015 18:31_

I'm trying to implement "pull-to-refresh" myself with native iOS UIRefreshControl, however, I could not `addTargetActionForControlEvents`, the app crashes with segfault every time i pull
```typescript

class RefreshHandlerImpl extends NSObject {
	static new(): RefreshHandlerImpl {
		return <RefreshHandlerImpl>super.new();
	}

	private _owner: UIRefreshControl;

	public initWithOwner(owner: UIRefreshControl): RefreshHandlerImpl {
		this._owner = owner;
		return this;
	}

	public "myRefresh" (sender: UIRefreshControl) {
		//console.log("refresh")
		//this._owner._emit(eventReload);


	public static ObjCExposedMethods = {
		'myRefresh': { returns: interop.types.void, params: [UIRefreshControl] }
	};
}


// Event handler for Page "loaded" event attached in main-page.xml
export function pageLoaded(args: EventData) {
	var page: Page = <any>args.object;
	var postListView: ListView = view.getViewById(page, "PostListView");


	var refreshControl = new UIRefreshControl();
	var rHandler = RefreshHandlerImpl.new().initWithOwner(refreshControl);
	refreshControl.addTargetActionForControlEvents(rHandler, "myRefresh", UIControlEvents.UIControlEventValueChanged)
	postListView.ios.addSubview(refreshControl);


	page.bindingContext = mainViewModel;
}

```

_Copied from original issue: NativeScript/cross-platform-modules#278_

enchev:
_From @hshristov on April 13, 2015 7:41_

Hi truongsinh,
Could you try to remove the "" around myRefresh method? e.g.

public myRefresh (sender: UIRefreshControl) {
...
}

enchev:
_From @truongsinh on April 13, 2015 18:38_

@hshristov I tried that before. The only reason I used "" is to have "myRefresh:" instead of "myRefresh". However, I believe when transpiling to ES5, it's the same!

enchev:
_From @truongsinh on April 13, 2015 18:40_

Take a look at http://docs.nativescript.org/runtimes/ios/how-to/ObjC-Subclassing, search for
```typescript
// A native class with the name "JSObject" will be registered, so it should be unique
class JSObject extends NSObject implements NSCoding {
    public encodeWithCoder(aCoder) { /* ... */ }

[...]

PR #126, opened 2015-05-07
pull-to-refresh feature for iOS

truongsinh:
https://github.com/NativeScript/NativeScript/issues/51
optional pull-to-refresh feature for iOS (those for Android is being developed).
When defining `ListView` XML, if there is xml attribute `refresh`, `ListView` automatically add `UIRefreshControl` view and handler.
Example:
```xml
<!-- main-page.xml -->
<Page xmlns="http://www.nativescript.org/tns.xsd" loaded="pageLoaded">
	<StackLayout>
		<ListView
			items="{{ postList }}"
			isScrolling="{{ isScrolling }}"
			loadMoreItems="listViewLoadMoreItems"
			refresh="myRefreshFunction"
			>
			<ListView.itemTemplate>
				<!-- truncated -->
			</ListView.itemTemplate>
		</ListView>
	</StackLayout>
</Page>
```
```ts
// main-page.ts
export function myRefreshFunction(args: RefreshEventData) {
  mainViewModel.refresh()
  .then(()=>{
    args.done();
    })
  .catch(function(e) {
    setTimeout(function() { throw e; });
  });
  ;
}
```

### DIFF
--- a/ui/list-view/list-view-common.ts
+++ b/ui/list-view/list-view-common.ts
@@ -43,6 +43,7 @@ export class ListView extends view.View implements definition.ListView {
     public static itemLoadingEvent = "itemLoading";
     public static itemTapEvent = "itemTap";
     public static loadMoreItemsEvent = "loadMoreItems";
+    public static refreshEvent = "refresh";
 
     public static separatorColorProperty = new dependencyObservable.Property(
         SEPARATORCOLOR,
--- a/ui/list-view/list-view.d.ts
+++ b/ui/list-view/list-view.d.ts
@@ -126,4 +126,14 @@ declare module "ui/list-view" {
          */
         view: view.View;
     }
+
+    /**
+     * Event data containing information for "refresh" with done callback
+     */
+    export interface RefreshEventData extends observable.EventData {
+        /**
+         * Callback when refresh is done to hide the loading icon
+         */
+        done: () => void;
+    }
 }
--- a/ui/list-view/list-view.ios.ts
+++ b/ui/list-view/list-view.ios.ts
@@ -10,6 +10,7 @@ import color = require("color");
 var CELLIDENTIFIER = "cell";
 var ITEMLOADING = common.ListView.itemLoadingEvent;
 var LOADMOREITEMS = common.ListView.loadMoreItemsEvent;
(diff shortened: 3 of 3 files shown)


truongsinh:
CLA signed

ErjanGavalji:
@truongsinh, thanks for the valuable input!

We are looking forward to the android implementation. Don't hesitate to ask for help if you hit a roadblock...

Just to remind, don't forget to rebase onto master too. We are constantly adding functionality and fixing glitches.

truongsinh:
:+1:

truongsinh:
Actually we do have have a question. In iOS, `UITableView` is parent view, `UIRefreshControl` is child view. In Android, it's the other way around, i.e. `ListView` is child view, `SwipeRefreshLayout`. When user access `myListview._ios`, they get `UITableView`; myy question is, when user access `myListview._android`, whether they should get `ListView ` or `SwipeRefreshLayout`.

hshristov:
@truongsinh, In such cases we use complex object. In Frame class we expose ios property of type iOSFrame which contains the viewController and few more iOS specific properties.

truongsinh:
However, if I follow such implementation, it would be a breaking change. Is it still okay?

ErjanGavalji:
@truongsinh Indeed! Initially we thought of accepting the breaking change, but due to the wide usage of the ListView module, it might be too harsh. The team decided to check some more options, which will come into play as soon as the roadmap allows it. Pull to refresh is a widely wanted feature, so it will not be delayed.

Kind regards,
Erjan Gavalji

truongsinh:
How about I clone `ListView`, and make it `ListView2` or `RefreshableListView`, so we can follow such suggested implementation without any breaking changes (I think this is Java has so much deprecated classes/methods)

ligaz:
@truongsinh Is it possible this to be abstracted in a reusable piece of code without inheritance? For example using something like [mixins](http://www.typescriptlang.org/Handbook#mixins) or [monkey patching](http://me.dt.in.th/page/JavaScript-override/)

truongsinh:
Try to open PR

truongsinh:
Cant re-open this PR, continue at #231

PR #210, opened 2015-05-24
Override addChild/removeChild on contentView

NathanaelA:
This overrides the addChild/removeChild to fix a issue where addChild / removeChild is called on a contentView or a CV descendant it will then crash the client.   This basically adds a wrapper to override the native implementation so that it won't crash.   addChild will basically set the new child as the primary content.   removeChild will basically set the content to null.

### DIFF
--- a/apps/tests/ui/scroll-view/scroll-view-tests.ts
+++ b/apps/tests/ui/scroll-view/scroll-view-tests.ts
@@ -313,5 +313,5 @@ export function test_scrollView_addChild_removeChild() {
     TKUnit.assertEqual(scrollView.content, btn, "scrollView.content addChild");
 
     scrollView.removeChild();
-    TKUnit.assertEqual(scrollView.content, null, "scrollView.content removeChild");
+    TKUnit.assertNull(scrollView.content, "scrollView.content removeChild");
 }
--- a/apps/tests/ui/scroll-view/scroll-view-tests.ts
+++ b/apps/tests/ui/scroll-view/scroll-view-tests.ts
@@ -75,7 +75,7 @@ export function test_default_TNS_values() {
     TKUnit.assertEqual(scrollView.horizontalOffset, 0, "Default scrollView.horizontalOffset");
 }
 
-export function test_vertical_oriantation_creates_correct_native_view() {
+export function test_vertical_orientation_creates_correct_native_view() {
     scrollView.orientation = enums.Orientation.vertical;
 
     if (app.android) {
@@ -86,7 +86,7 @@ export function test_vertical_oriantation_creates_correct_native_view() {
     }
 }
 
-export function test_horizontal_oriantation_creates_correct_native_view() {
+export function test_horizontal_orientation_creates_correct_native_view() {
     scrollView.orientation = enums.Orientation.horizontal;
 
     if (app.android) {
@@ -97,7 +97,7 @@ export function test_horizontal_oriantation_creates_correct_native_view() {
     }
 }
 
-export function test_scrollabeHeight_vertical_orientation_when_content_is_small() {
+export function test_scrollableHeight_vertical_orientation_when_content_is_small() {
     scrollView.orientation = enums.Orientation.vertical;
     scrollView.width = 200;
     scrollView.height = 300;
@@ -112,7 +112,7 @@ export function test_scrollabeHeight_vertical_orientation_when_content_is_small(
     TKUnit.assertEqual(scrollView.scrollableWidth, 0, "scrollView.scrollableWidth");
 }
 
-export function test_scrollabeHeight_vertical_orientation_when_content_is_big() {
+export function test_scrollableHeight_vertical_orientation_when_content_is_big() {
     scrollView.orientation = enums.Orientation.vertical;
 
     scrollView.width = 200;
@@ -129,7 +129,7 @@ export function test_scrollabeHeight_vertical_orientation_when_content_is_big() 
 
 }
 
-export function test_scrollabeWidth_horizontal_orientation_when_content_is_small() {
+export function test_scrollableWidth_horizontal_orientation_when_content_is_small() {
     scrollView.orientation = enums.Orientation.vertical;
     scrollView.width = 200;
     scrollView.height = 300;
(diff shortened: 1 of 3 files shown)


Nathanaela:
The test passes now.  If you run this test w/o the contentView changes; the test will cause the app to crash.  So this fixes it so the app won't crash and so the code is valid just like it is else where...

ErjanGavalji:
run ci

hshristov:
@Nathanaela what is the purpose of these new methods? Why don't you directly set content property?

Nathanaela:
Primarily for consistency sake.  Because people are used to using addChild/removeChild for anything that has children.  In this specific case If you attempt to use either of those on a contentView (or a descendant like scrollView) the android runtime will CRASH as addChild/removeChild is valid from somewhere in the prototype chain but it tries to do something that is invalid.

hshristov:
Well for me adding new methods that duplicates existing functionality is unnecessarily. This means more code to cover with tests, more maintenance, less options for API changes, etc.

Nathanaela:
I understand your point.  However, having a consistent api is also pretty important.   Not only did it bite someone on the forums (which is how I learned about it) -- but it bit me for a while while I was trying to help him until I finally opened up the scrollview and its parent contentview code and realized that it only had one child element.     

In this specific case I'm actually using the "content' variable to set/clear; so I'm not duplicating any functionality.   If the functionality changes in the content function (which is where any code should change); it is already future proofed since these two functions are just a wrapper around setting the "content" variable.

hshristov:
Hi @Nathanaela,

I think we have pretty consistent APIs. Only Layout descendants can have multiple childs so there we have addChild, removeChild methods. On elements that can have single child we expose it using the content property. This way it is clear that Border, ScrollView, Page, etc. cannot have multiple children. 

Exposing addChild on ScrollView could confuse someone that ScrollView could have more than 1 child (which is not the case). That is why I think it is wrong to expose such methods on ScrollView (or ContentView).

Nathanaela:
I don't feel that strongly about this patch -- it was just meant to help eliminate a mysterious crash and add what I felt was some additional consistency..      If you believe the api is much better without these helpers, then feel free to close the issue.

ligaz:
I think I will vote for what @Nathanaela is proposing here. First it will provide a consistent API and will let the customers fall into the pit of success. Second we are actually adding/removing the child of the content view. However we should throw an Error when the content is already set.

@hshristov I see your point and just to give you a reference the [ContentControl in WPF already implements the IAddChild interface](http://referencesource.microsoft.com/#PresentationFramework/Framework/System/Windows/Controls/ContentControl.cs,f8ee23c685054e71) :smile:

atanasovg:
@Nathanaela The proposed implementation does not suggest it will override the `content` property each time `addChild` is called. Also, calling `addChild(null/undefined)` will have the same effect as `removeChild`.
@ligaz This is somewhat religious :smiley:. For me, having an `addChild` method, which throws when more than one child is added seems somewhat odd. I would vote for Hristo's opinion. However, we are trying to create the best API's for our users, so if we have more votes on this we will definitely add it.

Nathanaela:
@atanasovg - The implementation was done more to solve a crashing issue that newer developers would hit.   Having it throw an error (i..e **This control only supports one child**) on adding a second child would at least be something that even though will crash the application would be useful and at least tell them they are attempting to use the control incorrectly and why.   Throwing on trying to add a null/undefined children, also probably should throw a error message just to be helpful.   =

 I have no issues making those changes and pushing another patch.       The biggest thing for me is that if addChild throws a error; at least the callstack & error should makes some sense.    if I recall correctly the error didn't really make any sense for someone new to the framework (it makes a lot more sense now that I've spent a lot of time with the framework).

atanasovg:
@Nathanaela I am missing the user scenario here - why would one call `addChild` on a `ContentView`? You are saying that the `addChild` method is in the prototype chain of the `ContentView` class but it actually isn't. `ContentView` derives from `CustomLayoutView`, inheriting `View` directly. The only class that has these children methods is the `Layout` one and its inheritors. So, the error produced by calling `addChild` on a `ContentView` should be `Undefined is not a function`. If one needs to traverse the visual tree there are the `viewModule.eachDescendant` and `View._eachChildView` methods.

Nathanaela:
@atanasovg - the reason someone would use addChild would be because you used it on other controls that had a child control.   As I said before, the only reason for the patch is consistency.   
 As for it being in the prototype chain -- I'm wrong,  :-)       I thought the original error message had a call stack way up into the java code before it crashed.    I went and retested it yesterday and realized that I miss-read the call stack.

dtopuzov:
run ci

hdeshev:
Just noticed this PR and it crossed my mind that having a unified `addChild`/`removeChild`/`insertChild` interface for view descendants would make monstrosities such as [the one](https://github.com/NativeScript/nativescript-angular/blob/master/src/nativescript-angular/view_node.ts#L137-L153) in the Angular 2 renderer codebase get better or completely disappear.

:+1:

enchev:
Hey folks,

If this can help @hdeshev and our Angular support I suggest to merge this pull request. @atanasovg @hshristov what do you think?

Thanks

atanasovg:
As far as I understand, Deshev's request is to introduce these `add/removechild` methods on a per `View` instance basis. Either way we have the private [_addView](https://github.com/NativeScript/NativeScript/blob/master/ui/core/view.d.ts#L451) and [_removeView](https://github.com/NativeScript/NativeScript/blob/master/ui/core/view.d.ts#L454) methods, which do similar things. So we basically will need to make these public and documented. Generally the approach for structuring the public API of a Visual Tree and how children are inserted is either based on `Containers` (`Layouts` in our case) or each base `View` (`Element`) has this notation - like the suggested one.

This however, while easy to implement, will be a fundamental breaking change. 

@hdeshev Wouldn't the private `_addView/_removeView` methods suffice for the Angular case?

hdeshev:
@atanasovg `_addView/_removeView` is probably good enough.

enchev:
In this case should we create issue related to public exposure of _addView and _removeView and close this pull request? @NathanaelA @atanasovg @hshristov @hdeshev what do you think?

atanasovg:
>This however, while easy to implement, will be a fundamental breaking change.

I think we should not change the current API - the Layout approach works fine and we have the private methods to match @hdeshev request.

PR #211, opened 2015-05-24
fix view name in trace msg when style property is not handled

lighta971:


ErjanGavalji:
@lighta971, thanks!

Would you please, sign the [CLA](http://www.nativescript.org/cla)?

Cheers,
Erjan

lighta971:
Hi, I did.

enchev:
@lighta971 It seems you are using different email address similar to this case:

PR #225, opened 2015-05-27
Fixed #222: ListPicker showing number of items in list

### LINKED ISSUE #222: ListPicker showing number of items in list (Ticket938420)
Hello,

I am having strange issues with the ListPicker UI element.  It shows the correct items, but sometimes also an additional item which is the number of items in the list.  In the attached project, it will start out showing "0".  It should be blank - because the representative list contains no elements.  If you select the first passenger (John Smith) all is well and good. Then if you select both passengers, it will show them in the ListPicker, with a third element that says "2".  Then unselect John Smith and it will correctly show Jane Smith, with an item that says "1". 

What is going on here and how can I stop these number items appearing?

Thanks!

https://admin.telerik.com/?Context=Support&Control=~%2fUserControls%2fSupport%2fThreadReply.ascx&ST=938420&SM=0

Sample Project: https://www.dropbox.com/s/k2srnjhq9mtyiqr/Ticket938420.zip?dl=0

hamorphis:


nsndeck:
:+1:

enchev:
:+1:

PR #226, opened 2015-05-27
separator is no longer forced to the edge

enchev:
Use this code in ListView.itemLoading event if you want to control the
inset:
```JavaScript
function listViewItemLoading(args) {
if (args.ios) {
   args.ios.separatorInset = UIEdgeInsetsZero;
   args.ios.preservesSuperviewLayoutMargins = false;
   args.ios.layoutMargins = UIEdgeInsetsZero;
 }
}
```


nsndeck:
:+1:

enchev:
Fixed #227.

PR #232, opened 2015-05-28
Fixed #228: ListPicker for Android shows text after bound to an empty…

### LINKED ISSUE #228: ListPicker for Android shows text after bound to an empty array.
If the ListPicker was originally bound to an array of strings and subsequently bound to an empty/null/undefined array, the last selected item could be still seen. The problem comes from the fact that the Android NumberPicker has no concept of zero items, i.e. it always has to have at least one item which is zero (minValue = value = maxValue = 0).

hamorphis:
… array.


enchev:
:+1:

hshristov:
:+1:

ErjanGavalji:
:+1:

ErjanGavalji:
:+1:

nsndeck:
:+1:

PR #235, opened 2015-05-28
css import support added

enchev:


ErjanGavalji:
:+1:

hshristov:
Do we cache CSS selectors if you navigate twice to the same page?

enchev:
@hshristov Not sure. Maybe this can be the next task.

ErjanGavalji:
@hshristov @enchev Caching is an interesting domain. We must be careful not to have too much stuff cached. A user might never use the cached item(s), thus cache only pointlessly eating memory.

dtopuzov:
run ci

dtopuzov:
run ci

nsndeck:
:+1:

PR #323, opened 2015-06-15
Resolved  #322: Creating an Observable by passing a JSON object in th…

### LINKED ISSUE #322: Creating an Observable by passing a JSON object in the constructor does not define the respective properties on the Observable object instance.
As a result, when someone changes the value of one of those properties through direct JavaScript notation (i.e. without calling Observable.set method), no notifications are fired and binding are not updated. For example:

```JavaScript
var viewModel = new observableModule.Observable({ count:5 });
viewModel.count++; 
```

will not update any bindings to **count**.

hamorphis:
…e constructor does not define the respective properties on the Observable object instance.


hshristov:
:+1:

PR #334, opened 2015-06-16
relative path for web view

truongsinh:
if url is, for example "~/my/path.ext", then realUrl is "file://bundle/id/location/my/path.ext"


ErjanGavalji:
@truongsinh, thanks for the contribution!

There is the `src` property that accepts value of variety of types (ImageSource, relative path, URL...) which can be used for the purpose. The [URL property is obsolete](https://github.com/NativeScript/NativeScript/blob/master/ui/web-view/web-view.d.ts#L39).

I'm closing the pull request now.

truongsinh:
Thanks.

truongsinh:
@duongnhatduy

Issue #363, opened 2015-06-26
Expose native Android activity events


valentinstoychev:
Often is the case when using 3rd party native library that a developer want to use the native application events. They are now available as overrides, but may be set only once - for example - ```application.android.onActivityDestroyed```.

The next time this is set it will override the previous caller. That is why we need these to be implemented as events.

PR #374, opened 2015-07-01
Android activity events

enchev:


hshristov:
Are we going to deprecate callback methods?

enchev:
@hshristov  Good idea! maybe we can write in the console from the callbacks that they are deprecated?

ligaz:
When you switch to a newer version of TypeScript compiler you can create a `@Depricated` decorator that can write to the console whenever the method is called.

enchev:
@ligaz We use already TypeScript 1.5 and we can definitely use decorators. I'll make different pull request later.

nsndeck:
:+1:

ErjanGavalji:
:+1:

enchev:
Implemented https://github.com/NativeScript/NativeScript/issues/363

Issue #391, opened 2015-07-06
Implement cancellable back button pressed for Android.


hamorphis:
Implement cancellable back button pressed for Android.

hamorphis:
var application = require("application");
var frame = require("ui/frame");
application.mainModule = "cuteness.io/main-page";
application.on(application.androidActivityBackPressedEvent, function (args) {
    // You can assign your pages an id to distinguish between them.
    if (frame.topmost().currentPage.id === "detailsPage") {
        // Cancel the back navigation
        args.cancel = true;
        // do your thing here, i.e. frame.topmost().navigate(...); 
    }
});
application.start();

PR #392, opened 2015-07-06
Resolved Issue #391: Added cancellable Android activityBackPressed ev…

hamorphis:
…ent + fixed the names of the other activity events because they were wrong.


atanasovg:
How about exposing these events on the `application.android` object? This seems architecturally more correct to me. I know the `.android` object is currently lazy initialized but there is no concrete reason for this, so it may safely be available from the very beginning of the `application` module initialization.

hamorphis:
:+1: @atanasovg

Issue #191, opened 2015-05-20
Extend the set of support CSS properties in {N}


valentinstoychev:
Currently we have a limited set of [CSS style properties support](http://docs.nativescript.org/styling) and we are planning to extend the list of supported properties with our next releases.

Please comment below what additional properties you would like to see supported as part of {N}.

vakrilov:
Properties related to backgrounds:
 * `background-position`
 * `background-repeat`

And some font properties:
 * `font-family`
 * `font-style`
 * `font-weight`

bundyo:
These would be nice in no particular order:
* inherit/initial/default
* overflow/-x/-y
* box-shadow
* border (current XML element is rather strange, maybe the element can be automatically wrapped with it upon encounter of this property?)
* background-size
* left/right/top/bottom with width/height auto for absolute layouts? Width and height can't help much when the View changes its size.
* text-transform/text-decoration
* text-shadow
* transform?

Low priority, for bling mostly:
* [filter](https://developer.mozilla.org/en-US/docs/Web/CSS/filter)
* @supports() type of queries

RangerMauve:
Would Flexbox ever be considered, maybe using Facebook's existing [css-layout](https://github.com/facebook/css-layout) code? As I understand it, however. The absolutely positioned css wasn't considered because it it'd require NativeScript to implement its own layout engine.

emiloberg:
Not so much a property as a value: Sizes (width/height) as percentages

jlooper:
I would really like to have background-image property options supported, especially  no-repeat and -x/-y. This is a showstopper for an app I'm working on now.

bundyo:
Also, when adding support for background-position, please consider the little known right/bottom offset positioning, for instance:

    background-position: right 10px bottom 20px;

emiloberg:
`font-weight` because currently the only way of setting font-weight is
```
        <Label>
            <Label.formattedText>
                <FormattedString>
                    <FormattedString.spans>
                        <Span text="My bold text" fontAttributes="Bold" />
                    </FormattedString.spans>
                </FormattedString>
            </Label.formattedText>
        </Label>
```

Or am I missing something?

manijak:
What about state-selectors? To be able to style pressed or hover effects? 
```css
button:pressed{
  background-color:#FED700;
}
.mybutton:pressed{
  background-color:#FED700;
}
```

alejonext:
@manijak seudo-class

[...]

PR #394, opened 2015-07-06
Feature/css background props

vakrilov:
Progress on #191 
Rework the background/border support. All of the following CSS properties are now supported:
* `background-color`
* `background-image`
* `background-position`
* `background-repeat`
* `background-size`
* `border-color`
* `border-width`
* `border-radius`


nsndeck:
:+1:

PR #396, opened 2015-07-06
Moved Activity related Android-specific events to application.android…

hamorphis:
… instance. Renamed cross-platform events by deleting their "event" suffix.

### DIFF
--- a/ui/frame/frame.android.ts
+++ b/ui/frame/frame.android.ts
@@ -427,8 +427,8 @@ var NativeActivity = {
             result(requestCode, resultCode, data);
         }
 
-        application.notify(<application.AndroidActivityResultEventData>{
-            eventName: application.androidActivityResultEvent,
+        application.android.notify(<application.AndroidActivityResultEventData>{
+            eventName: "activityResult",
             object: application.android,
             activity: this,
             requestCode: requestCode,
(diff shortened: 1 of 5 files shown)


nsndeck:
:+1:

vakrilov:
:+1:

enchev:
:+1:

ui/frame/frame.android.ts
```diff
@@ -496,12 +496,12 @@ var NativeActivity = {
...
-            eventName: application.androidActivityBackPressedEvent,
+            eventName: "activityBackPressed",
             object: application.android,
             activity: this,
             cancel: false,
         };
-        application.notify(args);
+        application.android.notify(args);
 
         if (args.cancel) {
```
ligaz: I know this comment is not related to this PR but I was unable to comment to the previous one. So here it goes:

Consider adding a [`preventDefault()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) method on the args because this is what JavaScript web developers will expect from this API.

hamorphis: I will merge the branch now and will discuss the preventDefault suggestion with the team, since it might affect other events as well. We might need to change this in a couple of places.

PR #400, opened 2015-07-07
Fixed wrong event names in application.d.ts + added constants for the…

hamorphis:
… Android Activity Events + reordering of members + a small tslint fix.


nsndeck:
:+1:

PR #402, opened 2015-07-07
Feature/inline style only from xml

vakrilov:
In this PR:
 * You should be able to set inline style only form XML. Doing `view.style="color: red"` from code should throw.
 * `BorderView` fixed to fallback to the generic borders


enchev:
:+1:

Issue #389, opened 2015-07-03
Add dynamic Widget/Component creation


patrickseda:
### What’s the problem?
The current UI definition paradigm requires us to know exactly which UI elements are needed before a `Page` creation. This is often ***not*** the case for larger apps. Although data binding handles dynamic UI updating nicely for tabular data and other small items, it is not the right solution for large dynamically generated UI compositions.

### What do we have now?
The existing capabilities of {N} ***do*** allow for us to specify isolated groupings of UI components (*or widgets*) referenced into the XML of the parent `Page`.
```
// Define our Widget in file "xml-declaration/widget1".
<StackLayout id="stackLayout" >
    <Label id="label1" text="Label1" />
    <Button id="button1" text="Button1" /> 
</StackLayout>
```
```
// Embed our custom Widget into our Page.
<Page xmlns:widgets="xml-declaration/widget1">
     <widgets:Widget1 />
</Page>
```

### What do we need?
We need a dynamic widget builder, providing full control over ***when*** a custom widget gets created.

Ideally we’d expect its usage and API to be simple, perhaps something like:
*(using the widget XML declared above)*
```
var widget1 = viewBuilder.build({ name: ‘xml-declaration/widget1’ });
someView.addChild(widget1);
```

Currently we ***do*** have the ability to dynamically create `Page` views/controllers from an XML file.
This is done via the `Frame` class using the static method `resolvePageFromEntry()`:
```
var widget1 = frameCommon.resolvePageFromEntry({ moduleName: 'xml-declaration/widget1' });
someView.addChild(widget1);
```
where the XML would be declared something like:
```
<Page xmlns="http://www.nativescript.org/tns.xsd" loaded="pageLoaded" >
    <StackLayout id="stackLayout" >
        <Label id="label1" text="Label1" />
        <Button id="button1" text="Button1" /> 
    </StackLayout>
</Page>
```
Note that the widget must be wrapped in a `Page` element or this will ***not work***!
But making every widget a `Page` would incur excessive overhead, and quite frankly you could create some nonsensical compositions, like a `Page` with a `StackedLayout` parent.

This is not the right solution.

### Technical considerations
The main *benefit* of having a `Page` as our root in a widget is contains the internal method `addCssFile()` which applies styling during the `resolvePageFromEntry()` call, where the css file name is dynamically determined from the target module name.

[...]

PR #403, opened 2015-07-07
Custom component loading improved + tests

enchev:


enchev:
Implemented part of https://github.com/NativeScript/NativeScript/issues/389

hamorphis:
I like
:+1:

vakrilov:
:+1:

ui/builder/builder.ts
```diff
@@ -202,9 +175,65 @@ function parseInternal(value: string, exports: any): componentBuilder.ComponentM
...
+    componentPath = componentPath.replace("~/", "");
+
+    var fileName = componentPath;
+    console.log("BEFORE: " + componentPath)
+
+    if (!fs.File.exists(fileName)) {
+        fileName = fs.path.join(fs.knownFolders.currentApp().path, componentPath, componentName) + ".xml";
+    }
+
+    console.log("AFTER: " + fileName)
```
ErjanGavalji: If we want these logs, we can replace them with trace statements...

ErjanGavalji:
:+1:

ui/builder/builder.ts
```diff
@@ -112,40 +113,12 @@ function parseInternal(value: string, exports: any): componentBuilder.ComponentM
 
                 var componentModule: componentBuilder.ComponentModule;
 
-                if (args.prefix) {
+                if (args.prefix && args.namespace) {
```
hshristov: I think this check should be removed because it is not universal. In xml I can prefix with the default namespace even our components so this is not the correct check.

PR #404, opened 2015-07-07
Remove NSError** parameters

jasssonpet:
See https://github.com/NativeScript/ios-runtime/issues/186

~~Some of the changed methods should be wrapped in `try/catch` an the `onError` callback should be called.~~


fealebenpae:
@ligaz: in what way is this patch a breaking change, what is needed for it to be compatible?

ligaz:
The main point is will pre 1.2 tns modules work with 1.2 version of the iOS runtime? If the answer is no then this is a breaking change. If the answer is yes then everything is fine and we can safely remove the tag.

fealebenpae:
We are marking the `NSError**` parameters as optional - omitting them will enable the new throwing behavior, but code that still uses them will activate the legacy path. You can check out the linked issue for more details.

However, we think it best to promote the new behavior in order to force developers to handle errors from native API's, hence this patch. This patch should not affect the public API surface of the cross-platform modules and thus I don't think it's a breaking change.

We have engineered the throwing path to be backwards-compatible with older JavaScript code and so we should support running an older version of the modules on the new runtime. Is running newer modules on an older runtime such a strict requirement that breaking it merits the tag on this pull request?

ligaz:
My point is that we are not thinking deep enough about breaking changes. It is super easy to introduce them and they can bite the community in an unexpected ways. If we are open to them no matter how small and tiny they could be and just communicate them well in our release notes we will be fine.

fealebenpae:
I understand, we all are extremely aware of this at the runtime level, and that is why @jasssonpet has gone through tremendous hoops to keep things as backwards-compatible as possible, even at the expense of possible fast-paths the code could have used.

In this particular case we are trying to amend a lapse at a fundamental level - the fact of the matter is that up to now several critical layers of client-facing code were ignoring some of the errors returned by the underlying native APIs. While I agree that client-facing code that suddenly starts returning errors can constitute a breaking change, it is at most a change that breaks already broken behavior in order to make it right - potentially painful, like having to re-set a bone that has healed wrong, but vital nonetheless.

In any case, this will in no way impact existing NativeScript projects, because users will have to specifically opt in by updating the modules and runtime of their projects. Thankfully we are not bound by the compatibility constraints of system-wide frameworks like .Net and Node.js are. If we favor compatibility over correctness too zealously, we risk ending up with win32.

atanasovg:
I am confused - is this or is this not a breaking change? As far as I understand from @fealebenpae, the  error parameter just becomes optional, which does not break existing code. @ligaz?

PanayotCankov:
**This is not a breaking change**, but someone from the modules team will have to do that:

> @jasssonpet: Some of the changed methods should be wrapped in try/catch an the onError callback should be called.

I strongly recommend adopting the new version (that throws). And since it throws you can not "if" the method result. Instead you should try/catch it. NativeScript for iOS will try to keep the Objective-C API bridging to JavaScript as close as possible to the Objective-C API bridging to Swift.

jasssonpet:
run ci

jasssonpet:
I think this PR is finished now.

dtopuzov:
run ci

Issue #406, opened 2015-07-08
Prompt dialog - iOS7: okButton returns result equal to false, cancelButton returns result equal to true


vchimev:
The issue is replicable on iOS7 and NOT replicable on iOS8.

PR #413, opened 2015-07-09
dialogs result in iOS7 fixed

enchev:
Fixed https://github.com/NativeScript/NativeScript/issues/406


enchev:
run ci

enchev:
run ci

ErjanGavalji:
:+1: 

Add a comment that this is due to the API nuisance in iOS7 and that this will get removed in a year or two when ONLY a nice API is available?

Issue #304, opened 2015-06-10
Expose additional properties of NavBar/ActionBar


hshristov:
- [x] BackButton
- [x] Title
- [x] Icon
- [x] CenterView
- [x] ActionItems/optionsMenu

kvelikov:
Here are some additional suggestions
- [ ] Background color
- [ ] FontSize
- [ ] FontFamily

PR #433, opened 2015-07-14
Action bar

vakrilov:
Implementation for #304


nsndeck:
:+1:

ui/action-bar/action-bar.d.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        navigationButton: NavigationButton;
+        actionItems: ActionItems;
+        titleView: view.View;
+
+        page: pages.Page;
+
+        shouldShow(): boolean
+
+        updateActionBar();
```
atanasovg: ActionBar.updateActionBar - there is tautology here. Only `update` sounds better to me.

ui/action-bar/action-bar.d.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        //@private
+        _updateAndroidActionBar(menu: android.view.IMenu);
```
atanasovg: Same here - `updateAndroid` sounds better to me.

PR #495, opened 2015-07-27
Rename the tns-core package to tns-core-modules.

ErjanGavalji:
That will make its name easily understandable. It will be used as a plugin in the future.


enchev:
tumblr_inline_mh3clp8yer1r3kdj6

PR #507, opened 2015-07-29
Stop __extends function output

ErjanGavalji:
The __extends function is declared in the runtimes. Thus, the one
that the TypeScript compiler generates is reduntant and only
causes unnecessary marshaling delay.


enchev:
:+1:

PR #514, opened 2015-07-29
Add the __decorate function

ErjanGavalji:
Removing __extends function generation by setting the
noEmitHelpers flag we removed the __decorate as well.
While the __extends exists in the runtimes, __decorate
is an EcmaScript-related functionality, hence its place
is in the modules.


vakrilov:
:+1:

PR #573, opened 2015-08-12
Search bar font

enchev:


ui/styling/stylers.ios.ts
```diff
@@ -437,6 +437,33 @@ export class SearchBarStyler implements definition.stylers.Styler {
         }
     }
 
+    // font
+    private static setFontInternalProperty(view: view.View, newValue: any, nativeValue: any) {
+        var bar = <UISearchBar>view.ios;
+        var sf = <UITextField>bar.valueForKey("_searchField");
```
fealebenpae: I think using private fields of UIKit classes has the risk of breaking in a future OS release.

enchev: Do we have any alternative? I'll gladly replace this!

enchev:
run ci

ui/styling/stylers.ios.ts
fealebenpae: Ideally we'd base our CSS on UIAppearance, which is the iOS HIG-compliant way of doing things, but that would mean supporting things only UIAppearance does, or using custom controls. [Pixate](https://github.com/Pixate/pixate-freestyle-ios) actually does this with dynamic subclassing.

In this particular instance we can do something like `[[UITextField appearanceWhenContainedIn:[UISearchBar class], nil] setFont:font]`, but that would change all text fields in search bars across the application, so it wouldn't be as granular as CSS with its ids and classes.

fealebenpae:
:+1:

apps/tests/ui/search-bar/search-bar-tests.ts
```diff
@@ -58,6 +58,22 @@ export var testSearchBarHintColorAndroid = function () {
...
+
+        searchBar.text = "";
+        searchBar.hint = "hint font-size test";
+
+        var expectedValue = 30;
+        var actualValue;
+
+        searchBar.style.fontSize = expectedValue;
+        actualValue = searchBarTestsNative.getNativeFontSize(searchBar);
+        TKUnit.assert(actualValue === expectedValue, "Actual: " + actualValue + "; Expected: " + expectedValue);
```
hshristov: We have TKUnit.assertEqual - method for easier/better error reporting :)

hshristov:
:+1:

Issue #587, opened 2015-08-14
Animation promise in iOS may never be resolved nor rejected.


genexliu:
https://github.com/NativeScript/NativeScript/blob/master/ui/animation/animation.ios.ts#L99

For iOS, if an animation object has some of its animations finished and some cancelled, _animationFinishedPromise will never be resolved nor rejected.

I have a fix for this and will try to submit a PR later.

PR #588, opened 2015-08-14
fix the issue that an iOS animation promise may never be resolved nor rejected

genexliu:
Before the fix, if an iOS animation object has some of its animations finished and some cancelled, that._cancelledAnimations and that._finishedAnimations will both be smaller than that._mergedPropertyAnimations.length and thus _animationFinishedPromise will never be resolved nor rejected. This PR tries to fix this bug.


genexliu:
Note: this is for issue #587

hamorphis:
Merged!

PR #591, opened 2015-08-14
Fix XHR statusText property to return just the text

ligaz:
In order to follow the spec the XHR statusText property should
return just the status text without the status code. Adding
tests to guard this behavior.

### DIFF
--- a/apps/tests/xhr-tests.ts
+++ b/apps/tests/xhr-tests.ts
@@ -221,6 +221,24 @@ export function test_raises_onload_Event(done) {
     xhr.send();
 }
 
+export function test_sets_status_and_statusText(done) {
+    let xhr = new XMLHttpRequest();
+    xhr.onreadystatechange = () => {
+        if (xhr.readyState > 3) {
+            try {
+                TKUnit.assertEqual(xhr.status, 200);
+                TKUnit.assertEqual(xhr.statusText, 'OK');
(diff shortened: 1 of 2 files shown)


moll:
Umm, I don't think XMLHttpRequest was wrong, again. :P Last I remember, XMLHttpRequest was supposed to return the status code in `statusText`. Which spec did you spotted that noted the opposite? https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/statusText lists it _with_ status.

moll:
Whoops, wrong link above. The correct one: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest.

ligaz:
Just tested this in Chrome and it returns the text without the status code :smile:

ligaz:
And [here](http://www.w3.org/TR/XMLHttpRequest/#the-statustext-attribute) is the spec that mentions that it should return just the HTTP status text.

ligaz:
Just tested in Firefox and it behave the same. I guess the bug is in the MDN docs.

moll:
You're right. I stand corrected. Great. At least something's consistent in this world. :)

enchev:
:+1:

dtopuzov:
run ci

dtopuzov:
:+1:

PR #614, opened 2015-08-23
Fix the 9 misspellings of _detach in the two Gestures files.

NathanaelA:
Nothing major; While perusing the source code I found this simple misspelling in the _detach function in the gestures files.    Would be worth fixing before it becomes a fixed point in time and space.


atanasovg:
run CI

dtopuzov:
run ci

dtopuzov:
run CI

dtopuzov:
run ci

NathanaelA:
Is their a way that the reports could be automatically forwarded to the public github by the ns-bot; or on that reports server, that specific path could be public?  Because doing a search through the entire source code base I only find those 9 misspellings of it...

vakrilov:
:+1:

enchev:
:+1:

Issue #583, opened 2015-08-13
Page onNavigatingFrom should allow for canceling


jkappel71:
Say you have a data entry form that a user has filled out, but they have not filled out all required fields.  Pressing the back button (on android) causes the app to go back, no way to cancel it at all.  If the form is not properly filled out, you can't save the data so it's just lost.

We need to be able to respond to the onNavigatingFrom event in some manner that will cause the navigation event to be cancelled.

stonecourier:
Was just about to raise this as an issue also. I would like to see this implemented.

sitefinitysteve:
+1

NathanaelA:
Whoever works on this one; read the notes in #618; we had a long discussion on implementation.

spierala:
I think on Android there are (in the meantime) some good ways to cancel back navigation:

- Use of activityBackPressed: https://docs.nativescript.org/core-concepts/application-lifecycle
- Use of NavigationButton and overwrite the tap function: https://docs.nativescript.org/ui/action-bar

On iOS as far as I know the only solution is to hide the NavigationButton with ``visibility="collapsed`` and mimic the Back Button by adding a custom ActionItem which can have an icon and tap action of your choice.
The tap action can then cancel or allow the navigation based on some condition. 
You can navigate programatically by using ``this.routerExtensions.backToPreviousPage()`` (also see: https://docs.nativescript.org/core-concepts/angular-navigation).
The biggest challenge is then to style the custom back button the same way as the default back button. E.g. I am still on the search for the back button icon on iOS :)

PR #618, opened 2015-08-23
Created a new event to allow blocking of navigation.

NathanaelA:
This creates an event that gets ran before navigation is even tried.  This allows a page to cancel navigating from a page. Fix for #583 

Example:
Page.on('canNavigateFrom', function(args) {   args.canNavigate = false; });

### DIFF
--- a/ui/frame/frame-common.ts
+++ b/ui/frame/frame-common.ts
@@ -130,6 +130,9 @@ export class Frame extends view.CustomLayoutView implements definition.Frame {
             // TODO: Do we need to throw an error?
             return;
         }
+        if (!this._onCanNavigateFrom()) {
+            return;
+        }
 
         var backstackEntry = this._backStack.pop();
         var navigationContext: NavigationContext = {
@@ -149,6 +152,9 @@ export class Frame extends view.CustomLayoutView implements definition.Frame {
 
     public navigate(param: any) {
         trace.write(this._getTraceId() + ".navigate();", trace.categories.Navigation);
+        if (!this._onCanNavigateFrom()) {
+            return;
+        }
 
         var entry = buildEntryFromArgs(param);
         var page = resolvePageFromEntry(entry);
@@ -230,6 +236,13 @@ export class Frame extends view.CustomLayoutView implements definition.Frame {
 
     public _navigateCore(backstackEntry: definition.BackstackEntry) {
         //
+    }
+
+    public _onCanNavigateFrom() {
+        if (this.currentPage) {
+            return this.currentPage.onCanNavigateFrom();
+        }
+        return true;
     }
 
     public _onNavigatingTo(backstackEntry: definition.BackstackEntry) {
--- a/ui/page/page-common.ts
+++ b/ui/page/page-common.ts
@@ -27,6 +27,7 @@ function onActionBarHiddenPropertyChanged(data: dependencyObservable.PropertyCha
 
 export class Page extends contentView.ContentView implements dts.Page {
     public static actionBarHiddenProperty = actionBarHiddenProperty;
+    public static canNavigateFromEvent = "canNavigateFrom";
     public static navigatingToEvent = "navigatingTo";
     public static navigatedToEvent = "navigatedTo";
     public static navigatingFromEvent = "navigatingFrom";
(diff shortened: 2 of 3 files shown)


Nathanaela:
A couple notes; the naming I was unsure of as I haven't seen this technique applied anywhere else in the TNS code base.   It was a toss up on if I should use:
canNavigateFrom  or "allowNavigation"  ; the Pull request can easily be updated to switch the naming if the allowNavigation is preferred.   

 The args.canNavigate by default is passed in as true, so ANY event handler that is listening to the canNavigateFrom event, is allowed to switch that to false to cancel the navigation.     This even gets called on both a navigate(x) and a goBack() request before any other events are fired; and if the result is false; then it will cancel it right their before any frame stack modification is done.

atanasovg:
Hey @NathanaelA, thanks a bunch for your contribution. We discussed your pull internally and our suggestion is to use the existing `onNavigatingFrom` event rather than creating a new event. The event data may be extended with a new `cancel` property, which, if set to `true`, will prevent the page from being navigated from.

A similar implementation may be found in the application module for android - as stated in the [definition file](https://github.com/NativeScript/NativeScript/blob/master/application/application.d.ts#L286).

Additionally, my thinking is that it is a better idea to add a more general approach to canceling back/forward navigation. Because it is the `ui/frame` module that manages the navigation, it seems naturally to me to extend the `Frame` class with, for example, a `previewNavigation` event, which will provide the following data:

```
interface PreviewNavigationData {
    entry: NavigationEntry,
    isBackNavigation: boolean
    cancel: boolean
}
```

What do you think?

NathanaelA:
Well, their was a couple reasons why I did it this way (because initially I looked at using the OnNavigatingFrom event).  :grinning:    The onNavigatingFrom actually fires VERY late in the navigation process.  You would have to refactor a lot of things to use this event as this event is fired AFTER; loading and the new page, after pushing and popping the back stack...  So either that event has to be moved up a lot sooner in the code; or you have to have a lot of code added to "undo" the changes if cancel = true.

As for cancel in the app module; awesome -- I didn't realize there was somewhere else you had this ability.  I would definitely go with using "cancel" to keep things consistent.

As for using the page component; the only reason I used it was because that was where and how you do all your other page related events.   You don't do them off the frame.  I don't want this routine to run on every single navigation; I would only want this currently only on some pages.  So you actually want to do **page.on("canNavigateFrom", function (args) { args.cancel = true; });** not **frame.on(...)**   

Actually to be honest, as I dug into this and other issues over the week -- I think ultimately it would be much better to extend the event system to support **stopPropagation()** and **preventDefault()**.    This would eliminate you having to create custom event handlers for situations like this.     stopPropagation would eliminate the rest of the event chain from running.   preventDefault would allow me to cancel the default behavior.  So, when the navigatingFrom (or canNavigateFrom) event is fired, I would just say args.preventDefault() and be done.   As then it is up to the event firer to check the return value.   Both could be easily added to the Observable component at this point and would only require minor changes to the App module -> back cancel code and my above PR to support this.

dtopuzov:
run ci

enchev:
Hey @NathanaelA, @atanasovg, @hshristov,

What the status of this pull request? Is it still valid or the requested functionality is already available?

Thanks!

atanasovg:
Well, as I already stated in my previous comment, our suggestion is to **not** introduce new event but rather use the existing `onNavigatingFrom`. Still, @NathanaelA has a valid point where the proposed approach requires significantly more work (and low-level navigation changes) to be implemented. So, probably we will need to plan this issue and implement it internally.

enchev:
@atanasovg In this case should we close this pull request and open issue? @NathanaelA what do you think?

Thanks

NathanaelA:
@enchev - I have no issues with you closing the pull request; and us opening a issue.  However -- I know of several people who were asking about this functionality (hence the reason I coded it and did a pull request) and so I would recommend it be put close to the top of the 1.5 list of fixes that need to be done since the NS team wants to refactor the OnNavigatingFrom event to add this feature rather than use a new event..

enchev:
Ok, in this case let's close this and open an issue describing in details the problem with `onNavigatingFrom` event and what needs to be refactored.

PR #620, opened 2015-08-24
frame-common.ts is missing the code to automagically load the page.css

NathanaelA:
frame-common.ts is missing the code to automagically load the page.css file if it exists on a page that is created via the JavaScript createPage() function.  The expected behavior is that the page.css file is always loaded if it exists, no matter how the page is created.

### DIFF
--- a/apps/tests/ui/page/page-tests-common.ts
+++ b/apps/tests/ui/page/page-tests-common.ts
@@ -247,8 +247,37 @@ export function test_LoadPageFromModule() {
     try {
         var topFrame = FrameModule.topmost();
         TKUnit.assert(topFrame.currentPage.content instanceof LabelModule.Label, "Content of the test page should be a Label created within test-page-module.");
-        var testLabel = <LabelModule.Label>topFrame.currentPage.content
+        var testLabel = <LabelModule.Label>topFrame.currentPage.content;
         TKUnit.assert(testLabel.text === "Label created within a page module.");
+    }
+    finally {
+        helper.goBack();
+    }
+}
+
+export function test_LoadPageFromDeclarativeWithCSS() {
+    helper.navigateToModule("ui/page/test-page-declarative-css");
+    try {
+        var topFrame = FrameModule.topmost();
+        TKUnit.assert(topFrame.currentPage.content instanceof LabelModule.Label, "Content of the test page should be a Label created within test-page-module-css.");
+        var testLabel = <LabelModule.Label>topFrame.currentPage.content;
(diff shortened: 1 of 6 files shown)


NathanaelA:
The second patch technically is not needed since the expectation is that the createPage will return a valid page object.  But rather than crash at that line, in that case -- that second patch will allow the css addition check to fail and then the already existing error check about 10 lines later be able to throw the error about the page not being a page object.

atanasovg:
I would extract the css code from the `pageFromBuilder` method and add it to the `resolvePageFromEntry` one. This will save you copy-and-paste:

```javascript
if (moduleExports && moduleExports.createPage) {
    trace.write("Calling createPage()", trace.categories.Navigation);
    page = moduleExports.createPage();
}
else {
    page = pageFromBuilder(moduleNamePath, moduleExports);
}

if (!(page && page instanceof pages.Page)) {
    throw new Error("Failed to load Page from entry.moduleName: " + entry.moduleName);
}

// Possible CSS file path.
var cssFileName = fileResolverModule.resolveFileName(moduleNamePath, "css");
if (cssFileName) {
    page.addCssFile(cssFileName);
}
```

NathanaelA:
@atanasovg good idea, that actually is a better way because then then code is located ony in one place again.   I'll make that change.

NathanaelA:
Any other suggestions?   Do you need the patches rebased/ squashed?

enchev:
:+1:

hshristov:
@NathanaelA,

patch looks good. Could you add some simple unit test that verifies this change?
You could place it in https://github.com/NativeScript/NativeScript/tree/master/apps/tests/ui/page

NathanaelA:
@hshristov I don't see any unit tests for any page.css in /apps/tests/ui/page -->  Do you just want me to create a two new pages; one that has a Declarative UI xml file and loads the page.css and one that uses createPage.   Do I just need to visually show the CSS is loading, or is their some framework that can be ran to check the differences...

hshristov:
Simple createPage method with some css in the same folder which applies Page.background would be fine :)

NathanaelA:
Too late.  Already did more than that.  :grinning: I've added two tests.   I added the missing Declarative UI XML w/ CSS file and I added the new createPage w/ CSS.   I am running the background color test against the Label component though since having the Label defined properly is what verifies the page itself actually loaded fine.

hshristov:
Great. Thank you for your continuous support :)
I though that we had Declarative UI with CSS test but apparently we didn't.

Issue #590, opened 2015-08-14
CSS not working on Repeater when bindingContext set on 'navigatingTo'


emiloberg:
Easiest described with an example.

Look at this very simple app. It has a `Repeater` where every item will be a label with the cssClass `red`, and below them there's a "fixed" label with the text `This is always red` and the same cssClass.

```
<Page xmlns="http://www.nativescript.org/tns.xsd" loaded="pageLoaded">
	<StackLayout>
	    <Repeater items="{{ items }}">
	        <Repeater.itemTemplate>
	            <Label text="{{ title }}" cssClass="red" />
	        </Repeater.itemTemplate>
	    </Repeater>
	    <Label text="This is always red" cssClass="red" />
	</StackLayout>
</Page>
```

```
function pageLoaded(args) {
    var page = args.object;
    page.bindingContext = {
		items: [
			{
				title: 'Test1'
			},
			{
				title: 'Test2'
			}
		]
    };
}
exports.pageLoaded = pageLoaded;
```

```
.red {
	color: #ff0000;
}
```

So far so good. However, if I change from setting the bindingContext on `loaded` to `navigatingTo`, e.g. changing:

```
<Page xmlns="http://www.nativescript.org/tns.xsd" loaded="pageLoaded">
```
to
```
<Page xmlns="http://www.nativescript.org/tns.xsd" navigatingTo="pageLoaded">
```

then no CSS will be applied on the `itemTemplate` of the Repeater. Instead it will look like this. Note that CSS is still applied to the "fixed" label.

![skarmavbild 2015-08-14 kl 16 52 06](https://cloud.githubusercontent.com/assets/4673048/9276530/d5106d98-42a4-11e5-8f89-aaf9707c46ee.png)

PR #630, opened 2015-08-27
repeater children now created no matter if isLoaded is true

enchev:
Fixed https://github.com/NativeScript/NativeScript/issues/590


vakrilov:
:+1:

Issue #619, opened 2015-08-24
View.style is not consistent with Declarative UI .style property.


NathanaelA:
If you look at the view-common.ts class; attempting to assign anything via JavaScript to the **style** property of the class, will throw a runtime error about the style being read only.   However in the Declarative UI, you use the **style** property to assign a text style string to the component.     

This is inconsistent, Anything I can do in the Declarative UI; should be equally do-able via JavaScript.   Yes, I realize I can modify individual **View.style.someProperty**; but if I can use a String in the Declarative UI; I should be able to use a string via JavaScript.  

I should be able to do view.style = "padding-left: 2; background-color: red" and get this applied.   Just like if I put it in the Declarative UI XML.   

I believe what needs to happen if I apply a inline style string to the **style**; that I need to reset the styles variable; re-apply App.css, then Page.css; then run the same code that is inside **this._applyInlineStyle(inlineStyleString);**.   This way the style is reset back to scratch each time I apply a inline style string.  

Any arguments against?   If my position and assumptions on what needs to happen is correct, I can work on a commit patch later in the week...

enchev:
Hey @NathanaelA,

Very good question! 

Style property works very similar to HTML/JavaScript - in HTML/XML you can set the in-line string and in JavaScript you can manipulate directly the Style object using desired properties/methods. What we are missing in {N} is something similar to [setAttribute](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute) where you can use the same syntax as in HTML/XML.

If we have this method you will be able to set in-line style like this:
```
var lbl = new Label();
lbl.setAttribute("style", "color: red"):
```
What do you think?

NathanaelA:
Howdy @enchev,

setAttribute would be awesome addition.  

I realize this idea of setting style directly diverges from how the browser works (and I think the browser is wrong too.  :grinning: ).  But since this is a new platform, and maybe I'm different, but I prefer that the NativeScript itself is as internally consistent as it can to itself.  This divergence won't break any code.  If I can do style in the Declarative UI; I should be able to do it in JavaScript.    However, having a setAttribute would allow me to work around the issue much easier....

hshristov:
In my opinion setAttribute is too generic. This would mean that user will be able to set other properties as well (not only style property). I believe that there should be only one way of doing things. This will give us smaller test surface, consistent API, less confusion.
[...]

PR #632, opened 2015-08-28
setInlineStyle implementation + test

enchev:
Implemented https://github.com/NativeScript/NativeScript/issues/619

### DIFF
--- a/ui/core/view-common.ts
+++ b/ui/core/view-common.ts
@@ -923,7 +923,7 @@ export class View extends proxy.ProxyObject implements definition.View {
         return false;
     }
 
-    public setStyle(style: string): void {
+    public setInlineStyle(style: string): void {
         if (!types.isString(style)) {
             throw new Error("Parameter should be valid CSS string!");
         }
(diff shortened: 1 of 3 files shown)


ui/core/view-common.ts
```diff
@@ -923,6 +923,12 @@ export class View extends proxy.ProxyObject implements definition.View {
         return false;
     }
 
+    public setStyle(style: string): void {
+        if (types.isString(style)) {
```
hshristov: if not string - I think we should thrown an exception.

ui/core/view-common.ts
```diff
@@ -923,6 +923,12 @@ export class View extends proxy.ProxyObject implements definition.View {
         return false;
     }
 
+    public setStyle(style: string): void {
```
hshristov: And name it setInlineStlye to make sure that it will override CSS style.

hshristov:
:+1:

dtopuzov:
run ci

dtopuzov:
run ci

PR #637, opened 2015-08-28
Fix for Button padding, broderWidth & background in android.

hshristov:
Fix for types - isString and isNumber.
Fix page background to be applied in onLoaded method.
Fix for SearchBar reseColorProperty method.
Modified several tests to pass on VS Emulator for Android.
Enhancements for dependency-observable - it is now possible to add defaultValueGetter function to extract default value for a property. It can also be cached or not.
Remove android_constants because they are per theme and we cannot use them safely.


nsndeck:
:+1:

hshristov:
run ci

vakrilov:
:+1:

vchimev:
run ci

PR #653, opened 2015-09-01
Changed createSelectorsFromSyntaxTree to use a loop rather than filter.

### LINKED ISSUE #623: CSS Inconsistency...
The documentation specifies correctly that the css type-selectors are case in-sensitive; but does not say anything about the properties case sensistivity.   Since, in the browser I can do 
div { baCKgrouND-colOr: blUe; }  and it works; you would naturally assume that in NativeScript you could do the same thing...   Nope; apparently the property names are case sensitive and must be all lower case.  So why don't we have the css parser automatically lowercase the property name so that Button { Background-Color: Blue; } then works properly as expected!

Now to fix this we have a couple of choices:
1. css\reworkcss.js on line 232 we can append a .toLowerCase() to the place the property is assigned and returned.
2. In Style-scope, I can update the createSelectorsFromSyntaxTree function and add the toLowerCase() in their.
3. I haven't traced this down, but where ever it goes to set the style property when the rule matches; make it case automatically lowercase the property before it assigns it.

The quickest fix is to do it in reworkcss; but that means you have your own non-stock reworkcss.js file.   I hate when that happens in my projects...
So I am leaning toward 2 -- so that way the fix is in the code you control -- I can do the style-scope change fairly easily and that routine should actually be more performant when I'm done.  :grinning: 

Thoughts?

NathanaelA:
1. Changed code to lowercase the declaration property
2. Minimize the data pushed to each rule (by about half)

Benchmark on v8 shows this new code is about a 1/3 faster than the original code.  (Not to mention it should cause less memory pressure as it saves memory per css rule).

This is to fix #623

### DIFF
--- a/apps/tests/ui/style/style-tests.ts
+++ b/apps/tests/ui/style/style-tests.ts
@@ -816,6 +816,26 @@ export function test_set_invalid_CSS_values_dont_cause_crash() {
     }, invalidCSS);
 }
 
+// Check Mixed, Upper and lower case properties
+var casedCSS = ".cased {" +
+    "cOlOr: blue; " +
+    "FONT-SIZE: 30; " +
+    "background-color: red; " +
+    "}";
+
+export function test_set_mixed_CSS_cases_works() {
+    var testButton = new buttonModule.Button();
+    testButton.text = "Test";
+    testButton.cssClass = "cased";
+
+    helper.buildUIAndRunTest(testButton, function (views: Array<viewModule.View>) {
+        TKUnit.assertEqual(30, testButton.style.fontSize);
+        helper.assertViewBackgroundColor(testButton, "#FF0000");
+        helper.assertViewColor(testButton, "#0000FF");
+    }, casedCSS);
+}
+
+
 // <snippet module="ui/styling" title="styling">
 // For information and example how to use style properties please refer to special [**Styling**](../../../styling.md) topic. 
 // </snippet>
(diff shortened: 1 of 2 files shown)


enchev:
:+1:

vakrilov:
@NathanaelA 
Can you please verify this with a test. 
There are already some style test [here](https://github.com/NativeScript/NativeScript/blob/master/apps/tests/ui/style/style-tests.ts).

NathanaelA:
@vakrilov I added a test to that file to test to verify that lower case property names still work, and that Upper Case, and mixed case property values now work properly.   Anything else you need?

vakrilov:
@NathanaelA 
Thanks for the contribution
:+1:

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

enchev:
run ci

apps/tests/ui/style/style-tests.ts
```diff
@@ -816,6 +816,26 @@ export function test_set_invalid_CSS_values_dont_cause_crash() {
...
+    testButton.cssClass = "cased";
+
+    helper.buildUIAndRunTest(testButton, function (views: Array<viewModule.View>) {
+        TKUnit.assertEqual(30, testButton.style.fontSize);
+        helper.assertViewBackgroundColor(testButton, "#FF0000");
+        helper.assertViewColor(testButton, "#0000FF");
+    }, casedCSS);
+}
+
+
```
enchev: Hi @NathanaelA,

Please remove the second blank line!

Thanks

NathanaelA:
@enchev - Blank line deleted...

enchev:
run ci

enchev:
run ci

dtopuzov:
run ci

PR #661, opened 2015-09-02
Atanasovg/navigation backstack

### LINKED ISSUE #548: Expose public API controlling whether a Page should be added to the navigation backstack or not
It is a common use case to have some initial (and probably not only) page that the user should NOT be able to navigate back to - for example a LoginPage. Or an Authentication page. We should enable an API that allows for a page to be explicitly excluded from the back navigation. See [this thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/nativescript/eEDQJ4VinZw) for more information.

As a proposal, we may add an additional parameter to the [NavigationEntry API](https://github.com/NativeScript/NativeScript/blob/master/ui/frame/frame.d.ts#L127-L147) called `backstackVisible` (or alike).

atanasovg:
**IMPORTANT** Do not merge yet - we need unit tests & functional tests

# API:

### Exclude the main page from the backstack:

```js
var application = require("application");

// mainEntry is of type "NavigationEntry"
application.mainEntry = {
	moduleName: "./main-page",
	backstackVisible: false
}
```

### Navigate to a Page that should not be in the backstack
```js
var frame = require("ui/frame");
var topFrame = frame.topmost();

topFrame.navigate({
	moduleName: "./some-inner-page",
	backstackVisible: false
});
```

# TODO

* Write unit tests
* Write functional tests
* Update the navigation article or create a new one, explaining how the feature works

Resolves #548

### DIFF
--- a/CrossPlatformModules.csproj
+++ b/CrossPlatformModules.csproj
@@ -143,6 +143,7 @@
     </TypeScriptCompile>
     <TypeScriptCompile Include="apps\animations\model.ts" />
     <TypeScriptCompile Include="apps\orientation-demo\main-page.ts" />
+    <TypeScriptCompile Include="apps\tests\navigation-tests.ts" />
     <TypeScriptCompile Include="apps\tests\ui\animation\animation-tests.ts" />
     <TypeScriptCompile Include="apps\tests\ui\search-bar\search-bar-tests-native.android.ts" />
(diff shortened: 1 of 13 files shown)


dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

hamorphis:
run ci

Issue #69, opened 2015-04-24
Add rotate, translate & scale transforms properties on View


enchev:
_From @hshristov on March 11, 2015 13:41_

Rotate, translate & scale transforms.

_Copied from original issue: NativeScript/cross-platform-modules#163_

PR #665, opened 2015-09-03
Added View.translateX, View.translateY, View.scaleX, View.scaleY, Vie…

hamorphis:
…w.rotate properties.


enchev:
Implemented #69

PR #678, opened 2015-09-04
Dialog action fix

PeterStaev:
If the user wants to display only action sheet buttons w/o any text and
he sends null as message, the empty string causes an empty title to be
shown above the buttons. Having null instead, completely eliminates the
empty space.

### DIFF
--- a/ui/dialogs/dialogs.android.ts
+++ b/ui/dialogs/dialogs.android.ts
@@ -209,7 +209,7 @@ export function action(arg: any): Promise<string> {
 
     var options: dialogs.ActionOptions;
 
-    var defaultOptions = { cancelButtonText: dialogsCommon.CANCEL };
+    var defaultOptions = { title: null, cancelButtonText: dialogsCommon.CANCEL };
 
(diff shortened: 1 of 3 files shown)


hshristov:
@PeterStaev 
You are absolutely correct. The problem I see is that there should be another parameter for title and we should call the native api with the correct parameters. Right now we pass the message as Title in both platforms and we pass "" as message.

PeterStaev:
@hshristov  I agree this is a much better idea!

PeterStaev:
Added the title option. In order to not cause much problems for existing code using the dialog.action:
For **iOS7** - Continue to send the message as the title of the UIActionSheet since there is no message there. 
For **iOS8+** - Send both title and message to the UIAlertController
For **Android** - If there is a title sent, pass both title and message to the AlertDialog. If there is no title, pass the message as title for the AlertDialog

enchev:
run ci

enchev:
run ci

enchev:
:+1:

PR #679, opened 2015-09-04
Fix for cancel backswipe gesture. It was broken with the implementati…

hshristov:
…on of navigation to a page without adding it to backstack.


dtopuzov:
:+1:

vakrilov:
:+1:

PR #686, opened 2015-09-06
Added method setError to the textfield && Added the options passwordHint and userNameHint

maury91:
Hi,
i've added on the android branch of the text-field the method setError,
in android this method is used to validate the fields
i don't know the equivalent on ios, but i found that simple script that can be implemented https://github.com/tomkowz/TSValidatedTextField

Edit :

I've discovered that the options password and userName in the Login Dialog, set the defaults values for this fields.

I've added this two options to help the user to understand what the fields mean


dtopuzov:
run ci

dtopuzov:
run ci

enchev:
run ci

enchev:
Hi @maury91,

Will you add the same for iOS?

Thanks

PR #690, opened 2015-09-07
Link click detection

NathanaelA:
This makes the <HtmlView> clickable and automatic parsing.

Please note; their is one **possible negative** to this code; on iOS to allow the clicking ability the label has to allow selection, so the user can actually select text (i.e. you can actually select & copy the "Go to") in the label.   However, for the minor drawback; you can now do:

\<HtmlView html="Go to http://www.master-technology.com or email me at me@somewhere"/> and both links will just work.   :grinning:  (And \<a href="...">Link Text\</a> now also works... on both platforms)

### DIFF
--- a/ui/html-view/html-view.android.ts
+++ b/ui/html-view/html-view.android.ts
@@ -30,5 +30,7 @@ export class HtmlView extends common.HtmlView {
 
     public _createUI() {
         this._android = new android.widget.TextView(this._context);
+        this._android.setAutoLinkMask(15); // 15 (0x0f) = Linkify All
     }
+
 }
--- a/ui/html-view/html-view.android.ts
+++ b/ui/html-view/html-view.android.ts
@@ -10,6 +10,13 @@ function onHtmlPropertyChanged(data: dependencyObservable.PropertyChangeData) {
     }
 
     if (types.isString(data.newValue)) {
+        // If the data.newValue actually has a <a...> in it; we need to disable auto-linking
+        // as it internally disables anyways; and then links won't work..
+        if (data.newValue.toLowerCase().indexOf("<a ") > 0) {
+            view.android.setAutoLinkMask(0);
+        } else {
+            view.android.setAutoLinkMask(15); // 15 (0x0f) = Linkify All text links
+        }
         view.android.setText(<any>android.text.Html.fromHtml(data.newValue));
     } else {
         view.android.setText("");
@@ -30,7 +37,10 @@ export class HtmlView extends common.HtmlView {
 
     public _createUI() {
         this._android = new android.widget.TextView(this._context);
-        this._android.setAutoLinkMask(15); // 15 (0x0f) = Linkify All
+        // This makes the <a href...> work
+        this._android.setLinksClickable(true);
+        this._android.setMovementMethod(android.text.method.LinkMovementMethod.getInstance());
+
     }
 
 }
--- a/ui/html-view/html-view.android.ts
+++ b/ui/html-view/html-view.android.ts
@@ -10,13 +10,14 @@ function onHtmlPropertyChanged(data: dependencyObservable.PropertyChangeData) {
     }
 
     if (types.isString(data.newValue)) {
-        // If the data.newValue actually has a <a...> in it; we need to disable auto-linking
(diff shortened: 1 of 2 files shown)


NathanaelA:
I should also add; that we have to use the UITextView on iOS rather than the UILabel; as the UILabel doesn't support multiple links without making some modifications to the UILabel; and at that point you might as well use a third party component that is designed to be a Clickable Label.

enchev:
run ci

ui/html-view/html-view.android.ts
```diff
@@ -10,6 +10,13 @@ function onHtmlPropertyChanged(data: dependencyObservable.PropertyChangeData) {
     }
 
     if (types.isString(data.newValue)) {
+        // If the data.newValue actually has a <a...> in it; we need to disable auto-linking
+        // as it internally disables anyways; and then links won't work..
+        if (data.newValue.toLowerCase().indexOf("<a ") > 0) {
```
ligaz: The condition can be simplified to the following:
```ts
let mask = data.newValue.toLowerCase().indexOf("<a ") >= 0 ? 0 : 15;
view.android.setAutoLinkMask(mask)
```

dtopuzov:
tslint fails with:

Running "tslint:build" (tslint) task
Fatal error: Maximum call stack size exceeded

dtopuzov:
run ci

dtopuzov:
run ci

ui/html-view/html-view.android.ts
NathanaelA: I'm not a big fan of ternary ops, I find them in most places harder to read quickly.   But if you really want I can convert it to one.

NathanaelA:
@dtopuzov -- Are you sure some other code isn't causing the stack overflow.   The code change is rather simple, and I've ran the grunt multiple times while working on this and never saw any issues...

ui/html-view/html-view.android.ts
ligaz: The main issue is not the ternary but that you are doing the same operation (`setAutoLinkMask(...)`) in both bodies of the expression just passing different parameter. Imagine that in the future you have to change this code which will yield that you have to change it on two places instead of one. If you want to go without the ternary I would suggest the following refactoring:
```ts
let mask = 15;
if (data.newValue.toLowerCase().indexOf("<a ") >= 0) {
  mask = 0;
}
view.android.setAutoLinkMask(mask)
```

ui/html-view/html-view.ios.ts
```diff
@@ -27,21 +27,24 @@ function onHtmlPropertyChanged(data: dependencyObservable.PropertyChangeData) {
...
+        this._ios = UITextView.new();
 
-        this._ios = new UILabel();
-        this._ios.userInteractionEnabled = true;
-        this._ios.numberOfLines = 0;
+        this._ios.scrollEnabled = false;
+        this._ios.editable = false;
+        this._ios.selectable = true;
+        this._ios.userInteractionEnabled  = true;
+        this._ios.dataDetectorTypes = 255; // Future Proof this; valid max in iOS is currently 15; however ALL = MAX UNSIGNED INT which isn't valid in JavaScript
```
ligaz: Can the enum declaration (`UIDataDetectorTypes.UIDataDetectorTypeAll`) be used here instead?

NathanaelA: Ah, yes -- My note is incorrect; I was thinking JS was limited to under 32 bits.  But it is actually under 53 bit.  So it won't support a 64 bit number; but it will support a 33 bit number fine.  Is their a thunking performance issue for pulling in UIDataDetectorTypes.UIDataDetectorTypeAll rather than using a hard coded issue.

ligaz: The code readability should be the one we should optimize for because this is not a compute intensive code :smile:  

Regarding the question itself we should ask someone with deeper knowledge about iOS runtime. @fealebenpae can you elaborate if there will be any marshalling penalty if the enum is used instead of the integer value?

enchev:
run ci

ligaz:
:+1: well done.

enchev:
:+1:

enchev:
run ci

Issue #675, opened 2015-09-04
Allow CSS & JS files to be set in XML


hshristov:
Allow CSS file as well as JS code-behind file to be specified in the XML. This will allow CSS & JS to be set in different folder from the XML.

atanasovg:
+1, this is an easy-to-implement feature that is extremely useful when people need custom folder structure.

Plamen5kov:
:+1:  it would give us a project structure, so bigger projects are easier to read.

PR #706, opened 2015-09-10
CSS and XML code files now can be specified in XML

enchev:
Implemented #675

Example:
#### XML
```XML
<Page codeFile="~/custom-code-file"
           cssFile="~/custom-css-file.css" loaded="loaded">
  <Label cssClass='MyClass' />
</Page>
```
#### Code
```TypeScript
export function loaded(args) {
    console.log("Custom code file loaded!")
}
```
#### CSS
```CSS
.MyClass {
    background-color: green;
}
```

### DIFF
--- a/CrossPlatformModules.csproj
+++ b/CrossPlatformModules.csproj
@@ -79,6 +79,7 @@
     <TypeScriptCompile Include="apps\action-bar-demo\pages\data-binding.ts">
       <DependentUpon>data-binding.xml</DependentUpon>
     </TypeScriptCompile>
+    <TypeScriptCompile Include="apps\tests\xml-declaration\custom-code-file.ts" />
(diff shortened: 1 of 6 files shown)


enchev:
Example:
```XML
<Page codeFile="~/custom-code-file"
           cssFile="~/custom-css-file.css">
  <Label cssClass='MyClass' />
</Page>
```
```JavaScript
export function loaded(args) {
    args.object.customCodeLoaded = true;
}
```
```CSS
.MyClass {
    background-color: green;
}
```

hshristov:
:+1: 
It will be even better if we can move the 'implicit require' (the one that happens before calling component-builder) inside the component builder and don't require it if there is explicit 'codeFile' attribute set.

Plamen5kov:
:+1:

PR #723, opened 2015-09-13
Fixes the XML Entities issues on Android 

NathanaelA:
This patch adds two features -- during parsing of the XML it will now parse &amp;#xZZZZ;  and &amp;#YY; properly.       The place you normally notice this as being broken (and it totally is) is when you add any symbolic font that uses characters symbols that can't easily be accessed like Fonttastic or Google Material Icons.  In the current version of NS you currently can't use these fonts without remapping the font with a font program/service to move the glyphs from the FF00 range down to the normal text range.    With this patch, you can now access them just like you would in HTML.  &amp;#xFF00; would show that glyph.   (This is a VERY easy issue to run into with fonts)

Second it adds several hundred valid entity &names; html entities like &amp;copy; &amp;trade; &amp;euro; etc...

### DIFF
--- a/node-tests/test-xml.ts
+++ b/node-tests/test-xml.ts
@@ -32,4 +32,15 @@ describe("xml parser", () => {
         parser.parse("<element>&lt;&gt;&quot;&amp;&apos;</element>");
         assert.equal("<>\"&'", last_data);
     });
+
+    it("resolves greek letter entities in attributes", () => {
+        parser.parse("<element text='&Omega;'>blah</element>");
+        assert.equal("Ω", last_attrs.text);
+    });
+
+    it("resolves entities in element text", () => {
+        parser.parse("<element>&Omega;</element>");
+        assert.equal("Ω", last_data);
+    });
+
 });
--- a/xml/xml.ts
+++ b/xml/xml.ts
@@ -62,35 +62,35 @@ export class ParserEvent implements definition.ParserEvent {
     }
 }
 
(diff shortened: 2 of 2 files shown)


NathanaelA:
Please note this also gives the great side effect that it allows the &amp;#xXXXX; and entity. codes to be used everywhere rather than just in HtmlViews.     Simple Benchmarks show no speed difference on all my screens even with multiple codes per line.

enchev:
run ci

atanasovg:
@NathanaelA I have one question on this - why do you use arrays (`_amp160List`, `_ampGreekUpper` and `_ampGreekLower`) instead of Object literals (like `_ampCodes`) where the key is the word and the value is the actual index? Doing linear search in an array is much slower than a by-key (dictionary) look-up.

NathanaelA:
@atanasovg trying to limit the amount of memory it affects.   Adding all three of those entity lists adds more memory footprint.  And I was trying to limit the memory footprint as much as possible.  (Hence the reason they are declared out at the global scope; so we don't keep creating larger temporary variables that have to be cleaned up each time we call that routine) 

The array list of words is (from everything I understand on v8) will use a less memory than the object.   So I only used the object for the non-sequential words.    However with all that being said; I didn't actually measure it -- which maybe I should do.   Because I also do prefer the dictionary lookup.

NathanaelA:
I should also add; how complete do we want the html/xml entity list to be; http://dev.w3.org/html5/html-author/charref has the most complete list of which their appears to be a couple thousand of them.  My understanding is Chrome & Firefox support that entire list...    I pulled the most common ones from multiple places; but I didn't want to do that whole list if the patch didn't get accepted because putting in several thousand of them is A LOT OF WORK...  ;-D

NathanaelA:
@atanasovg -- ok, I ran the tests to verify my understanding.   I took the Object _ampCodes and measured exactly how much memory it currently takes as is (1,208) and converted it into a straight array and then measured it (32).  The amount of memory saved is pretty large.   

Now as for speed; I ran it through a search of 10 (good) & 1 (bad) different items 100,000 times for each one.   _ampCodes took about half the time (Dictionary: 458ms) vs _ampList (Array: 731ms) for doing a total 1,100,000 lookups each. Speed per entry is approx .00041ms vs .00066ms    I think we can spend the extra .couple nano-seconds for that large of a retained memory savings.  ;-D

dtopuzov:
run ci

hdeshev:
It just dawned on me that the XML entity code will be invoked only when we encounter an entity string, i.e. `_entitySearchRegEx` matches something that looks like `&blah;`. XML entities are a rarely-used feature and choosing either approach shouldn't affect the parsing speed much.

NathanaelA:
@atanasovg -- My earlier benchmark was wrong.   Sorry about the incorrect info; I had a typo in the Object look up so it was causing a massive slowdown in the object loop.   You are correct the Object lookup is considerably faster.    (100,000 loops with 10 good & 1 bad lookup)
The speed difference is:  Object: 19sec,  Map: 24ms, and List: 355ms.    
The Memory difference is: Object: 6088, Map: 7272, List: 1896.

So based on the information; I still think using the List is the best choice.

NathanaelA:
@hdeshev the _entitySearchRegEx will match:
 &amp;#DD; / &amp;#xHHHH; / &amp;ENTITYNAME;     

However, the regex automatically passes in a variable of which one of those three regex strings it matched; so that we can run separate code paths for each one.  This keeps the code fairly optimized depending on what it detects.

atanasovg:
@NathanaelA I was just about to comment on your metrics but you were faster :) Yes, the Object lookup should be constant and the array one - linear. In fact I am a bit surprised by the memory footprint of the JS array since it is nothing more but an extended Object. It seems the VM manages its memory more efficiently. Still, let me put it this way - we "sacrifice" 3KB of memory for the sake of a constant lookup - I would say it is worthy.

@hdeshev For this particular case it really doesn't significantly matter, still, it is important we fight for every possible millisecond, especially with code, which executes upon application start-up.

hshristov:
:+1:  on what @atanasovg said. We need every millisecond which we can get. So my choice goes for Map.

In fact I was considering different approach (not sure if it would be possible) - to expose different control which will have single property - the Unicode character. Something like ```<FontIcon Glyph="&#x03A3" />```.
Still I have to check if the parser will support this.

NathanaelA:
@hshristov - NOOOOOO, that would be a significant downgrade in my humble opinion!    Letting me put the character in any of my controls is WAY better.  For example I did this <Button text="&amp;#xE163" cssClass="icon"> and now my button has the nifty send icon that is standard across multiple communication apps.     Allowing me to embed the font characters anywhere I want is WAY better than a specifc <FontIcon> control.

NathanaelA:
@atanasovg -- Well when you put it that way; 4k doesn't sound like much.  :-)   If you would rather have it all in a object; I can do so.   Also, how many of the http://dev.w3.org/html5/html-author/charref do we want to support.   

As for Map, I don't believe iOS JavaCore on 7 supports Map; mdn says it is a Safari 8 Mobile feature.

NathanaelA:
@atanasovg & @ligaz --- I've updated the patch to make the changes you guys requested.     This includes the patches and tests from #729 

The only thing I am unsure of is line 68 of the patch, I had to use **any** as the return type.  When I attempted to use Map<string, number> I couldn't figure out how to get it to compile, sorry not a pro at typescript yet...  :grinning:

atanasovg:
@NathanaelA Thanks, the pull seems good to me. I personally would have it done without the temporary object `var objCodes =...` but rather call directly `ampCodes.set(key, value);` for each entry on a new line - for example `ampCodes.set("Tab", 9);`. This will save:

1. The temporary object allocation
2. The `for-each` cycle
3. The `hasOwnProperty` check

Additionally, it would make the code more readable (for me). It will however make the file larger (due to the new lines).

So, it is up to you :) I am :+1: for the pull.

@hshristov what do you think?

hshristov:
@NathanaelA I've checked the code and saw ```s = String(s)``` in the ```_dereferenceEntities``` method. Why do you need to create new String object?

NathanaelA:
@hshristov - that code is/was already present.  I figured you guys had a good reason for it, because I didn't see any obvious reason why it was their.   I decided to leave it anyways as I didn't want removing it to break something because I removed something that was actually needed.

hshristov:
@NathanaelA Well my bad. I was reviewing the whole file (instead of changes only). I will check if this is needed and remove it later.
:+1:

PR #726, opened 2015-09-13
This fixes the issue with the android Dialogs 

NathanaelA:
On Android either the dialogs can be dismissed by clicking anywhere elsewhere on the screen on Android.  (Also a couple other things can cause the OS to auto-dismiss the dialog) In all these cases the dialog will now call the callback with a null value as the first parameter (and finish calling the unloaded and other code that is supposed to be ran when the dialog is closed).   

This patch also adds the convenience function; closeDialog, that you can call from the dialog screen page component which will also do all the same cleanup and call the close callback for you.  (i.e. var page = object.args;  page.closeDialog("Value1", "Value2"...);

### DIFF
--- a/ui/page/page-common.ts
+++ b/ui/page/page-common.ts
@@ -33,9 +33,10 @@ export class Page extends contentView.ContentView implements dts.Page {
     public static navigatedFromEvent = "navigatedFrom";
     public static shownModallyEvent = "shownModally";
 
+    protected _closeModalCallback: Function;
+
     private _navigationContext: any;
 
-    private _closeDialogCallback: Function;
     private _cssApplied: boolean;
     private _styleScope: styleScope.StyleScope = new styleScope.StyleScope();
     private _actionBar: actionBar.ActionBar;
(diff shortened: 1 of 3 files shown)


enchev:
run ci

dtopuzov:
run ci

dtopuzov:
:+1:

NathanaelA:
I've updated the pull request to add some internal consistency.    We use "showModal" so I changed my helper function to "closeModal" rather than "closeDialog" as it was in the first version.
I also add a fix for the modal screen being full screen on Android.  It will detect that it is a dialog and automatically be sized properly.

hshristov:
Hi @NathanaelA,

Sorry for the late reply. @dtopuzov confirmed that your patch fixes the described issue but the root issue about dialogs taking full screen in Android when fullScreen parameter is false is that GridLayout works incorrectly. At least I think so but I haven't got time to dig deeper. Once I check it I will write back.

enchev:
run ci

hamorphis:
I will checkout your branch and test it locally. Meanwhile, since you have added a new public member called, you will need to add its definition to the page.d.ts file and document it. Also, you need to update the Modal Pages section of the Navigation help article located [`here`](https://github.com/NativeScript/docs/blob/production/navigation.md#modal-pages)

hamorphis:
Nevermind, I will get your branch locally, add the documentation and create a new pull request with your changes.

hamorphis:
Here is the new pull request: https://github.com/NativeScript/NativeScript/pull/864

hamorphis:
I merged pull request #864

Issue #720, opened 2015-09-12
Alert doesn't show up in the first run of the App


atanasovg:
_From @AntonioCuevaUrraco on September 8, 2015 6:44_

I found a weird behaviour in our app related to the Alert.
If we try to do and alert for Android the very first time that the App is running this don't show up it get rejected with this trace.

```

V/JS      ( 3658): the problem is Error: java.lang.reflect.InvocationTargetException
V/JS      ( 3658):     java.lang.reflect.Constructor.newInstance(Native Method)
V/JS      ( 3658):     java.lang.reflect.Constructor.newInstance(Constructor.java:288)
V/JS      ( 3658):     com.tns.Platform.createInstance(Platform.java:241)
V/JS      ( 3658):     com.tns.Platform.callJSMethodNative(Native Method)
V/JS      ( 3658):     com.tns.Platform.dispatchCallJSMethodNative(Platform.java:727)
V/JS      ( 3658):     com.tns.Platform.callJSMethod(Platform.java:606)
V/JS      ( 3658):     com.tns.Platform.callJSMethod(Platform.java:590)
V/JS      ( 3658):     com.tns.Platform.callJSMethod(Platform.java:580)
V/JS      ( 3658):     com.tns.gen.android.app.Fragment_frnal_prepareExtend_l60_c37__PageFragmentBody.onOptionsItemSelected(android.app.Fragment.java)
V/JS      ( 3658):     android.app.Fragment.performOptionsItemSelected(Fragment.java:2159)
V/JS      ( 3658):     android.app.FragmentManagerImpl.dispatchOptionsItemSelected(FragmentManager.java:1990)
V/JS      ( 3658):     android.app.Activity.onMenuItemSelected(Activity.java:2895)
V/JS      ( 3658):     com.tns.NativeScriptActivity.onMenuItemSelected(NativeScriptActivity.java:1887)
V/JS      ( 3658):     com.android.internal.policy.impl.PhoneWindow.onMenuItemSelected(PhoneWindow.java:1127)
V/JS      ( 3658):     com.android.internal.view.menu.MenuBuilder.dispatchMenuItemSelected(MenuBuilder.java:761)
V/JS      ( 3658):     com.android.internal.view.menu.MenuItemImpl.invoke(MenuItemImpl.java:152)
V/JS      ( 3658):     com.android.internal.view.menu.MenuBuilder.performItemAction(MenuBuilder.java:904)
[...]

PR #728, opened 2015-09-14
startActivity used if no foregroundActivity

enchev:
Fix for https://github.com/NativeScript/NativeScript/issues/720


atanasovg:
:+1:  great

PR #736, opened 2015-09-14
use isNullOrUndefined instead of isUndefined

slavchev:
Due a bug in the Android runtime when a Java method returns `null` value it was propagated as `undefined` in JavaScript. This is no longer the case. As a result, the current loop check never holds `true`. The proposed workaround is to use `isNullOrUndefined` instead of `isUndefined`. Another, yet simpler solution is to use `line === null` check.


enchev:
:+1:

PR #752, opened 2015-09-16
Extracted view-related utility methods from "utils/utils" to "ui/util…

hamorphis:
…s" so that the "utils/utils" module does not depend on the view module.


enchev:
:+1:

hdeshev:
:+1:

hshristov:
If this is going to be iOS specific module then you don't need ```.d.ts.``` file.
And you can call it iosUtils to make it clear that it contains iOS specific code only.

hamorphis:
run ci

Issue #774, opened 2015-09-18
If an Animation instance is played more than once, the same promise is resolved each time leading to unexpected results.


hamorphis:
Since the finished promise of an Animation is one and the same during the life of the animation, the Animation cannot be properly played multiple times, i.e. the following code fails:

    animation1.play().finished
        .then(function () { return animation1.play().finished; })
        .then(function () { return animation1.play().finished; })

We should make the **play()** method return a new Promise each time it is played. 

In other words, there should be a new promise returned for each play invocation -- not one promise per Animation instance.

hamorphis:
### Breaking changes
-  [(#774)](https://github.com/NativeScript/NativeScript/issues/774) Animation class no longer has a **finished** property because an animation can be played multiple times. The **play** method now returns a new Promise each time it is invoked. Use this to listen for the animation finishing or being cancelled. When upgrading to version 1.4.0 or above simply remove **.finished** from your code.
**Old Code (JavaScript)**:
```JavaScript
animation1.play().finished.then(function () { console.log("Finished"); });
```
**New Code (JavaScript)**:
```JavaScript
animation1.play().then(function () { console.log("Finished"); });
```
**Old Code (TypeScript)**:
```JavaScript
animation1.play().finished.then(()=>console.log("Finished"));
```
**New Code (JavaScript)**:
```JavaScript
animation1.play().then(()=>console.log("Finished"));
```

PR #791, opened 2015-09-21
Fixed: If an Animation instance is played more than once, the same pr…

hamorphis:
…omise is resolved each time leading to unexpected results. #774


ErjanGavalji:
:+1:

PR #821, opened 2015-09-24
Extracts common logic into web-view-common. The check for http/https …

### LINKED ISSUE #819: WebView check for http/https should be case insensitive
The [check here](https://github.com/NativeScript/NativeScript/blob/master/ui/web-view/web-view.android.ts#L97) should be case insensitive, otherwise urls like "HTTP://xxxx" are treated as data.

atanasovg:
…is now case insensitive. Fixes #819

### DIFF
--- a/ui/web-view/web-view-common.ts
+++ b/ui/web-view/web-view-common.ts
@@ -2,6 +2,9 @@
 import view = require("ui/core/view");
 import dependencyObservable = require("ui/core/dependency-observable");
 import proxy = require("ui/core/proxy");
+import utils = require("utils/utils");
+import fs = require("file-system");
+import trace = require("trace");
 
 var urlProperty = new dependencyObservable.Property(
     "url",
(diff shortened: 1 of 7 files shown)


enchev:
:+1:

ui/web-view/web-view-common.ts
```diff
@@ -101,7 +123,19 @@ export class WebView extends view.View implements definition.WebView {
         throw new Error("This member is abstract.");
     }
 
-    public _loadSrc(src: string) {
+    public _loadFileOrResource(path: string, content: string) {
+        throw new Error("This member is abstract.");
```
ligaz: There is an `abstract` class support in TypeScript 1.6. It might be better to use it here to better donate the semantics.

enchev:
run ci

ui/web-view/web-view-common.ts
atanasovg: @ligaz That's a good idea. @hdeshev Do we support 1.6 already?

ligaz: It looks like this: https://github.com/NativeScript/NativeScript/blob/master/package.json#L28

enchev: Yep, we support 1.6!

enchev:
run ci

vchimev:
run ci

dtopuzov:
run ci

enchev:
I've just noticed that there is no test. Can we add one?

enchev:
run ci

dtopuzov:
CONSOLE ERROR Test: --- [WEB-VIEW.testLoadUpperCaseSrc]  FAILED: args.url should equal 'HTTP://nsbuild01.telerik.com/docs/'

PR #832, opened 2015-09-28
Transformations such as scale, translate, rotate won't be incorrectly affected by the layout

PanayotCankov:
In iOS we now set `bound` along the `frame` property on the native views.

The easiest way to reproduce this is to display an UI similar to:
```
<GridLayout width="50" height="50">
    <Image src="res://img" rotate="25" backgroundColor="red" />
</GridLayout>
```
The image will be rotated but will also somewhat scale since the transform was not reapplied after the frame was set. After applying this fix, the image will appear correctly rotated.


hshristov:
run ci

hshristov:
:+1:

PR #842, opened 2015-09-29
Change from "classic" TS 1.6 imports to the default resolution scheme.

hdeshev:
- Use relative imports in place of most of our absolute ones.
- Add "private" ambient modules for modules that we need to import using
an absolute path (e.g. when app/.../test-something.ts needs to import
ui/styling/style-scope)


ErjanGavalji:
:+1:

hshristov:
:+1:

hdeshev:
run ci

hdeshev:
run ci

dtopuzov:
run ci

PR #845, opened 2015-09-29
web-view loads local data with UTF-8

emiloberg:
This PR fixes that UTF-8 characters (such as the Nordic `åäöæø`) does not show correctly in a Web-view, if content is loaded from local file or local html string.

content-type is changed from `text/html` to `text/html; charset=utf-8` as this is needed for newer Android (where older just rely on the mime-type `utf-8`)


emiloberg:
CLA signed.

enchev:
run ci

enchev:
:+1:

Issue #857, opened 2015-09-30
Android action dialog actions are not shown if message is provided


enchev:
Example:
```JavaScript
dialogs.action({
            title: "Title",
            message: "Action",
            cancelButtonText: "Close",
            actions: ["One", "Two", "Three"]
}).then(function (r) { return console.log("Action\u00A0result:\u00A0" + r); });
```

PR #858, opened 2015-09-30
Message will be set only if you do not have actions since message and actions are mutually exclusive.

enchev:
Fix for #857


vchimev:
:+1:

PR #865, opened 2015-10-01
Fixed:  iOS animations combining several affine transform properties …

hamorphis:
…set only the first property on our view after they finish. #835


enchev:
:+1:

dtopuzov:
:+1:

PR #892, opened 2015-10-07
Add itemLongPress event to ListView

nedium:
In my application i needed to show an array of items in a listview and when user longpresses listview item a popup menu would show. 
I have searched this feature, but couldn't locate it, so i changed the source code as simple as possible.

After this change user can reference itemLongPress event in xml like this:
```xml
<ListView items="{{ TaskList }}" itemLongPress="TaskListAction">
```

Codebehind would be:
```javascript
exports.TaskListAction = function(args) {
    var Task = viewModel.TaskList.getItem(args.index);
}
```

This change is android only, i really dont know the ios equivalent so i was unable to type it.

Note this is my first time using github (yes i was living under a rock last 5 years) so if i am doing something wrong please imform me.

### DIFF
--- a/ui/list-view/list-view.android.ts
+++ b/ui/list-view/list-view.android.ts
@@ -12,6 +12,7 @@ import utils = require("utils/utils")
 var ITEMLOADING = common.ListView.itemLoadingEvent;
 var LOADMOREITEMS = common.ListView.loadMoreItemsEvent;
 var ITEMTAP = common.ListView.itemTapEvent;
+var ITEMLONGPRESS = common.ListView.itemLongPressEvent;
 var REALIZED_INDEX = "realizedIndex";
 
 global.moduleMerge(common, exports);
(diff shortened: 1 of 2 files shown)


ui/list-view/list-view.android.ts
```diff
@@ -89,6 +90,16 @@ export class ListView extends common.ListView {
...
             }
         }));
+        
+        this.android.setOnItemLongClickListener(new android.widget.AdapterView.OnItemLongClickListener({
+            onItemLongClick: function (parent, convertView, index, id) {
+                var owner = that.get();
+                if (owner) {
+                    owner.notify({ eventName: ITEMLONGPRESS, object: owner, index: index, view: owner._getRealizedView(convertView, index) });
+                }
+                return true;
```
nedium: This was a shortcut to make it work, return value can be read from final event handler, but i dont know how achive this in this context...

enchev:
Hi @nedium,

Thank you very much for your contribution however we want to keep our modules cross-platform as possible and that is why every feature should be present in both Android and iOS versions.

In your case this can be done very easily without modifying our core code by accessing native Android ListView in loaded event. Here is an example:

```XML
<ListView loaded="listViewLoaded">
   ...
</ListView>
```

```JavaScript
exports.listViewLoaded = function(args){
   var listView = args.object;
   var that = new WeakRef(listView);
   listView.android.setOnItemLongClickListener(new android.widget.AdapterView.OnItemLongClickListener({
            onItemLongClick: function (parent, convertView, index, id) {
                var owner = that.get();
                if (owner) {
                    owner.notify({ eventName: ITEMLONGPRESS, object: owner, index: index, view: owner._getRealizedView(convertView, index) });
                }
                return true;
            }
        }));
}
```

PR #912, opened 2015-10-09
Fixed #909: Android ListView will not update its UI after refresh is …

### LINKED ISSUE #909: Android ListView will not update its UI after refresh is called until it is scrolled up and down.
When bound to a simple array of simple items, the ListView does not update its UI after refresh is called.

hamorphis:
…called until it is scrolled up and down.


hshristov:
:+1:

hamorphis:
@arover  I am not missing else. If the data item is not observable I need to reset the binding context so it can be refreshed.

hamorphis:
run ci

dtopuzov:
run ci

PR #918, opened 2015-10-11
Refer to current _map, not most recent _map in Observable

jjd314:
The `Observable` module uses `defineProperty` every time a new `Observable` is made, and the get/set methods refer to the `that` variable in the closure.  The result is that If you call `new Observable()` multiple times on objects that have identical properties (like `Observables` in an `ObservableArray`, as in [this StackOverflow question](http://stackoverflow.com/questions/31898061/observable-in-listview-binds-elements-incorrectly)), all the `Observables` will access/update only the last object's properties.

I changed the get/set methods to use `this`, which solves the immediate problem.  It still seems inefficient to redefine the get/set methods every time an `Observable` is created.  Also, every `Observable` will have every observable property from any other `Observable` defined for it, which seems odd, but not the worst possible problem, I suppose.

### DIFF
--- a/apps/tests/observable-tests.ts
+++ b/apps/tests/observable-tests.ts
@@ -1,0 +1,26 @@
+import TKUnit = require("./TKUnit");
+require("globals");
+
+// <snippet module="data/observable" title="observable">
+// # Observable module
+// ``` JavaScript
+import observableModule = require("data/observable");
+// ```
+// </snippet>
+
+require("globals");
+
+export var test_Observable_shouldDistinguishSeparateObjects = function () {
+    // <snippet module="data/observable" title="observable">
+    // ### Create two Observables from different objects.
+    // ``` JavaScript
+    var obj1 = {val: 1};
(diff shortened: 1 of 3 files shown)


hamorphis:
Thank you for your work. We will need some kind of a unit test which we can add to our test suite. The test must fail before the suggested changes and it must pass after we make the suggested changes.

jjd314:
Thanks @hamorphis.  I took a stab at adding the test to the test suite.  I apologize in advance if I did not follow your conventions for it.  If it is not correct as is, the test is very simple and I hope it won't be too much trouble for you to move it to where it should be.

hamorphis:
@jjd314 I will take a look at it today. Thank you so much.

dtopuzov:
run ci

hamorphis:
I have extended the unit test a little bit, made some refactoring and created a new pull request: https://github.com/NativeScript/NativeScript/pull/927

PR #933, opened 2015-10-14
Css upgrade

hdeshev:
Upgraded the CSS parse libs and made them use the same file structure as upstream. That should fix a problem in the Angular codebase that requires a submodule of the css lib directly: `css/lib/parse/index`


hdeshev:
run ci

PR #935, opened 2015-10-14
Fixed #801: Chained animations lose state on iOS.

### LINKED ISSUE #801: Chained animations lose state on iOS
The code below chains two animations—one that rotates an image 90 degrees, and a second that scales it to twice its size.

``` xml
<Page loaded="loaded">
    <AbsoluteLayout>
        <Image src="~/logo.png" id="logo" />
    </AbsoluteLayout>
</Page>
```
```js
exports.loaded = function(args) {
    var page = args.object;
    var message = page.getViewById("logo");
    message.animate({
        rotate: 90,
        duration: 1000
    }).then(function() {
        return message.animate({
            scale: { x: 2, y: 2 },
            duration: 1000
        });
    });
};
```

This works as intended on Android, but on iOS the second animation incorrectly rotates the image during the scaling, and then reverts to the appropriate value when the animation completes. You can see this behavior in the gif below:

ios-bug

hamorphis:


hshristov:
:+1:

PR #936, opened 2015-10-14
Subsequent animation of transition and rotation or scale will appear jumpy in iOS

PanayotCankov:
Move from begin/commit animation blocks to keyframe animations, they handle continuation from the current state better. There is a new "Sequence" animation in the "animations" app that displays a sequence of animations that used to be buggy in iOS (the animated object was jumping around) but should now display a nice continuous animation. The problem that has been fixed appeared only when multiple properties that affect transform are animated: rotate + translate or scale + translate.


dtopuzov:
run ci

enchev:
:+1:

PR #947, opened 2015-10-15
TabView color and selectedColor

vakrilov:


ui/styling/stylers.android.ts
```diff
@@ -643,6 +644,51 @@ export class ActionBarStyler implements definition.stylers.Styler {
...
+        var tabLayout = tab._getAndroidTabView();
+
+        for (var i = 0; i < count; i++) {
+            tabLayout.getTextViewForItemAt(i).setTextColor(nativeValue);
+        }
+    }
+
+    private static getColorProperty(view: view.View): any {
+        var tab = <tabView.TabView>view;
+        var tv: android.widget.TextView = tab._getAndroidTabView().getTextViewForItemAt(0);
```
hshristov: Do we always have at least 1 item? I think a check for item count will prevent eventual exception here although I'm not sure we can extract the value if there is no TabItem.
Eventually new TextView() could work.

ui/tab-view/tab-view-common.ts
```diff
@@ -233,4 +251,13 @@ export class TabView extends view.View implements definition.TabView, view.AddAr
             }    
         }
     }
+    
+    public _getAndroidTabView(): org.nativescript.widgets.TabLayout {
+        // Android specific
+        return undefined;
+    }
+
+    public _updateIOSTabBarColors() {
```
hshristov: missing return type - : void

ui/tab-view/tab-view.ios.ts
```diff
@@ -295,4 +307,39 @@ export class TabView extends common.TabView {
...
+    }
+}
+
+function getTitleAttributesForStates(tabView: TabView): { normalState: any, selectedState: any } {
+    var normalState = <any>{};
+    if (tabView.color instanceof color.Color) {
+        normalState[UITextAttributeTextColor] = tabView.color.ios;
+    }
+
+    var selectedState = <any>{};
```
hshristov: do you need the cast to <any>? Do new Object() works?

hshristov:
:+1:

Issue #948, opened 2015-10-15
Can not 'import gestures from "ui/gestures"' in 1.4


emiloberg:
This is merely "FYI" thing as this involves ES6 transpiled code.

I have a ES6 based project which is run though babel. I just upgraded to 1.4 and can no longer write:
```
import gestures from 'ui/gestures'
```

doing that gives a runtime error:
```
undefined:12:24: JS ERROR TypeError: undefined is not an object (evaluating 'Base.prototype')
1   0x8f7f1 -[TNSRuntime executeModule:]
2   0x5019e main
3   0x3122ac9 start
4   0x1
Session ended without errors.
```

However, requiring it with the require syntax `const gestures = require('ui/gestures');` works as it should. All other modules (at least the ones used in my project which is quite some) can be `import`-ed

Sorry for not being able to dig deeper and figure out the real problem, However, it  could have something to do with [this change](https://github.com/NativeScript/NativeScript/commit/77838ae9c6e2120a02c6ba6a900f6a87828ea32f#diff-9abab60af7821740c613343766ce5b4dL1)

hdeshev:
Hi Emil,

I tried reproducing the crash using a [fork](https://github.com/hdeshev/nativescript-babel-es6import) of a Babel project I [found](https://github.com/emiloberg/nativescript-crashing-listview) on the internet :) and , but I couldn't reproduce the crash after upgrading to {N} 1.4.

I tried both Android and iOS, and this code:
```javascript
import gestures from 'ui/gestures';

//...

console.log('tap gesture: ' + gestures.GestureTypes.tap);
```

... runs fine when added to `app/src/views/main-page.js`. Am I missing something?

I then tried running your `oppna-program-reklistan-app` and got a different error. The `__extends` function throws an error in the content-view about trying to extend an `undefined` base class. This appears to be caused by a circular import in the `tns-core-modules` package that is triggered if something imports `ui/core/view` before importing `application`.

Luckily, you can work around the problem by forcing an `application` import early on. When I did that, the `import * as gestures from 'ui/gestures';` import started working. I wouldn't be surprised if the circular import breaks the ES6/Babel import as well.

I'm logging a separate issue for the circular import...

emiloberg:
Commenting on a closed issue here, but I'm pretty sure the `__extends` and the Runtime error is stem from the same common problem. And I think you're bang on the target about Babel not being able to figure out how to handle the circular imports.

[...]

PR #951, opened 2015-10-16
Break a circular import view-application-view.

hdeshev:
By late-importing `application` in `bindable` and `stylers-common`.

It all started in [this issue](https://github.com/NativeScript/NativeScript/issues/948).

Note that requiring `ui/core/view` early on may still break if "globals" hasn't been imported since it requires that the `moduleMerge` function be available.

### DIFF
--- a/ui/core/bindable.ts
+++ b/ui/core/bindable.ts
@@ -2,12 +2,21 @@
 import definition = require("ui/core/bindable");
 import dependencyObservable = require("ui/core/dependency-observable");
 import weakEvents = require("ui/core/weak-event-listener");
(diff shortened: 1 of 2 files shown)


hdeshev:
run ci

enchev:
:+1:

atanasovg:
This way we implicitly make the `appModule()` return type `any`, meaning no type checking. But if this is the only possible solution then it is obviously a necessary evil.

hdeshev:
I thought about the `any` issue as well, but couldn't think of a better solution. Let's leave it like that.

PR #952, opened 2015-10-16
Fixed #820: The x and y components of an Android translate or scale a…

### LINKED ISSUE #820: The x and y components of an Android translate or scale animation are not animated together when delay is specified.
Suppose I want to make an image huge with the following animation:

```js
imageReference.animate({
    scale: { x: 5, y: 5 },
    duration: 1000,
    delay: 1000
});
```

This works on iOS, but on Android the x and y scaling happen independently (see gif below). If I remove the `delay` the problem goes away.

android-bug

hamorphis:
…nimation are not animated together when delay is specified.


hamorphis:
run ci

dtopuzov:
run ci

PR #956, opened 2015-10-19
dismissSoftInput implemented

### LINKED ISSUE #849: Implement dismissSoftInput() for search-bar
Feature request:

The search-bar has no way of dismissing an open keyboard. This makes the search-bar a quite unusable as the normal user pattern is that the user search for something and then press the item and gets navigated there. On Android (at least on >= 5.x), the open keyboard will continue to stay open, even on the new page. On iOS it will close.

the `editable-text-base`, and therefore the `text-field`, has a `dismissSoftInput`. Can this be implemented on the search-bar as well?

Related, it should probably have the `focus()` method as well. 

Or am I missing a workaround?

enchev:
Fixed #849


ErjanGavalji:
:+1:

PR #961, opened 2015-10-20
Fixed #779: frame.topmost().currentPage is not the same instance as t…

### LINKED ISSUE #779: [iOS]: frame.topmost().currentPage is not the same instance as the page being loaded in page.loaded event handler.
During page.loaded event, frame.topmost() does not have the proper information for getViewById to work.  In fact currentPage = null.   However on Android this does work during the page.loaded event.  

This also causes a Page.showModal(x) to fail during the page.loaded event.

hamorphis:
…he page being loaded in page.loaded event handler.


hamorphis:
run ci

enchev:
:+1:

PR #970, opened 2015-10-21
Fixed two propertyChange event notifications.

nsndeck:


nsndeck:
run ci

nsndeck:
run ci

vakrilov:
:+1:

hshristov:
:+1:

rclai:
@nsndeck @vakrilov , this broke a functionality of mine where I have a setter that emits multiple `notifyPropertyChange` for different non-persistent fields:

```js
Object.defineProperty(deviceViewModel.prototype, "myfield", {
  get: function () {
    return this._myfield;
  },
  set: function (value) {
    if (this._myfield !== value) {
      this._myfield = value;
      this.notifyPropertyChange("field2", value);
      this.notifyPropertyChange("field3", value);
      this.notifyPropertyChange("field4", value);
    }
  },
  enumerable: true,
  configurable: true
});
```

Because of `disableNotifications`, those notifications that I'm doing inside the setter will not work.

rclai:
In addition, there is a lot of example code that you guys have that needs to be updated because notifications no longer work inside a setter.

rclai:
I see that based on the changes you guys have made that there's no longer a need to check for inequality and notify a change inside the setter since it is now done under the hood. This means that if a value changes and other values depend on it, they now need to be set separately outside the setter individually. This should've been mentioned as a breaking change.

PR #973, opened 2015-10-22
iOS owner pattern changed to use WeakRef in order to prevent memory l…

hshristov:
…eaks.

Optimized list-view-tests.
Added time per test.


enchev:
:+1:

nsndeck:
:+1:

PR #974, opened 2015-10-22
Switch color and background-color css support added

### LINKED ISSUE #642: Ability to style other native widgets
It would be great to be able to use CSS to style the ProgressBar and Switch.

The progress bar could have:
* backgroundColor
* color (the bar itself)
* width and height
* borderRadius (possibility to square it off)

The switch would be much the same, but not sure how possible it is.

enchev:
Fixed #642


dtopuzov:
:+1:

PR #976, opened 2015-10-22
WebView html with images fixed

### LINKED ISSUE #963: Always set base URL in WebView
Hi,

This issue is highly related to #241 (and fix #625 for that issue). However, as they're closed and delivered I'm opening a new one.

#### Use case
In my app I'm downloading JSON which is then turned into HTML with a little help from Handlebars - and then served in a WebView. I'm also downloading some images which is shown in that HTML file. 

So basically:
```
wv = new webViewModule.WebView();
wv.src = `
<html>
	<body>
		Hi there, here's a picture
		<img src="some/path/to/my/previously/downloaded/local/image.png">
	</body>
</html>
`
```

#### Problem
Setting the src of the WebView to a html string invokes `_loadData ` which invokes
```
this._android.loadData(src, "text/html; charset=utf-8", "utf-8");
```
Code here: https://github.com/NativeScript/NativeScript/blob/master/ui/web-view/web-view.android.ts#L97-L103

The problem with WebView's `loadData` method is that Android doesn't allow retrieving the image in the `` src with `documents()` (and possible subdirectories) if the files are stored there, or just prepend it with a fixed path if the files are included in the actual app.

enchev:
Fixed #963

### DIFF
--- a/apps/ui-tests-app/web-view/webviewhtmlwithimages.xml
+++ b/apps/ui-tests-app/web-view/webviewhtmlwithimages.xml
@@ -1,0 +1,6 @@
+﻿<Page>
+  <GridLayout>
+    <WebView src="&lt;html&gt;&lt;body&gt;&lt;img src='./tests/logo.png'&gt;&lt;/img&gt;&lt;/body&gt;&lt;/html&gt;" />
+  </GridLayout>
+</Page>
+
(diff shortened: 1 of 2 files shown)


dtopuzov:
:100:

PR #994, opened 2015-10-28
lower-case-dashed components css support fixed

### LINKED ISSUE #993: CSS does not support new lowercase element names
Since {N} 1.4 I can now use `<text-field>` instead of `TextField`, but if I try to style `text-field` the CSS rules are not applied; I have to use `TextField`.

Can the CSS implementation be updated to support the new lowercase, hyphen-separator element names?

Thanks.

enchev:
Fixed #993


vakrilov:
:+1:

PR #1001, opened 2015-10-30
Fonts fixes

enchev:


nsndeck:
:100:

manijak:
How does this work? It was added the the 1.5 release? I removed my own fontawesome.ttf file in the app/fonts/ folder, and after a build they all show as [?].

NathanaelA:
No this looks to be a bug fix, an new test and an example file.   They aren't including the fonts with NS, they are only using them to test them to make sure fonts don't break again.

If you need fonts in your app; you need to continue to include them in your fonts folder...

manijak:
Oh ok. I was confused by it since they added it as a "new" addition to the 1.5 release.
https://github.com/NativeScript/NativeScript/releases
The first one under ###New

NathanaelA:
Yeah, I saw that too.  The actual font support had several really GOOD fixes in v1.4.   It looks like this fix was more of a corner case.  As far as I know font support is really good now on both iOS and Android; so I'm not sure what more they can add...

I suspect that someone else went through and tagged stuff for the readme and didn't realize this was more a cleanup than anything new, Oh well.   Some of the other features in v1.5 are REALLY awesome, I've been playing with them for a couple weeks now (Using the nightly masters on http://nativescript.rocks), ...

PR #1006, opened 2015-10-30
Don't expose platform types in public d.ts files.

hdeshev:
Use `any`. Add a comment with the real type e.g. `/* UIColor */`

This is a performance optimization for faster client code compiles. I also added a `compile-dts` grunt target that runs a check.


enchev:
:100:

PR #1017, opened 2015-11-02
application should not crash when start() method called multiple times

### LINKED ISSUE #1007: When application.start() is called twice on iOS, a meaningful error should be thrown
iOS apps which have a double initialization of application crash with no error.

When you inspect the system log for the device (in my case the iOS Simulator), you see `Oct 30 17:27:00 KOEVSKAMAC.local jasmine[25320]: *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'There can only be one UIApplication instance.'`

However, in such cases, some kind of error should be shown in the console of the CLI or on the device, or both. 

In my case this occurred while I was playing with unit testing. I had a second require for app.js inside my tests and when I tried running the unit tests on app, the app kept crashing with no meaningful error.

enchev:
Fixed #1007

### DIFF
--- a/application/application.ios.ts
+++ b/application/application.ios.ts
@@ -217,7 +217,11 @@ global.__onUncaughtError = function (error: Error) {
     definition.notify({ eventName: definition.uncaughtErrorEvent, object: <any>definition.ios, ios: error });
 }
 
+var started: boolean = false;
 exports.start = function () {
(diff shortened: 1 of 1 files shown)


enchev:
run ci

hamorphis:
:+1:

PR #1030, opened 2015-11-03
Add the typings hint to nativescript-dev-typescript.

hdeshev:
I forgot it in my previous PR. `nativescript-dev-typescript` works without it, but it would be best if we keep it in our `package.json`.


tailsu:
:+1:

PR #1042, opened 2015-11-05
binding to special properties fixed

### LINKED ISSUE #1041: Support binding for `class` property of views.
Currently `class` property does not support binding. However `cssClass` does.

Layout properties like `row` and `col` also do not support binding - it would be great if they do.

enchev:
Fixed #1041


vakrilov:
:+1:

hdeshev:
:+1: 

machine

PR #1047, opened 2015-11-06
Web view utf fix

### LINKED ISSUE #1038: WebView with HTML string source does not render on old Android
It seems like older versions of Android doesn't render the WebView correctly when given a WebView with a `src` set to a HTML string.

Easiest describe with an example, given a simple page page like this:

```
<Page xmlns="http://www.nativescript.org/tns.xsd" loaded="pageLoaded">
	<GridLayout rows="*" columns="*" id="container" />
</Page>
```

And a Javascript file which just sets the `src` of the WebView to an HTML string:

```
var webViewModule = require('ui/web-view');
var gridLayout = require('ui/layouts/grid-layout');

function pageLoaded(args) {
	page = args.object;
	var wv = new webViewModule.WebView();
	var container = page.getViewById('container');
	gridLayout.GridLayout.setColumn(wv, 0);
	gridLayout.GridLayout.setRow(wv, 0);
	container.addChild(wv);

	wv.src = '<!DOCTYPE html>' +
		'<html lang="en">' +
		'<head>' +
			'<meta charset="utf-8">' +
			'<meta http-equiv="content-type" content="text/html; charset=UTF-8">' +
			'<meta name="viewport" content="width=device-width, height=device-height, initial-scale=1.0, user-scalable=no, user-scalable=0" />' +
			'<title>Test Page</title>' +
		'</head>' +
		'<body>' +
			'<h1>Hi there</h1>' +
			'<h2>Here are some Swedish characters: ÅÄÖ</h2>' +
		'</body>' +
		'</html>';
}

exports.pageLoaded = pageLoaded;
```

The above works perfectly on all my devices. However, I've got a bug report from a user running Android 4.3/API 18 (on a Samsung Galaxy S3). On her device, the WebView displays the HTML code:

4216470493659screenshot_2015-10-29-12-08-51

After a long time debugging, it looks like her 4.3 device doesn't understand the MIME type `text/html; charset=utf-8` set in [web-view.android.ts](https://github.com/NativeScript/NativeScript/blob/master/ui/web-view/web-view.android.ts#L101-L125), and renders it as `text/plain` instead.

When removing the `; charset=utf-8` part on both places in the file linked above, it worked and rendered as it should.

However. The `; charset=utf-8` part was added (by me in [this PR](https://github.com/NativeScript/NativeScript/pull/845)) because internet suggested it was needed for UTF-8 characters to work on some API versions. (E.g. [StackOverflow](http://stackoverflow.com/questions/14474223/utf-8-not-encoding-html-in-webview-android)).

As I don't have a battery of devices/tests I don't feel comfortable sending a PR removing this.

### Imho what needs to be done is:

1) Remove the `; charset=utf-8`
2) Run tests to see that UTF-8 characters (such as ÅÄÖ) are still displayed on all devices from API 17 up to todays number
  * when given a HTML **string**.
  * when given a the URI to a **local HTML file** inside the {N} app
  * when given a the URI to a **web page on the internet** (such as [https://en.wikipedia.org/wiki/Ö](https://en.wikipedia.org/wiki/%C3%96) )

enchev:
Attempt to fix #1038

### DIFF
--- a/apps/tests/ui/web-view/test.html
+++ b/apps/tests/ui/web-view/test.html
@@ -5,6 +5,6 @@
 	<meta charset="utf-8" />
 </head>
 <body>
-    <span style="color:red">Test</span>
+    <span style="color:red">TestÖ</span>
 </body>
 </html>
--- a/apps/tests/ui/web-view/web-view-tests.ts
+++ b/apps/tests/ui/web-view/web-view-tests.ts
@@ -75,7 +75,7 @@ export class WebViewTest extends testModule.UITest<webViewModule.WebView> {
             // <hide>
             let actual;
             let expectedTitle = 'MyTitle';
-            let expectedHtml = '<span style="color:red">Test</span>';
+            let expectedHtml = '<span style="color:red">TestÖ</span>';
 
             if (webView.ios) {
                 actual = webView.ios.stringByEvaluatingJavaScriptFromString("document.body.innerHTML").trim();
@@ -120,7 +120,7 @@ export class WebViewTest extends testModule.UITest<webViewModule.WebView> {
 
             if (webView.ios) {
                 actual = webView.ios.stringByEvaluatingJavaScriptFromString("document.body.innerHTML").trim();
-                expected = '<span style="color:red">Test</span>';
+                expected = '<span style="color:red">TestÖ</span>';
             } else if (webView.android) {
                 actual = webView.android.getTitle();
                 expected = 'MyTitle';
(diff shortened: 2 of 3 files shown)


hshristov:
:+1:

vchimev:
:+1:

emiloberg:
@enchev The MIME now reads `text/html;`, notice the semicolon, shouldn't it read `text/html` without it?

enchev:
@emiloberg  Will this be a problem? According to our tests everything worked as expected.

emiloberg:
My guess is as good as yours but I'd say **maybe**. I wouldn't at all be surprised if older Android versions didn't understand it. Just like Android  SDK 18 it didn't understand `text/html; charset=utf-8` as described in #1038.

The [default value is `text/html`](http://developer.android.com/reference/android/webkit/WebView.html#loadDataWithBaseURL(java.lang.String, java.lang.String, java.lang.String, java.lang.String, java.lang.String))

All in all: Leaving the semicolon there **could** render a problem, and I can't see no reason to keep it.

vchimev:
@enchev @emiloberg - my fault here guys,
I unconsciously removed the semicolons while giving a try to the PR, so yes - they cause the problem on Api18 and I just made a PR in addition to this one in order to remove them - https://github.com/NativeScript/NativeScript/pull/1053.

PR #1051, opened 2015-11-06
Grunt: add an update-external-app task

hdeshev:
Builds the modules in bin/dist/modules and updates an app's
node_modules/tns-core-modules dir.

grunt update-external-app --path=/some/path/myapp1

/cc @nsndeck


hshristov:
:+1:

PR #1063, opened 2015-11-09
Fix dts file exclusion problems when packing tns-core-modules package.

hdeshev:
Exclude unneeded stuff like apps, node_modules, .baseDir.ts


hdeshev:
run ci

PR #1074, opened 2015-11-10
Reload app.css and clear file resolver cache on livesync.

vakrilov:


hdeshev:
:+1: Can you squash the commits, please?

Fatme:
:+1:

hshristov:
:+1:

NathanaelA:
@vakrilov -- 
LOL, I ran into the exact same issue with my LiveSync plugin (Crash while Minimize).   However, one thing you are not taking in account here is that when the app is restored; don't you want to clearCache, loadCSS, and reloadPage?   Your patch will allow the new files to get put up to the device; but they won't ever be actived if the app is minimized and then restored.

Out of curiosity; is this the new code for the XML/CSS based livesync code you guys are building; the resolver/clearCache has no effect on JavaScript files correct?

PR #1083, opened 2015-11-11
Fixed #1082: Null reference exception in view.android.setOnTouchListe…

### LINKED ISSUE #1082: Null reference exception in view.android.setOnTouchListener method
`return owner._nativeView.onTouchEvent(motionEvent);` can throw if _nativeView is undefined.

hamorphis:
…ner method


hshristov:
Can you add test?

PR #1085, opened 2015-11-11
Frame reload fix

NathanaelA:
I'm not sure why the existing code is the way it is; but if I call "reload" to reload the current page, I really don't want my history nuked, and I really don't need two copies of the current page in the page history...

This patch eliminates the history nuking, and sets it so that the reload'd page is not included in the page history a second time...


NathanaelA:
Ergh, got to commits merged into one.  Will re-do this one...

PR #1086, opened 2015-11-11
This will now allow the dialog to call the resolve function when cancelled by clicking outside.

NathanaelA:
All the dialogs also suffer on Android from the click outside and then not having the promise resolved or rejected.   
This code does two different things:
1. The promises to be resolved if clicked outside of dialog.
2. The ability to disable the click outside to cancel.  (This might be worth setting as default on the login dialog)

### DIFF
--- a/ui/dialogs/dialogs.android.ts
+++ b/ui/dialogs/dialogs.android.ts
@@ -13,7 +13,7 @@ function createAlertDialog(options?: dialogs.DialogOptions): android.app.AlertDi
     var alert = new android.app.AlertDialog.Builder(appmodule.android.foregroundActivity);
     alert.setTitle(options && types.isString(options.title) ? options.title : "");
     alert.setMessage(options && types.isString(options.message) ? options.message : "");
-    if (options && (options.cancelable === false || options.cancelable === 0)) {
+    if (options && options.cancelable === false) {
         alert.setCancelable(false);
     }
     return alert;
--- a/ui/dialogs/dialogs.android.ts
+++ b/ui/dialogs/dialogs.android.ts
@@ -275,7 +275,7 @@ export function action(arg: any): Promise<string> {
             var alert = new android.app.AlertDialog.Builder(activity);
             var message = options && types.isString(options.message) ? options.message : "";
             var title = options && types.isString(options.title) ? options.title : "";
-            if (options && (options.cancelable === false || options.cancelable === 0)) {
+            if (options && options.cancelable === false) {
                 alert.setCancelable(false);
             }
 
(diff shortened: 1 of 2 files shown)


enchev:
Hey @NathanaelA,

According to your code you should have `cancelable` property for [dialog options](https://github.com/NativeScript/NativeScript/blob/master/ui/dialogs/dialogs.d.ts#L113-L123). Will you add iOS implementation also?

Thanks

NathanaelA:
Erps, I always forget about those files...  

No cancelable is an Android only option, to my knowledge you can't cancel a dialog on ios by just clicking outside of the dialog.  On Android this is the default behavior, and should remain so as this is the normal behavior.  

However, sometimes you might not want that to be the default, and since this is a dialog that you have to have access to the dialog creation variable to change the cancelable before you show it.   This is really the only way to expose it during creation.    Trust me I'm not big on single platform options; but since this is a Android only default feature the cancel ends up being a Android only fix.

enchev:
Hey @NathanaelA,

Maybe it will be useful if we can have `ActionOptions` configurable as well? Something like this:
```TypeScript
    /**
     * Provides options for the dialog.
     */
    export interface CancelableOptions {
        /**
         * Gets or sets if the dialog can be canceled by taping outside of the dialog.
         */
        cancelable?: boolean;
    }

    /**
     * Provides options for the dialog.
     */
    export interface ActionOptions extends CancelableOptions {
        /**
         * Gets or sets the dialog title.
         */
        title?: string;

        /**
         * Gets or sets the dialog message.
         */
        message?: string;

        /**
         * Gets or sets the Cancel button text.
         */
        cancelButtonText?: string;

        /**
         * Gets or sets the list of available actions.
         */
        actions?: Array<string>;
    }

    /**
     * Provides options for the dialog.
     */
    export interface DialogOptions extends CancelableOptions {
        /**
         * Gets or sets the dialog title.
         */
        title?: string;

        /**
         * Gets or sets the dialog message.
         */
        message?: string;
    }
```

NathanaelA:
@enchev, LOL, your right.  Sorry about that oversite.  I already added the code to it so it works; but I rarely use TypeScript so I always forget to modify the .d.ts files...  I'll get that updated...

NathanaelA:
@enchev, LOL, your right.  Sorry about that oversite.  I already added the code to it so it works; but I rarely use TypeScript so I always forget to modify the .d.ts files...  I'll got it updated...

ui/dialogs/dialogs.android.ts
```diff
@@ -262,6 +275,9 @@ export function action(arg: any): Promise<string> {
             var alert = new android.app.AlertDialog.Builder(activity);
             var message = options && types.isString(options.message) ? options.message : "";
             var title = options && types.isString(options.title) ? options.title : "";
+            if (options && (options.cancelable === false || options.cancelable === 0)) {
```
enchev: Hey @NathanaelA,

Please remove `options.cancelable === 0`  - this cannot be compiled since the property is boolean.

Thanks

ErjanGavalji:
@NathanaelA: well, there is a problem with the grunt-ts task, where it just does not stop the entire build when broken (I'm sorry I did not spare the time to either fix that or find a workaround for it)

I believe this is what you hit. Try running `grunt ts:build` and check its result.

enchev:
run ci

enchev:
:+1:

PR #1087, opened 2015-11-11
The Frame.reloadPage function is wrong.  ;-)

NathanaelA:
The way the Frame.reloadPage works is wrong.  If I am trying to reload the current page; I don't want the entire application history nuked.  And since I don't want the history nuked; I don't need an extra copy of this page in the history.   This fixes both those bad defaults.

### DIFF
--- a/ui/frame/frame-common.ts
+++ b/ui/frame/frame-common.ts
@@ -34,11 +34,11 @@ export function reloadPage(): void {
         let currentEntry = frame._currentEntry.entry;
         let newEntry: definition.NavigationEntry = {
             animated: false,
-            clearHistory: true,
+            clearHistory: false,
             context: currentEntry.context,
             create: currentEntry.create,
             moduleName: currentEntry.moduleName,
-            backstackVisible: currentEntry.backstackVisible
+            backstackVisible: true
         }
 
         frame.navigate(newEntry);


hshristov:
Hi @NathanaelA,

when you call navigate({  backstackVisible: true }) this makes the new page not going into the backstack. So what will happen is that once your mainPage is shown and you LiveSync-ed you will navigate to the modified mainPage but will have the original in the backstack. Subsequent livesync will keep only the inital mainPage (due to the new one not going into backstack - your modification).

For next version we will improve this behavior but for now it seems safer to clear the whole history so that you cannot see the previous page.

NathanaelA:
I'm looking at this from a framework perspective not a livesync/liveedit perspective.   If I programmatically make changes to my page like delete some components; then by calling frame.topmost().reloadPage() I want to Reload my page and reset the page back to the original defaults.    In this case I do NOT expect my entire history to be deleted and cleared.   All I want is my page reloaded.   

If this is supposed to be a LiveSync/LiveEdit function, then it probably should be marked as _ReloadPage or LiveSyncReloadPage; because as it stands now; this function is very misleading in its effects, and in the case you are mentioning (and the way it is developed) it is not a general purpose function...

hshristov:
No, it is definitely not a general purpose function at the moment. This idea was exactly this - to reload current page without affecting history/backstack but we are in the process of code freeze for 1.5 so I've decided that for the moment this is good enough. Once we branch 1.5 I will research how to best implement it.

enchev:
Hey @NathanaelA,

Unfortunately we've not managed to improve this for the upcoming release however we will do our best to do it right after. 

Thanks

NathanaelA:
@enchev - Yeah, I understand why; I had to revert my LiveEdit to my old method because the new history option works opposite to what I had "assumed" and now I understand why this "reloadPage" does it the way they do it.  (which is still broken) 

  For this to work, what needs to happen is the ability to "replace" the last item in the back buffer with the current loaded.   At this moment, not putting the new item in the back buffer, doesn't stop going back to the prior one.  So "the" solution in this pull wouldn't fix the issue anyways.

enchev:
In this case maybe we should close this pull request?

NathanaelA:
Yes, we probably do -- however then we still need an "issue" to fix this as the issue is still wrong; but this is not the solution.   Do you want me to open a new issue; or leave this to document that their is an issue that needs to be fixed.

enchev:
Maybe we can use [this one](https://github.com/NativeScript/NativeScript/issues/1417)?

NathanaelA:
Ah yes, that one will work.  Closing as this PR doesn't actually fix the issue properly.

Issue #1048, opened 2015-11-06
Event for start/end of an event


EnguGH:
Hi,

is it on the roadmap to implement the broadcast of events at the beginning / end of any gesture ?
Let's say I want to catch a pan : I'd like to know where exactly the initial tap happened. And I'd even more like to know when it ends so I can put all my elements back in place.

Thanks !

N3ll:
If there is enough demand we will consider implementing a cross-platform abstraction of low level events like touchDown, touchUp, etc.
Until then you can use the native APIs and those low level events.

EnguGH:
OK thanks !
I hadn't seen the native APIs things in the docs.

x4080:
I think its important for the touch gesture detection in multi platform mode

PR #1088, opened 2015-11-12
states for various gestures added

enchev:
Implemented #1048

### DIFF
--- a/ui/gestures/gestures.d.ts
+++ b/ui/gestures/gestures.d.ts
@@ -40,6 +40,28 @@ declare module "ui/gestures" {
     }
 
     /**
+     * Defines an enum with supported gesture types.
+     */
+    export enum GestureStateTypes {
+        /**
+         * Gesture cancelled.
+         */
+        cancelled,
+        /**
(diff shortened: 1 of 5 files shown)


enchev:
run ci

ui/gestures/gestures.d.ts
```diff
@@ -40,6 +40,28 @@ declare module "ui/gestures" {
...
+     */
+    export enum GestureStateTypes {
+        /**
+         * Gesture cancelled.
+         */
+        cancelled,
+        /**
+         * Gesture began.
+         */
+        began,
```
hshristov: Why not using - start, end, change, cancel?

hshristov:
:+1:

rclai:
What is the API for tap down and up? Also, how does this look on XML?

enchev:
I've just updated our docs with examples how to use gestures with state (tap, double tap and long press gestures are discreet). Here is an example with pan gesture:
https://github.com/NativeScript/docs/blob/production/gestures.md#pan

Issue #1018, opened 2015-11-02
BackgroundPosition & negative values


manijak:
I am trying to use the background-image & background-position in my css. The Image part works, but I am having issues with the background-position. 

My bg-image is a sprite image, and I use the position to focus on a specific part of the sprite. It does not seem to work at all with negative values for x & y. 

```css
.my-sprite{
background-image:url('~/images/bg-sprite.png');
background-position: -36 -120;
}
```

Just to test, I also tried to use center/center but it still did not move
```css
.my-sprite{
background-image:url('~/images/bg-sprite.png');
background-position: center center;
}
```

My guess is it cannot move the bg-image outside the screen bounds?

PR #1107, opened 2015-11-16
Fix: background-position with negative values

vakrilov:
Related to #1018. In this PR:
* Support for negative values in `background-position`.
* Handle the case when only one value(ex. "left") is specified in `background-position`.
* UI - test cases for the fixes

### DIFF
--- a/ui/styling/background-common.ts
+++ b/ui/styling/background-common.ts
@@ -3,7 +3,14 @@ import colorModule = require("color");
 import types = require("utils/types");
 import enums = require("ui/enums");
 import dts = require("ui/styling/background");
-var cssValue = require("css-value");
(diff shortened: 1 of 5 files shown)


enchev:
:+1:

ui/styling/background-common.ts
```diff
@@ -164,6 +168,50 @@ export class Background implements dts.Background {
         return res;
     }
 
+    private static parcePosition(pos: string): { x: CSSValue, y: CSSValue } {
```
ErjanGavalji: Let's rename `parcePosition` to `parsePosition`...

ErjanGavalji:
:+1: after the typo fix

PR #1110, opened 2015-11-17
Added test to verify that page layout is correct when page properties…

hshristov:
… like spanUnderStatusBar, ActionBar are changed.


enchev:
:+1:

Issue #685, opened 2015-09-05
Support CSS text-transform


sitefinitysteve:
Would be great to support text-transform in the css

http://www.w3schools.com/cssref/pr_text_text-transform.asp
(Sorry for w3schools link, dont hate me :)

PR #1120, opened 2015-11-18
text-transform 

enchev:
Implemented #685


ui/styling/style.ts
```diff
@@ -566,6 +577,12 @@ export class Style extends DependencyObservable implements styling.Style {
         }
     }
 
+    public _updateTextTransform() {
+        if (this._getValue(textTransformProperty) !== enums.TextTransform.none) {
```
hshristov: Will it works if you set textTransform to 'uppercase' and then to 'none'?

enchev: This code is not related to changing textTransform property but to perform some other stuff if the  textTransform is set (not none).

hamorphis:
:+1:

nsndeck:
:+1:

PR #1134, opened 2015-11-23
Unify the Promise definition with the one in es6-shim.

hdeshev:
Making it possible to compile code referencing Angular Promise-based APIs.


dtopuzov:
:+1:

PR #1136, opened 2015-11-23
Add Template factory function and use it in the ui/builder, Repeater and ListView components

PanayotCankov:
This will allow more flexible `itemTemplate` to be used in the Repeater and ListView including factory functions written in pure JavaScript or originating from other framework such as Angular.

This may be perceived as a breaking change, since the template builder will now return a `function` instead of `string`. These strings however, usually round-trip back to "ui/builder".parse, and the parse now has overload to accept both `string` and `function` types.

It is recommended in future to use `"ui/core/view".Template` for `itemTemplate` properties in child controls, but ours will still accept strings so they can be as backward compatible as possible.

Some existing controls that expose `itemTemplate` string property, that are not updated after this change, may get `function` for value instead of `string` and trigger errors elsewhere should the template be used as a `string`.

The ListView and Repeater no longer pass themselves as context when the template is realized, instead the template captures the context from the page when the template is parsed.

### DIFF
--- a/apps/tests/ui/list-view/list-view-tests.ts
+++ b/apps/tests/ui/list-view/list-view-tests.ts
@@ -8,6 +8,7 @@ import observable = require("data/observable");
 import types = require("utils/types");
 import platform = require("platform");
 import utils = require("utils/utils");
+import { Label } from "ui/label";
 
 // <snippet module="ui/list-view" title="list-view">
 // # ListView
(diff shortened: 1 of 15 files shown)


enchev:
:100:

hshristov:
:+1:

hdeshev:
:100: 

immortal

PR #1141, opened 2015-11-24
API for setting origin (a.k.a. pivot or anchor) point for view transfroms

vakrilov:
Related to #1140
screen


vakrilov:
run ci

ligaz:
Isn't it better to name this `originX` and `originY`? The `origin` name looks like the most neutral name and is used in a couple of other frameworks including the web itself.

enchev:
:+1:

PR #1144, opened 2015-11-24
Kvelikov listview separator ios

kvelikov:


ui/list-view/list-view-common.ts
```diff
@@ -85,6 +85,10 @@ export class ListView extends view.View implements definition.ListView {
             )
     );
 
+    public get iosSettings(): definition.IOSListViewSettings {
```
hshristov: Why do we have iosSettings in the common file? My guess is for intellisense.
Wouldn't it be better if we have settings object where we expose .ios & .android?

ui/list-view/list-view.ios.ts
```diff
@@ -7,6 +7,8 @@ import proxy = require("ui/core/proxy");
 import dependencyObservable = require("ui/core/dependency-observable");
 import color = require("color");
 
+declare var UIEdgeInsetsMake: (top: number, left: number, bottom: number, right: number) => UIEdgeInsets;
```
hshristov: This should goes into ios.d.ts

ui/list-view/list-view.ios.ts
```diff
@@ -218,6 +240,10 @@ export class ListView extends common.ListView {
             this.refresh();
         }
         this._ios.delegate = this._delegate;
+
+        if (this._ios.separatorInset) {
```
hshristov: So we won't support change after ListView is shown?

ui/list-view/list-view.ios.ts
atanasovg: See this: https://github.com/NativeScript/ios-runtime/issues/405. Wouldn't it be better (and more JS-friendly) to use JSON literals to describe insets?

hshristov:
@atanasovg What do you think about current implementation - in order to avoid breaking changes we expose new property settings which holds .ios & .android platform dependent settings?

atanasovg:
My opinion is that we have to do the breaking change now in order to avoid API pollution in the future. What I imagine is the following convention for each such `ios` and `android` properties:

```
interface NativeObject {
    view: any; // might be typed - e.g. android.view.View
}

class iOSListViewNativeObject implements NativeObject {
    view: UITableView;
    public separatorInsetLeft: number;
    public separatorInsetRight: number;
}

public class ListView {
    public ios: iOSListViewNativeObject;
}
```

Let's discuss this further and see how to proceed. We may open a separate issue to get the users feedback before we decide.

ErjanGavalji:
:+1:

ligaz:
Hey @kvelikov,

We discussed this today and brainstormed a couple of solutions taking all aspects into account:
* Preserve backwards compatibility.
* Provide consistent API.
* Flexibility of the solution itself.

The solution that looks to fulfill all these is the following: We can extend the existing native objects returned from the `ios` and `android` properties with custom properties, because they are expandable JavaScript objects.

There are two tricky parts to this solution that comes first to mind: 
* The first one is that for Android we will have to make sure that those properties are preserved because Android widgets are initialized lazily when the Activity is initialized.
* The second one is how are we going to define the TypeScript type defined by this property. The idea is to use [intersection types](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#intersection-types) available from TypeScript 1.6 and above.

Do you think you can spike the implementation of this? Feel free to ping me, @atanasovg or @hdeshev for assistance.

kvelikov:
Hi guys, 
I'll give it a try.

kvelikov:
run ci

kvelikov:
run ci

kvelikov:
Hey @ligaz, @atanasovg,
Can you check the latest version?

atanasovg:
@kvelikov I think this is way more elegant :) I have only one question - why do we need the `settings` object? I mean, wouldn't it be better to have the settings properties attached directly to the `ios` proxy - for example `this.ios.separatorInsetLeft` instead of `this.ios.settings.separatorInsetLeft`

kvelikov:
I find grouping these settings into one object more convenient. This will be harder to mistaken the native properties from the settings. 

The second and more important reason is that this object (settings) extends Observable and refreshes the ListView when there is a change.

atanasovg:
I see. OK then, :+1: from me, ping @hshristov for his opinion as well

ui/list-view/list-view.ios.ts
```diff
@@ -95,6 +95,16 @@ class UITableViewDelegateImpl extends NSObject implements UITableViewDelegate {
         if (owner && (indexPath.row === owner.items.length - 1)) {
             owner.notify(<observable.EventData>{ eventName: LOADMOREITEMS, object: owner });
         }
+
+        if (cell.separatorInset) {
+            cell.separatorInset = UIEdgeInsetsMake(cell.separatorInset.top, owner.ios.settings.separatorInsetLeft, cell.separatorInset.bottom, owner.ios.settings.separatorInsetRight);
+        }
+
+        if (cell.preservesSuperviewLayoutMargins) {
```
ligaz: Do you need this check? You can always set it to `false` :smile:

ligaz:
I'm also not a fan of grouping those properties in a `settings` object. :smile: Imagine that we want to add a function to this object as well. It will be strange to place this function in `settings`, so we will either have to place it on the `ios` object or provide another property like `functions` which will be even stranger.

In order to achieve "observability" you should be able to use `Object.defineProperty()` and provide your own setter for the property that will refresh the `ListView`.

hshristov:
After internal discussion we agreed that this approach is not good enough because:
- we expand a native object but in some cases the native object could die and be recreated from the OS. In this case expanded properties won't be set
- it is hard to document single property which have union type (e.g. UITableView & iOSListViewSettings)
- there might be collision with property names if in later versions native views expose properties with same name as in our setting (e.g. ios) object. Which one should take precedense?
- we don't want the internal settings object

So in next major version - 2.0.0 we will change all ```.android``` and ```.ios``` properties to be of some different type.

atanasovg:
@hshristov :+1: 
I agree on all points. Plus, to me it seems just wrong to mess with the API of the native widgets. If we are not happy with the API design and see architectural inconsistencies then I'd prefer the breaking change vs. a compromise in the API that may lead to problems in future.

vchimev:
run ci

vchimev:
run ci

vchimev:
run ci

vchimev:
run ci

vchimev:
run ci

vchimev:
run ci

vchimev:
run ci

Issue #1166, opened 2015-11-27
Icon in Action Bar


amjd:
I have been trying to add icons in the Action Bar, such as menu, back etc., on the left of the title text. This is the code I'm using:

```xml
  <Page.actionBar>
      <ActionBar>
        <ActionBar.title>
          <Label text="&#xe200;" cssClass="icon" />
          <Label text="SupportIT" style="font-size: 20; font-weight: bold" />
        </ActionBar.title>
        <ActionBar.actionItems>
            <ActionItem text="Settings" android.position="popup" />
            <ActionItem text="Profile" android.position="popup" />
            <ActionItem text="Logout" android.position="popup" />
        </ActionBar.actionItems>
      </ActionBar>
  </Page.actionBar>
```

The `icon` class has just one CSS property, `font-family: icomoon`. I'm using a custom generated icon font.

The same label displays the icon when used outside the Action Bar.

bradmartin:
@amjd - just to be certain, do you have that font installed in your project under 'app/fonts'? If so, then it's possible that using the icon fonts (Ionicons, FontAwesome, etc.) with the escaped HTML strings is not supported. If that's the case, hopefully the team can extend the use of the icon fonts on any element via the text property.

amjd:
@bradmartin Yes, the font is exists in 'app/fonts'. The icons show up in the rest of the page both when used directly or as an escaped HTML string. The problem seems to be with the Action Bar.

bradmartin:
Okay well hopefully the team will extend the usage of the icon fonts inside any view with a text property.

enchev:
Hey folks,

While you can set font-family for any view in titleView property of the ActionBar, currently font-family is not supported for the ActionBar items. Here is an example for titleView:
```XML
    <ActionBar>
      <ActionBar.titleView>
          <Label text="&#xf17b;" style="font-family: FontAwesome;"/>
      </ActionBar.titleView>
    </ActionBar>
```

We will do our best to add support for ActionBar items soon.

Thanks

amjd:
@enchev That worked, but when I add another label with the title of the page only the latter displays, not the icon. I could combine them into one label but that would mean messing up the font of the title and I won't be able to use the icon alone to perform some action like back, menu, etc.

Also, the icons added using a label don't have a ripple effect on Android when selected, like they have in most apps. Case in point, the action items icon (three dots) has the effect. The ripple effect isn't really a priority, the icon should just act like its native counterpart.

[...]

PR #1190, opened 2015-12-01
Support for android.systemIcon in NavigationBar in Android

vakrilov:
Related to #1166


hshristov:
:+1: 
Just mark is as a breaking change.

PR #1191, opened 2015-12-01
Check for UTF-8 BOM's in package.json files

hdeshev:
package.json files must NOT contain the UTF-8 BOM headers as that breaks node.js imports and other tools (such as webpack).

I added a grunt task that fails the build if a package.json file contains a BOM, and set it to run on compile. I also removed the old BOM-removal task as  that ran on the wrong set of files, and didn't cover package.json's.

I also removed BOM's for all broken files.


hdeshev:
run ci

PR #1203, opened 2015-12-04
WebViewClientClass exported

### LINKED ISSUE #1196: onReceivedSslError is not exposed via the WebViewClientClass
On Android (not sure about iOS) You are missing the following class function.
```
WebViewClientClass.prototype.onReceivedSslError = function(webView, handler, error) {
       // Code should call a user callback obviously; and if that callback returns true you proceed
       // otherwise  you do cancel (the default)
      //  handler.cancel() or  handler.proceed();
};
```
And since WebViewClientClass is NOT exposed via an export; you can't monkey patch it in; leaving the worst possible choice of having to manually modify a runtime class.   :angry:

enchev:
Fixed #1196

### DIFF
--- a/ui/web-view/web-view.android.ts
+++ b/ui/web-view/web-view.android.ts
@@ -4,7 +4,7 @@ import fs = require("file-system");
 
 global.moduleMerge(common, exports);
 
-class WebViewClientClass extends android.webkit.WebViewClient {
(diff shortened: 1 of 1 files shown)


enchev:
run ci

dtopuzov:
run ci

enchev:
run ci

vakrilov:
:+1:

PR #1205, opened 2015-12-04
Upgrade tslint to latest version.

hdeshev:
Still crashing with stack overflow when enabling the unused variables check, but at least it's newer :disappointed_relieved:


enchev:
3m-lint-roller-cat
:100:

vakrilov:
:+1:

PR #1215, opened 2015-12-07
FIX: Platform specific _addViewCore, ignoring index parameter

vakrilov:
`_addViewCommon` accepts index param in [view-common.ts](https://github.com/NativeScript/NativeScript/blob/master/ui/core/view-common.ts#L985) but it is ignored in ios/android implementations.


enchev:
:+1:

dtopuzov:
:+1:

PR #1219, opened 2015-12-08
Made two memory leak tests async so that GC could be triggered.

hshristov:


dtopuzov:
:100:

hshristov:
run ci

vchimev:
run ci

dtopuzov:
:+1:

PR #1220, opened 2015-12-08
timer and dialogs modules will be loaded lazy

enchev:


hshristov:
:+1:

globals/globals.ts
```diff
@@ -38,14 +16,51 @@ if (platform.device.os === platform.platformNames.android) {
     global.console.dump = function (args) { c.dump(args); };
 }
 
-global.XMLHttpRequest = xhr.XMLHttpRequest;
-global.FormData = xhr.FormData;
-global.alert = dialogs.alert;
+global.setTimeout = function (callback, milliseconds) {
+    let tm = require("timer");
```
PanayotCankov: You can `var tm;` outsied and `if (!tm) tm = require("timer");`. May be faster, depends on how the modules are cached in the runtimes. Otherwise :+1:

PanayotCankov:
:+1:

dtopuzov:
:+1:

globals/globals.ts
blagoev: +1 for this one. its better to skip the require except the first one on all the places here.

maybe do it this way

if (global.alert !== dm.alert) { global.alert = dm.alert; }

PR #1221, opened 2015-12-08
Fixed issue with two way binding update.

nsndeck:
Fixes issue with updating another property within a property setter.

set value1(value) {
    if (this._value1 !== value) {
        this._value1 = value;
        this.value2 = value + " something";
        this.notifyPropertyChange("value1", value);
    }
}

set value2 (value) {
...
}

The problem is that notification of `value2` is not raised.


enchev:
:+1:

PR #1228, opened 2015-12-10
url with % fixed

enchev:
It seems that this url is a special case since url encoding techniques
failed for both ios and android.
For ios:
stringByAddingPercentEscapesUsingEncoding
stringByAddingPercentEncodingWithAllowedCharacters
For android:
java.net.URLEncoder

### DIFF
--- a/apps/tests/http-tests.ts
+++ b/apps/tests/http-tests.ts
@@ -534,7 +534,8 @@ export var test_getString_FromVariousUrls_ShouldWorkProperly = function (done) {
     var urls = [
         "https://it-ebooks-api.info/v1/book/1615005640",
         "https://www.telerik.com",
-        "https://spreadsheets.google.com/tq?key=1tJ64Y8hje0ui4ap9U33h3KWwpxT_-JuVMSZzxD2Er8k"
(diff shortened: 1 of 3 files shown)


hshristov:
:+1:

AntonioCuevaUrraco:
I noticed that after updating to 1.5.1 my http requests were failing and I dig till here.
If the original url pass to options is already encoded this should not apply or is going to happen something like this.
Employee Advocay get encoded into Employee%20Advocacy
The replacement change it into Employee%2520Advocacy

I propose something like this
```ts
 var originalUrl= options.url;
        try{
            var decodedUrl = decodeURI(originalUrl);
            if (decodedUrl === originalUrl) {
                //the url is not encoded, fix the %
                options.url.replace("%", "%25");
            }
        }
        catch(e){
            //This means something crash with js decode encode (url malformed)
            // if we don't catch it will be and infinite loop, revert changes
            options.url=originalUrl;
        }
    javaOptions.url = options.url;

enchev:
Maybe it will be better to simply "unfix" this and and leave the developer to fix any failing url with stringByAddingPercentEscapesUsingEncoding and java.net.URLEncoder. I'm submitting new pull request: https://github.com/NativeScript/NativeScript/pull/1464

aylesm:
+1 on taking this fix out.  If you've already been a good citizen and encoded your URIs.  This will cause a double encode which you can't work around using this module.

PR #1231, opened 2015-12-10
Fix the test run on iOS

ErjanGavalji:
The test failed due to a huge output by the `tns platform add ios`
command. Change the grunt `exec` task to `shell` and specify max
length.


vchimev:
:+1:

PR #1236, opened 2015-12-11
Fixed Issue ##1235: ObservableArray does not raise its `change` event…

hamorphis:
… when its `length' property is set.


enchev:
:+1:

PR #1237, opened 2015-12-11
fetch and xhr lazy loaded into global context

enchev:


globals/globals.ts
```diff
@@ -62,12 +62,22 @@ global.prompt = function (args) {
     return getDialogs().prompt(args);
 }
 
-var xhr = require("../xhr/xhr");
-global.moduleMerge(xhr, global);
+class XMLHttpRequest {
+    constructor() {
+        var xhr = require("../xhr/xhr");
+        global.moduleMerge(xhr, global);
```
hshristov: Won't this `moduleMerge` on every `new XMLHttpRequest()`?
Better merge it once only when `xhr` is required initially.
```
var xhr;
class XMLHttpRequest {
    constructor() {
        if (!xhr) {
           xhr = require("../xhr/xhr");
           global.moduleMerge(xhr, global);
        }
        return new xhr.XMLHttpRequest();
   }
}
```

globals/globals.ts
```diff
@@ -62,12 +62,22 @@ global.prompt = function (args) {
...
+    }
+}
+
+global.XMLHttpRequest = XMLHttpRequest;
 
 // Fetch module should be after XMLHttpRequest/FormData! 
-var fetchModule = require("fetch");
-global.moduleMerge(fetchModule, global);
+global.fetch = function (args) {
+    var fetchModule = require("fetch");
```
hshristov: same caching of the require could be applied here (see the previous comment)

fealebenpae:
I think it'd be better to use `Object.defineProperty` on this one to define a getter on the global object:
```
Object.defineProperty(global, "XMLHttpRequest", {
    get: function() {
        var xhr = require("../xhr/xhr").XMLHttpRequest;
        Object.defineProperty(this, "XMLHttpRequest", { value: xhr });
        return xhr;
    }
});
```

globals/globals.ts
fealebenpae: `global.moduleMerge(xhr, global)` will clobber the value of `XMLHttpRequest` on the global object so this constructor will only run once, but it might break in case someone caches the value of `global.XMLHttpRequest` before `new`-ing it.

vakrilov:
:+1:

hdeshev:
:+1: 

wtf

globals/globals.ts
```diff
@@ -5,6 +5,34 @@ global.moduleMerge = function (sourceExports: any, destExports: any) {
...
+
+function registerOnGlobalContext(name, module) {
+    Object.defineProperty(global, name, {
+        get: function () {
+            // We do not need to cache require() call since it is already cached in the runtime.
+            let m = require(module);
+            global.moduleMerge(m, global);
+
+            // Redefine the property to make sure the above code is executed only once.
+            Object.defineProperty(this, name, { value: m[name], configurable: true, writable: true });
```
vakrilov: maybe set "writable: false" after the redefine

nsndeck:
:+1:

PR #1248, opened 2015-12-14
webpack  improvements

hdeshev:
Fix several issues I encountered while bundling modules code.

The biggest one being that webpack can't handle dynamic `require` calls at compile time and needs a hint.

The way to make this work is to expose an API that registers well-known modules such as:

```typescript
global.registerModule("ui/dialogs", () => require("ui/dialogs"));
```

The latter `require` gets changed by webpack to its own require function, and we can easily resolve it at runtime.

### DIFF
--- a/file-system/file-system.ts
+++ b/file-system/file-system.ts
@@ -442,10 +442,17 @@ export module knownFolders {
 
     export var currentApp = function (): Folder {
         if (!_app) {
-            var currentDir = __dirname;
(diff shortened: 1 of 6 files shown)


hdeshev:
Ping @enchev -- I changed the globals lazy loading code to use `utils/module-loader` too.

enchev:
:+1:

hdeshev:
run ci

hdeshev:
Killed the utils/module-loader module and moved the 3 functions to globals.

PR #1256, opened 2015-12-15
lazy requires implemented in various modules

enchev:
the required dependency will be executed (required) only when needed in the code not as usual pre-required in the beginning

### DIFF
--- a/application/application.ios.ts
+++ b/application/application.ios.ts
@@ -1,13 +1,12 @@
-﻿import appModule = require("./application-common");
+﻿import common = require("./application-common");
 import frame = require("ui/frame");
-import types = require("utils/types");
-import view = require("ui/core/view");
 import definition = require("application");
-import enums = require("ui/enums");
-import uiUtils = require("ui/utils");
-import fileResolverModule = require("file-system/file-name-resolver");
+import * as uiUtilsModule from "ui/utils";
+import * as typesModule from "utils/types";
+import * as fileResolverModule  from "file-system/file-name-resolver";
+import * as enumsModule from "ui/enums";
 
-global.moduleMerge(appModule, exports);
+global.moduleMerge(common, exports);
 
 class Responder extends UIResponder {
     //
@@ -15,7 +14,7 @@ class Responder extends UIResponder {
 
 class Window extends UIWindow {
 
-    private _content: view.View;
+    private _content;
 
     initWithFrame(frame: CGRect): UIWindow {
         var window = super.initWithFrame(frame);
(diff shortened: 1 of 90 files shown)


PanayotCankov:
:+1:

atanasovg:
My major concern here is with losing all the typed information - TS considers all such lazy modules as `any`. This in long term means that any API changes will not be caught by the compiler and will need to be verified in some other way(s). If this can be done via some custom build step at JavaScript level only that will be perfect.

PanayotCankov:
@atanasovg that are some valid concerns. However I'd like to lazy load at least the poly-fills that add to the global and slow the startup time for now.

enchev:
Hey folks,

I've made this pull request just to test with all apps not only with our unit tests (with help from our QAs), still I don't mind the discussion! :)
Indeed we will lose build-time checkes after this switch from TypeSccript import to plain JavaScript require however we will win up to 30% improved loading time of the app.

atanasovg:
Yep - 30% is pretty significant improvement, I am definitely not against it :) All I am wondering is whether the same effect may be achieved in some other way, without losing the typed information.

enchev:
@atanasovg Unfortunately I'm not aware of any other way where you can require on demand but have TypeScript types available. Anyone have any idea?

hdeshev:
I found a way to have typed lazy imports. It's ugly, but it works:

1. Write a `d.ts` file for your module using a global module path e.g. "my-module". We do that already for our modules.
2. Add a "normal" global import, to get the type: `import * as myModule from "my-module".
3. Add your lazy require call using a **relative** module path, and cast the require return value to your module type: `var lazyHello: typeof myModule = require("./my-module");`

You can also do it for single exports:
```
import {hello} from "my-module";
...
var h: typeof hello = require("./my-module").hello;
h();
```

Full project here: https://github.com/hdeshev/typescript-typed-lazy-import

enchev:
WOW! I'll rework my changes! Tanks @hdeshev!

enchev:
run ci

enchev:
run ci

application/application.ios.ts
```diff
@@ -196,7 +199,7 @@ class IOSApplication implements definition.iOSApplication {
             }
 
             exports.notify(<definition.OrientationChangedEventData>{
```
hshristov: definition leftover. but why do you change definition with exports?
It doesn't save anything because application.ios.ts is in fact - application.js so it requires it self.

camera/camera.ios.ts
```diff
@@ -71,8 +75,8 @@ class UIImagePickerControllerDelegateImpl extends NSObject implements UIImagePic
 
 var listener;
 
-export var takePicture = function (options?: definition.CameraOptions): Promise<imageSource.ImageSource> {
```
hshristov: Same as in application module - removing definition does not get performance improvement.

PR #1273, opened 2015-12-18
Fix the travis build

ErjanGavalji:
Add an npm install step so that local grunt gets found
Update the nodejs version used to 4.2.2


enchev:
:+1:

PR #1274, opened 2015-12-18
Make the modules correct NativeScript plugin

rosen-vladimirov:
By definition NativeScript plugins have "nativescript" key in their `package.json`.
When there's such key, CLI allows you to execute `tns plugin add tns-core-modules`.
Also this gives you the ability to specify minimum required runtime version.
I've set it to 1.3.0 as I couldn't find any information which is the current supported runtimes version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,5 +37,11 @@
   "typings": "tns-core-modules.d.ts",
   "dependencies": {
     "tns-core-modules-widgets": "1.6.0"
+  },
+  "nativescript": {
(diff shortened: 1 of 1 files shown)


ligaz:
It will be a very good practice to update those versions when we are taking dependencies on new features from the runtimes. @NativeScript/tns-modules

enchev:
Moved to https://github.com/NativeScript/NativeScript/pull/1276

PR #1279, opened 2015-12-21
More reliable global objects lazy loader.

hdeshev:
Not redefining global properties to avoid breaking monkey patchers such as
zone.js (Angular 2).

### DIFF
--- a/globals/globals.ts
+++ b/globals/globals.ts
@@ -31,17 +31,24 @@ global.registerModule("ui/dialogs", () => require("ui/dialogs"));
 global.registerModule("../xhr/xhr", () => require("../xhr/xhr"));
 global.registerModule("fetch", () => require("fetch"));
 
-function registerOnGlobalContext(name, module) {
+const __tnsGlobalMergedModules = new Map<string, boolean>();
+
+function registerOnGlobalContext(name: string, module: string): void {
+
     Object.defineProperty(global, name, {
         get: function () {
             // We do not need to cache require() call since it is already cached in the runtime.
             let m = global.loadModule(module);
-            global.moduleMerge(m, global);
+            if (!__tnsGlobalMergedModules.has(module)) {
+                __tnsGlobalMergedModules.set(module, true);
+                global.moduleMerge(m, global);
+            }
 
             // Redefine the property to make sure the above code is executed only once.
-            Object.defineProperty(this, name, { value: m[name], configurable: true, writable: true });
+            let resolvedValue = m[name];
+            Object.defineProperty(this, name, { value: resolvedValue, configurable: true, writable: true });
 
-            return m[name];
+            return resolvedValue;
         },
         configurable: true
     });


vakrilov:
:+1:

fealebenpae:
I propose the following: leave the getter on the property descriptor as-is, and in the body of the setter add `Object.defineProperty(this, name, { value: value, configurable: true, writable: true })`.

This should enable monkey patching in the form of
```
global.XMLHttpRequest = function MyFakeXHR() { ... };
```
while preserving the crucial feature of the property descriptor to replace itself with a data descriptor on first access. This is important on iOS where the engine can't JIT-inline the property accessor.

hdeshev:
@fealebenpae Looks great! I changed the setter, and it seems to work fine in ng2 projects.

fealebenpae:
@hdeshev why change the getter, though? At worst, you'd need to remove the call to`global.moduleMerge`. How does it break zone.js?

globals/globals.ts
```diff
@@ -34,14 +34,20 @@ global.registerModule("fetch", () => require("fetch"));
...
         get: function () {
-            // We do not need to cache require() call since it is already cached in the runtime.
-            let m = global.loadModule(module);
-            global.moduleMerge(m, global);
-
-            // Redefine the property to make sure the above code is executed only once.
-            Object.defineProperty(this, name, { value: m[name], configurable: true, writable: true });
-
-            return m[name];
+            if (this['__' + name] === undefined) {
```
ligaz: You can cache the indexer access to a variable in order to avoid it being called twice.

hdeshev:
> @hdeshev why change the getter, though? At worst, you'd need to remove the call toglobal.moduleMerge. How does it break zone.js?

It seems the second `Object.defineProperty` doesn't **immediately** replace the property, and the zone `global["setTimeout"] = patchedFunction` effectively does nothing. You can test it with the original code by adding a set function in the outer `Object.defineProperty` that throws an exception, and call the setter after the inner/second `Object.defineProperty` that is supposed to remove the setter. You'd still get the exception though. ¯\\\_(ツ)_/¯ At least that's what it does on Android.

hdeshev:
run ci

fealebenpae:
@hdeshev: unfortunately I don't have an Android environment on hand, but I tried to reproduce this with node.js:
```
var name = "setTimeout";
Object.defineProperty(global, name, {
	get: function() {
		function lazySetTimeout() {
			console.log("lazySetTimeout called");
		}
		
		Object.defineProperty(this, name, { value: lazySetTimeout, configurable: true, writable: true });
		return this[name];
	},
	set: function(value) {
		Object.defineProperty(this, name, { value: value, configurable: true, writable: true });
	},
	configurable: true
});

setTimeout();

global["setTimeout"] = function fakeSetTimeout() {
	console.log("fakeSetTimeout called");
};

setTimeout();
```

Running this gives me
```
lazySetTimeout called
fakeSetTimeout called
```
and if I comment out the first call to `setTimeout` I only get `fakeSetTimeout called` as expected.

Perhaps the culprit is the call to `global.moduleMerge`? It occurred to me it wouldn't be right to call `moduleMerge` in the getter as that would overwrite any previous monkey patching:
```
global["setInterval"] = function fakeSetInterval() { ... };
setTimeout();
```
would effectively overwrite the value of `global["setInterval"]` because the getter of `global.setTimeout` will have called `moduleMerge`.

hdeshev:
> Perhaps the culprit is the call to global.moduleMerge?

@fealebenpae, you are 100% correct. What was really happening was that zone.js was patching `setTimeout`, and then `clearTimeout`, which in turn called `moduleMerge` again and "unpatched" the `setTimeout` function.

I fixed it by keeping track of global merges and calling `moduleMerge` just once.

piledriver

PR #1291, opened 2015-12-23
Improve default XML namespace skips.

hdeshev:
- Check every parsed namespace instead of running a global regex on the
whole input
- Check for namespaces ending in tns.xsd and allow schemas hosted on
different URL's than the default one.


nsndeck:
:+1:

PanayotCankov:
:+1:

Issue #1280, opened 2015-12-21
text-transform breaks child colors


sitefinitysteve:
http://www.screencast.com/t/ywVfBrLx

So this works okay, text though isn't uppercase
```
<Label cssClass="logo-label" textWrap="false">
	<Label.formattedText>
		<FormattedString>
			<FormattedString.spans>
				<Span text="{{ BrandPrefix }}" fontAttributes="Bold" cssClass="prefix" />
				<Span text="{{ BrandPostfix }}" fontAttributes="Bold" foregroundColor="#bd5680" />
			</FormattedString.spans>
		</FormattedString>
	</Label.formattedText>
</Label>
```

So when I add in text-transform: uppercase it's now uppercase,  but I lose the purple on "Help"

pointdeveloper:
yes text-transform does not work..but text-transform is not mentioned in the docs http://docs.nativescript.org/ui/styling so i guess it is not meant to work in the first place..kindly suggest any other way to transform the case of text...

sitefinitysteve:
It might not be in the docs, but it is there... Just maybe borked which is why i opened the issue 

https://twitter.com/nativescript/status/677116835398053888

pointdeveloper:
okay i found this in the docs http://docs.nativescript.org/ApiReference/ui/enums/TextTransform/README i have no idea how to implement this in the view

sitefinitysteve:
Just add it in the css like you would in html

pointdeveloper:
okay i found this in the docs http://docs.nativescript.org/ApiReference/ui/enums/TextTransform/README i have no idea how to implement this in the view

pointdeveloper:
okay..i have another issue regarding this..i am loading a list from a http call and the list is loading fine and i can getting the desired text-transform, BUT i only see the text in the visible area as capitalized the rest of the text is not, so when i scroll down i see non-styled text

sitefinitysteve:
I'm not using it on that context so i can't comment, but if you believe this to be a problem, open a new issue for it... Its not related to this one so it might get overlooked

priyangamani:
Text Transform is not working in tns 3.0

PR #1345, opened 2016-01-11
text-transform and text-decoration with formattedText fixed

enchev:
Fix for #1280

### DIFF
--- a/utils/utils.ios.ts
+++ b/utils/utils.ios.ts
@@ -49,108 +49,103 @@ export module ios {
         }
     }
 
-    export function setTextDecoration(view: dts.ios.TextUIView, value: string) {
-        var attributes: NSMutableDictionary = NSMutableDictionary.alloc().init();
-        var values = (value + "").split(" ");
+    export function setTextDecorationAndTransform(v: any, decoration: string, transform: string) {
+        if (v.formattedText) {
+            if (v.style.textDecoration.indexOf(enums.TextDecoration.none) === -1) {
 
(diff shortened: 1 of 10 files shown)


utils/utils.ios.ts
```diff
@@ -49,108 +49,103 @@ export module ios {
...
+
+            for (let i = 0; i < v.formattedText.spans.length; i++) {
+                let span = v.formattedText.spans.getItem(i);
+                span.text = getTransformedText(v, span.text, transform);
+                console.log("HERE: " + span.text);
+            }
         } else {
-            setTextDecorationNative(view, view.text || view.attributedText, NSMutableDictionary.alloc().init());
+            let source = v.text;
+            let attributes = NSMutableArray.array();
```
hshristov: According to ios-runtime it will be much faster if we pass JS array here.

utils/utils.ios.ts
```diff
@@ -49,108 +49,103 @@ export module ios {
...
+            let range = { location: 0, length: source.length };
+
+            var decorationValues = (decoration + "").split(" ");
+
+            if (decorationValues.indexOf(enums.TextDecoration.none) === -1) {
+                let dict = NSMutableDictionary.new();
```
hshristov: Same change for Map<> here.

hshristov:
:+1: but it would be better if we switch NSArray & NSDictionary to JS Array and JS Map.

dtopuzov:
run ci

PR #1381, opened 2016-01-15
Fix originX/Y not applied when set before layout in android

PanayotCankov:
Use the OriginPoint class in android widgets for the originX/Y implementation.
It will listen for layout events and recalculate the pivot point in pixels based on the origin point that is relative to the view size.


vakrilov:
:+1:

hshristov:
:+1:

PR #1398, opened 2016-01-19
Add an additional notification for propChange event (required for angular2).

nsndeck:


ui/core/dependency-observable.ts
```diff
@@ -364,6 +364,16 @@ export class DependencyObservable extends observable.Observable {
             var changeData = super._createPropertyChangeData(property.name, newValue);
             this.notify(changeData);
         }
+
+        if (this.hasListeners(property.name + "Change")) {
```
hshristov: Why don't use `let name = property.name + "Change"` and then use the name field?
This way there will be only one string concatenation.

enchev:
:+1:

PR #1407, opened 2016-01-20
Atanasovg/snapshot refactorings

atanasovg:
This pull applies the following refactoring:

- Some circular dependencies removed - for example in the application module, which has the following require graph:
    - `application-common` requires `application`
    - `application` requires `application-common`
    - `application` requires `application`
- All of the `android.xxx` calls are now wrapped in functions in order to be lazily evaluated. This enables initial snap-shotting of the modules.
- Modified the `file-system.currentApp` implementation to use Android APIs rather than string operations.
- Removed `definition.xxx` calls for all BUT types calls. This fixes cases where the module requires itself. For example, the TS compiler will generate a require call for the same module
    ```js
    import definition = require("myModule");
    exports class Class1 implements definition.Class1 {
    }

    var a;
    if (a typeof definition.Class1) { // wrong, generates require call for the same module
    }

    if (a typeof Class1) { // correct, uses the definition import only for type reference
    }
    ```

### DIFF
--- a/application/application.d.ts
+++ b/application/application.d.ts
@@ -118,8 +118,9 @@ declare module "application" {
 
     /**
      * Loads css file and parses to a css syntax tree.
+     * @param cssFile Optional parameter to point to an arbitrary css file. If not specified, the cssFile property is used.
      */
-    export function loadCss(): void;
+    export function loadCss(cssFile?: string): Array<cssSelector.CssSelector>;
 
     /**
      * Call this method to start the application. Important: All code after this method call will not be executed!
@@ -329,6 +330,11 @@ declare module "application" {
          * The name of the application package.
          */
         packageName: string;
+
+        /**
+         * True if the application is not running (suspended), false otherwise.
+         */
+        paused: boolean;
 
         /**
          * This method is called by the JavaScript Bridge when navigation to a new activity is triggered.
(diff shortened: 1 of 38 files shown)


application/application.d.ts
```diff
@@ -331,6 +332,11 @@ declare module "application" {
         packageName: string;
 
         /**
+         * True if the application is not running (suspended), false otherwise.
+         */
+        paused: boolean;
```
hshristov: This field is not used on iOS. Do we need it in the .d.ts?

atanasovg: Well, it is part of the `AndroidApplication` class.

ui/image-cache/image-cache.android.ts
```diff
@@ -1,35 +1,45 @@
...
+        //}
+    };
+
+    LruBitmapCacheClass = LruBitmapCache;
+}
 
 export class Cache extends common.Cache {
     private _callback: any;
-    private _cache: LruBitmapCache;
+    private _cache;
```
hshristov: missing type (: LruBitmapCacheClass)

hshristov:
:+1:

enchev:
run ci

atanasovg:
run ci

dtopuzov:
run ci

vchimev:
run ci

vchimev:
run ci

vchimev:
run ci

PR #1453, opened 2016-01-28
Register ProxyViewContainer children with parent layout.

hdeshev:
Fixes a crash when used in a GridLayout on iOS since the layout needs to be
notified of any layout children changes.

/CC: @vakrilov, @hshristov


hdeshev:
run ci

vchimev:
run ci

PR #1460, opened 2016-02-01
Array.create used instead reflection

enchev:


hdeshev:
:+1: 

Since we're using `"byte"` and `"int"` as literals it would make sense to do so for `"float"` instead of using the `floatType` constant. We can create constants for byte and int, but that seems to be of very little value.

You can go further and extract a method or pull the one-liner that creates the single-element float array before the switch statement instead of replicating it in [almost] every case. :neckbeard:

PR #1470, opened 2016-02-02
Enable build times profiling

PanayotCankov:
Will output build times for the grunt tasks by using:
```
grunt --profile
```
Profile will not print very small tasks. To list all tasks use:
```
grunt --profile=true --verbose=true
```


vakrilov:
:+1:

PR #1472, opened 2016-02-02
Refactored animations in iOS to allow 360 degrees rotation and custom timing functions.

tzraikov:


ui/animation/animation.ios.ts
```diff
@@ -16,30 +16,59 @@ class AnimationDelegateImpl extends NSObject {
         return <AnimationDelegateImpl>super.new();
```
hshristov: We agreed to use the newer approach where we don't have `static new(): AnimationDelegateImpl` but instead `static initWithFinishedCallback` function which internally casts the result to `AnimationDelegateImpl`. This reduce the method calls and workarounds an issue with the runtime

enchev:
run ci

enchev:
:+1:

PR #1486, opened 2016-02-04
Fix paths in iOS, some devices and iOS combinations use /private/var paths while some /var/ this was normalized by resolvin symlinks

PanayotCankov:


enchev:
:100:

file-system/file-system-access.ios.ts
```diff
@@ -215,7 +215,9 @@ export class FileSystemAccess {
     }
 
     public getLogicalRootPath(): string {
-        return NSBundle.mainBundle().bundlePath;
+        let mainBundlePath = NSBundle.mainBundle().bundlePath;
+        let resolvedPath = NSURL.fileURLWithPath(mainBundlePath).URLByResolvingSymlinksInPath.path;
```
fealebenpae: You can call `-[NSString stringByResolvingSymlinksInPath]` which should do the same while skipping the extra `NSURL` allocation.

PanayotCankov: I will,
Thanks :)

PanayotCankov: It will become
```
let mainBundlePath = NSBundle.mainBundle().bundlePath;
let resolvedPath = NSString.stringWithString(mainBundlePath).stringByResolvingSymlinksInPath;
return resolvedPath;
```
The bundlePath is marshalled to JavaScript string. We need NSString. Is this better?

fealebenpae: It's cheap enough. We could have gotten away with it easier if `stringByResolvingSymlinksInPath` had been a method - `NSString.prototype.stringByResolvingSymlinksInPath.call(mainBundlePath)`, but that would have only netted us only one less FFI call, the actual count of allocations would have been the same.

Issue #1371, opened 2016-01-13
listview android bug and fix


x4080:
Background :
- Load item 50, "realizedItems" is 6, this is the item cached to display at the screen
- Filter the array to 5, scroll up, it will crash because realizedItems>new array

Solution : (list-view.android.js)
```
    ListView.prototype._notifyScrollIdle = function () {
        var keys = Object.keys(this._realizedItems);
        var i;
        var length = keys.length;
        var view;
        var key;
        for (i = 0; i < length; i++) {
            key = keys[i];
            view = this._realizedItems[key];
			if (view[REALIZED_INDEX]<this.items.length) // FIX
            this.notify({
                eventName: ITEMLOADING,
                object: this,
                index: view[REALIZED_INDEX],
                view: view
            });
        }
    };

```

IOS is working fine

(EDITED because the first fix is still getting error @lollipop)

enchev:
Hey @x4080,

Why not submit pull request with the fix?

x4080:
because i dont know how :)

ErjanGavalji:
@x4080, I can lend a hand with that or I can do the pull request for you. I'd recommend you do it to get the full credits of contributing to the open-source project. I'll be glad to assist... Let me know which one you prefer.

x4080:
just do it, you guys have this wonderful work and let others to use it, already great things by itself, this is a little tiny contribution for ns community, so you have my blessing :)

ErjanGavalji:
@x4080, thanks for the kind words :)

Can I only ask you to sign our [CLA](http://www.nativescript.org/cla)? (Sorry for having to cause that inconvenience...)

Cheers,
Erjan

PR #1490, opened 2016-02-04
Fix ListView crash in android when realized items more than left items

ErjanGavalji:
Apply the fix, suggested by Github user @x4080, issue #1371


enchev:
:+1:

PR #1494, opened 2016-02-04
Add a homepage entry to the package.json

ErjanGavalji:


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "tns-core-modules",
   "description": "Telerik NativeScript Core Modules",
   "version": "1.6.0",
+  "homepage":"http://www.nativescript.org",
```
ligaz: You can use `https` here.

ErjanGavalji: makes sense, applying a change

ligaz:
:+1: Outstanding work!

PR #1502, opened 2016-02-05
Fixed crash where activity was incorrectly using existing initialized frame.

hshristov:
Fixed TabView to use _nativeView instead of Frame.
Added an option to change the rootView.

### DIFF
--- a/application/application.d.ts
+++ b/application/application.d.ts
@@ -5,7 +5,7 @@ declare module "application" {
     import cssSelector = require("ui/styling/css-selector");
     import observable = require("data/observable");
     import frame = require("ui/frame");
-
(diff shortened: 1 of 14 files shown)


application/application.d.ts
```diff
@@ -130,7 +141,7 @@ declare module "application" {
     /**
      * The main entry point event. This method is expected to use the root frame to navigate to the main application page.
      */
-    export function onLaunch(context: any): void;
+    export function onLaunch(context?: any): void;
```
vakrilov: Should we use name the parameter intent instead of context and specify that it is undefined for iOS.

vakrilov:
:+1:

PR #1512, opened 2016-02-08
Fix android corner radius in cases when the radius is less than half the border width

PanayotCankov:
Degraded after adding antialiasing to the rounded corners drawing.


hshristov:
:+1:

enchev:
:+1:

PR #1531, opened 2016-02-10
Replace -[NSObject performSelector:] with NSInvocation

fealebenpae:
It appears that `-[NSObject performSelector:]` returns a dummy value when calling a void method on the iOS Simulator, whereas it correctly returns `nil` on a device. Since `-[NSObject performSelector:]` is declared as returning `id`, the iOS runtime tries to marshal the dummy value to a JavaScript object and fails because the dummy value is not a valid Objective-C object.

The fix is to use `-[NSObject valueForKey:]` for non-void zero-parameter methods and `NSInvocation` for everything else.

### DIFF
--- a/ui/transition/transition.ios.ts
+++ b/ui/transition/transition.ios.ts
@@ -3,6 +3,18 @@ import frame = require("ui/frame");
 import types = require("utils/types");
 import trace = require("trace");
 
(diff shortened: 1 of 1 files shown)


enchev:
run ci

enchev:
:+1:

PR #1533, opened 2016-02-11
Fixed null and undefined as Label text

### LINKED ISSUE #1524: Label text="{{ anyField }}" should not print out "null" if the field is "null"
Label text="{{ anyField }}" should not print out "null" if the field is "null", I think this should be changed, otherwise we need to have a converter to every single field.

Let me know if I am missing something here.
Thanks

enchev:
Fixed #1524


hshristov:
:+1:

PR #1539, opened 2016-02-11
Fix clearEvent after programmatically setting text

speigg:
searchBarTextDidChange only fires when the user changes the text. 
In the following scenario, clearEvent was not being fired:
1. User clears the text (clearEvent is fired from searchBarTextDidChange)
2. Text is programmatically set to a non-empty string (searchBarTextDidChange is not called)
3. User clears the text again (searchBarTextDidChange is called, but clearEvent is not fired)

This PR allows the clearEvent to fire at step 3. 

I would also suggest that a separate event be used when the cancel button is clicked (cancelEvent), as the cancel button being clicked does not imply that the text is cleared.


enchev:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

enchev:
:+1:

Issue #1534, opened 2016-02-11
FileSystemAccess.fileExists returns true for directory as well


atanasovg:
Since in Java a `File` object represents both a directory and a file, we need to check not only for `exists` but for `isDirectory` as well:
https://github.com/NativeScript/NativeScript/blob/master/file-system/file-system-access.android.ts#L78

enchev:
Hey folks,

It seems we were too fast! :)

`File.exists` will return true for folder in both iOS and Node.js as well. It seems that this is the expected behavior!

cokobware:
Thanks for fixing this guys!

lhitc:
This is completely contrary to virtually every other programming language or framework, not only Java or Node.js.
The various "exists" functions check for existance regardless of file type, and directories are just files of a special type. Typically you check that at the second stage.

So, this breaks intuitively expected behaviour and possibly also some already existing code out there.

Just my 5 cent.

NathanaelA:
I happen to agree with @lhitc -- exists is one of those unique cases where most developers know that fs.exists = true for both folders and files.    Take this psuedo code for example:
if (!fs.exists('blah')) {
  var file = fs.create('blah');   // Fails here, but the assumption is that it shouldn't...
  file.write('wow');
}

enchev:
Well, I was afraid of this. :) Possible option will be to restore the original behavior and provide additional optional parameter with `true` as default value:
```JavaScript
 export class File extends FileSystemEntity {
        static exists(path: string, includeFolder?: boolean): boolean;
}
```

In this case you will be able to set this to `false` if you want to check only for files not for folders. 
```JavaScript
var fileExists = fs.File.exists(path, false);
```
Hey @atanasovg, @cokobware, what do you think?

lhitc:
If you want to implement both variants for convenience, it might be more consistent to have 3 different methods, like:
static exists(path: string): boolean; // classic behaviour
static isExistingRegularFile(path: string): boolean; // true if plain file
static isExistingFolder(path: string): boolean; // true if folder

atanasovg:
@lhitc .NET for one will NOT return true for a directory: https://msdn.microsoft.com/en-us/library/system.io.file.exists(v=vs.110).aspx. But I guess our thinking is biased due to the fact that many of the core engineering team have extensive experience with .NET.

[...]

PR #1544, opened 2016-02-12
File.exists fixed for ios as well + tests

enchev:
Update for #1534


file-system/file-system-access.ios.ts
```diff
@@ -148,7 +148,7 @@ export class FileSystemAccess {
 
     public fileExists(path: string): boolean {
         var fileManager = NSFileManager.defaultManager();
-        return fileManager.fileExistsAtPath(path);
+        return fileManager.fileExistsAtPath(path) && !this.folderExists(path);
```
fealebenpae: You can call `-[NSFileManager fileExistsAtPath:isDirectory:]` like so: https://github.com/NativeScript/ios-runtime/blob/v1.5.2/src/NativeScript/require.js#L75.

enchev: Yep, this is what this.folderExists is doing

dtopuzov:
:+1: ok to merge

file-system/file-system-access.ios.ts
fealebenpae: Yes, except you use two native calls to accomplish the same task a single native call can perform just as well. Not to mention that calling both methods requires `path` to be copied to `NSString` so that's not only one extra FFI operation, but an extra allocation as well.

enchev: Is it better? https://github.com/NativeScript/NativeScript/pull/1545

PR #1572, opened 2016-02-18
Add global __metadata function

rosen-vladimirov:
When `emitDecoratorMetadata` compiler option is set to true in `tsconfig.json`, TypeScript generates some calls to `__metadata` function.
By default in our NativeScript TypeScript projects the `noEmitHelpers` option is set to true, so TypeScript compiler does not generate the `__metadata` function.
So the application crashes at runtime. Add the missing method, so we can have both `noEmitHelpers` and `emitDecoratorMetadata` set to true in `tsconfig.json`.

### DIFF
--- a/globals/globals.ts
+++ b/globals/globals.ts
@@ -119,6 +119,14 @@ if (typeof global.__decorate !== "function") {
     }
 }
 
+if (typeof global.__metadata !== "function") {
+    global.__metadata = function (k, v) {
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
run ci

enchev:
:+1:

nsndeck:
:+1:

vchimev:
run ci

PR #1575, opened 2016-02-18
Unify our Map/Set typings with the ones in es6-shim (angular)

hdeshev:
Right now we are extending the types in d.ts files included in angular projects, but it would be easiest for users if those were just included in the core typings.


enchev:
:+1:

hdeshev:
run ci

ErjanGavalji:
:+1:

PR #1616, opened 2016-02-23
Upgrade to TypeScript 1.8.2

hdeshev:
# WARNING! BREAKING CHANGE!

This pull request deletes the `ui.ts` and `ui.d.ts` files, which means you can no longer do `import {Page} from "ui"`. I believe this is a relatively unpopular feature, and few people (if any) will miss it.

The reason behind the above is that we have modules exporting symbols with the same name e.g. `Options` in `ui/core/view` and `ui/editable-text-base`. When you reexport those modules from `ui.ts`, you get a name clash. This code no longer works:

```typescript
export * from "ui/core/view";
...
export * from "ui/editable-text-base";
```

TypeScript 1.7 did not detect this problem (I doubt you could reliably use any of the clashing symbols anyway).

I looked for ways to hide or rename a single symbol while reexporting the rest as *, but it seems there is no way to do that (confirmed in the TypeScript gitter chat room). The only alternative is to list **every** exported symbol in the minor module when reexporting it and rename the conflicting ones. And, of course, keep them in sync with every release. :scream: 

CC @enchev, @atanasovg, @hshristov


enchev:
:+1:

dtopuzov:
run ci

hshristov:
:+1:

ErjanGavalji:
:+1:

vakrilov:
:+1:

hdeshev:
run ci

atanasovg:
:100: 
Besides the described problems, requiring this shortcut module requires virtually all the modules and degrades the loading time significantly - for example the [cuteness demo](https://github.com/NativeScript/NativeScript/blob/master/apps/cuteness.io/main-page.ts#L2)

PR #1627, opened 2016-02-24
Allow the build to produce js files near the ts files

PanayotCankov:
There is a `grunt inplace` task now that outputs the js files near the ts instead of in the bin folder.
The `npm pack` after `grunt inplace` produces a package similar to the one produced in `build/modules` with the only difference that it includes all readme and licence files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # generated files
 *.map
+*.tgz
 .baseDir.ts
 .sublime-grunt.cache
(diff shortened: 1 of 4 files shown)


hdeshev:
:+1:

hshristov:
:+1:

ErjanGavalji:
:+1:

dtopuzov:
run ci

dtopuzov:
run ci

PR #1647, opened 2016-02-25
Frame.navigate while in background

### LINKED ISSUE #1389: Frame navigate, while android Activity is paused, crash
##### Prio* Doing frameModule.topmost().navigate() when the app is not in the foreground crashes Java

The following page-1.js will crash Nativescript if you press the phone's "home" button as to switch activity away from the program within 5 seconds.

How can this be worked around?

var timerModule = require("timer");

exports.loaded = function(args) {
     timerModule.setTimeout(function () {
          frameModule.topmost().navigate({ create: function () {
     		return new pageModule.Page();
          }});
     },5000);
};

PanayotCankov:
Frame.navigate used to crash when in background, now it delays the navigation until activity resume.
Fixes #1389
The Android app should be sent to background to reproduce the issue so it should have ui tests.

### DIFF
--- a/ui/frame/frame-common.ts
+++ b/ui/frame/frame-common.ts
@@ -127,7 +127,7 @@ function pageFromBuilder(moduleNamePath: string, moduleExports: any): Page {
     return page;
 }
 
-interface NavigationContext {
+export interface NavigationContext {
     entry: definition.BackstackEntry;
     isBackNavigation: boolean;
 }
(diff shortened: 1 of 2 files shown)


enchev:
:+1:

ErjanGavalji:
:+1:

calebeaires:
:+1:

PR #1652, opened 2016-02-25
Fix language tags in doc-snippets

ErjanGavalji:
The doc-snippets were inconsistent
(no spaces before title, small-caps, JS
instead of JavaScript, etc.)

Not bound to an issue. Change related to generated How-to articles.

The change itself is in the unit test comments.


hshristov:
:+1:

PR #1679, opened 2016-03-01
Fix a crash when using a ProxyViewContainer in ListView item templates.

hdeshev:
Not a regular {N} usage scenario, but it can happen with Angular components
which get wrapped in a proxy container.


enchev:
:+1:

PR #1682, opened 2016-03-02
Introducing DatePicker date property.

nsndeck:
Added a cumulative `date` property - a result from already existing year, month and day properties. The reason for adding such property is to allow Angular2 `ngModel` support via the new property and also its type - `JavaScript Date type` allows using Angular2 built-in date pipes.

### DIFF
--- a/apps/tests/ui/date-picker/date-picker-tests.ts
+++ b/apps/tests/ui/date-picker/date-picker-tests.ts
@@ -1,6 +1,5 @@
 ﻿import TKUnit = require("../../TKUnit");
-import helper = require("../helper");
-import viewModule = require("ui/core/view");
+import testModule = require("../../ui-test");
 import datePickerTestsNative = require("./date-picker-tests-native");
 import color = require("color");
 import platform = require("platform");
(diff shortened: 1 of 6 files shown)


apps/tests/ui/date-picker/date-picker-tests.ts
```diff
@@ -92,6 +92,15 @@ export function test_WhenCreated_DayIsUndefined() {
     });
 }
 
+export function test_WhenCreated_DateIsUndefined() {
+    helper.buildUIAndRunTest(_createDatePicker(), function (views: Array<viewModule.View>) {
+        var datePicker = <datePickerModule.DatePicker>views[0];
+        var actualValue = datePicker.date;
+        var expectedValue = undefined;
+        TKUnit.assert(actualValue === expectedValue, "Actual: " + actualValue + "; Expected: " + expectedValue);
```
hshristov: AssertEqual give you better message when test fails.

apps/tests/ui/date-picker/date-picker-tests.ts
```diff
@@ -92,6 +92,15 @@ export function test_WhenCreated_DayIsUndefined() {
     });
 }
 
+export function test_WhenCreated_DateIsUndefined() {
+    helper.buildUIAndRunTest(_createDatePicker(), function (views: Array<viewModule.View>) {
```
hshristov: Instead of buildUIAndRunTest - the new approach with UITest base class leads to faster unit-tests execution.

apps/tests/ui/date-picker/date-picker-tests.ts
```diff
@@ -122,6 +131,42 @@ export function testDayFromLocalToNative() {
     });
 }
 
+export function test_DateIsSetCorrectlyWhenDayIsSet() {
+    helper.buildUIAndRunTest(_createDatePicker(), function (views: Array<viewModule.View>) {
+        var datePicker = <datePickerModule.DatePicker>views[0];
+        var expectedValue = 19;
+        datePicker.day = expectedValue;
+        let expectedDate = new Date(0, 0, 19);
```
hshristov: Is 0, 0, 19 valid Date?

ui/date-picker/date-picker.android.ts
```diff
@@ -90,14 +103,17 @@ export class DatePicker extends common.DatePicker {
 
                     if (year !== this.owner.year) {
                         this.owner._onPropertyChangedFromNative(common.DatePicker.yearProperty, year);
+                        this.owner._onPropertyChangedFromNative(common.DatePicker.dateProperty, new Date(year, this.owner.month - 1, this.owner.day));
                     }
 
                     if ((month + 1) !== this.owner.month) {
```
hshristov: Why do we use month + 1?
What will happen if month is 12?

ui/date-picker/date-picker.ios.ts
```diff
@@ -61,6 +61,18 @@ function onMinDatePropertyChanged(data: dependencyObservable.PropertyChangeData)
 
 (<proxy.PropertyMetadata>common.DatePicker.minDateProperty.metadata).onSetNativeValue = onMinDatePropertyChanged;
 
+function onDatePropertyChanged(data: dependencyObservable.PropertyChangeData) {
+    var picker = <DatePicker>data.object;
+
+    if (picker.ios) {
+        var comps = NSCalendar.currentCalendar().componentsFromDate(NSCalendarUnit.NSCalendarUnitYear | NSCalendarUnit.NSCalendarUnitMonth | NSCalendarUnit.NSCalendarUnitDay, picker.ios.date);
```
hshristov: Why not change the name to something more meaningful?

ui/date-picker/date-picker.ios.ts
```diff
@@ -61,6 +61,18 @@ function onMinDatePropertyChanged(data: dependencyObservable.PropertyChangeData)
...
+        comps.date = data.newValue;
+        picker.ios.setDateAnimated(NSCalendar.currentCalendar().dateFromComponents(comps), false);
```
hshristov: Why we don't use animations?

PR #1707, opened 2016-03-07
Allow multiple headers with the same name to be retrieved from a HTTP response

### LINKED ISSUE #1695: http module does not support multiple headers with same name
In the implementation of http module, the headers are parsed into an object where the keys are header names and the values are header values. If there are multiple headers with same name in the response like multiple `Set-Cookie` headers, only the last header survive. It overwrites the previous headers.

Example:

These headers: 
````
Header-A : content1
Header-B : content2
Header-B : content3
````
will result in:
````js
{
    "Header-A" : "content1",
    "Header-B" : "content3"
}
````

How can I reach the lost headers? If there is not a way, you may also provide the headers as an array next to the current headers object in http module.

### PR DESCRIPTION
**Pull request by enchev**
Fixed #1695

### DIFF
--- a/apps/tests/http-tests.ts
+++ b/apps/tests/http-tests.ts
@@ -535,6 +535,27 @@ export var test_request_headersSentAndReceivedProperly = function (done) {
     });
 };
 
+export var test_request_headersWithSameKeyAddedProperly = function (done) {
+    var keyName = "key";
+    var value1 = "value1";
+    var value2 = "value2";
+    
+    var headers = {};
+    
+    (<any>http).addHeader(headers, keyName, value1);
+    (<any>http).addHeader(headers, keyName, value2);
+    
+    try {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: apps/tests/http-tests.ts [hunk 1]
```diff
@@ -535,6 +535,21 @@ export var test_request_headersSentAndReceivedProperly = function (done) {
...
 };
 
+export var test_request_headersWithSameKeyAddedProperly = function (done) {   
+    var headers = {};
+    
+    (<any>http).addHeader(headers, "key", "value1");
+    (<any>http).addHeader(headers, "key", "value2");
+    
+    try {
+        TKUnit.assert(Array.isArray(headers["key"]), "Headers with same key not added properly!");
```
hdeshev: Why not check for both values here instead of just verifying the result as being an array?

hdeshev:
:+1:

vchimev:
run ci

PR #1712, opened 2016-03-08
Introducing time property for TimePicker.

nsndeck:
Added `time` property for TimePicker. This property is cumulative for both existing properties (hour and minute) and will be used as a single input option in `Angular2 ngModel support` (two-way data binding).


dtopuzov:
:+1:

PR #1714, opened 2016-03-08
Update webView.url due to tests timeout

vchimev:


vchimev:
Timeout probably caused by the network connectivity.

PR #1717, opened 2016-03-08
Hot fix for DatePicker to emit date change only once.

nsndeck:
Fixed an issue with emitting date change twice. When month is changed from Dec to Jan DatePicker by default also increments year. Therefore {N} emits two changes which in some reason cannot handled by Angular2 change detection infrastructure.


nsndeck:
run ci

PR #1724, opened 2016-03-09
Change time only once due to Angular issues.

nsndeck:
A little change to the way `time` property is changed, since within Angular2 scenario too close (in time) changes are not handled correctly. The problem comes from the fact that android native time picker widget when minutes change from (0 to 59 and vice versa) also changes hour - which in previous implementation results in two changes of the `time` property.


dtopuzov:
run ci

PR #1730, opened 2016-03-09
Code file exports fixed

### LINKED ISSUE #1667: Tap Event Does Not Fire When Using codeFile In Different Directory
Yes.

### Tell us about the problem
`tap` event does not fire when the main XML's codeFile is in a different directory than the XML.  Page element's `loaded` event seems to work fine.

For example, if the main module is `views/launches/default` and it loads the XML.  The `Page` tag has a `codeFile` attribute pointing to `controllers/launches/default`.

The `tap` event exported from this codeFile never runs.  If the `codeFile` is in the same directory as the XML it works without any problems.

Both.

- CLI: 1.6.1

```
<!-- file: views/launches/default.xml -->
<Page codeFile="~/controllers/launches/default">
    <Button text="Tap" tap="tapp" />
</Page>
```

```
// file: views/launches/default.js

"use strict";

exports.tap = function(args) {
	console.log("tap event fires!"); // never fires...
};
```

enchev:
Fixed #1667

### DIFF
--- a/apps/tests/xml-declaration/xml-declaration-tests.ts
+++ b/apps/tests/xml-declaration/xml-declaration-tests.ts
@@ -166,6 +166,13 @@ export function test_parse_ShouldResolveExportsFromCodeFile() {
     TKUnit.assert((<any>page).customCodeLoaded, "Parse should resolve exports from custom code file.");
 }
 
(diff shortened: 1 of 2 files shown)


hdeshev:
:+1:

Issue #1599, opened 2016-02-20
Expose Android and iOS type definitions


losttime:
I'm working on updating a plugin to work with 1.6.0. I've discovered that the Android and iOS type definitions, which are used in the plugin, are no longer included when I run `npm install tns-core-modules`. From what I can tell, the grunt file is specifically excluding them from the build that gets published.

Can those type definition files get published to the npm repo, or is there another preferred way to utilize them (preferably without copying them into my working directory)?

losttime:
After further diffing I've realized that the Android and iOS weren't included with `tns-core-modules` before. That said, I still think it would be useful to have them exposed so plugin developers can extend them more easily.

enchev:
Hey @losttime,

You are absolutely right! We have plans to add support for automatic generation of TypeScript definitions when you add for example platform using our CLI. In this case you will have the correct definitions for the current platform (for example iOS 9.2), otherwise we need to ship different definition for every Android API level and iOS version.

Milewski:
by the way if i want to download those typing manually how could i do? where can i find them?

fealebenpae:
@Milewski: on iOS you can generate them today: http://docs.nativescript.org/runtimes/ios/how-to/Use-Native-Libraries#troubleshooting. We're still working to make this seamless on Android, though.

enchev:
Hey @ErjanGavalji can you help here?

ErjanGavalji:
@Milewski, for the time being you can also download them from [the respective tags in the NativeScript repo in GitHub](https://github.com/NativeScript/NativeScript/blob/v1.6.0/android17.d.ts). There is some automation done for creating these as a package, but still figuring out how to deal with the dependencies.

PeterStaev:
@enchev , from a plugin developer standpoint generating those from the CLI is not much of an option. I think the definitions of the latest stable version for each platform should be included either as a separate package or as part of the current tns-core-modules package.

ErjanGavalji:
@PeterStaev: Well, the thing is, (at least on android) the definitions are different depending on the minimum version the plugin supports. That said, it might make a better sense of having these generated, addressing the minimum platform version the plugin or application developer intends supporting. (By the way, the core team are already working on implementing the generator @enchev mentioned earlier).

How do you feel about that? These can easily get published as an NPM package, but won't they be misleading?

This decision must be carefully made for once published it would be bad for the community to just unpublish or stop publishing them...

PeterStaev:
[...]

PR #1736, opened 2016-03-10
Update iOS TypeScript definitions for the NativeScript modules

vbresults:
https://github.com/NativeScript/ios-runtime/issues/523

Logic for enum name changes:
* Strip `k` prefix from property name
* Compact property = property name - enum name
* If Compact property doesn't start with `[A-Za-z_]`, prefix with `t`

CI builds will fail unless the runtime maps or intercepts the old definitions (or we preserve the old definitions) -- not sure how this area is to be handled.

No, just updated definitions.

### DIFF
--- a/ios.d.ts
+++ b/ios.d.ts
@@ -24644,24 +24644,24 @@ declare enum CGColorRenderingIntent {
 	kCGRenderingIntentSaturation
 }
 declare enum CGColorSpaceModel {
-	kCGColorSpaceModelUnknown,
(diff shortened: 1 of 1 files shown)


enchev:
Hey @vbresults,

Instead of changing the current definitions maybe we should proceed with the ideas in this thread:
https://github.com/NativeScript/NativeScript/issues/1599

What do you think?

vbresults:
@enchev :+1:

PR #1739, opened 2016-03-11
event check

enchev:


ui/editable-text-base/editable-text-base.android.ts
```diff
@@ -84,7 +84,7 @@ export class EditableTextBase extends common.EditableTextBase {
         var editorActionListener = new android.widget.TextView.OnEditorActionListener({
             onEditorAction: function (textView: android.widget.TextView, actionId: number, event: android.view.KeyEvent): boolean {
                 var owner = that.get();
-                if (owner && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER) {
+                if (owner && event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER) {
```
hshristov: If there is no `event` should we `dismissSoftInput`?

enchev: I guess yes.

dtopuzov:
:+1:

PR #1761, opened 2016-03-15
Add source information on the view

### LINKED ISSUE #828: Create a 'source'-like property on the View class for debugging
Create a `source`-like property on the View class, that can be set to the location in JavaScript or XML location (file, line, column) that created the View.

It would be helpful for debugging.

We had a support issue where the ListView's `refresh` method was set to `false` in an app with multiple ListViews it was hard to track it down.

Related:
 - Better XML declaration image source errors reporting #503
 - Better migration and source/line/column when logging usage of deprecated members #502

PanayotCankov:
Closes #828
`ui/builder` will now attach source url, line and column to the created views.
view.toString() will now append `@file://file:line:column` to the view information.
The utils/debug.Source.get method can be used to extract SourceCode information from view instances.


hdeshev:
:+1:

Issue #1660, opened 2016-02-27
Feature Request: Add css ability for other properties


NathanaelA:
### Did you verify this is a real problem by searching 
yes

### Tell us about the problem
I'd like to be able to do
```css
StackLayout {
  orientation: horizontal;
}
```

### Which platform(s) does your issue occur on?
Both

### Please provide the following version numbers that your issue occurs with:
v1.6x (and before)

### Please tell us how to recreate the issue in as much detail as possible.
CSS does not appear to be able to set orientation on StackLayout.

Grid: Rows, Columns, row, column, colSpan, rowSpan would also be nice to be able to control in CSS

enchev:
Hey @NathanaelA,

At the moment we are trying to avoid CSS properties that are supported only for specific components. For example you cannot set orientation for Button.

NathanaelA:
Hey @enchev,

1. I realize the property is only used for some components; that doesn't stop its usefulness.  Right this minute instead of being able to control this via CSS, I have to create a function in the page it is on to handle this. ;-)

2. I realize orientation is not valid HTML CSS, but using it as a valid NS CSS should not matter.  As it is NS has several other CSS fields that aren't standard and that seems really silly to prefix a fully valid NS CSS value -- you really don't want to be constrained by HTML.  You need to do what is right for NS.

enchev:
Hey @atanasovg, @hshristov, @ligaz, @tjvantoll what do you think? Should we:
1. Expose CSS properties that will work only on some of the elements similar to browsers
2. Expose custom CSS properties valid for NativeScript only with a `-ns-` prefix similar to other vendor prefixes: https://developer.mozilla.org/en-US/docs/Glossary/Vendor_Prefix

atanasovg:
That's definitely an interesting discussion. As of now we **do** have properties that do not make sense for some views - for example `font-size` does not affect most of the views directly but rather the text-displaying ones. Still, if the property is inheritable - like the `font-size` one - then it is useful to apply it at root level and let interested descendants inherit and utilize the value. With that said:

- My vote goes for adding CSS properties regardless of whether they are applicable to each view type or not.
- IMO @NathanaelA has a point as to prefixing non-CSS standard properties with `-ns-`. Since {N} is the only CSS parsing engine such an additional separation seems artificial to me.

tjvantoll:
[...]

PR #1773, opened 2016-03-16
Plain component properties now can be applied from CSS

enchev:
Implemented #1660

### DIFF
--- a/ui/styling/css-selector.ts
+++ b/ui/styling/css-selector.ts
@@ -5,6 +5,7 @@ import * as trace from "trace";
 import * as styleProperty from "ui/styling/style-property";
 import * as types from "utils/types";
 import * as utils from "utils/utils";
+import {getSpecialPropertySetter} from "ui/builder/special-properties";
 
 var ID_SPECIFICITY = 1000000;
 var ATTR_SPECIFITY = 10000;
@@ -38,7 +39,7 @@ export class CssSelector {
     get expression(): string {
         return this._expression;
     }
-    
+
     get attrExpression(): string {
         return this._attrExpression;
     }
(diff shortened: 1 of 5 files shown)


hshristov:
:+1:

ui/styling/css-selector.ts
```diff
@@ -57,11 +57,15 @@ export class CssSelector {
 
     public apply(view: view.View) {
         this.eachSetter((property, value) => {
-            try {
-                view.style._setValue(property, value, observable.ValueSource.Css);
-            }
-            catch (ex) {
-                trace.write("Error setting property: " + property.name + " view: " + view + " value: " + value + " " + ex, trace.categories.Style, trace.messageType.error);
+            if (property in view) {
```
vakrilov: This will probably not handle `specialProperties`. For example `row`, `column` properties from GridLayout

vakrilov:
Perhaps verify that special properties work too.
Otherwise :+1:

atanasovg:
Guys, in my [comment here](https://github.com/NativeScript/NativeScript/issues/1660#issuecomment-191207755) I meant that we may enable component properties by following our current path - by creating new `style-property` within the `style` module - like `color`  for example: https://github.com/NativeScript/NativeScript/blob/master/ui/styling/style.ts#L937. Sorry for not being clear enough. 

This new implementation treats applying a property through CSS with a `local` value priority, while the implementation of the [style property](https://github.com/NativeScript/NativeScript/blob/master/ui/styling/style-property.ts#L85-L109) adds new modifiers like `visualState` and `css` to the `effectiveValue` calculation.

Besides, I remember us talking some time ago on this very topic and we decided against enabling each and every property applicable through CSS mainly due to some properties simply not fitting within the styling paradigm - like `text`, `value`, `items`, etc. Can we reiterate this and evaluate the pros and cons?

Ping @enchev, @hshristov, @vakrilov

enchev:
We've talked with @hshristov and @vakrilov about the inheritance and we've concluded that such non style properties are non inheritable and can be set as local values. For example `StackLayout` `orientation` cannot and should not be inherited. 

Personally I support this since it will give developers a lot more power.

NathanaelA:
@enchev - Point of confusion at least for me.   Are you saying **StackLayout.orientation** is not inheritable as in the property doesn't get inherited by its children; or are you saying that it isn't going to be settable at all via the css mechanism?

enchev:
@NathanaelA Not inheritable in the CSS world. No CSS cascading as well.

PR #1780, opened 2016-03-16
Fix the leave comments option of the build

ErjanGavalji:
The grunt file respected the --leavecomments option, used when
generating the How-To articles of the API Reference. This was broken with
the passthrough option to grunt-ts.


PanayotCankov:
:+1:

PR #1789, opened 2016-03-17
Go back with no animation on iOS using the stock "Back" button

### LINKED ISSUE #1787: Going back with the stock "Back" button on iOS always animates the transition.
When you navigate to a page on iOS by setting `animated: false` in the navigation entry, the forward navigation is not animated which is expected. However, if you go back with the stock iOS "Back" button located in the NavigationBar, the back transition is always animated regardless of the `animated: false` setting when coming to this page.

hamorphis:
Going back with the stock "Back" button on iOS will now respect the animated setting used when coming to the current page

Fixes #1787


hshristov:
:+1:

PR #1817, opened 2016-03-21
Disable custom transition test on Android 6.9 (API 23)

hamorphis:
http://stackoverflow.com/questions/33188485/resultindex-is-1-the-polygon-must-be-invalid-adter-addview


enchev:
:+1:

Issue #1811, opened 2016-03-20
Element.Animate vs Animation.Animate differences


NathanaelA:
## Please, provide the details below:
If you use Element.Animate; you lose the ability to cancel the animation, because it returns a promise from the .play() event and not the animation object.

### Did you verify this is a real problem by searching
yes

### Tell us about the problem
No animation cancelling allowed on a Element.Animate

### Which platform(s) does your issue occur on?
All

### Please provide the following version numbers that your issue occurs with:
1.7

---

I have developed a very simple "patch" that will monkey patch the Promise returned by the Element.animate (and any .then's) to have a .cancel on it (i.e. to match the same name as the animation.cancel function).   This will then allow you to do:

```
var x = Element.animate({blah}).then(function() { more blah });
if (something_occured) x.cancel();
```

Is this something that would be acceptable to be put into core, if so I'll convert it from JS into TS and push a pull request.   But if you are opposed to something "monkey" patching the promise to add a cancel function, then I won't bother doing the translating back into TS, and just add it to my own little bag of tricks.

hamorphis:
Go ahead and create a pull request with the new method. You can also add a test similar to `test_CancellingAnimation` located [here](https://github.com/NativeScript/NativeScript/blob/master/apps/tests/ui/animation/animation-tests.ts)

NathanaelA:
@hamorphis Added pull request #1821 -- I really hate TS & v8 today.  Literally had JS working 100% in 5 minutes.   Took multiple hours to get TS and v8 to play nicely.   Oy, this turned into a fiasco time wise; but it is done w/ tests.

PR #1821, opened 2016-03-22
This adds the ability to Cancel the Animation via the Promise returned

NathanaelA:
Fixed: https://github.com/NativeScript/NativeScript/issues/1811

Yes, updated the animation-tests to have a test to test this feature (and fixed the test for the cancelling of the normal animation object)

### Notes
When attempting to switch this from JS to TS -- I originally created a class that extended from the Promise; unfortunately; the v8 engine actually has an issue with an descendant promise class and will throw the error **Uncaught TypeError: #<AnimationPromise> is not a promise** when you try in create it.  So, the implementation I had to use was to create a dummy class (so that TS would allow the users to be able to use Promise.cancel() in TS ).  Then I create a new normal Promise; but then use casting to make TS think it is a AnimationPromise so that everywhere else you can use the new .cancel feature.   

The trick to make a Animation Promise cancellable is to use monkey patching of the promise after it is created so that it adds the .cancel to that specific instance of the promise, and so that any .then/.catch also wrap their returned promise with a .cancel prototype on them...   So it shouldn't matter how deep you go; each level will self wrap themselves.

### DIFF
--- a/apps/tests/ui/animation/animation-tests.ts
+++ b/apps/tests/ui/animation/animation-tests.ts
@@ -83,21 +83,69 @@ export var test_CancellingAnimation = function (done) {
     // <snippet module="ui/animation" title="animation">
     // # Cancelling animation
     // ``` JavaScript
-    var animation1 = label.createAnimation({ translate: { x: 100, y: 100 } });
+    var animation1 = label.createAnimation({ translate: { x: 100, y: 100}, duration: 500 });
     animation1.play()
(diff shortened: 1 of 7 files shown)


hamorphis:
I will take a look.

hamorphis:
:+1:

PanayotCankov:
:+1:

hshristov:
:+1:

Issue #1822, opened 2016-03-22
verticalAlignment "middle" and "center" crash the app on Android


alg:
We discovered that these two values crash the app on Android. I found that it happens because of bad case-statement that looks like this:

```
case VerticalAlignment.center || VerticalAlignment.middle:
```

The default case picks up everything that doesn't match and raises an error for an unrecognized option.

Affected files:

* ui/core/view.android.ts
* ui/core/view-common.ts
* ui/layouts/stack-layout/stack-layout.ios.ts

I'm submitting PR to resolve this in a minute.

PR #1823, opened 2016-03-22
Fixes incorrectly formed switch-case statements

alg:
Resolves issue #1822.

Replaces incorrectly formed switch-case statements with multiple options on the same case-line:

    case option1 || option2:
      ...
      break;

with correctly formed multiple case statements:

    case option1:
    case option2:
      ...
      break;

This bug prevented from using "middle" and "center" options for `verticalAlignment` style.

I can include tests, but it's a trivial change. Let me know if you need them.

### DIFF
--- a/ui/core/view-common.ts
+++ b/ui/core/view-common.ts
@@ -720,7 +720,8 @@ export class View extends ProxyObject implements definition.View {
                 childTop = top + marginTop * density;
                 break;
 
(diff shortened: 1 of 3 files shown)


enchev:
run ci

enchev:
run ci

enchev:
:+1:

PR #1827, opened 2016-03-22
Page.showModal API extended to accept page instance

vakrilov:
Added a `showModal` method overload that accepts page instance as a first argument


vakrilov:
run ci

enchev:
:+1:

PR #1830, opened 2016-03-23
Add `LayoutBase.clipToBounds` property

### LINKED ISSUE #1829: `LayoutBase.clipToBounds=false` does not work on Android [CAN"T FIX]
Add a child with negative margin to a layout. Part of the child should be rendered outside its parent layout and it is, but the part that is outside is clipped even if the layout's clipToBounds property is set to false.

hamorphis:
Resolves #1829


hshristov:
:+1:

hamorphis:
run ci

hdeshev:
:+1:

PR #1834, opened 2016-03-23
Allow updates to application CSS rules.

hdeshev:
Expose an `application.addCss` method and hide `application.cssSelectors`
from end-user typings. Update StyleScope's with new application selectors
on style application.

NOT updating existing controls/views unless reloaded, or forced (via a
change to className, etc).

This feature grew out of a demand for the Angular renderer, where we need
to apply component styles globally and have them take effect on all pages.

Part of a fix for the second issue discussed in NativeScript/nativescript-angular#115

### DIFF
--- a/application/application-common.ts
+++ b/application/application-common.ts
@@ -6,6 +6,8 @@ import cssSelector = require("ui/styling/css-selector");
 import * as fileSystemModule from "file-system";
 import * as styleScopeModule from "ui/styling/style-scope";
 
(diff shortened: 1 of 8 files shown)


hdeshev:
run ci

hamorphis:
:100:

nsndeck:
:+1:

vakrilov:
:+1:

PR #1837, opened 2016-03-24
Attempt to load at least jpg images for versions lower than 9

### LINKED ISSUE #1788: Difference in jpg resource syntax on iOS/Android
In order to use a `.jpg` file that lives in `App_Resources` on iOS I must use `res://my_file.jpg`, whereas on Android I need to use `res://my_file` (no extension). This forces me to needlessly fork my code to use an image with the same name.

You can find a live example in the [Groceries repo’s angular branch](https://github.com/NativeScript/sample-Groceries/tree/angular). See https://github.com/NativeScript/sample-Groceries/blob/5f3c416567ab77b605c90775708a4ae5fcb758e8/app/platform.android.css#L2 and https://github.com/NativeScript/sample-Groceries/blob/5f3c416567ab77b605c90775708a4ae5fcb758e8/app/platform.ios.css#L2.

Thanks.

enchev:
Related to #1788

### DIFF
--- a/image-source/image-source.ios.ts
+++ b/image-source/image-source.ios.ts
@@ -11,7 +11,7 @@ export class ImageSource implements definition.ImageSource {
     public ios: UIImage;
 
     public loadFromResource(name: string): boolean {
-        this.ios = UIImage.imageNamed(name);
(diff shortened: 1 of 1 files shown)


vchimev:
:+1:

PR #1838, opened 2016-03-24
ShownModally fired when the dialog is actually shown in Android

vakrilov:
Fix: The `shownModally` event should be fired when the dialog is shown in Android.


hshristov:
:+1:

Issue #1820, opened 2016-03-21
all topmost().navigate instances stop working after Admob Interstitial close


calebeaires:
When Admob Interstitial is called and after user close the ad, all ```frameModule.topmost().navigate()``` stop working.

This problem occurs on [Nativescript blog implementation](https://www.nativescript.org/blog/details/monetize-your-nativescript-apps-with-admob-(part-2---android)) and nativescript-admob plugin too. from the blog content.

Example code I am using:
```
exports.buttonClicked = function(){
    frameModule.topmost().navigate({
        moduleName: "views/home",
        animated: true,
        transition: {
            name: "slideTop"
        }
    });
}
```
Platform: Android 5;
Nativescript: 1.7;

NickIliev:
@calebeaires 
Yes indeed this is existing issue and the problem is scheduled for fixing. We think it was introduced with the #1647 PR. Thank you for noticing this problem - we will do our best to resolve it in a timely matter

calebeaires:
Very thanks. 

Is there some workaround code to work with before you fix it?

PR #1839, opened 2016-03-24
Resumed activities should be marked as foreground

PanayotCankov:
Fixes/Implements #1820 .

### Does your pull request have [unit tests]
It can't have unit tests, since it involves sending the app to the background and starting multiple activities.

### DIFF
--- a/application/application.android.ts
+++ b/application/application.android.ts
@@ -108,13 +108,11 @@ function initEvents() {
             androidApp.paused = false;
             androidApp.foregroundActivity = activity;
 
-            if (activity === androidApp.foregroundActivity) {
-                if (typedExports.onResume) {
-                    typedExports.onResume();
-                }
-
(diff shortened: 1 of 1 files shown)


hshristov:
:+1:

application/application.android.ts
```diff
@@ -106,6 +106,7 @@ function initEvents() {
 
         onActivityResumed: function (activity: any) {
             androidApp.paused = false;
+            androidApp.foregroundActivity = activity;
 
             if (activity === androidApp.foregroundActivity) {
```
NathanaelA: If the fix in line 108 is to set foregroundActivity = activity; then why check it in 110?

PanayotCankov: @NathanaelA You are correct. I also think it would be the correct behavior to call "onResume" of the module right? I guess we should just delete the `if` and preserve its body?

calebeaires:
I add this line, then the problem gone away. It`s fixed.

PR #1841, opened 2016-03-24
The 'grunt inplace' build generate source maps

PanayotCankov:
Allows an app to be debugged using the TS modules.
Allows breakpoints to be set in TS modules source code, locals to be inspected in modules TS etc.

### Does your pull request have [unit tests]
No, this affects builds, not the delivery.


hshristov:
:+1:

hdeshev:
:+1:

PR #1842, opened 2016-03-24
fix(es6): objectcontstructor assign

NathanWalker:
Silences TypeScript compiler warning:
```
error TS2339: Property 'assign' does not exist on type 'ObjectConstructor'
```

This is especially problematic for `nativescript-angular` projects that utilize various es6 features and including the `es6-shim` definition file is not an option as it creates numerous duplicate warnings due to same definitions found throughout tns-core-modules (es6-promise.d.ts, es-collections.d.ts, etc.).

A better fix down the line would be to remove all the individual es6 definitions (es6-promise.d.ts, es-collections.d.ts, etc.) in this repo and replace with the single `es6-shim` found here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/es6-shim/es6-shim.d.ts

### DIFF
--- a/es6.d.ts
+++ b/es6.d.ts
@@ -1,1 +1,9 @@
 declare var Symbol: any;
+
+interface ObjectConstructor {
+    assign(target: any, ...sources: any[]): any;
+    is(value1: any, value2: any): boolean;
+    setPrototypeOf(o: any, proto: any): any;
+}
+
+declare var Object: ObjectConstructor;


NathanWalker:
I would take the step of removing the individual definitions and replacing with `es6-shim` but it would require more knowledge about the build/publishing setup that I am unsure of at the moment.

hdeshev:
I'm not sure that we should use es6-shim instead of our own typings. I've been trying to play it safe and make our typings as compatible with es6-shim as possible. There is some talk (angular/angular#5755) about removing es6-shim from the Angular dependencies, and it would be really sad if we introduce it to NativeScript right before Angular removes it :)

hdeshev:
run ci

hdeshev:
I also need to verify if the {N} runtimes actually support `Object.assign` and `Object.is`. Other than that I like the PR, and I think we should merge it. :+1:

hdeshev:
Er, tslint checking fails - complains about some indentation issue.

NathanWalker:
@hdeshev Thanks for pointing that out. Maybe we run with [core-js](https://github.com/zloirock/core-js) then :+1: 

In meantime, yeah this might help ease some warnings for some. If swapping everything to use `core-js` is in near future (like next week or 2), then I would ditch this PR, and just do that. I might could tackle the swap, just need a lil insight on how the `tns-core-modules` build is setup.

hdeshev:
I'm not sure about the move to core-js. It's a huge library. I think we should just merge the updated typings for ObjectConstructor at this point (the API is supported by both the Android and iOS runtimes).

By the way, where did you get that compile error? Is it in some part of the Angular codebase? I'd never seen it before.

Issue #1608, opened 2016-02-22
Add support for CSS 3 animations & transitions


tzraikov:
This feature will allow adding animations directly through CSS code. Currently adding animations is possible only with JS(TypeScript) code. Making this feature available will separate the animation logic from the application one and the process of adding new animations will be easier. Moreover, it will be theoretically easy to use existing CSS3 animation libraries (e.g. [animate.css](https://daneden.github.io/animate.css/)). Adding support for CSS pseudo selectors will streamline the process of triggering animations by adding visual states (e.g. button1:highlighted).

In order to add initial support for CSS animations, the following CSS properties should be supported:
 
- [x] animation
- [x] animation-name
- [x] animation-delay
- [x] animation-duration
- [x] animation-iteration-count
- [x] animation-timing-function
- [x] @keyframes

The following animation properties could be implemented at later stages:
- [x] animation-direction
- [x] animation-fill-mode

In addition to animation properties, the following CSS property should be supported:
- [x] transform

Currently only the following properties are animatable in {N}:
- [x] opacity
- [x] backgroundColor
- [x] scaleX
- [x] scaleY
- [x] translateX
- [x] translateY
- [x] rotate
- [x] rotateX
- [x] rotateY

Example: 

```CSS
.button1 {
    background-color: white;
}

.button1:highlighted {
    animation-duration: 0.75s;
    animation-name: test;
}

@keyframes test {
from { background-color: red; }
  to { background-color: white; }
}
```

CSS animations can be improved further by implementing #1894 and #1895

mikeres0:
+1

myninka:
#1767

PanayotCankov:
The following animation:
```
@keyframes intro-logo-bg-enter {
    from {
        transform: scale(10, 10);
    }
    to {
        transform: translate(0, -40);
    }
}
#intro-logo-bg {
    animation-name: intro-logo-bg-enter;
    animation-duration: 3;
    animation-timing-function: ease-out;
    animation-fill-mode: forwards;
}
```
Must leave the object after the animation scaled 1, 1 and translated 40 pixels down.
Currently it animates it properly but at the end the object jumps to scale 10 10 and probably 40 pixels down.

PanayotCankov:
Consider allowing multiple animations to be chained using the `animation:` shorthand property, such as:
```
@keyframes anim1 {
    to {
         background-color: red;
    }
}
@keyframes anim2 {
    from {
        background-color: red;
    }
    to {
[...]

PR #1843, opened 2016-03-24
implemented CSS animations

tzraikov:
#1608


apps/tests/ui/animation/css-animation-tests.ts
```diff
@@ -0,0 +1,316 @@
...
+        "animation-duration: 4s; " +
+        "animation-timing-function: ease-in; " +
+        "animation-delay: 1.5; " +
+        "animation-iteration-count: 10; " +
+        "animation-direction: reverse; " +
+        "animation-fill-mode: forwards; " +
+        " }";
+    let animation = createAnimationFromCSS(css);
+
+    TKUnit.assert(animation.name === "first");
```
hshristov: assertEqual produce better error message

apps/tests/ui/view/view-tests-common.ts
```diff
@@ -201,6 +201,7 @@ export var test_addView_WillThrow_IfView_IsAlreadyAdded = function () {
         }
 
         TKUnit.assert(thrown);
+        views[1]._removeView(newButton);
```
hshristov: If this test throws - view won't be added to parent, why removing it here?

ui/animation/keyframe-animation.d.ts
```diff
@@ -0,0 +1,94 @@
+declare module "ui/animation/keyframe-animation" {
+
+import view = require("ui/core/view");
+
+    export class KeyframeDeclaration {
```
hshristov: Do we need class here or interface will work too?
Same for KeyframeInfo.

ui/animation/keyframe-animation.ts
```diff
@@ -0,0 +1,193 @@
...
+            }
+        }
+        for (let index = 1; index < length; index++) {
+            let a = animations[index];
+            if (a["curve"] === undefined) {
+                a["curve"] = info.curve;
+            }
+        }
+        if (animations.length > 0) {
+            animations[0]["duration"] = 0.001;
```
hshristov: Do we need this duration here?

ui/animation/keyframe-animation.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    private animate(view: view.View, index: number, iterations: number) {
+        if (index === 0) {
+            let animation = this.animations[0];
+            let modifier = animation["valueSource"];
+
+            if (animation["backgroundColor"] !== undefined) {
+                view.style._setValue(style.backgroundColorProperty, animation["backgroundColor"], modifier);
+            }
+            if (animation["scale"] !== undefined) {
```
hshristov: Do we always have to set both 'x' and 'y'?
What will happen if I set only 'x'? Will this reset my 'y' value?

hshristov:
:+1:

PanayotCankov:
:+1:

tzraikov:
run ci

Issue #1757, opened 2016-03-14
Support letter-spacing in CSS


tjvantoll:
Low priority, but I would like {N} to support the `letter-spacing` CSS property. At the moment I’m manually calling `setLetterSpacing()` on Android, and trying (and failing) to use this technique for iOS http://stackoverflow.com/questions/20580736/how-to-increase-the-character-spacing-in-uilabel.

And, just as an idea, this might be a good first issue for someone looking to contribute to {N}.

tjvantoll:
Just fyi this is how I’m currently accomplishing this in Groceries: https://github.com/NativeScript/sample-Groceries/blob/be06c99c138ff0c214576ff71066be985955c967/app/views/login/login.js#L54-L68

lawrencetaur:
Line-height would be a great addition as well. refer: #1664

sitefinitysteve:
I would argue not low priority as these seemingly "fluffy" features all go towards making nice looking apps, and in this case groceries is a showcase app so it would go towards reducing the code complexity for a new user looking at it.

vchimev:
There is an issue with the Label, TextField and TextView elements on iOS that *text-transform* does not apply when *letter-spacing* is set. For example:
```xml
<Label text="labelLabel" style="text-transform: uppercase; letter-spacing: 10;" />
<TextField text="textField" style="text-transform: uppercase; letter-spacing: 10;" />
<TextView text="textView" style="text-transform: uppercase; letter-spacing: 10;" />
```
It works as expected for Button.

vchimev:
The letter-spacing property for Android is supported on API levels >= 21.

hamorphis:
**Have in mind that letter spacing should be specified in EM units, not pixels!**

https://github.com/NativeScript/NativeScript/issues/2528

priyangamani:
letter-spacing in CSS not supported ios..

PR #1855, opened 2016-03-28
Letter spacing CSS support added

enchev:
Implemented #1757

### DIFF
--- a/ui/text-base/text-base-styler.android.ts
+++ b/ui/text-base/text-base-styler.android.ts
@@ -108,6 +108,23 @@ export class TextBaseStyler implements style.Styler {
         utils.ad.setWhiteSpace(view._nativeView, enums.WhiteSpace.normal);
     }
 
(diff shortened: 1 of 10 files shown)


enchev:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

ui/text-base/text-base-styler.android.ts
```diff
@@ -108,6 +108,19 @@ export class TextBaseStyler implements style.Styler {
         utils.ad.setWhiteSpace(view._nativeView, enums.WhiteSpace.normal);
     }
 
+    // letter-spacing
```
hshristov: These are API 21 methods so we should check if methods is available at runtime before using them.

hshristov:
:+1:  After check for API 21

enchev:
run ci

PR #1856, opened 2016-03-28
Platform module refactoring

vakrilov:
Refactored `platform` module structure. Introduced `Screen` and `Device` interfaces.
Now the module export instance of these interfaces instead of classes with static methods.


enchev:
:+1:

PanayotCankov:
:+1:

PR #1860, opened 2016-03-29
Text transform and decoration updated on binding context changed

### LINKED ISSUE #1859: CSS text-transform is not correctly applied in a virtualized ListView
If you have a Label inside a ListView.itemTemplate and this Label has its text-transform set to uppercase from CSS, only the labels that are initially realized in the ListView get their text with capital letters. If you scroll down, the newly realized labels do not capitalize their text.

enchev:
Fixed #1859


hamorphis:
:100:

PR #1861, opened 2016-03-29
Work around a webpack bundling fail in transitions by using full imports

hdeshev:
It seems Webpack can't discover aliased modules that relative-import other
aliased modules e.g. "ui/transition", aliased to
"ui/transition/transition.android", not discovering "./slide-transition"
which is aliased to "ui/transition/slide-transition.android.js"


PanayotCankov:
:+1:

hdeshev:
run ci

PR #1868, opened 2016-03-30
No iOS UINavigationBar animation for the first Page

PanayotCankov:
The nativescript-marketplace-demo has an ActionBar defined on its first page and hideActionBar set to true. When the app is launched the action bar appears for milliseconds displaying its animation when disappearing. It sounds reasonable to disable the action bar animations when showing/hiding when the main page is still loading, and only use animation for subsequent navigation or dynamically setting hideActionBar=true/false after onLoaded.


PanayotCankov:
run ci

PR #1877, opened 2016-03-31
Zone-aware versions of certain APIs: setTimeout/setInterval mostly.

hdeshev:
Instead of waiting for zone.js to patch our global declarations and break
the lazy module loading optimizations there, we'll let it patch stuff
before "globals" gets loaded, so that it doesn't touch (and break) the code
there.

Of course, that requires that functions that need legit patching need to be
made aware of the zone, or get patched later on. The global.zonedCallback
function makes that easy, and we use it for setTimeout/setInterval.

TODO: Think this over again and identify other functions that need zone awareness.
//cc @vakrilov, @nsndeck

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -146,6 +146,10 @@ declare function setInterval(callback: Function, milliseconds?: number): number;
  */
 declare function clearInterval(id: number): void;
 
+//@private
+declare function zonedCallback(callback: Function): Function;
+//@endprivate
+
 declare class WeakRef<T> {
     constructor(obj: T);
     get(): T;
--- a/globals/globals.ts
+++ b/globals/globals.ts
@@ -29,6 +29,14 @@ global.loadModule = function(name: string): any {
     }
 }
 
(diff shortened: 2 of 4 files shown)


vakrilov:
:+1: 
Looks safe to me.
About other candidates fro wrapping:
  * [XHR's `onload` and `onerror`](https://github.com/NativeScript/NativeScript/blob/1c7c4813c7157f084ad0179472c682006b586b36/xhr/xhr.ts#L17-L18).
  * Should we wrap all Promises - that should take care of `http` and `fetch` modules. Actually it will take fix `xhr` as well as it uses the `http`

hdeshev:
run ci

hdeshev:
It turns out zone.js already patches our Promise implementations. I think it's best if we leave the XHR API alone -- last time zone.js patched it, it broke our unit test runner. :frowning: 

I tested this PR with HTTP requests, and other sample projects, and they seem fine. Tests run successfully too.

Issue #1905, opened 2016-04-07
Visual states not working properly when the animation is canceled in Android


tzraikov:
1. Add a highlighted state animation and set it to a button:

.button {
    background-color: green;
}

.button:highlighted {
    animation-name: highlight;
    animation-duration: 2s;
    animation-fill-mode: forwards;
}

@keyframes highlight {
    from { background-color: yellow; }
    to { background-color: red; }
}

2. Play the animation by touching the button.
3. Touch up before the animation ends. The button will remain in some undetermined state. The correct behavior is to reset its background to green (like in iOS).

PR #1906, opened 2016-04-07
Fixed: Visual states not working properly when the animation is canceled

tzraikov:
#1905

### DIFF
--- a/ui/animation/animation.ios.ts
+++ b/ui/animation/animation.ios.ts
@@ -15,14 +15,28 @@ let FLT_MAX = 340282346638528859811704183484516925440.000000;
 
 declare var CASpringAnimation:any;
 
-class AnimationInfo
-{
+class AnimationInfo {
     public propertyNameToAnimate: string;
     public fromValue: any;
     public toValue: any;
     public duration: number;
     public repeatCount: number;
     public delay: number;
+}
+
+interface PropertyAnimationInfo extends common.PropertyAnimation {
+    _propertyResetCallback?: any;
+    _originalValue?: any;
+}
+
+interface AnimationDefinitionInternal extends definition.AnimationDefinition {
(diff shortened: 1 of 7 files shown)


ui/animation/animation.ios.ts
```diff
@@ -144,9 +144,10 @@ export class Animation extends common.Animation implements definition.Animation
         let i = 0;
         let length = this._mergedPropertyAnimations.length;
         for (; i < length; i++) {
-            (<UIView>this._mergedPropertyAnimations[i].target._nativeView).layer.removeAllAnimations();
-            if ((<any>this._mergedPropertyAnimations[i])._propertyResetCallback) {
-               (<any>this._mergedPropertyAnimations[i])._propertyResetCallback((<any>this._mergedPropertyAnimations[i])._originalValue);
+            let propertyAnimation = this._mergedPropertyAnimations[i];
+            propertyAnimation.target._nativeView.layer.removeAllAnimations();
+            if ((<any>propertyAnimation)._propertyResetCallback) {
```
hshristov: why do we cast to `any`?

PanayotCankov:
:+1:

ui/animation/keyframe-animation.ts
```diff
@@ -98,6 +100,16 @@ export class KeyframeAnimation {
...
+            if (this._currentAnimation && this._currentAnimation.isPlaying) {
+                this._currentAnimation.cancel();
+            }
+            this._isPlaying = false;
+            this._rejectAnimationFinishedPromise();
+        }
+    }
+
     public play(view: view.View): Promise<void> {
         if (this._isPlaying) {
```
hshristov: Do we need to throw exception here? Can we just return if running?

hshristov:
:+1:

PR #1913, opened 2016-04-08
Style properties are now applied on Buttons

### LINKED ISSUE #1902: Css text properties not applying on Button when no other text-element is used in page
When trying to set **color** , **font-size** , **font-weight** and **font-style** on **Button** styles are not applied if no other text-based element is used (Label, TextField, ect.) in the current xml file. 
All other css-properties are working.

_Example:_

in app.css (or main-page.css)
```
Button {
    color:red; 
    font-style:italic;
    font-size: 50;
    font-weight: bold;
}
```

in main-page.xml
```
<!-- in this case Css works
<Page xmlns="http://schemas.nativescript.org/tns.xsd">
  <StackLayout>
    <Label text="Tap the button" />
    <Button text="TAP" tap="{{ onTap }}" />
  </StackLayout>
</Page>-->


<!-- here Css doesn't work
<Page xmlns="http://schemas.nativescript.org/tns.xsd">
  <StackLayout>
    <Button text="TAP" tap="{{ onTap }}" />
  </StackLayout>
</Page>-->

```

related to [#1891](https://github.com/NativeScript/NativeScript/issues/1891)

hamorphis:
Resolves #1902

### DIFF
--- a/ui/button/button.android.ts
+++ b/ui/button/button.android.ts
@@ -1,6 +1,8 @@
 ﻿import common = require("./button-common");
 import utils = require("utils/utils")
 import dependencyObservable = require("ui/core/dependency-observable");
+import style = require("ui/styling/style");
+import { TextBaseStyler as TBS } from "ui/text-base/text-base-styler";
 
 global.moduleMerge(common, exports);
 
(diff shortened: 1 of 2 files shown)


hshristov:
:+1:

PR #1930, opened 2016-04-11
Run tests in travis-CI

ErjanGavalji:
Add android definitions to check if they will get downloaded
Set the language to android
Disable the CLI install prompt
Add buildOnly- and runOnly- commands
Update the version of the packages used
Install android 23 to make the app build
Build the test app with the @next android runtime
Specify oracle java version 8
Use switcher to switch to oraclejdk8
Make the final build a single step so that it breaks on first failure
Reorder attempting to use higher nodejs
Use nodejs 5.10.1
Detect test failure
Decrease verbosity
Unignore the build folder recursively
Remove the redundant node-version messages
Run the tests with Api level 21

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@ package/
 !css-value/**/*.*
 !fetch/**/*.*
 !apps/TelerikNEXT/lib/**/*.*
-!build/*.*
+!build/**/*.*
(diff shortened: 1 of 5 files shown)


dtopuzov:
:+1:

PR #1960, opened 2016-04-13
Aligned the rotate animation behaviour between Android and iOS

tzraikov:
Animation behavior was different in Android and iOS when applying a rotate animation on 360 degrees. 
In Android you should reset the rotate property to 0 in order to reapply the animation. In iOS this is not necessary. The correct behavior is the Android one.


hshristov:
:+1:

PR #1964, opened 2016-04-14
SegmentedBar item titles wrapping fixed. Now similar to iOS.

### LINKED ISSUE #1963: Default SegmentedBar items to textwrap false, expose option to configure
Yeppers

### Tell us about the problem
Default segmented bar on Android is just really bad looking... I find it hard to use it without one of the tabs wrapping.  Would be better to not wrap by default, and enable via a property on the segmented bar or on each SegmentedBarItem

image

Android

- CLI: 1.7.1
- Cross-platform modules: 1.7.1
- Runtime(s): 
```  js
  "tns-ios": {
      "version": "1.7.0"
    },
    "tns-android": {
      "version": "1.7.1"
    }
```
- Plugin(s): None
``` xml
            <SegmentedBar selectedIndex="{{ filterIndex }}">
                <SegmentedBar.items>
                    <SegmentedBarItem title="Incomplete" />
                    <SegmentedBarItem title="Complete" />
                    <SegmentedBarItem title="Something Else" />
                    <SegmentedBarItem title="All" />
                </SegmentedBar.items>
            </SegmentedBar>
```

enchev:
Fixed #1963

### DIFF
--- a/apps/ui-tests-app/tab-view/nowrap.xml
+++ b/apps/ui-tests-app/tab-view/nowrap.xml
@@ -1,0 +1,12 @@
+<Page>
+      <StackLayout>
+            <SegmentedBar>
+                  <SegmentedBar.items>
+                        <SegmentedBarItem title="Incomplete" />
+                        <SegmentedBarItem title="Complete" />
+                        <SegmentedBarItem title="Something Else" />
+                        <SegmentedBarItem title="All" />
+                  </SegmentedBar.items>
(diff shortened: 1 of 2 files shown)


hshristov:
Can we expose these settings so that user could change them?

enchev:
I'm not sure if wrapping should be allowed at all in this component.

ErjanGavalji:
:+1:

PR #1967, opened 2016-04-14
Fix: editable not working when keyboardType is set

### LINKED ISSUE #1966: TextView editable=false does not work when keyboardType is set
The `editable = false` is not respected when `keyboardType` is set on android textView:
```
<TextField keyboardType="email" editable="false" ></TextField>
```

Reported by @dreambo8563 in NativeScript/nativescript-angular#170

vakrilov:
Resolves #1966


enchev:
:+1:

Issue #515, opened 2015-07-29
CSS Style Property "z-index"


bfattori:
I know that `z-index` is a web specific property, but I was thinking today about using an `AbsoluteLayout` to handle the positioning of elements in a view. One of the things that I'm thinking could potentially be a problem is z-indexing. Defining the items in source-order will probably function as a stop-gap, but I was wondering if I were to position a floating element higher in the order, and then a background element lower down in the order, would it overlap the floating element?

I haven't tried this myself - just thinking out loud.

rclai:
I'm trying to create a dropdown menu in iOS by using an `AbsoluteLayout` and using the animation framework to make it show, however, I cannot get it to show up in front of everything. How do I make it go in front of things?

stevetayloruk:
+1

hshristov:
I think z-index could be added but it won't help with dropdown menu. You can implement it now using AbsoluteLayout added as a last child in the root view. Then change its visibility.

stevetayloruk:
@rclai please see here: #1222

patrickseda:
+1 ... Having z-index is **critical** for more advanced UI. Most "business" apps won't ever need it, but when you get into dynamically-generated or gesture-based controls, it is a _must have_.

terreb:
z-index doesn't work in 2.0.0. Please try this xml:

`<Page xmlns="http://schemas.nativescript.org/tns.xsd" >
    <AbsoluteLayout>
      <GridLayout zIndex="99" height="100" width="100" backgroundColor="red"></GridLayout>
      <GridLayout zIndex="1" height="200" width="200" backgroundColor="green"></GridLayout>
    </AbsoluteLayout>
</Page>`

The red rectangle should be on top, but it's hidden behind the green one or am I doing something wrong?

FireSt-dead:
This works in 2.2.0 with angular 0.3.0:
```
<Image class="asignee" *ngFor="let assignee of issue.assignees; let i = index" [index]="i" [src]="assignee.avatar_url" [style.zIndex]="-i"></Image>
```

prolink007:
I am still not able to get this to work. Why is this issue closed?

sittingbool:
I agree, where is the problem in that?

mudlabs:
At current zIndex appears to be constrained to its parent container. Not the page.

```xml
<!-- Just Ignoring that the desired layout can be achieved by other means.  -->
<Page>
    <Label id="topZ" text="Be on top" />
[...]

PR #1993, opened 2016-04-19
z-index support added

enchev:
Implemented #515

### DIFF
--- a/ui/core/view.android.ts
+++ b/ui/core/view.android.ts
@@ -700,6 +700,23 @@ export class ViewStyler implements style.Styler {
     private static getTranslateYProperty(view: View): any {
         return view.translateY;
     }
+    
+    // z-index
+    private static getZIndexProperty(view: View): any {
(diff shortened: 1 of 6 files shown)


enchev:
run ci

vakrilov:
👍 
However we should state explicitly that this will not work on Android API < 21

ui/core/view.android.ts
```diff
@@ -700,6 +700,23 @@ export class ViewStyler implements style.Styler {
...
     }
+    
+    // z-index
+    private static getZIndexProperty(view: View): any {
+        return view.android.getZ ? view.android.getZ() : 0;
+    }
+
+    private static setZIndexProperty(view: View, newValue: any) {
+        if (view.android.setZ) {
+            view.android.setZ(utils.layout.toDeviceIndependentPixels(newValue));
```
hshristov: Converting toDIP - will probably result to non Int value. We need to remove it.

vchimev:
run ci

PR #1998, opened 2016-04-20
Fix broken navigation after clear history with transition

### LINKED ISSUE #1948: Navigation stops working after navigating with clearHistory and a transition
Hi,

After navigate to page using clearHistory = true the navigation not work anymore. 

**Steps:**
1 - Open app
2 - Navigate to page two  using { clearHistory: false }
3 - Navigate to page one using { clearHistory: true }
4 - Navigate to page two again - navigaton not work anymore

I'm using last version android@next. This bug has been reported?

hamorphis:
Resolves #1948


hamorphis:
run ci

PR #2001, opened 2016-04-20
Fix: Android pages disappear before their exit transition has finished

### LINKED ISSUE #2000: Android pages disappear before their exit transition has finished
For example, if you navigate with a slide transition, you can see how the new page slides into view. The old page disappears immediately instead of gradually sliding out of view.

hamorphis:
Resolves #2000


ui/transition/transition.android.ts
```diff
@@ -13,6 +13,7 @@ var ENTER_POPEXIT_TRANSITION = "ENTER_POPEXIT_TRANSITION";
 var EXIT_POPENTER_TRANSITION = "EXIT_POPENTER_TRANSITION";
```
hshristov: Can we remove these fields and use private/local interface? This way you will get some intellisense.

enchev:
:+1:

hshristov:
:+1:

Issue #1797, opened 2016-03-18
Feature: Allow <> inside of strings in Declarative UI XML


NathanaelA:
### Tell us about the problem
I fully 100% realize this is AGAINST the XML specifications; and that for XML to be properly formed you are supposed to use &amp;gt; and &amp;lt; instead of > and <.   However, since this XML parser is only used for NativeScript; I would highly recommend we allow these characters to be treated as valid because of the Binding system we use.  I think everyone has been bitten at least once by doing something like:   
**"x > 1 ? 'hi' : 'hello'"**    
I've been doing NS for almost a year; and I still get bit by this issue.   It is such a waste of time trying to figure out why your xml is no longer working.

I realize this is a polarizing issue; and &amp;gt; / &amp;lt; would STILL work.     All that would happen is the rules would be relaxed in strings for < > characters.

Please note I am volunteering to do this patch; but I do not want to waste my time if it will not be accepted.   So let me know if this is something that the NS team would be willing to accept into core.

sitefinitysteve:
I am in favour, I get hit by it as well...both would still "work" though right, if someone wants to stay strict they can?

NathanaelA:
@sitefinitysteve yes, the strict &amp;gt; and &amp;lt; would still work just as usual -- we don't want to break any existing code.  :grinning:

hamorphis:
Duplicate of https://github.com/NativeScript/NativeScript/issues/1118

NathanaelA:
@hamorphis the issue you linked to is also closed -- my issue is slightly different in that I was **volunteering** to do this feature; are you saying that if I was willing to do this feature (and it was just as performant); it would not be accepted in core?   Or can we re-open this and get a consensuses for the team on if this would be something accepted into core?

NathanaelA:
@hamorphis @valentinstoychev 

I was looking into #1937 (\<import\> feasibility) so I was in easysax and realized just how easy it would be to test this issues.   So I have a 10 or so line patch to easysax that allows the <> to be inside either set of quotes.  Benchmarks show that on a 100meg file looped 20 times; before the patch averages 27.79 seconds.  After the patch average time is: wait for it...  :grinning:  **26.67**...   By doing the '>' search a different way; I chopped a whole second off AND it now supports everything inside both single and double quotes.   On the little tiny templates like NS normally deals with the time change is negligible (which is why I went to a 100mb xml file, so I could actually see a change).

[...]

PR #2004, opened 2016-04-20
Add ability to add & < > inside XML quoted strings

NathanaelA:
Fixes/Implements #1797 

Yes, updated the node-tests/test-xml.ts test suite to now test this feature.

---

Please note; One of the issues that was raised when I proposed this feature was that it might be slower.   

I have tested the speed of this using a 112 meg XML file, in a parsing loop of 20 times, and the speed went from 27.79 seconds to 26.67 seconds.   On average it saved slightly over 1 second each time I ran the tests.   This implementation is FASTER than the original and now supports any characters inside quoted strings.   

  If you would like a copy of the benchmark & xml file; feel free to ask and I can drop them somewhere as I didn't figure we would want to add a 112 meg xml file added to the repo...  :grinning: 

   To keep this patch simple; I didn't do any other optimizations beyond just the quote loop code.  There are a couple other places that we could probably enhance the easysax parser and speed it up even more....

### DIFF
--- a/js-libs/easysax/easysax.js
+++ b/js-libs/easysax/easysax.js
@@ -462,7 +462,7 @@ EasySAXParser.prototype.parse = function(xml) {
     , elem
     , tagend = false
     , tagstart = false
-    , j = 0, i = 0
+    , j = 0, i = 0, k = 0, len
     , x, y, q, w
     , xmlns
     , stopIndex = 0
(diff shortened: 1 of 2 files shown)


enchev:
run ci

enchev:
:+1:

vakrilov:
:+1:

PR #2013, opened 2016-04-22
Setting bindingContext to null fails in some cases

PanayotCankov:
Our QSF list view example selection fails when navigating away from the example when the bindingContext is set to null, but I couldn't reproduce it in a clean unit test.

We can create automated test in the QSF


vakrilov:
:+1:

hshristov:
:+1:

PR #2015, opened 2016-04-22
Remove redundant comments from the gruntfile

ErjanGavalji:
No related issue, this is a small refactoring only.

No tests, the change is in the build script itself.


enchev:
:+1:

Issue #1563, opened 2016-02-16
Enable modules snapshot for Android


atanasovg:
# The Story
In a nutshell - since in Android application loading time is one of the areas that need improvement,
we've made a POC implementation that takes advantage of V8's [startup snapshots](http://v8project.blogspot.bg/2015/09/custom-startup-snapshots.html) feature to see what may be achieved with it. The results are quite promising and we may gain literally more than **1 second !** of an improvement by saving all the modules.

# Technical Details
Due to the V8 API specifics, we need to bundle the entire modules JS into one single file and pass it to the [V8::CreateSnapshotDataBlob](http://v8.paulfryzel.com/docs/master/classv8_1_1_v8.html#a05f6c30582cc654b5fb09fab98a0dbbb) method. What V8 does when making the snapshot is to parse, compile and run this script into a new Context and then to save the state of the heap into a binary representation. Then, upon next application runs, this binary file may be used to load the whole representation of the modules directly within memory. 

Here is my proposal for taking advantage of this feature:

# Distribute pre-generated BLOBs
The snapshot is CPU-architecture dependent. Hence, if we want to distribute pre-generated version of the snapshot we will need to package three files, saved against the three available architectures - `armeabi-v7a`, `x86` and `arm64-v8a`. The average size of one file is ~3 MB but it compresses very well and an archived version is about 400 KB. 

**This is the most efficient way performance-wise. It adds further optimization by skipping the extraction of numerous JS files initially**

# Tasks:
- [x] Ensure modules are snapshot-ready. This is already done - https://github.com/NativeScript/NativeScript/pull/1407
- [x] Write a custom bundler that puts all the modules content within a single JS file. I've already done a custom Node task for the POC but we may go with [webpack](https://webpack.github.io/) for example (as @hdeshev) suggested.
- [x] Think how to automate snapshot generation against the three CPU architectures.
- [x] Think how to distribute (package) these three binary files. For example we may have two packages - `tns-core-modules` and `tns-core-modules-snapshot`. Or we may package all into one package.
- [x] Since the debugger does not work with snapshots, we will need to rely on the snapshot for release builds ONLY. Hence, there will be some effort on the CLI side. Plus, for release builds the CLI should not pack the modules JavaScript files but only the BLOBs  (ping @teobugslayer, @ligaz).
- [x] Enable the Android Runtime to consume such a BLOB directly, depending on the current CPU architecture. We may use the same convention as for the native part of the Runtime itself, for example:

    ```
[...]

PR #2028, opened 2016-04-25
Make FileSystem.currentAppPath platform specific

jasssonpet:
When running in the case of snapshot `__dirname` is not defined. Circumvent this by using platform-specific logic for finding the correct path.

Related to: https://github.com/NativeScript/NativeScript/issues/1563

ping @atanasovg, @KristinaKoeva


atanasovg:
:+1: Seems completely safe. This is not needed to go into the 2.0 release, we may have it applied locally on our machines prior to snapshot generation, right?

jasssonpet:
Yes, I think it's safer that way.

jasssonpet:
run ci

jasssonpet:
run ci

PR #2030, opened 2016-04-25
Fixed pseudoselectors to work with Angular

tzraikov:


hshristov:
I think we need to expose 2 new functions:
- `applyStyles`
- `removeStyles`

which should be called after CSS is applied and once element is removed from visual tree.
This way we won't expose private functions `_apply...` and we will be able to stop the visualStateWatcher.

What do you think?

vakrilov:
IMO - `removeStyles` is misleading for a public method. In fact it resets all properties with CSS value source - this won't affect any properties that have local value - for example if inline style was used.
Maybe we can a protected method that is called after CSS styles have been applied and use it in the ios button to start the listener.

hshristov:
:+1:

PR #2031, opened 2016-04-25
Make JavaProxy extends lazy

jasssonpet:
Related to: https://github.com/NativeScript/NativeScript/issues/1563

ping @atanasovg, @KristinaKoeva 

**P.S.** To view the diff without whitespaces: https://github.com/NativeScript/NativeScript/pull/2031/files?w=1

### DIFF
--- a/tns-core-modules/utils/lazy.ts
+++ b/tns-core-modules/utils/lazy.ts
@@ -2,3 +2,17 @@ export default function lazy<T>(action: () => T): () => T {
     let _value: T;
     return () => _value || (_value = action());
 }
+
+export function lazyExtend<T>(classExtend: () => T): { value: T } {
+    let result = { value: null };
+    let register = () => result.value = classExtend();
+
+    if (!global.__snapshot) {
+        register();
+    } else {
+        global.__pendingJavaExtendCalls = global.__pendingJavaExtendCalls || [];
+        global.__pendingJavaExtendCalls.push(register);
+    }
+
+    return result;
+}
--- a/tns-core-modules/application/application.android.ts
+++ b/tns-core-modules/application/application.android.ts
@@ -3,37 +3,33 @@ import definition = require("application");
 import frame = require("ui/frame");
 import observable = require("data/observable");
 import * as typesModule from "utils/types";
+import {lazyExtend} from "utils/lazy";
 
 global.moduleMerge(appModule, exports);
 var typedExports: typeof definition = exports;
 
-@JavaProxy("com.tns.NativeScriptApplication")
-class NativeScriptApplication extends android.app.Application {
+lazyExtend(() =>
+    (<any>android.app.Application).extend("com.tns.NativeScriptApplication", {
+        onCreate(): void {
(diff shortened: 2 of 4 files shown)


hshristov:
@jasssonpet I though we won't sacrifice TypeScript in the process. When TS Class is wrapped in function you cannot instantiate in. I guess that is why you need
`declare module com { ... `

application/application.android.ts
```diff
@@ -4,37 +4,40 @@ import frame = require("ui/frame");
...
 
-@JavaProxy("com.tns.NativeScriptApplication")
-class NativeScriptApplication extends android.app.Application {
-
-    constructor() {
-        super();
-        return global.__native(this);
-    }
+lazyExtend(exports, () =>
+    __decorate([JavaProxy("com.tns.NativeScriptApplication")],
```
hdeshev: :-1: can we find a way to not use `__decorate` here? I suggest we kill the `JavaProxy` decorator completely, and provide an API that adds a Java-JavaScript mapping. The result could probably look like:

```TypeScript
javaClass("com.tns.NativeScriptActivity", () =>
    class NativeScriptActivity {
    //...
    }
);
```

This way the async thing needed by the snapshot will amount to a change in the `javaClass` code.

Oh, and I like the benefit of defining "Java" classes in closures to prevent access to those from the rest of the JS code.

jasssonpet:
@hshristov I understand it's tough to use TypeScript in such edge cases. According to the ES7 draft specifications "A decorator is an expression", but TypeScript is emitting the following error: `error TS1109: Expression expected`.

I'm opening this PR to raise a discussion about the issue so that we can find the best solution for it.

hshristov:
If we are going to have such 'workarounds` I would vote for writing the whole class in Java and exposing some listener so that we don't need to extend/override it in JavaScript. In my opinion this is probably the best solution.

jasssonpet:
Hello there,

@PanayotCankov Helped me to update this old pull request with your suggestions. Please review it one more time and say whether this is acceptable.

tns-core-modules/utils/lazy.ts
```diff
@@ -1,4 +1,18 @@
 export default function lazy<T>(action: () => T): () => T {
     let _value: T;
     return () => _value || (_value = action());
-}
\ No newline at end of file
+}
+
+export function lazyExtend<T>(nativeName: string, classExtend: () => T): { value: T } {
+    let result = { value: null };
+    let register = () => result.value = global.__decorate([JavaProxy(nativeName)], classExtend());
```
jasssonpet: Sadly, this won't work because the static binding generator isn't able to make the link.

jasssonpet:
Updating the PR one more time.

hshristov:
-100
We are again falling back to JavaScript. You can't/shouldn't force people to write such code. If goes against our principle that TypeScript is first class citizen. I'm 100% we will forget about this hacks and will break the snapshot with some new check-in. A more robust approach needs to be implemented here.

PR #2063, opened 2016-05-03
Split layout-base to iOS and Android files

### LINKED ISSUE #1829: `LayoutBase.clipToBounds=false` does not work on Android [CAN"T FIX]

hamorphis:
We can't implement clipToBounds for Android without calling setClipChildren(false) on every ancestor up in the visual tree, which will kill performance. It will also lead to unwanted side effects such as other totally unrelated views being affected by setting the parents' setClipChildren to false. The problem in Android is that a ViewGroup either clips ALL of its children or it does not. Unlike iOS, the clipping cannot be controlled on a per view basis. So clipToBounds=false will have to be somehow achieved with stacking different views on top of one another in an AbsoluteLayout or GridLayout. There is always a workaround when playing with layouts.
The following article explains this in detail: http://stackoverflow.com/questions/25044085/when-drawing-outside-the-view-clip-bounds-with-android-how-do-i-prevent-underli

Resolves #1829


hamorphis:
run ci

hamorphis:
run ci

hamorphis:
run ci

PR #2067, opened 2016-05-03
Add __extends in case it's missing

jasssonpet:
Related to #1563

### DIFF
--- a/globals/globals.ts
+++ b/globals/globals.ts
@@ -42,6 +42,16 @@ global.zonedCallback = function (callback: Function): Function {
     }
 }
 
+global.__extends = global.__extends || function (d, b) {
+    for (var p in b) {
+        if (b.hasOwnProperty(p)) {
+            d[p] = b[p];
+        }
+    }
+    function __() { this.constructor = d; }
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
+};
+
 global.registerModule("timer", () => require("timer"));
 global.registerModule("ui/dialogs", () => require("ui/dialogs"));
 global.registerModule("xhr", () => require("xhr"));


enchev:
run ci

enchev:
Lint errors
https://travis-ci.org/NativeScript/NativeScript/builds/127524699#L2257
https://travis-ci.org/NativeScript/NativeScript/builds/127524699#L2258

NathanaelA:
@jasssonpet Why would __extends be missing; it is part of the js (ts_helpers) file that is loaded before any of the normal code?

jasssonpet:
Hey @NathanaelA,

You are right that the Android Runtime already has TypeScript extends functionality built in. But this is only the case when the NativeScript modules are running in the default environment.

In this case, we are running the code through the [V8 snapshot API](https://github.com/NativeScript/android-v8/blob/4.7.80/include/v8.h#L5998-L6003) and there are no runtime intrinsic modules available there. Refer to [this commit](https://github.com/NativeScript/nativescript-angular-snapshot/commit/3b2fe6b902a73eae1358c82cef33a521f281a92e).

Also, I would like to add, that our modules shouldn't expect to be running in any particular environment and (in theory) the non-ui stuff should be running in node.js and the web browser where TypeScript isn't nothing but just another JavaScript transpiler.

enchev:
:+1:

NathanaelA:
@jasssonpet - Ah that makes some :grinning: sense.

 On your last use-case, If I include a non-ui module in node, I wouldn't even know to be bringing in global also, so the module will crash anyways.  :grinning: 

In addition just skimming the code for global; I really wouldn't want to bring in global in node, as it will probably mess up several actual node built-ins...

Thanks for the explanation for the primary use case...

PR #2070, opened 2016-05-04
Fixed css specificity levels for type selectors.

nsndeck:
Fixes css selectors specificity when mixed type and class selector is used (Label.red).

### DIFF
--- a/ui/styling/css-selector.ts
+++ b/ui/styling/css-selector.ts
@@ -143,7 +143,12 @@ export class CssSelector {
 
 class CssTypeSelector extends CssSelector {
     get specificity(): number {
-        return TYPE_SPECIFICITY;
+        let result = TYPE_SPECIFICITY;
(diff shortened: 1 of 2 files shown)


nsndeck:
run ci

ui/styling/css-selector.ts
```diff
@@ -143,7 +143,12 @@ export class CssSelector {
 
 class CssTypeSelector extends CssSelector {
     get specificity(): number {
-        return TYPE_SPECIFICITY;
+        let result = TYPE_SPECIFICITY;
+        let dotIndex = this.expression.indexOf(DOT);
```
NathanaelA: What about #Id support?  This appears to only support **.Class**

nsndeck:
Indeed this will support only `.Class`. We discussed (in the team) the option with `#`, but since it is too specific and connected with an unique control team decide that there is no need to support it for the moment (no real business case).

Issue #2059, opened 2016-05-02
Camera module: Save images to camera roll


tjvantoll:
This feature request has come up several times, for example see https://www.thepolyglotdeveloper.com/2015/11/using-the-device-camera-in-your-nativescript-mobile-app/#comment-2653570744.

Essentially it would be nice for the camera module to provide the ability to save images to the camera roll out of the box. This would essentially be the equivalent of the [Cordova camera plugin’s `saveToPhotoAlbum` option](https://www.npmjs.com/package/cordova-plugin-camera).

I’ve marked this as a good first task, as I think it’s something a new contributor could tackle—either directly in the camera module or through a plugin.

Daxito:
+1

victorsosa:
Hi @tjvantoll,

I like to take this good first task; but I need help find out two thing.

1. How I can get this Android call from NativeScript:

```
File storageDir = Environment.getExternalStoragePublicDirectory(
            Environment.DIRECTORY_PICTURES);
```
I need this because the normal call like:
`utils.ad.getApplicationContext().getExternalFilesDir(null)`
Only works for folder private to this app; So if you want to share you need to call the one before. (Permissions included)

2. How I can send a broadcast call, like it is done in an android app using the Activity class; but on NativeScript there is not Activity class. I have the intent object to be send I just need the broadcast method to be call.

`Activity.sendBroadcast(IntentObject);`


The rest of the code I can figure out how to do it.

victorsosa:
Never mine there is a PR#2069 created already.

enchev:
run ci

victorsosa:
HI, @tjvantoll 

this was implemented and you can close the issue

tjvantoll:
@victorsosa Will do. Nice work on this 😀

PR #2078, opened 2016-05-06
Added save to gallery option - IOS

victorsosa:
This is to completed the PR#2069; now Also IOS will save the picture in the camera roll

Implements #2059.

`saveToGallery` returns the image path defaults to false.

```
cameraModule.takePicture({ saveToGallery: true }).then(function (path) {
            console.log(path)
        });

```
Updated the default option to return the image path also

```
cameraModule.takePicture().then(function (picture,path) {
            console.log(picture,path)
        });
```


enchev:
run ci

enchev:
:+1:

PR #2082, opened 2016-05-09
Call row/column removed callbacks on removeRows/removeColumns.

hdeshev:
Changed callback names to _onRowRemoved, etc and changed visibility to
public to work around a TypeScript "protected" limitation.

Originally reported in NativeScript/nativescript-angular#211.


hdeshev:
run ci

PR #2099, opened 2016-05-13
Typo in Pointer.getY method for iOS

covex-nn:
Pointer.getY method for iOS returned this.location.x instead of this.location.y


enchev:
run ci

enchev:
:+1:

PR #2100, opened 2016-05-13
Fix navigatedFrom event raised when Activity is destroyed without act…

hshristov:
navigatedFrom event was raised when Activity is destroyed without actual navigation.
This was breaking application suspend/resume when `Don't keep activities` is `on`.


hshristov:
run ci

PR #2105, opened 2016-05-13
add native script key in package.son files

Plamen5kov:
https://github.com/NativeScript/android-runtime/issues/436 - more optimized build.

In order for a more optimized build, we need to make a difference between pure javascript modules and native script specific ones. We will make the difference when we read the `package.json` file of each plugin, including the native script modules. That's why we need to add a native script key inside each modules `package.json`.
Cli already has this logic for plugins and uses the "nativescript" key of the plugins `package.json`. This PR will simply follow the established practice.

ping: @enchev @vakrilov @hshristov @PanayotCankov

### DIFF
--- a/application-settings/package.json
+++ b/application-settings/package.json
@@ -1,2 +1,5 @@
-{ "name" : "application-settings",
-  "main" : "application-settings.js" }
+{
+	"name" : "application-settings",
+  	"main" : "application-settings.js",
+	"nativescript": {}
+}
--- a/application/package.json
+++ b/application/package.json
@@ -1,2 +1,5 @@
-{ "name" : "application",
-  "main" : "application.js" }
+{
+	"name" : "application",
+	"main" : "application.js",
+	"nativescript": {}
+}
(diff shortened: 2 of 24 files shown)


hshristov:
Modules are not stand alone packages so this seems unnecessary. why not using *.android. convention?

Plamen5kov:
@hshristov 
* ` why not using *.android. convention`
* As it is discussed in the issue the .android convention is great, but it's optional. Currently we also allow a not specified android or iOS js file to be used for android specific code. In other words if someone decides to write 
```
if(android) {
    //android speciffic code
}
```
and this file needs to be analyzed, we have no means of recognizing it as android specific.

In fact, modules are not standalone packages, but they should be, so this would be one step towards the final goal. Furthermore this PR wouldn't be necessary if we didn't have special logic when preparing the native script modules.

Plamen5kov:
run CI

hshristov:
:+1:

PR #2112, opened 2016-05-14
Fixed breaking change in camera for android.

nsndeck:
Generally reverted some breaking change from a previous commit with some other adjustments like returning scaled image for all cases (saveToGallery and not). Also changed all vars to lets.


enchev:
:+1:

Issue #2113, opened 2016-05-14
Panning gesture in Android creates non-smooth delta coordinates. 


TheOriginalJosh:
if you add a panning event and add a console.log  message outputting the delta coordinates to the console in iOS you will see a  smooth coordinates transition like below: 
```
May 14 07:31:31 Joshs-MacBook-Pro slide[64850]: CONSOLE LOG file:///app/nativescript-intro-slides.js:190:28: deltaX:-19.5  deltaY:138.5
May 14 07:31:31 Joshs-MacBook-Pro slide[64850]: CONSOLE LOG file:///app/nativescript-intro-slides.js:190:28: deltaX:-18.5  deltaY:138.5
May 14 07:31:31 Joshs-MacBook-Pro slide[64850]: CONSOLE LOG file:///app/nativescript-intro-slides.js:190:28: deltaX:-17.5  deltaY:138.5
May 14 07:31:31 Joshs-MacBook-Pro slide[64850]: CONSOLE LOG file:///app/nativescript-intro-slides.js:190:28: deltaX:-16.5  deltaY:138.5
``` 

however on Android in the console you will see occasional coordinates that jump back and forth such as the deltaX coordinates below:

```
JS: deltaX:-90.51925223214283  deltaY:-2.868350074404759
JS: deltaX:-86.28752790178572  deltaY:-2.868350074404759
JS: deltaX:-90.90020461309524  deltaY:-2.868350074404759
JS: deltaX:-85.31352306547615  deltaY:-2.868350074404759
JS: deltaX:-90.73135230654759  deltaY:-2.868350074404759
JS: deltaX:-90.73135230654759  deltaY:-2.868350074404759
```

I believe this is because in the gestures.android.ts file for the CustomPanGestureDetector it is is using getX(int) and getY(int). I've tested changing this to getRawX() and getRawY() and seen a noticeable improvement in the coordinate bounce. 

here is the PR with a potential fix. https://github.com/NativeScript/NativeScript/pull/2114

vjoao:
I can confirm this issue as well.

PR #2114, opened 2016-05-14
changed getX and getY to getRawX and getRawY for CustomPanGestureDetector 

TheOriginalJosh:
Relates to issue https://github.com/NativeScript/NativeScript/issues/2113

getRawX & Y always return the X & Y coordinates relative to the view
getX and getY sometimes return coordinates that are relative to the
previous motion. this update will allow smoother coordinates on panning
instead of occasional  jumping back and fourth of the delta X and Y
coordinates

### DIFF
--- a/ui/gestures/gestures.android.ts
+++ b/ui/gestures/gestures.android.ts
@@ -110,7 +110,7 @@ export class GesturesObserver extends common.GesturesObserver {
                 this._eventData = new TouchGestureEventData();
             }
 
-            this._eventData.prepare(this.target, motionEvent);          
+            this._eventData.prepare(this.target, motionEvent);
             _executeCallback(this, this._eventData);
         }
 
@@ -489,9 +489,9 @@ class CustomPanGestureDetector {
     private getMotionEventCenter(event: android.view.MotionEvent): { x: number, y: number } {
         let count = event.getPointerCount();
         let res = { x: 0, y: 0 };
-        for (var i = 0; i < count; i++) {
-            res.x += event.getX(i);
-            res.y += event.getY(i);
+        for (let i = 0; i < count; i++) {
+            res.x += event.getRawX();
+            res.y += event.getRawY();
         }
 
         res.x /= (count * this.density);


enchev:
run ci

vakrilov:
Hey @TheOriginalJosh - thanx for the PR,
However, I have concerns about it.
First of all the loop variable `i` is not used inside the loop. The idea of the method is to get the `center` of the gesture - which is median point of all the current pointers. That's why the original code sums the coordinates of all the pointers and the takes the average of them. The code in the PR will probably take into account only one of the pointers.

Also - according to the [docs](http://developer.android.com/reference/android/view/MotionEvent.html#getRawX()) - `getRawX/Y()` method will return coordinates relative to the screen, not to the current view. This means that you will almost always do some translation of the coordinates you get from the method.

TheOriginalJosh:
I'm sorry, I don't follow. What is CI?

enchev:
Hey @TheOriginalJosh,

I've just ran (`run ci` command) our local CI (continues integration) to check if everything is running properly with your pull request. :)

TheOriginalJosh:
@enchev Cool, if you have any problems or questions. Don't hesitate to ask.

vakrilov:
Hey @TheOriginalJosh - thanx for the PR,
However, I have concerns about it.
The loop variable `i` is not used inside the loop. The idea of the method is to get the `center` of the gesture - which is median point of all the current pointers. That's why the original code sums the coordinates of all the pointers and the takes the average of them. The code in the PR will take into account only one of the pointers - so if you are panning with more than one finger it will not work correctly.

Actually - the [`getRawX()`](http://developer.android.com/reference/android/view/MotionEvent.html#getRawX()) methods will return screen coordinates. That said, I agree that using the `getRaw` methods will be more stable, especially if you are changing the layout of the element that is being panned.

vakrilov:
Closing the in favor of #2118 which includes the multi-touch scenario.

TheOriginalJosh:
@vakrilov Np, when I tested it on my phone, which is a Nexus 5x with Android Marshmellow, I was able to use multi-touch without a problem. Changing the layout is where I had the problems, not so much problems but just occasional inconsistency with the delta coordinates that end up causing a jumpiness​ that i wasn't getting on ios.

PR #2120, opened 2016-05-16
Export single CSS property resolution API.

hdeshev:
Extract single style property resolution to an exported API.

Used by CSS selectors, and now the Angular `setElementStyle` renderer API.


hdeshev:
run ci

enchev:
:+1:

Issue #2084, opened 2016-05-09
App crashes when animating ActionBar


speigg:
### Did you verify this is a real problem by searching [Stack Overflow](http://stackoverflow.com/questions/tagged/nativescript) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)? 
yes

### Tell us about the problem
The app crashes when attempting to animate the actionbar.

### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 1.7.0
- Cross-platform modules: 1.7.1
- Runtime(s): tns-ios@1.7.0
- Plugin(s): n/a

### Please tell us how to recreate the issue in as much detail as possible.

When attempting to animate the actionbar, the app crashes with the following stacktrace:
```
May  9 12:26:10 gs app[75152]: 2   0x70ccc0 ffi_closure_inner
May  9 12:26:10 gs app[75152]: 3   0x70d612 ffi_closure_i386
May  9 12:26:10 gs app[75152]: 4   0x14d5e38 -[UIApplication sendAction:to:from:forEvent:]
May  9 12:26:10 gs app[75152]: 5   0x14d5db7 -[UIApplication sendAction:toTarget:fromSender:forEvent:]
May  9 12:26:10 gs app[75152]: 6   0x1679f3b -[UIControl sendAction:to:forEvent:]
May  9 12:26:10 gs app[75152]: 7   0x167a2d4 -[UIControl _sendActionsForEvents:withEvent:]
May  9 12:26:10 gs app[75152]: 8   0x16792c1 -[UIControl touchesEnded:withEvent:]
May  9 12:26:10 gs app[75152]: 9   0x155652e -[UIWindow _sendTouchesForEvent:]
May  9 12:26:10 gs app[75152]: 10  0x15575cc -[UIWindow sendEvent:]
May  9 12:26:10 gs app[75152]: 11  0x14f8be8 -[UIApplication sendEvent:]
May  9 12:26:10 gs app[75152]: 12  0x16686369 -[UIApplicationAccessibility sendEvent:]
May  9 12:26:10 gs app[75152]: 13  0x14cd769 _UIApplicationHandleEventQueue
May  9 12:26:10 gs app[75152]: 14  0x3739e5f __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__
[...]

PR #2124, opened 2016-05-17
Fixed: App crashes when animating ActionBar

tzraikov:
#2084

### DIFF
--- a/ui/action-bar/action-bar.ios.ts
+++ b/ui/action-bar/action-bar.ios.ts
@@ -29,7 +29,21 @@ export class NavigationButton extends ActionItem {
 }
 
 export class ActionBar extends common.ActionBar {
-    public update() {
+
+     get ios(): UIView {
+
+        if (!(this.page && this.page.parent)) {
+            return;
+        }
+
+        let viewController = (<UIViewController>this.page.ios);
+        if (viewController.navigationController !== null) {
+            return viewController.navigationController.navigationBar;
+        }
+        return null;
+     }
+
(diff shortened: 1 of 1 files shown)


enchev:
:+1:

ui/action-bar/action-bar.ios.ts
```diff
@@ -29,7 +29,18 @@ export class NavigationButton extends ActionItem {
...
 export class ActionBar extends common.ActionBar {
-    public update() {
+
+     get ios(): UIView {
+
+        if (!(this.page && this.page.parent)) {
+            return;
+        }
+
+        let viewController = (<UIViewController>this.page.ios);
```
hshristov: Probably a check if `navigationController !== null` will be good. We will support pages that are not hosted in Frame.

hshristov:
:+1:

ui/action-bar/action-bar.ios.ts
```diff
@@ -38,7 +52,7 @@ export class ActionBar extends common.ActionBar {
         var viewController = (<UIViewController>this.page.ios);
         var navigationItem: UINavigationItem = viewController.navigationItem;
         var navController = frameModule.topmost().ios.controller; 
-        var navigationBar = <UINavigationBar>navController.navigationBar;
```
vakrilov: `navigationBar` is used later on - so this will probably cause a crash. If there is no `navigationController` - probably it is just safer to return - we cannot do much anyway.

PR #2127, opened 2016-05-17
Add a flag when logging to skip string template calculations in release

PanayotCankov:
When logging non errors wrap guard them:
``` TypeScript
if (tracer.enabled) {
    tracer.write("some " + heavy + "calculations", nonErrorCategory);
}
```
Even if the tracer is disabled, the arguments will be calculated, sometimes calling heavy toString() implementations.

The guard is supposed to provide the most efficient way to skip any tracing overhead when running in release.


dtopuzov:
run ci

PR #2133, opened 2016-05-18
Removed unnecessary view state creation.

nsndeck:
Deleted code within `desctoyItem` method, since it leads to some problems when trying to create new contents in a TabView.


vchimev:
👍

enchev:
:+1:

PR #2148, opened 2016-05-19
Use global.require in loadModule

jasssonpet:
This is required when preparing only the tns-core-modules for snapshotting. Otherwise webpack replaces this with custom webpack contexts (https://webpack.github.io/docs/configuration.html#automatically-created-contexts-defaults-module-xxxcontextxxx) which is not what we want.

Related to #1563

@hdeshev Do you think this change is going to conflict in any way with the webpack plugin?

### DIFF
--- a/globals/globals.ts
+++ b/globals/globals.ts
@@ -25,7 +25,7 @@ global.loadModule = function(name: string): any {
     if (loader) {
         return loader();
     } else {
-        return require(name);
+        return global.require(name);
     }
 }
 


hshristov:
:+1:

enchev:
:+1:

hdeshev:
Better late than never -- the change doesn't impact the bundling plugin (much). It will work most of the time, and the only thing that can possibly go south is forwarding "unknown" imports to the underlying `require` implementation instead of throwing the usual webpack exception. That may actually be a good thing. :wink:

PR #2182, opened 2016-05-26
Add API that resolves the native android.app.Fragment instance created for a Page

atanasovg:
The specific use case is to enable the YouTube player plugin in a {N} app. The Google API works either through extending [YouTubeBaseActivity](https://developers.google.com/youtube/android/player/reference/com/google/android/youtube/player/YouTubeBaseActivity) or by creating an instance of the [YouTubePlayerFragment](https://developers.google.com/youtube/android/player/reference/com/google/android/youtube/player/YouTubePlayerFragment) class. Since navigation in the core Modules is fragment-based, it is more naturally to use the second approach of the player API.

This pull adds a method on the `AndroidFrame` interface that returns the native `android.app.Fragment` created for a Page instance. This native Fragment instance is needed so that we can take advantage of the [nested fragments](https://developer.android.com/about/versions/android-4.2.html#NestedFragments) functionality and integrate a YouTube fragment instance.


ui/frame/frame.d.ts
```diff
@@ -294,6 +294,12 @@ declare module "ui/frame" {
          * Gets or sets whether the page UI will be cached when navigating away from the page.
          */
         cachePagesOnNavigate: boolean;
+
+        /**
+         * Finds the native android.app.Fragment instance created for the specified Page.
+         * @param page The Page instance to search for.
+         */
+        fragmentForPage(page: pages.Page): android.app.Fragment;
```
hshristov: Don't return native type or clients will need the big andoid.d.ts in order to compile.

ui/frame/frame.android.ts
```diff
@@ -509,6 +509,38 @@ class AndroidFrame extends Observable implements definition.AndroidFrame {
         // can go back only if it is not the main one.
         return this.activity.getIntent().getAction() !== android.content.Intent.ACTION_MAIN;
     }
+    
+    public fragmentForPage(page: pages.Page): android.app.Fragment {
+        if(!page || !this._owner) {
+            return undefined;
+        }
+        
+        var tag;
```
hshristov: All other fields are defined with `let`

hshristov:
Btw I got another idea. If there is YouTubeActivity can we use `startActivity` method instead of using fragments?

atanasovg:
The approach with the Activity will be far more limited and will require more native code to implement. The approach with Fragments will allow us to create a full-blown plugin that exposes a, say `YouTubeView`, module that will allow us to plug it into our Visual Tree, enabling all the benefits like layouts, data-bindings, styling, etc.

atanasovg:
run ci

PR #2211, opened 2016-05-30
Fixed:  Android layerType should not be changed if there is no need

### LINKED ISSUE #1899: Android layerType should not be changed if there is no need
Don't change android layer type If possible because it cause different type of issues on different devices. If change is mandatory first cache current layer type using `nativeView.getLayerType()` so a reset will return it to its default value - not like now setting it always to `software_layer`.

Check https://github.com/NativeScript/NativeScript/issues/1874 for one such problem.

hamorphis:
Resolves #1899


hamorphis:
run ci

hamorphis:
run ci

PR #2214, opened 2016-05-30
[Second Try] Add API that resolves the native android.app.Fragment instance created for a Page

atanasovg:
The specific use case is to enable the YouTube player plugin in a {N} app. The Google API works either through extending [YouTubeBaseActivity](https://developers.google.com/youtube/android/player/reference/com/google/android/youtube/player/YouTubeBaseActivity) or by creating an instance of the [YouTubePlayerFragment](https://developers.google.com/youtube/android/player/reference/com/google/android/youtube/player/YouTubePlayerFragment) class. Since navigation in the core Modules is fragment-based, it is more naturally to use the second approach of the player API.

This pull adds a method on the `AndroidFrame` interface that returns the native `android.app.Fragment` created for a Page instance. This native Fragment instance is needed so that we can take advantage of the [nested fragments](https://developer.android.com/about/versions/android-4.2.html#NestedFragments) functionality and integrate a YouTube fragment instance.

### DIFF
--- a/tests/app/ui/page/page-tests.android.ts
+++ b/tests/app/ui/page/page-tests.android.ts
@@ -143,18 +143,22 @@ export var test_ChangePageCaching_BeforeNavigated_DoesNotThrow = function () {
 
 export var test_Resolve_Fragment_ForPage = function () {
     var testPage: PageModule.Page;
+    var navigatedTo = false;
+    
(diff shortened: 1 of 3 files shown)


atanasovg:
run CI

hshristov:
:+1:

PR #2233, opened 2016-06-02
Expose the zonedCallback function

vakrilov:
Expose the `zonedCallback` function so that it can be used from plugin developers to wrap callbacks that come from native inside zone. This is needed in Angular project to trigger the change detection.


hshristov:
👍

vchimev:
run ci

vakrilov:
run ci

PR #2244, opened 2016-06-03
Remove the android.app.Application extend from the core modules

atanasovg:
This pull removes the Android Application object extend from the core modules. The change enables the following scenarios:

- Make the modules Heap snap-shottable
- Allow users to provide custom `android.app.Application` implementation. This includes
    - Entirely custom application type
    - Implement an additional interface on the default `com.tns.NativeScriptApplication` class

The change also introduces the following paths for application initialization:

- User provides custom application extend. In the `onCreate` method he should manually initialize the application module via `application.android.init(this)`. The static-binding-generator will take care of the `*.JAVA` file generation.
- No custom extend is defined. Application will use the default `com.tns.NativeScriptApplication` implementation provided with the Android Runtime project template. Application module is initialized upon `application.start` call.
- User provides custom extend but does not initialize the application module explicitly. This falls back to a somewhat hacky implementation, that retrieves the native `android.app.Aplication` object through reflection. **TODO: We may throw an error instead of doing this**.

Note: This pull will work when we merge this one as well: https://github.com/NativeScript/android-runtime/pull/471

### DIFF
--- a/tns-core-modules/application/application.android.ts
+++ b/tns-core-modules/application/application.android.ts
@@ -120,7 +120,6 @@ function initLifecycleCallbacks() {
 
 let currentOrientation: number;
 function initComponentCallbacks() {
-    // TODO: Verify whether the logic for triggerring application-wide events based on Activity callbacks is working properly
     let componentCallbacks = new android.content.ComponentCallbacks2({
         onLowMemory: function() {
             gc();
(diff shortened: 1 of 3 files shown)


hshristov:
👍 
My only concern is how will this affect existing applications? I guess they will need to manually add this `android.application.js` in their project templates?

atanasovg:
@hshristov They are not forced to do so - that's the second initialization path, as listed in the pull's description. The extend will happen only if they need to add some custom initialization logic.

In other words - existing applications shouldn't be affected.

vchimev:
run ci

atanasovg:
run ci

hshristov:
👍

PR #2250, opened 2016-06-06
Ignore the case when getting a response header

emelois:
According to the RFC: https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
"Field names are case-insensitive".
For instance, Google Cloud Endpoints use "content-type" in the response.

### DIFF
--- a/tns-core-modules/xhr/xhr.ts
+++ b/tns-core-modules/xhr/xhr.ts
@@ -119,7 +119,7 @@ export class XMLHttpRequest {
         this._responseTextReader = () => r.content.toString();
         this._response = JSON.parse(this.responseText);
 
-        // Add toString() method to ease debugging and 
+        // Add toString() method to ease debugging and
         // make Angular2 response.text() method work properly.
         Object.defineProperty(this._response, "toString", {
             configurable: true,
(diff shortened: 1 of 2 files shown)


tns-core-modules/xhr/xhr.ts
```diff
@@ -194,10 +194,13 @@ export class XMLHttpRequest {
     public getResponseHeader(header: string): string {
         if (types.isString(header) && this._readyState > 1
             && this._headers
-            && this._headers[header]
             && !this._errorFlag
         ) {
-            return this._headers[header];
+            for (var i in this._headers) {
+                if (i.toLowerCase() === header.toLowerCase()) {
```
hshristov: You could cache the call to `header.toLowerCase()` instead of make it always.

hshristov:
Is it possible to add a unit-test?

emelois:
@hshristov I cached the lowercase header and added a unit test.

vchimev:
run ci

hshristov:
@emelois Thank you for your contribution. PR will be live with the next official release - 2.1.0

Issue #2225, opened 2016-06-01
The exit transition of the current page is not played when navigating with clearHistory.


bradmartin:
### Did you verify this is a real problem by searching [Stack Overflow](http://stackoverflow.com/questions/tagged/nativescript) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?

Yes

### Tell us about the problem
Please, ensure your title is less than 63 characters long and starts with a capital
letter.

```JS
public static goToMainPage() {
        frame.topmost().navigate({
            moduleName: "views/main-page/main-page",
            animated: true,
            clearHistory: true,
            // backstackVisible: false,
            transition: {
                name: "fade",
                duration: 350,
                curve: "easeIn"
            }
        })
    }

```

### Which platform(s) does your issue occur on?

* Android tested only*

### Please provide the following version numbers that your issue occurs with:
- CLI: (run `tns --version` to fetch it) 
- Cross-platform modules: (check the 'version' attribute in the
`node_modules/tns-core-modules/package.json` file in your project)
- Runtime(s): (look for the `"tns-android"` and `"tns-ios"` properties in the
`package.json` file of your project)
- Plugin(s): (look for the version number in the `package.json` file of your
project)

- 2.0.0
- 2.0.1
- 2.0.0
- "dependencies": {
		"nativescript-cardview": "^1.1.7",
		"nativescript-floatingactionbutton": "^2.2.2",
		"nativescript-parallax": "^1.1.0",
		"nativescript-platform-css": "^1.0.0",
		"nativescript-pulltorefresh": "^1.1.4",
		"nativescript-slides": "^1.4.5",
		"nativescript-swiss-army-knife": "0.1.1",
		"nativescript-telerik-ui": "^1.1.1",
		"tns-core-modules": "2.0.1"
	}

### Please tell us how to recreate the issue in as much detail as possible.

Comment out `clearHistory` on a page transition and you'll see the animation. I'm specifically testing the **fade** animation on a Droid Turbo 2 running Android 6.0

If I leave the `clearHistory="true"` the transition never animates and the next page just appears after what seems like the duration.

hamorphis:
I found the problem. When we navigate with clearHistory=true, the entire FragmentManager back-stack is popped to its bottom in order to clear the history and the current page is immediately destroyed instead of being given the chance to play its exit transition before it gets destroyed. I will see whether this can be fixed.

PR #2261, opened 2016-06-07
Fixed: Exit transition on API Level 23 with cachePagesOnNavigate=true

hamorphis:
When we are on API Level 23 and we have cachePagesOnNavigate=true, the exit transition of the current (disappearing) page ends immediately after its start, which in turn removes the page UI from the fragment and leaves a white spot in place of the current page while the enter transition of the new page is being played.

Related to #2225


hamorphis:
run ci

hamorphis:
cuteness

hamorphis:
uitests

PR #2269, opened 2016-06-08
Layout will not requestLayout when its text is changed if it has fixed size

PanayotCankov:
This is related to the list view speed improvement in the QSF.


hshristov:
:+1:

PR #2270, opened 2016-06-08
Avoid excessive requestLayout calls

hshristov:
Added View.isLayoutRequired - to check if Layout call is needed after a measure call. Use in list-view.
Measure pass now get all its properties from nativeLayoutParams property instead of using instance dependency properties (e.g. perf improvement).
List-view now layouts cells only if there is need to.
List-view now measure rows with the specified rowHeight.


PanayotCankov:
:+1:

PR #2271, opened 2016-06-08
Extract the Activity implementation logic in a separate class.

atanasovg:
This pull removes the prerequisite for a specific Activity type for the core modules and enables scenarios where users will need custom Activity base type - e.g. `YouTubeBaseActivity`. The modules now depend on being notified externally for an `Activity` event - like `onCreate`.

Steps for providing extended Activity at user level:

1. Add a new file, having the `*.android.*` suffix.
2. Implement/override the needed methods.
3. User must notify the core modules for the needed callbacks through the `ui/frame/AndroidActivityCallbacks` interface.
4. The static-binding-generator will take care of generating the new Activity class. The default one provided by the modules will be either overridden or will not be used (in case of a new Activity name).
5 User must update the manifest file if the Activity name, defined through the `JavaProxy` decorator, is changed to a new one.

**Cons:** The biggest concern with this approach is that if the user misses to call to the modules upon some needed event then the behavior of the entire application will become indeterminate and the app will either crash or perform in an unexpected way.

### DIFF
--- a/tests/app/ui/dependency-observable-tests.ts
+++ b/tests/app/ui/dependency-observable-tests.ts
@@ -232,7 +232,6 @@ export function test_PropertyMetadata_Options() {
 }
 
 export function test_PropertyEntry_effectiveValue() {
-    var p = new dependencyObservableModule.Property(generatePropertyName(), "testOwner", new dependencyObservableModule.PropertyMetadata(0));
     var entry = new dependencyObservableModule.PropertyEntry();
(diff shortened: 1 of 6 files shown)


jasssonpet:
👍 from the snapshot point of view

hshristov:
👍 after successful build.

vchimev:
👍 
merge in order to have green build

PR #2280, opened 2016-06-09
Fix setting src on async image mismatching imageSource

PanayotCankov:
There is a test, but due to the async nature of the problem it should have quite some timeout and may not be very reliable. I've added the test but it won't be executed on builds as we need stable unit testing.


hshristov:
👍

PR #2288, opened 2016-06-10
Decouple Fragment implementation logic from the Extend call.

atanasovg:
This pull decouples the android.app.Fragment extend in a separate file, which allows it to be safely skipped from the snapshot bundle. It also allows users to provide custom extend logic, given they utilize the `frame` module as needed (calling back through the provided `_callbacks` member).


hshristov:
👍

PR #2298, opened 2016-06-13
Move UILableImpl as TNSLabel in widgets.

hshristov:
Refactor Label class to the new use TNSLabel class.
Added definition class.


vakrilov:
👍

nsndeck:
:+1:

PR #2299, opened 2016-06-13
Fixed binding converter calls.

nsndeck:
Converter should be called even when source property value is null or undefined.


hshristov:
👍

PR #2306, opened 2016-06-14
Fix: Images have margin when added to Layout

### LINKED ISSUE #1425: Images have margin when added to Layout
When I add Image(s) to any Layout (tested on Stack and Grid), the images is added with margins on the left and right side (around 5 px on each side). Isn't it supposed to fill the width of the container and keep aspect? I tried also with stretch="AspectFill" same result. 

But as soon as I set a specific height to the image, the margins are gone.

hamorphis:
Resolves #1425


hamorphis:
run ci

hamorphis:
run ci

hshristov:
👍

hamorphis:
run ci

PR #2357, opened 2016-06-22
webpack fixes

hdeshev:
- Prefer registered modules in `builder` and `component-builder`.
- Implemented a bundle entrypoint registration mechanism using a package.json attribute.
- Register well-known modules used by the component builder as bundle entrypoints.
- Move text/(formatted-string|span) modules to subfolders, so we can register those as entry points (meh!)


enchev:
:+1:

hdeshev:
run ci

PR #2364, opened 2016-06-24
Export all fetch types to the global object when running in snapshot context

jasssonpet:
This comes from this test: https://github.com/NativeScript/NativeScript/blob/15df94ca829192f1d52796e19b6f41f599c53405/tests/app/fetch-tests.ts#L15

Ping @atanasovg, @enchev.

PR is for release branch.


atanasovg:
:+1: Looks good

PR #2392, opened 2016-06-30
TextWatchListener is removed at onDetach 

### LINKED ISSUE #2391: App crash when editing text and text-view is removed
If we remove TextView from UI and manually call `setText` on the native view (which was taken before TextView is removed from view hierarchy) then we crash the app (android only).

hshristov:
so that no exception is thrown if android raise textChanged but we destroyed our android object in onDestroy of edit-text.
Fixes #2391


hshristov:
run ci

vakrilov:
👍

PR #2431, opened 2016-07-08
Create a single NS app for cuteness.io, gallery-app and ui-tests-app

hamorphis:
We now have only two sample apps in our repo -- `apps/` and `tests/`. The `apps` application is a singe NS application that contains folders with cuteness.io, gallery-app and ui-tests-app. To run one of those apps you will have to change what is written in the `main` entry in `apps/app/package.json`. The tests app is as it was before. I have removed all LICENSE files from these apps since we are not distributing them to anyone.

All other small old apps were moved to https://github.com/NativeScript/tns-apps-graveyard. If we decide that we need one of those apps on a daily basis, we can move it in its own folder next to cuteness.io and the others.

The entire gruntfile.js was completely refactored to reflect the current state of our repository. Lots of redundant stuff from more than a year ago is now gone. The new build is tens of times faster than the old one. `gruntfile.js` went from 822 lines to 560 lines.

To add a new sample app to our build process, simply create it in a folder on the root level and then in `gruntfile.js` on line 129 add it to the array describing all sample apps:

```
var srcAppDirs = ["tests/app", "apps/app"];
```

That is all you need to do. The app will be transpiled, tslinted and packed in /bin/dist/tns-samples-myapp-x.x.x.tgz

A sample of what running 'grunt' in this branch will be produce can be found in this zip file: 
[bin.zip](https://github.com/NativeScript/NativeScript/files/353758/bin.zip)


vchimev:
👍

vchimev:
run ci

vchimev:
run ci

Issue #1278, opened 2015-12-21
TabView button icons (iOS)


manijak:
I am creating a separate issue based my comments here: https://github.com/NativeScript/NativeScript/issues/1269
The original question was answered. 

@vakrilov , @maknz
From what I can see in ```tab-view.ios.js``` under tns_core_modules, line 218 is reason why tab-view icons look different in the latest release.

```javascript
...
var originalRenderedImage = is.ios.imageWithRenderingMode(UIImageRenderingMode.UIImageRenderingModeAlwaysOriginal);
...
```
The icon image rendering mode is set to ```AlwaysOriginal```. This means the icons will keep their original rendering, this is not default iOS behaviour, I think...

If I override this to ```AlwaysTemplate``` or  ```Automatic```, it will use the iOS color-templating on the tabBar icons. And you can define both text and icon color with the ```selectedColor``` property. 

Most of the time it is easier to use ```AlwaysTemplate``` since iOS takes care of the normal and selected color for each button, and you can use one icon for both states. 

But if we use ```AlwaysOriginal``` then we must provide both normal and selected icon-images, since they are not colored by iOS anymore. So we need to provide to version of each icon (one for each state). And from the looks of it, TabView in nativescript does not provide properites to add icons for both states.

stevetayloruk:
+1

manijak:
This should have been fixed/improved for 1.7.0.  Very annoying issue where we have to manually change ```tab-view.ios.js``` every time there is a release/update to the core-modules.

Please change the image render mode to Automatic and give us the ability to set color to both selected and non-selected buttons. And/Or the possibility to set normal and selected icons.

hamorphis:
@manijak It's fixed. For simple stuff like this you can always change one two lines of code and then open a pull request.

toddanglin:
@manijak Thanks again for raising this issue and providing the suggested solution! Everyone will now benefit in the 2.2 release.

leocaseiro:
Perhaps, would be nice if we had the option to choose between `UIImageRenderingModeAlwaysOriginal` and `UIImageRenderingModeAutomatic`.

PR #1149, opened 2015-11-24
Correct TabView icon rendering mode

rclai:
Fixes #1148.

PR #2435, opened 2016-07-09
Update iOS TabView to use Automatic rendering mode

toddanglin:
Fixes/Implements #1278 

Very small change. Does not impact the TabView unit tests.

Changes the UIImageRenderingMode for tab icons from "AlwaysOriginal" to "Automatic." This allows icon images to inherit the `selectedColor` specified for the tabview (eliminating the need to provide two versions of every tab icon).

Unclear why this behavior was changed in NativeScript 1.5 (history of tab-view.ios.ts only goes back to May). This change restores behavior of tabview icons to pre-1.5 styling and is a huge time saver for iOS tabview styling.

### DIFF
--- a/tns-core-modules/ui/tab-view/tab-view.ios.ts
+++ b/tns-core-modules/ui/tab-view/tab-view.ios.ts
@@ -327,7 +327,7 @@ export class TabView extends common.TabView {
 
             var is = imageSource.fromFileOrResource(iconSource);
             if (is && is.ios) {
-                var originalRenderedImage = is.ios.imageWithRenderingMode(UIImageRenderingMode.UIImageRenderingModeAlwaysOriginal);
+                var originalRenderedImage = is.ios.imageWithRenderingMode(UIImageRenderingMode.UIImageRenderingModeAutomatic);
                 this._iconsCache[iconSource] = originalRenderedImage;
                 image = originalRenderedImage;
             }


toddanglin:
@tjvantoll @jlooper  Do either you have any additional context on why this `UIImageRenderingMode` was changed in the past? If there is no good reason to keep the change, the `Automatic` mode seems way more helpful as the default.

tjvantoll:
@toddanglin Unless I’m missing something, I don’t think `UIImageRenderingMode` actually changed in 1.5. Here’s the [commit in question](https://github.com/NativeScript/NativeScript/commit/2b1be2da43f49417808773903850959dc355323a) that landed in 1.5.1; [here’s the PR](https://github.com/NativeScript/NativeScript/pull/1149). It looks like the change was using `imageWithRenderingMode()` instead of setting the `renderingMode` property directly.

I see `UIImageRenderingModeAlwaysOriginal ` being used all the way back to [the original commit that introduced the ability to add images to tabs](https://github.com/NativeScript/NativeScript/commit/10a963e0a5fa4f290ba296bc691ae0d5d175fbb4#diff-5eb4ee77f5b2926368ab0d5c3192ca67) back in March of 2015.

Not sure whether changing `UIImageRenderingMode` would have any repercussions, but I personally can’t think of any, and I like the idea of making `selectedColor` easier 😄

toddanglin:
@tjvantoll Thanks for finding the old commit!

I recall in earlier versions of {N} the tab icon colors did change on iOS, so I wonder if something changed in iOS? Either way, this pull request would simply change the rendering mode to `UIImageRenderingModeAutomatic`, which _seems_ to be a superior option for styling tab icons.

Apps that have already added multiple icons to deal with `AlwaysOriginal` rendering should only have to ensure `selectedColor` is properly set in their `TabView` to avoid any visual change.

If this change is too heavy handed, the next best thing (I suppose) would be a new property on the `TabView` allowing developers to specify "rendering mode"...but long-term I can't see why anyone would want to import more images just to deal with selected state coloring.

hamorphis:
@toddanglin I merged the pull request. Thank you for this fix.

PR #2458, opened 2016-07-15
Unregister animations only once - on reset css

### LINKED ISSUE #2448: Keyframe animation not working on views, that has id selector applied.
Keyframe animation is not executed when an element has an id-selector applied on it:

Reproduce (delete the `#my-button` selector form CSS and the animation is triggered OK):

XML
```
<Page xmlns="http://schemas.nativescript.org/tns.xsd">
  <StackLayout>
    <Button text="TAP" tap="onTap" id="my-button" />
  </StackLayout>
</Page>
```

CSS
```
@keyframes rotate {
    from   { transform: rotate(0); }
    to  { transform: rotate(180); }
}
.animate {
   animation-name: rotate;
   animation-duration: 1s;
   animation-fill-mode: forwards;
}

button {
    font-size: 24;
    background-color: green;
}

#my-button {
    font-size: 24;
    background-color: red;
}
```

TS
```
import { EventData } from "data/observable";
import { View } from "ui/core/view";

export function onTap(args: EventData) {
    const view = <View>args.object;
    view.className = "animate";
}
```

vakrilov:
Resolves #2448 
cc @tzraikov

### DIFF
--- a/tns-core-modules/ui/page/page-common.ts
+++ b/tns-core-modules/ui/page/page-common.ts
@@ -10,14 +10,14 @@ import * as frameModule from "ui/frame";
 import proxy = require("ui/core/proxy");
 import keyframeAnimation = require("ui/animation/keyframe-animation");
 
-var fs: typeof fileSystemModule;
+let fs: typeof fileSystemModule;
 function ensureFS() {
     if (!fs) {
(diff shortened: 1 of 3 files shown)


tzraikov:
👍

PR #2459, opened 2016-07-15
fix: ActionBar's title not updating in OnLoaded event

zh-m:
ActionBar's Title is not updated when bound and bindingContext is set in OnLoaded event.


zh-m:
run ci

vakrilov:
👍

PR #2467, opened 2016-07-18
Fix: Set the binding context of a page automatically when navigated to.

### LINKED ISSUE #731: NavigationEntry.bindingContext property
When you navigate to a page and provide a **bindingContext**, it would be nice this context to become the bindingContext of the page. In this way, you would not have to subscribe to each page's **navigatedTo** event simply to set its bindingContext. Instead you can do it like this:

```
    frameModule.topmost().navigate({
        moduleName: "details-page",
        bindingContext: new DetailsViewModel(id),
        animated: true,
        transition: {name: "slide"}
    });
```

hamorphis:
Resolves #731


vakrilov:
👍

PR #2479, opened 2016-07-20
FIX: isBackNavigation wrong parameter 

vakrilov:
FIX: isBackNavigation wrong when navigation away form page that is not backstack visible

### DIFF
--- a/tns-core-modules/ui/frame/frame.ios.ts
+++ b/tns-core-modules/ui/frame/frame.ios.ts
@@ -10,15 +10,15 @@ import * as types from "utils/types";
 import application = require("application");
 import * as _transitionModule from "ui/transition";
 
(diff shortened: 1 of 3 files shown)


tns-core-modules/ui/frame/frame.ios.ts
```diff
@@ -145,6 +145,8 @@ export class Frame extends frameCommon.Frame {
             viewController.navigationItem.hidesBackButton = this.backStack.length === 0;
 
             // swap the top entry with the new one
+            var skippedNavController = newControllers.lastObject;
```
hshristov: You could use `let` here :)

vakrilov:
run ci

SvetoslavTsenov:
run ci

PR #2485, opened 2016-07-21
Note for deprecated module Border in HOW-TO.md

NickIliev:
related to [https://github.com/NativeScript/docs/issues/333](https://github.com/NativeScript/docs/issues/333)

### DIFF
--- a/tests/app/ui/border/border.md
+++ b/tests/app/ui/border/border.md
@@ -17,3 +17,5 @@ Using borders requires the Border module.
      </Border>
  </Page>
 ```
+
+> Note: Border module is **deprecated**! Use CSS properties instead.


NathanaelA:
@NickIliev -- What CSS property (in NativeScript) do you use if all you want is a single line on say the bottom of a StackLayout; or what item do you use if you want a border only on one side of a Label?    To my knowledge NativeScript's CSS doesn't support borders on a single side.

NickIliev:
@NathanaelA you are absolutely right.. will dismiss this PR than as the border module obviously still has its actual meaning.

NathanaelA:
Rats, I was really hoping I was wrong and missed a brand new cool css borders calls...  :grinning:

NickIliev:
For everyone who is interested in per-sed border applying, you can already do that with CSS styles.
e.g.

```
..btn {
    border-width: 5 10 15 20;
    border-color: red;
}
```

PR #2498, opened 2016-07-23
Add support for translucent ActionBar on iOS

dbbk:
Currently scrollviews get positioned below the ActionBar, meaning they do not appear underneath its translucent background. This pull request adds support for iOS's automaticallyAdjustsScrollViewInsets, which enables this.

See [before](https://drive.google.com/file/d/0B2gJIjdbhRp2SFNBdExfYkNmLVU/view) and [after](https://drive.google.com/open?id=0B2gJIjdbhRp2Skg5MjgzeE1vVkU).

Unfortunately this doesn't have unit tests because I'm not sure how to go about it - but if someone else could pick this up that'd be great.

### DIFF
--- a/tns-core-modules/ui/page/page.ios.ts
+++ b/tns-core-modules/ui/page/page.ios.ts
@@ -40,7 +40,7 @@ class UIViewControllerImpl extends UIViewController {
     public static initWithOwner(owner: WeakRef<Page>): UIViewControllerImpl {
         let controller = <UIViewControllerImpl>UIViewControllerImpl.new();
         controller._owner = owner;
-        controller.automaticallyAdjustsScrollViewInsets = false;
+        controller.automaticallyAdjustsScrollViewInsets = true;
         return controller;
     }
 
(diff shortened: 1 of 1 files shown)


hshristov:
Hi @dbbk,

Thank you for your contribution. I think it will be safer if we add new property on `ActionBar` or `Page` so that developers can change it. This way we won't introduce behaviour change that cannot be disabled.

The deleted code handles sizing page content and showing modal pages on iPad & iPhone in landscape & portrait? Have you checked the above cases  still works as expected with this change?

As for unit-tests we have https://github.com/NativeScript/NativeScript/blob/master/tests/app/ui/page/page-tests.ios.ts
where a test could be added. I think a test that assert the measure size of the content when translucent property is true should be good enough.

We are in a code freeze now but once 2.2.0 is out we could help with implementation & test.

dtopuzov:
run ci

dtopuzov:
Two unit tests fails:
```
iOS.SCROLL-VIEW.test_scrollView_persistsState_vertical
iOS.PAGE.test_page_no_anctionBar_measure_no_spanUnderBackground_measure_layout_size_isCorrect
```

DomGaud:
This modification worked for me, but when using a **_ListView_** within the Page the height of the listview container will always be to short by the height of the status bar and actionbar. 
So to correct the height calculation I also had to change the following in the **`onMesure`** function in the same file

From:
```Javascript
var heightSpec = utils.layout.makeMeasureSpec(height - actionBarHeight - statusBarHeight, heightMode);
```
To:
```Javascript
var heightSpec = utils.layout.makeMeasureSpec(height, heightMode);
```

valentinstoychev:
closing due to inactivity and lack of response.

PR #2500, opened 2016-07-25
Update grunt-typedoc to make it get new TS features

ErjanGavalji:
No tests. The change is in the build of the documentation, which was broken due to using new TypeScript features. The TypeDoc referred was old, hence the problem.


PanayotCankov:
:+1: :heart:

PR #2501, opened 2016-07-25
FIX: Wrong isBackNavigation value

vakrilov:
FIX: `isBackNavigation=true` in navigation event when navigating **forward** with clearHistory in ios.

Similar to #2479 
cc: @hshristov


vakrilov:
run ci

vakrilov:
run ci

PR #2505, opened 2016-07-25
Add npm script that generates ios .d.ts-es from the tests app

PanayotCankov:
The iOS typings in `tns-core-modules/ios/ios-*` can now be generated automatically using the tests app using:
```
npm run dts-ios
```
(mind the $PWD inside the script may brake should you have spaces in the path)

A lot of enum usages were updated. There are now type parameters to collections. ~~Promise differes due to the updated definitions from the typescript, PromiseLike replaces Thenable~~.

[x] We must still ensure the tests/ and app/ build properly.
[x] Angular builds properly.


PanayotCankov:
uitests

PanayotCankov:
cuteness

jasssonpet:
Ideally this would be generated on the user's machine so that differences between different iOS SDK versions in the runtime metadata and the declared types are caught at compile time.

PanayotCankov:
I agree, but it is a completely different story. For now we are looking to integrate the latest .d.ts-es and allow them to be easily updated. Dropping them from source control will require windows users to build the .d.ts-es on mac when working on windows + android. If everything is fine, when the time is right we will call the dts-ios on postinstall.

PanayotCankov:
This involves several changes in the modules so we want to have enough time to test it. We will merge this PR at the start of v2.3.0.

vchimev:
uitests

vchimev:
cuteness

PR #2512, opened 2016-07-26
Fix iOS font registration when bundling scripts.

hdeshev:
Use fs.knownFolders to get the app dir instead of the __dirname hack


hdeshev:
Fixes NativeScript/nativescript-dev-webpack#12

PR #2522, opened 2016-07-28
Fixed: CSS is not automatically livesynced in angular apps

tzraikov:
https://github.com/NativeScript/nativescript-cli/issues/1958

ping @vakrilov , @hshristov


vakrilov:
👍

PR #2539, opened 2016-08-01
Element that isn't a View should not be added to its parent

zh-m:
An element that is not an instance of a View should not be added directly to its parent and thus triggering the logic for setting any inherited or style properties. 
For example - ActionBar.actionItems in ng app.


vchimev:
run ci

Issue #2541, opened 2016-08-01
Cannot target es2015/es6 to use async/await


alexrainman:
When i change target to es2015/es6 in tsconfig.json i get lot of errors

```
node_modules/typescript/lib/lib.es6.d.ts(1342,16): error TS2314: Generic type 'Promise<T, R>' requires 2 type argument(s)
```

I also tried this tutorial without luck: https://www.nativescript.org/blog/details/using-babel-in-nativescript-apps

Any ideas?

hdeshev:
Hi @alexrainman,

This error is caused by a clash in the `es6-promise.d.ts` that NativeScript brings along and the TypeScript `lib.es6.d.ts` typings. We ship a separate ES6 typing that you need to reference in your `references.d.ts` -- `tns-core-modules.es6.d.ts`, and it should have be configured for ES6 builds.

Well, reality kicked in, and it seems we'd introduced public types that return promises, but are defined as using a promise-like interface type - `Thenable<T>`. You can work around that by aliasing the `Thenable<T>` type to `PromiseLike<T>`. The full `references.d.ts` file should now look like:

```TypeScript
/// <reference path="./node_modules/tns-core-modules/tns-core-modules.es6.d.ts" /> Needed for autocompletion and compilation.
declare type Thenable<T> = PromiseLike<T>;
```

Full demo project available [here](https://github.com/hdeshev/tns-ts-es2015-demo).

//cc @NickIliev

alexrainman:
@hdeshev this is cool, but as i am developing a plugin, i need other lines in my references files

```
/// <reference path="./node_modules/tns-core-modules/tns-core-modules.es6.d.ts" /> Needed for autocompletion and compilation.
/// <reference path="./node_modules/tns-platform-declarations/ios.d.ts" />
/// <reference path="./node_modules/tns-platform-declarations/android17.d.ts" />
/// <reference path="./node_modules/tns-platform-declarations/org.nativescript.widgets.d.ts" />
declare type Thenable<T> = PromiseLike<T>;
```

The project compiles, but i get exception. First it was showing that (android) namespace cannot be found and now i get this

```
Error: Trying to link invalid 'this' to a Java object
```

Any ideas?

Something else, my plugin will need async/await from es6. Will this be available by default in {N} anytime soon?

Thanks

hdeshev:
[...]

PR #2543, opened 2016-08-02
Rename Thenable<T> to PromiseLike<T>

hdeshev:
Also change the Promise generic param from R to T.

Make it similar to TypeScript's lib.es6.d.ts until we actually
remove those typings.

/cc @PanayotCankov


hdeshev:
Related to #2541

hdeshev:
run ci

PR #2549, opened 2016-08-04
Make package mains more webpack-friendly

hdeshev:
Chop off the `.js` ending in the `main` setting for all package.json files, so that it gets easier for webpack to redirect stuff like page.js to page.android.js.

I added a validation step that will fail the build if it finds a `"main": "blah.js"` setting.


enchev:
:+1:

vchimev:
👍

PR #2567, opened 2016-08-11
Added permission for location.

hshristov:
Without it Location tests fails on some devices (Samsung S4)

### DIFF
--- a/tests/app/App_Resources/Android/AndroidManifest.xml
+++ b/tests/app/App_Resources/Android/AndroidManifest.xml
@@ -18,6 +18,7 @@
 	<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
 	<uses-permission android:name="android.permission.INTERNET"/>
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+	<uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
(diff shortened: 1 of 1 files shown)


dtopuzov:
:100:

NathanaelA:
Why exactly would we be putting this as a DEFAULT permission; if my app doesn't use any "location" services; this is not a permission I need to have as a default permission.  Shouldn't the default permissions be minimal...    To be honest, imho the "internet" permissions should be stripped also.

hshristov:
@NathanaelA I'm not putting this in the default template. This is in the manifest.xml used by the unit-test app just to make sure unit-tests pass on Samsung devices.
As for the `Internet` permission you are correct. The best approach would be to make `http` module - plugin so that it brings its own permissions. But I'm not sure we want to make such fundamental change.

hshristov:
Same goes for camera module.

NathanaelA:
@hshristov Ah; I totally glossed over the "tests" as the prefix to the file location.  My Bad....   You had me a little concerned that we were going to start adding a ton of permissions to the default app.   Forgive me for reacting...

As for internet permission the purist in me say "remove it" -- but in reality it probably actually causes way less support issues by being left in, so ignore me... ;-)

hshristov:
@NathanaelA No problem :)

PR #2574, opened 2016-08-12
Fixed creating Observable object from nested JSON object.

nsndeck:

### DIFF
--- a/tns-core-modules/data/observable/observable.ts
+++ b/tns-core-modules/data/observable/observable.ts
@@ -51,18 +51,37 @@ export class Observable implements definition.Observable {
 
     private _observers = {};
 
+    public static fromJSON(json: any): Observable {
+        let observable = new Observable();
+        observable.addPropertiesFromJSON(observable, json, false);
+        return observable;
+    }
+
+    public static fromJSONRecursive(json: any): Observable {
+        let observable = new Observable();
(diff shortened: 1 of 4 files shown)


tns-core-modules/data/observable/observable.ts
```diff
@@ -51,21 +51,40 @@ export class Observable implements definition.Observable {
...
+        observable.addPropertiesFromJSON(observable, json, true);
+        return observable;
+    }
+
+    private addPropertiesFromJSON(observable: Observable, json: any, recursive?: boolean) {
+        let isRecursive = recursive || false;
+        observable._map = new Map<string, Object>();
+        for (var prop in json) {
+            if (json.hasOwnProperty(prop)) {
+                if (isRecursive) {
```
hshristov: You could use `recursive` param directly. If not set it will be `undefined` so it won't pass the check.

tns-core-modules/data/observable/observable.ts
```diff
@@ -51,21 +51,40 @@ export class Observable implements definition.Observable {
...
+                    if (!Array.isArray(json[prop]) && typeof json[prop] === 'object' && types.getClass(json[prop]) !== 'ObservableArray') {
```
hshristov: Can we use `instanceof ObservableArray` instead of this check:
```types.getClass(json[prop]) !== 'ObservableArray'```

PR #2591, opened 2016-08-16
Label zero height issue fixed.

nsndeck:
Rendering issue with a Label rendered for the first time with an empty string as text. This results in a zero height of the label and due to bug after changing text later layout request is rejected and text is not visible.


dtopuzov:
run ci

SvetoslavTsenov:
run ci

PR #2598, opened 2016-08-17
Fixed: There is an exception when removing a button

tzraikov:
1. Create a new {N} project.
2. On a button click add a new button
3. On a second button click, remove the button
4. Run and test the app

#2599


enchev:
:+1:

PR #2610, opened 2016-08-22
Color of buttons inside dialogs is not applied if there'are nested css styles

zh-m:
The color of buttons inside dialogs is not modified in case the following CSS is defined:

Page Button {
    background-color: #30BCFF;
    color: #FFFFFF;
}

 Button {
    font-size: 15;
    color: springgreen; 
}

In this case, the color of a button inside dialog will be still "#ffffff", while the desired one is "springgreen".

### DIFF
--- a/tns-core-modules/ui/dialogs/dialogs.android.ts
+++ b/tns-core-modules/ui/dialogs/dialogs.android.ts
@@ -39,6 +39,20 @@ function showDialog(builder: android.app.AlertDialog.Builder) {
             }
         }
     }
+
+    var backgroundColor = dialogsCommon.getButtonColor();
+    if (backgroundColor) {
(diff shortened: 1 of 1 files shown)


tns-core-modules/ui/dialogs/dialogs.android.ts
```diff
@@ -39,6 +39,20 @@ function showDialog(builder: android.app.AlertDialog.Builder) {
             }
         }
     }
+
+    var backgroundColor = dialogsCommon.getButtonColor();
```
hshristov: This looks like foregroundColor not background.

dtopuzov:
run ci

dtopuzov:
run ci

enchev:
run ci

PR #2617, opened 2016-08-23
Fix: Button with text-decoration can't change text

### LINKED ISSUE #2615: Button with text-decoration cannot change its text
The problem is that a Button which has text-decoration has its `setAttributedTitleForState` method called. When we change the text and call only the `setTitleForState` method nothing changes. We have to call the `setAttributedTitleForState` method as well.

hamorphis:
Resolves #2615


hamorphis:
uitests

PR #2643, opened 2016-08-30
Fixed issue with nested observables when null is given as property va…

nsndeck:


tns-core-modules/data/observable/observable.ts
```diff
@@ -249,30 +249,30 @@ export class Observable implements definition.Observable {
     }
 }
 
-function addPropertiesFromJSON(observable: Observable, json: any, recursive?: boolean) {
+function addPropertiesFromObject(observable: Observable, source: any, recursive?: boolean) {
     let isRecursive = recursive || false;
     observable._map = new Map<string, Object>();
-    for (var prop in json) {
-        if (json.hasOwnProperty(prop)) {
+    for (var prop in source) {
```
hshristov: use `let` instead of `var`

PR #2678, opened 2016-09-05
Fixes #2621: Add accessibility labels to views

### LINKED ISSUE #2621: Feature request: Add accessibility labels to views.
The App we're developing must be made accessible for our visually impaired users.
For that we'd like to have support for the relevant properties on Views on both `iOS` and `Android`.

For starters we need to be able to define the `accessibilityLabel` on a View, for iOS' VoiceOver and Android's TalkBack to read.

### Tell us about the problem
{N}'s `ui/core/view` have the property `automationText`, which maps to:
* on iOS: [UIAccessibilityIdentification.accessibilityIdentifier](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIAccessibilityIdentification_Protocol/#//apple_ref/occ/intfp/UIAccessibilityIdentification/accessibilityIdentifier)
* on Android: [android:contentDescription](https://developer.android.com/reference/android/view/View.html#attr_android:contentDescription)

As I see it the properties used on the two platforms doesn't provide the same functionality:
* `UIAccessibilityIdentification` on iOS is as I understand it used for UI Automation script, not for VoiceOver
* `contentDescription` on Android is used primarily for accessibility for views that doesn't have a textual representation, so TalkBack can tell the user what view is.

`automationText` is just mentioned with explanation in the API docs, so we wrongly guessed this was the property we could use, since it does what we need on Android. It doesn't on iOS.

So we suggest adding a property `accessibilityLabel` for the View-class, which would map:
* on iOS: [accessibilityLabel](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIAccessibility_Protocol/#//apple_ref/occ/instp/NSObject/accessibilityLabel)
* on Android: [android:contentDescription](https://developer.android.com/reference/android/view/View.html#attr_android:contentDescription)

We'd like to help implement support for accessibility in {N}, if you're interested in adding it upstream.

I know the labels are only part of what will be needed, but it'll be good place to start.
Things are like accessibility-events/notifications can come later.

Are you interested in us contributing this and do you have any suggestions?

m-abs:
### Fixes/Implements #2621

As suggested by @tsonevn I'm updating the `automationText` property on `ui/view` on iOS to set both `accessibilityIdentifier` and `accessibilityLabel`.

This fixes basic `VoiceOver` behavior on iOS.

### DIFF
--- a/tests/app/ui/view/view-tests.ios.ts
+++ b/tests/app/ui/view/view-tests.ios.ts
@@ -79,4 +79,5 @@ export function test_automation_text_set_to_native() {
     var newButton = new button.Button();
     newButton.automationText = "Button1";
     TKUnit.assertEqual((<UIView>newButton.ios).accessibilityIdentifier, "Button1", "accessibilityIdentifier not set to native view.");
+    TKUnit.assertEqual((<UIView>newButton.ios).accessibilityLabel, "Button1", "accessibilityIdentifier not set to native view.");
 }
--- a/tns-core-modules/ui/core/view.ios.ts
+++ b/tns-core-modules/ui/core/view.ios.ts
@@ -21,6 +21,7 @@ global.moduleMerge(viewCommon, exports);
 function onAutomationTextPropertyChanged(data: dependencyObservable.PropertyChangeData) {
     var view = <View>data.object;
     view._nativeView.accessibilityIdentifier = data.newValue;
+    view._nativeView.accessibilityLabel = data.newValue;
 }
 (<proxy.PropertyMetadata>viewCommon.View.automationTextProperty.metadata).onSetNativeValue = onAutomationTextPropertyChanged;
 


dtopuzov:
:100:

enchev:
run ci

csnthsh120:
Previously when we call accessibilityLabel property of the element based on automationText property assigned to it then it used to return the Displayed Text in that control but now it is returning automationText property of the control instead of displayed text. Could you please see whether this fix has anything to do with that?

m-abs:
Hi @csnthsh120,

Why would you expect the `automationText`-property to return the displayed text? Why not the `text`-property?

All this PR does is to set both `accessibilityLabel` and `accessilityIdentifier` on iOS, it doesn't change the getter for the value.
If no value i set, you should always have gotten the default value which is: `undefined`.

m-abs:
I don't see why value is set to the same value as `automationText`, I'd expect only `name` (assuming it's `accessibilityIdentifier`) and `label` (assuming it's `accessibilityLabel`)
I haven't used `appium inspector` so I don't know how it works.

A quick check could be, if you manually patch your local copy of `tns-core-modules`
Edit `node_modules/tns-core-modules/ui/core/view.ios.js` and comment out line 12:
```typescript
view._nativeView.accessibilityLabel = data.newValue;
```

But I think you need to create a bug report about this, preferably with a code example.

csnthsh120:
HI, @m-abs I already tried that commenting out the code but somehow it is messing up my whole project if I comment out the line of code you have added, will try again and if it not is not resolved by commenting out the line then I will create a bug for this.
Thanks for the quick response.

Issue #1808, opened 2016-03-19
Image out memory exception on Android


AtixD:
How to clear the memory. I dynamically change the image, and then crash. How to fix it?
![2016-03-19 17 36 07](https://cloud.githubusercontent.com/assets/17594120/13899614/abe4371e-ee03-11e5-9bae-201da7469773.png)
```

   getViewBtn(num,view)
    {
        switch (num) {
            case 0:
                return view.getViewById("first");
            case 1:
                return view.getViewById("second");
            case 2:
                return view.getViewById("third");
            case 3:
                return view.getViewById("fourth");
        
            default:
                break;
        }
    }
```
```
var arr = this.genForAnimate(this.index);
                
                var img_1 = <Image> this.getViewBtn(0,this.page);
                var img_2 = this.getViewBtn(1,this.page);
                var img_3 = this.getViewBtn(2,this.page);
                var img_4 = this.getViewBtn(3,this.page);
                
                allView.animate({
                    opacity : 0,
                    duration : 500
                }).then(function () {              
                    
 
                    img_1.src = arr[0];
                    img_2.src = arr[1];
                    img_3.src = arr[2];
                    img_4.src = arr[3];
                    
                    grid.parent.getViewById("enView").text = en;
                    grid.parent.getViewById("enView").opacity = 0;
                    
                    return allView.animate({
                        opacity : 1,
                        duration : 500
                    }).then(function() {
                       tts.speak(en); 
                    });
                });  
```

bradmartin:
What's your interface? Also what are the file sizes of the images? Last what's the memory on the device/emulator you're testing with?

enchev:
Hey @AtixD,

Please check my reply to this issue:
https://github.com/NativeScript/NativeScript/issues/1801

AtixD:
@bradmartin  There is an application crash after reloading images with alpha channel. Device Asus ZenFone 2 Laser (2GB).

Issue #123, opened 2015-05-06
Out of memory, Bitmap images on Android


duongnhatduy:
Hi,

I'm trying to make something similar to sample-cuteness, loading a listview of images and save it by using image-cache. It works fine on iOS but crashes on android. 

I/dalvikvm-heap( 3757): Forcing collection of SoftReferences for 4506492-byte allocation
I/dalvikvm-heap( 3757): Clamp target GC heap from 33.083MB to 32.000MB
D/dalvikvm( 3757): GC_BEFORE_OOM freed 71K, 3% free 31709K/32680K, paused 6ms, total 6ms
E/dalvikvm-heap( 3757): Out of memory on a 4506492-byte allocation.
I/dalvikvm( 3757): "AsyncTask #2" prio=5 tid=13 RUNNABLE
I/dalvikvm( 3757):   | group="main" sCount=0 dsCount=0 obj=0xb323b7c0 self=0xb8540bc0
I/dalvikvm( 3757):   | sysTid=3775 nice=10 sched=3/0 cgrp=[fopen-error:2] handle=-1202450384
I/dalvikvm( 3757):   | state=R schedstat=( 0 0 0 ) utm=20 stm=20 core=0
I/dalvikvm( 3757):   at android.graphics.BitmapFactory.nativeDecodeByteArray(Native Method)
I/dalvikvm( 3757):   at android.graphics.BitmapFactory.decodeByteArray(BitmapFactory.java:500)
I/dalvikvm( 3757):   at com.tns.Async$Http$RequestResult.readResponseStream(Async.java:185)
I/dalvikvm( 3757):   at com.tns.Async$Http$HttpRequestTask.doInBackground(Async.java:264)
I/dalvikvm( 3757):   at com.tns.Async$Http$HttpRequestTask.doInBackground(Async.java:214)
I/dalvikvm( 3757):   at android.os.AsyncTask$2.call(AsyncTask.java:288)
I/dalvikvm( 3757):   at java.util.concurrent.FutureTask.run(FutureTask.java:237)
I/dalvikvm( 3757):   at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
I/dalvikvm( 3757):   at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
I/dalvikvm( 3757):   at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
[...]

PR #2679, opened 2016-09-05
(Bugfix) Android Image memory leak, resolves #2571

### LINKED ISSUE #2571: [Android] memory leak for Images
In case an image source has been changed in several times there is java.lang.OutOfMemoryError exception.

java.lang.RuntimeException: An error occured while executing doInBackground()
at android.os.AsyncTask$3.done(AsyncTask.java:300)
at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:355)
at java.util.concurrent.FutureTask.setException(FutureTask.java:222)
at java.util.concurrent.FutureTask.run(FutureTask.java:242)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
at java.lang.Thread.run(Thread.java:818)
Caused by: java.lang.OutOfMemoryError: Failed to allocate a 8032012 byte allocation with 1496439 free bytes and 1461KB until OOM
at dalvik.system.VMRuntime.newNonMovableArray(Native Method)
at android.graphics.BitmapFactory.nativeDecodeByteArray(Native Method)
at android.graphics.BitmapFactory.decodeByteArray(BitmapFactory.java:634)
at org.nativescript.widgets.Async$Http$RequestResult.readResponseStream(Async.java:350)
at org.nativescript.widgets.Async$Http$HttpRequestTask.doInBackground(Async.java:428)
at org.nativescript.widgets.Async$Http$HttpRequestTask.doInBackground(Async.java:378)
at android.os.AsyncTask$2.call(AsyncTask.java:288)
at java.util.concurrent.FutureTask.run(FutureTask.java:237)
... 3 more

here is the code witch I used

```
import {Component} from "@angular/core";

@Component({
    selector: "my-app",
    template: `
        <StackLayout>
            <Image src="{{src}}" stretch="fill" (tap)="change()"></Image>
        </StackLayout>
    `
})
export class AppComponent {
    src;

    constructor() {
        this.change();
    }

    change() {
        this.src = 'https://placekitten.com/1000/200' + parseInt('' + Math.random() * 9);
    }
}
```

tns --version: 2.1.0
tns-core-modules: 2.2.0
tns-android 2.1.1

colinskow:
This PR fixes a severe memory leak in the Image component on Android. Images loaded via `<Image src="..." />` are kept in memory forever even after the source is changed or the component is unloaded. This causes image intensive applications to quickly run out of memory and crash.

This problem is discussed at length in #2571 and is likely the culprit behind #123, #1808, and nativescript/android-runtime#77.

My solution is to track Image sources loaded via resource, file or URL and call `recycle()` on the Bitmap when either the source is swapped or the component is unloaded. When an existing ImageSource or native image is passed in as a source it is left alone, as these are likely to be reused later.

I've created tests for this issue at:
https://github.com/colinskow/tns-image-test

I need help to turn them into automated unit tests in this repo, as I am not sure how to do that. But they pass with my pull request and fail on both 2.2.1 and master. By failure I mean the app crashes with Out Of Memory.

This PR handles the image problem and stops apps from crashing. But it may also be a bandaid for a more serious bug in the NativeScript Android runtime that is causing hard references for all images loaded to be held onto forever.

### DIFF
--- a/tns-core-modules/ui/image/image-common.ts
+++ b/tns-core-modules/ui/image/image-common.ts
@@ -86,7 +86,16 @@ export class Image extends view.View implements definition.Image {
     /**
      * @internal
      */
+    
+    // Track the source of the image to fix a memory leak on Android
+    // FALSE for file, resource, or URL
+    // These will be recycled when the src changes or the component is unloaded
+    // TRUE for a passed in ImageSource or NativeImage. These will be left alone.
+    _wasNativeSource: boolean;
+    _isNativeSource: boolean;
+
     _createImageSourceFromSrc(): void {
+        this._wasNativeSource = this._isNativeSource || false;
         var value = this.src;
         if (types.isString(value)) {
             value = value.trim();
@@ -101,6 +110,7 @@ export class Image extends view.View implements definition.Image {
                 if (!types.isString(this.src) || value !== currentValue.trim()) {
                     return;
                 }
+                this._isNativeSource = false;
                 this.imageSource = source;
                 this._setValue(Image.isLoadingProperty, false);
             }
@@ -145,11 +155,13 @@ export class Image extends view.View implements definition.Image {
             }
         }
         else if (value instanceof imageSource.ImageSource) {
+            this._isNativeSource = true;
             // Support binding the imageSource trough the src property
             this.imageSource = value;
             this._setValue(Image.isLoadingProperty, false);
         }
         else {
+            this._isNativeSource = true;
             this.imageSource = imageSource.fromNativeSource(value);
             this._setValue(Image.isLoadingProperty, false);
         }
--- a/tns-core-modules/ui/image/image.android.ts
+++ b/tns-core-modules/ui/image/image.android.ts
@@ -61,12 +61,32 @@ export class Image extends imageCommon.Image {
         return this._android;
     }
 
+    public onUnloaded() {
+        super.onUnloaded();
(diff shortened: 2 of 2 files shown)


enchev:
run ci

hshristov:
Hi @colinskow,

Thank you for your PR.
I've taken your approach and did simple version of your app without using Angular where ImageView `src` is changed in `setInterval` function. The test crash as you describe it.

While profiling another app we found two memory leaks that are fixed in master branch and will go live with 2.3.0. Still the case of changing ImageSource in a loop (or in `setInterval`) cause OOM exception. Calling `recycle` on the Bitmap solve this problem but there is no way to know when ImageSource can be recycled safely. For example on a second iteration I've changed the app to use image-cache module and it throws an error that bitmap is recycled.

ImageView current implementation uses image-source module to load images. This means that  native bitmap can be collected only after JavaScript imageSource instance is collected. This is not exactly memory leak but more garbage synchronisation issue.

With fixes in 2.3.0 if you perform navigation you probably won't see OOM Exception. With the next release (probably 2.3.1) we are going to change the ImageView implementation so it won't use image-source module but loads bitmaps directly in Java. This will allow Java garbage collector to release bitmaps once they are unused by ImageView widget.

I hope that this clarifies the problem.

PR #2702, opened 2016-09-12
Ios10 switch

PanayotCankov:
Add checks if some key iOS methods has been promoted to properties in the iOS 10 SDK.


dtopuzov:
💯

jasssonpet:
How would this work with a property that returns a function?

PanayotCankov:
It won't. It is mentioned in the "getter" docs it won't work with delegate type properties.
:"(

jasssonpet:
Can we check the Xcode version or the iOS SDK version being used and make that the condition for branching? I think it should be possible.

PanayotCankov:
I am not sure. We can't use the iOS version nor the Xcode, but if we have the SDK version it should be ok.

PR #2732, opened 2016-09-16
Add timeout warning log for test runner. Increase timeout to 1200

hdeshev:
Tinkering with the expect script...


PanayotCankov:
All tests ran in travis but test_AnimatingProperties failed. This fixes the global timeout and the t/o messaging is better.

Issue #2796, opened 2016-09-23
ActionItems lacks proper support for VoiceOver on iOS


m-abs:
The text-property on `ActionItem`s on the `ActionBar` isn't read by `VoiceOver` on `iOS`, only the icon name is and `ActionItem` doesn't have the property `automationText`. That's because `ActionItem` doesn't inherit from `ui/core/view`.

On iOS `UIBarButtonItem` is used to create the `ActionItem`, but `UIBarButtonItem` doesn't implement the `UIAccessibility`-protocol, but according to this stackoverflow http://stackoverflow.com/questions/23800764/ios-7-uibarbuttonitem-with-accessibility it should be possible it activate it anyways.

### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 2.3.0
- Cross-platform modules: 2.3.0
- Runtime(s): tns-ios 2.3.0

### Please tell us how to recreate the issue in as much detail as possible.
* Create an `ActionBar` with an `ActionItem` set a text-label on the ActionItem
* Run the app on iOS with VoiceOver enabled
* Listen to the text being read, it's the icon-name followed by "button"

m-abs:
# Update:
I think I see the problem:
https://github.com/NativeScript/NativeScript/blob/2.3.0/tns-core-modules/ui/action-bar/action-bar.ios.ts#L134-L163

If we have an icon on the `ActionItem` the text is never set on the `UIBarButtonItem`, so of cause iOS' VoiceOver can't read the text.

I'm working on a PR that hopefully solves the problem.

vchimev:
Hi @m-abs,

Thank you for the pull request.

Could you please send us a sample page which demonstrates the issue along with:
- how does VoiceOver read it currently and
- how is it expected to read it?

In this way, we would be able to review your pull request properly.

For example, my test page is as follow:
```
<Page>
  <Page.actionBar>
    <ActionBar >
      <ActionBar.actionItems>
        <ActionItem text="Item" />
        <ActionItem>
          <ActionItem.actionView>
            <Button text="View" automationText="test" />
          </ActionItem.actionView>
        </ActionItem>
      </ActionBar.actionItems>
    </ActionBar>
  </Page.actionBar>
  <StackLayout>
    <Label text="This is a test."/>
  </StackLayout>
</Page>
```
It is read as follows: *Item -> Button, Test -> Button*

m-abs:
HI @vchimev,

[...]

PR #2799, opened 2016-09-26
Fixes: #2796 - ActionItems lacks proper support for VoiceOver on iOS

m-abs:
fixes: #2796 - ActionItems lacks proper support for VoiceOver on iOS

The `UIBarButtonItem` element wasn't always created with a title/text so `VoiceOver` couldn't know what to read for the user.
This PR marks the element as an `AccessibilityElement` and sets the accessibilityLabel to the title/text if one is set.

I've chosen to mark the `UIBarButtonItem` with just `UIAccessibilityTraitButton`-trait, because I see it as the most general trait I could set.


dtopuzov:
run ci

vchimev:
uitests

vakrilov:
👍

PR #2822, opened 2016-09-29
Refactored image loading from camera.

nsndeck:
Added image-asset class (which instance will be returned from nativescript-camera plugin). Also some modifications to image-source (to support image-asset) and image (rotation logic moved from camera to image (native java)).


dtopuzov:
run ci

vchimev:
run ci

Issue #2640, opened 2016-08-29
Extending SegmentedBar to remove borders


gsaadeh:
I have a requirement where I want to remove the borders from the SegmentedBar on iOS. I figured the best way to do this is by extending the existing SegmentedBar and overriding the insert tab method. In the insert tab event, I can set the background to a clear image and even remove the separators. 

In my components folder, I've created a custom plugin as follows:

.common file:

```
import * as app from 'application';
import * as dialogs from 'ui/dialogs';
import definition = require("ui/segmented-bar");

export class MSegmentedBar extends definition.SegmentedBar {
    constructor() {
        super();
        console.log('Plugin is working!!');
    }
}
```

.ios file:

```
import common = require("./msegmentedbar.common");

export class MSegmentedBar extends common.MSegmentedBar {
      constructor() {
        super();
      }
}
```

and then reference it in my view:

```
<Border xmlns:x="components/msegmentedbar/msegmentedbar">
        <GridLayout rows="auto, *">
            <x:MSegmentedBar id="group-bar">
                <SegmentedBar.items>
                    <SegmentedBarItem title="Now" selectedTextColor="#fff" borderColor="transparent" />
                    <SegmentedBarItem title="Today" selectedTextColor="#fff" />
                    <SegmentedBarItem title="Week" selectedTextColor="#fff" />
                    <SegmentedBarItem title="Month" selectedTextColor="#fff" />
                </SegmentedBar.items>
            </x:MSegmentedBar>
...
```

Because not all the delegates are exported, the app runs, shows the console.log messages but doesn't render the items. If I do any sort of binding on the selectedIndex, it throws an exception. 

Is it possible at all to make this work or should we wait till the team completely exposes all the delegates?
### Additional info

NativeScript: 2.2.1
tns-core-modues: 2.2.1
Platform: iOS

enchev:
Hey @gsaadeh,

In my opinion you should extend first the native class (UISegmentedControl) and later use it in a JavaScript wrapper [simmilar to our SegmentedBar](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/segmented-bar/segmented-bar.ios.ts#L98). You can check various examples on how to remove UISegmentedBar borders: http://stackoverflow.com/questions/9314927/remove-uisegmentedcontrol-separators-completely-iphone

gsaadeh:
[...]

Issue #2471, opened 2016-07-19
Enhancement: Expose delegate classes externally from different modules


NathanaelA:
### Did you verify this is a real problem by searching 
yes

### Tell us about the problem
Frequently I would like to extend some behavior of NativeScript to handle a special case.   I've ran into this issue multiple times now.   The latest one is the TextEdit; on iOS you have to override the `shouldChangeCharactersInRange` to allow some additional functionality; however TNS currently creates its own delegate to handle several text edit items.    So I really can't create and add my own delegate....

So the solutions would be:
1. Telerik adds some sore of delegate splitter class (so multiple delegates could be attached).
2. Telerik just needs to expose the delegate class via the module.exports so I can monkey patch the delegate and then I'm responsible to add in the additional support if need be.
3. I have to clone any class I want to change and make my own "custom" version. 
4. I have to modify the actual tns-core-modules source directly.

For me, the simplest solution is if you started to expose all the delegate classes via exports so that I can monkey patch the prototype before it is created.

If the team is willing to take a patch to fix this issue; I'll work through them and create a patch that eliminates the private on them.

### Which platform(s) does your issue occur on?
iOS & Android

### Please provide the following version numbers that your issue occurs with:
All versions 2.1 and prior.

vip-git:
Hi @NathanaelA 

Do you know if this is going to resolved sooner or later ? Anyone from nativescript ?

Thanks,
V

NathanaelA:
@vip-git - Unfortunately the Core modules team disagreed to allow us access to these; so even though I did all the work in #2827 it never was accepted.  (Long discussion in 2827). 

I have a [TNS-Core-Patcher](https://github.com/NathanaelA/nativescript-dev-tnscorepatcher) project that I am working on that will allow us to bypass the issue; but I haven't had much time to work on it because of several other projects that I am in the process of finishing...

PR #2827, opened 2016-10-01
Export several more missing iOS delegates to allow modification

NathanaelA:
This will allow us to modify the delegates before they are used to add any missing functionality.  I have personally ran into this several times, UINavigationController, Label and WebView have all been issues because the delegates are hard coded and so you can't extend them to add new or fix functionality.

Fixes/Implements #2640, #2471
### You have [unit tests]

No added unit tests needed; just added the "export" to the typescript code so that the delegates are exported.

### DIFF
--- a/tns-core-modules/ui/animation/animation.ios.ts
+++ b/tns-core-modules/ui/animation/animation.ios.ts
@@ -41,7 +41,7 @@ interface IOSView extends viewModule.View {
     _isPresentationLayerUpdateSuspeneded();
 }
 
-class AnimationDelegateImpl extends NSObject implements CAAnimationDelegate {
+export class AnimationDelegateImpl extends NSObject implements CAAnimationDelegate {
 
     public nextAnimation: Function;
 
--- a/tns-core-modules/ui/button/button.ios.ts
+++ b/tns-core-modules/ui/button/button.ios.ts
@@ -9,7 +9,7 @@ import dependencyObservable = require("ui/core/dependency-observable");
 import types = require("utils/types");
 import {PseudoClassHandler} from "ui/core/view";
 
-class TapHandlerImpl extends NSObject {
+export class TapHandlerImpl extends NSObject {
     private _owner: WeakRef<Button>;
 
     public static initWithOwner(owner: WeakRef<Button>): TapHandlerImpl {
--- a/tns-core-modules/ui/date-picker/date-picker.ios.ts
+++ b/tns-core-modules/ui/date-picker/date-picker.ios.ts
@@ -99,7 +99,7 @@ export class DatePicker extends common.DatePicker {
     }
 }
 
-class UIDatePickerChangeHandlerImpl extends NSObject {
+export class UIDatePickerChangeHandlerImpl extends NSObject {
 
     private _owner: WeakRef<DatePicker>;
 
--- a/tns-core-modules/ui/dialogs/dialogs.ios.ts
+++ b/tns-core-modules/ui/dialogs/dialogs.ios.ts
@@ -10,7 +10,7 @@ import getter = utils.ios.getter;
 
 global.moduleMerge(dialogsCommon, exports);
 
-class UIAlertViewDelegateImpl extends NSObject implements UIAlertViewDelegate {
+export class UIAlertViewDelegateImpl extends NSObject implements UIAlertViewDelegate {
     public static ObjCProtocols = [UIAlertViewDelegate];
 
     static new(): UIAlertViewDelegateImpl {
@@ -29,7 +29,7 @@ class UIAlertViewDelegateImpl extends NSObject implements UIAlertViewDelegate {
     }
 }
 
-class UIActionSheetDelegateImpl extends NSObject implements UIActionSheetDelegate {
+export class UIActionSheetDelegateImpl extends NSObject implements UIActionSheetDelegate {
     public static ObjCProtocols = [UIActionSheetDelegate];
 
     static new(): UIActionSheetDelegateImpl {
--- a/tns-core-modules/ui/frame/frame.ios.ts
+++ b/tns-core-modules/ui/frame/frame.ios.ts
@@ -411,7 +411,7 @@ class TransitionDelegate extends NSObject {
 
 const _defaultTransitionDuration = 0.35;
 
-class UINavigationControllerAnimatedDelegate extends NSObject implements UINavigationControllerDelegate {
+export class UINavigationControllerAnimatedDelegate extends NSObject implements UINavigationControllerDelegate {
     public static ObjCProtocols = [UINavigationControllerDelegate];
 
     navigationControllerAnimationControllerForOperationFromViewControllerToViewController(
@@ -453,7 +453,7 @@ class UINavigationControllerAnimatedDelegate extends NSObject implements UINavig
     }
 }
 
-class UINavigationControllerImpl extends UINavigationController {
+export class UINavigationControllerImpl extends UINavigationController {
     private _owner: WeakRef<Frame>;
 
     public static initWithOwner(owner: WeakRef<Frame>): UINavigationControllerImpl {
@@ -665,7 +665,7 @@ export function _getNativeCurve(transition: definition.NavigationTransition): UI
 }
 
 /* tslint:disable */
-class iOSFrame implements definition.iOSFrame {
+export class iOSFrame implements definition.iOSFrame {
     /* tslint:enable */
     private _controller: UINavigationControllerImpl;
     private _showNavigationBar: boolean;
--- a/tns-core-modules/ui/gestures/gestures.ios.ts
+++ b/tns-core-modules/ui/gestures/gestures.ios.ts
@@ -7,7 +7,7 @@ import types = require("utils/types");
 
 global.moduleMerge(common, exports);
 
-class UIGestureRecognizerDelegateImpl extends NSObject implements UIGestureRecognizerDelegate {
+export class UIGestureRecognizerDelegateImpl extends NSObject implements UIGestureRecognizerDelegate {
     public static ObjCProtocols = [UIGestureRecognizerDelegate];
     public gestureRecognizerShouldRecognizeSimultaneouslyWithGestureRecognizer(gestureRecognizer: UIGestureRecognizer, otherGestureRecognizer: UIGestureRecognizer): boolean {
         return true;
@@ -15,7 +15,7 @@ class UIGestureRecognizerDelegateImpl extends NSObject implements UIGestureRecog
 }
 var recognizerDelegateInstance: UIGestureRecognizerDelegateImpl = <UIGestureRecognizerDelegateImpl>UIGestureRecognizerDelegateImpl.new();
 
-class UIGestureRecognizerImpl extends NSObject {
+export class UIGestureRecognizerImpl extends NSObject {
 
     private _owner: WeakRef<GesturesObserver>;
     private _type: any;
@@ -348,7 +348,7 @@ function _getRotationData(args: definition.GestureEventData): definition.Rotatio
     };
 }
 
-class TouchGestureRecognizer extends UIGestureRecognizer {
+export class TouchGestureRecognizer extends UIGestureRecognizer {
     public observer: GesturesObserver;
     private _eventData: TouchGestureEventData;
 
--- a/tns-core-modules/ui/list-picker/list-picker.ios.ts
+++ b/tns-core-modules/ui/list-picker/list-picker.ios.ts
@@ -47,7 +47,7 @@ export class ListPicker extends common.ListPicker {
     }
 }
 
-class ListPickerDataSource extends NSObject implements UIPickerViewDataSource {
+export class ListPickerDataSource extends NSObject implements UIPickerViewDataSource {
     public static ObjCProtocols = [UIPickerViewDataSource];
     
     private _owner: WeakRef<ListPicker>;
@@ -68,7 +68,7 @@ class ListPickerDataSource extends NSObject implements UIPickerViewDataSource {
     }
 }
 
-class ListPickerDelegateImpl extends NSObject implements UIPickerViewDelegate {
+export class ListPickerDelegateImpl extends NSObject implements UIPickerViewDelegate {
     public static ObjCProtocols = [UIPickerViewDelegate];
 
     private _owner: WeakRef<ListPicker>;
--- a/tns-core-modules/ui/list-view/list-view.ios.ts
+++ b/tns-core-modules/ui/list-view/list-view.ios.ts
@@ -26,7 +26,7 @@ global.moduleMerge(common, exports);
 
 var infinity = utils.layout.makeMeasureSpec(0, utils.layout.UNSPECIFIED);
 
-class ListViewCell extends UITableViewCell {
+export class ListViewCell extends UITableViewCell {
     public willMoveToSuperview(newSuperview: UIView): void {
         let parent = <ListView>(this.view ? this.view.parent : null);
 
@@ -50,7 +50,7 @@ function notifyForItemAtIndex(listView: definition.ListView, cell: any, view: vi
     return args;
 }
 
-class DataSource extends NSObject implements UITableViewDataSource {
+export class DataSource extends NSObject implements UITableViewDataSource {
     public static ObjCProtocols = [UITableViewDataSource];
 
     private _owner: WeakRef<ListView>;
@@ -87,7 +87,7 @@ class DataSource extends NSObject implements UITableViewDataSource {
     }
 }
 
-class UITableViewDelegateImpl extends NSObject implements UITableViewDelegate {
+export class UITableViewDelegateImpl extends NSObject implements UITableViewDelegate {
     public static ObjCProtocols = [UITableViewDelegate];
 
     private _owner: WeakRef<ListView>;
@@ -147,7 +147,7 @@ class UITableViewDelegateImpl extends NSObject implements UITableViewDelegate {
     }
 }
 
-class UITableViewRowHeightDelegateImpl extends NSObject implements UITableViewDelegate {
+export class UITableViewRowHeightDelegateImpl extends NSObject implements UITableViewDelegate {
     public static ObjCProtocols = [UITableViewDelegate];
 
     private _owner: WeakRef<ListView>;
--- a/tns-core-modules/ui/page/page.ios.ts
+++ b/tns-core-modules/ui/page/page.ios.ts
@@ -50,7 +50,7 @@ function isBackNavigationFrom(controller: UIViewControllerImpl, page: Page): boo
     return true;
 }
 
-class UIViewControllerImpl extends UIViewController {
+export class UIViewControllerImpl extends UIViewController {
 
     private _owner: WeakRef<Page>;
 
--- a/tns-core-modules/ui/scroll-view/scroll-view.ios.ts
+++ b/tns-core-modules/ui/scroll-view/scroll-view.ios.ts
@@ -6,7 +6,7 @@ import utils = require("utils/utils");
 
 global.moduleMerge(common, exports);
 
-class UIScrollViewDelegateImpl extends NSObject implements UIScrollViewDelegate {
+export class UIScrollViewDelegateImpl extends NSObject implements UIScrollViewDelegate {
     private _owner: WeakRef<ScrollView>;
 
     public static initWithOwner(owner: WeakRef<ScrollView>): UIScrollViewDelegateImpl {
--- a/tns-core-modules/ui/segmented-bar/segmented-bar.ios.ts
+++ b/tns-core-modules/ui/segmented-bar/segmented-bar.ios.ts
@@ -114,7 +114,7 @@ export class SegmentedBar extends common.SegmentedBar {
     }
 }
 
-class SelectionHandlerImpl extends NSObject {
+export class SelectionHandlerImpl extends NSObject {
 
     private _owner: WeakRef<SegmentedBar>;
 
--- a/tns-core-modules/ui/switch/switch.ios.ts
+++ b/tns-core-modules/ui/switch/switch.ios.ts
@@ -15,7 +15,7 @@ function onCheckedPropertyChanged(data: dependencyObservable.PropertyChangeData)
 
 global.moduleMerge(common, exports);
 
-class SwitchChangeHandlerImpl extends NSObject {
+export class SwitchChangeHandlerImpl extends NSObject {
 
     private _owner: WeakRef<Switch>;
 
--- a/tns-core-modules/ui/tab-view/tab-view.ios.ts
+++ b/tns-core-modules/ui/tab-view/tab-view.ios.ts
@@ -22,7 +22,7 @@ function ensureImageSource() {
     }
 }
 
-class UITabBarControllerImpl extends UITabBarController {
+export class UITabBarControllerImpl extends UITabBarController {
 
     private _owner: WeakRef<TabView>;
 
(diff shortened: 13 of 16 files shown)


NathanaelA:
Bot is apparently going crazy; I guess it REALLY wants someone to verify my patch.  ;-)

dtopuzov:
run ci

NathanaelA:
The CI failing makes no sense the entire patch just adds "exports" to all the delegates that I could find.   I'm sure I missed a couple of them, but I hit all the major ones -- especially the three that I have ran into several times over the last 18 months...

dtopuzov:
run ci

dtopuzov:
run ci

hshristov:
Hi @NathanaelA,

exposing delegate implementation is not safe in my opinion. We are using some private fields on the delegates, and some methods must call methods on the owner in order to work correctly. That is why we didn't expose them in the first place. And it make us less flexible - sometimes we change the delegate implementation in order to support different versions and features (e.g. ListView - rowHeight feature use different delegate). Exposing them would make such changes - breaking.

So I think a safer approach is to raise events on the controls that needs customisation through the delegates - like TextEdit, SegmentedBar, etc.

Do you think events will work for your cases?

NathanaelA:
Hey @hshristov 

Nope, sorry events WON'T work, in the long run imho.     If you think about how iOS delegates work, you will see several issue with delegates and events.   For you to do events rather than expose the delegates you (as the tns team) would have to override every single delegate option and then raise an event for each one, that would be very quickly unwieldy and the a large chunk of your source would become overriding a delegate functions and firing events and dealing with the return values.    Think of the UINavController delegates or WebView delegates, you really want to add over 20 events to each one just so that I can get access to one or two delegate methods that I need?     That is a lot of development effort and added code to the ios tns-core-modules side for imho a small gain.   

So lets look at real life scenarios -- lets take my #2640 - I need access to change: `shouldChangeCharactersInRange` in side the text delegate.  You don't even have it in the prototype chain.  In the frame.ios.js, I need access to the NavControllerDelegate for `shouldAutoRotate`. in the webview delegate I need access to several delegate functions.      The issue is that each program/plugin needs totally different delegate access.  In some cases this can be easily done because the delegates are already exposed; but in other cases (like the three mentioned above) because they are "private" you make me have to modify the tns-core-modules to expose it, or create some really crazy hacks to get access to it.    I HATE having to modify tns-core-modules...   ;-)   And weird hacks are much more likely to break...

I patch lots of "private" things in the TNS-core-modules; and I do _NOT_ expect them to stay and am willing to (and have) update my plugins when you break something.    That is just the risk I take as a plugin author.  But when I have access to the private stuff; then I can create the cool plugins like my ns-globalevents, ns-hidden, ns-scaledFonts, ns-dom, ns-orientation, etc.  They all tie into the private parts of the framework and modify things to attach there functionality to be able to handle things you didn't either expect to need or create hooks for.   And over the last 18 months, I've had to fix a change here and there; but it is well worth it as the abilities I have created in some of my plugins allow me to do things you never envisioned...

---

Now as to possible solutions:

1. Expose the delegates; and let me hack them as is.   The easiest solution for everyone.  It is my responsibility to keep my code working...  And if you break it, it is still my responsibility since I am tying into private implementations.  ;-)  This requires no work on your part, beyond my patch removing the private flag.

2. Expose events, but this requires a LOT of extra work on your part to add to the event system for each of the delegates and then adding all those missing functions to call events and then handling any return values back to the delegate function back to the framework.   This is a lot of work on your part; as you basically have to create delegate events to everything you even think might be used by someone in every single delegate...     

3. Create some sort of delegate subclass system.  
 Some sort of delegate registration/init system -- when you include my plugin I can call "global.delegateRegister('text-field', 'UITextFieldDelegate', function(del) {}, priority);"  to say that I am subclassing a specific delegate.   This way 10 different plugins can each subclass the same delegate and it would still all work (assuming everyone has good code).  Then in the text-field.ios.js file; it would do a 
` var delegateToUse = global.getDelegate('text-field','UITextFieldDelegate', UITextFieldDelegateImpl);`
The `getDelegate` code basically passes your UITextFieldDelegateImpl to the first function, its result it passes to the next function until all registered delegate functions are called.  Then the final result is passed back to the caller to be used.


4. Create some sort of delegate mixin system, like the above where you mixin my delegate functions, issues with this idea is that if we both need to have the same function name then I have to make sure MY function has all your code in it to make it work properly.

I think 3 is probably the best idea, as it allows you to keep your delegates private but allows me to override them to do my custom behavior.  However, it does require some revamping of a lot of files.   So in the meantime until a project can be done for it; I honestly would like the "hack" of 1 being done so I'm not STUCK while waiting for the core to be updated to a delegate sub classing system..

Thoughts?

hshristov:
What about a different approach - you create your delegate that accepts existing delegate. This existing instance will be our implementation without any details exposed. Then in your delegate you handle the method you need and in all other methods you call the default (e.g. our private implementation) delegate. I have this working in the case of ListView (e.g. UITableView delegate). 
This will require no change in the modules and most of the time it won't break your plugin.
Will this cover your plugin needs?

And btw - some functionality like `visibility='hidden'` will be much better in the core-modules :) (e.g. PR)

NathanaelA:
I don't understand what you are saying to do -- Lets take a real case where my patch would help.

-- 

How exactly do I get the `UINavigationControllerImpl` delegate from `frame.ios.js`?

Line: https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/frame/frame.ios.ts#L456 is where it is defined (and it isn't exported).  So I can't get the delegate there and extend / subclass it.

The next place it is referenced; You actually init it and assign its owner at https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/frame/frame.ios.ts#L681  -- So at this point you have created the class and initialized it and assigned the delegate to the iOSFrame...  

So at what point can I insert/subclass/superclass my code and your delegate?   

---

On the BTW, I totally agree `ns-hidden` would be much better served being inside the framework.  I have several plugins that would make a lot more sense being inside the framework.     Issue is I don't particularly like to waste my time to create a PR (like what is happening with this one) and go through the lengthy discussion phase to decide if it will be accepted.  It is a hell of a LOT easier for me to just to make a plugin that I can use right now, and skip the entire "lets discuss its merits of the PR"...  See the issue I have making enhancement PR's... :grinning:

NathanaelA:
@hshristov - Any chance you can explain how to do this; I've run into another one of these delegates (TabView) where I have no access to modify the delegate.

hshristov:
You implement a delegate like do now and in the constructor you pass an existing delegate. In method implementation you call either the delegate you got in the constructor or you implement it like you need. 

pseudo TS code.
```
class UINavigationControllerAnimatedDelegate extends NSObject implements UINavigationControllerDelegate {
    public static ObjCProtocols = [UINavigationControllerDelegate];
    public static initWithDelegate(defaultImplementation: UINavigationControllerAnimatedDelegate): UINavigationControllerAnimatedDelegate {
        // Init and save defaultImplementation here.
        this.defaultImplementation = defaultImplementation;
    }
    navigationControllerAnimationControllerForOperationFromViewControllerToViewController(params) {
        // either implement the method like you want to or call return this.defaultImplementation. navigationControllerAnimationControllerForOperationFromViewControllerToViewController(params);
    }
}
```

Then when you want to change the delegate of UINavigationController:
```
`var frame = new Frame()';
// take current delegate;
var currentDelegate = frame.ios.controller.delegate;
// and pass it to your implementation.
frame.ios.controller.delegate = UINavigationControllerAnimatedDelegate.initWithDelegate(currentDelegate);
```

Still not bullet-proof but it does not require any changes in modules.
The other approaches that you describe:

2. Events - I don't want to expose every delegate method as event so my thought was to expose only events that are needed (e.g. on demand).

3. Some sort of delegate registration - sound interesting. Definitely worth researching. 

4. Delegate mixin - it still tie us to preserve private methods so that delegate could call back to the owner.

In my opinion the safest approach is to create custom class with custom delegate. This way you will have total control. Everything else is not safe enough - we could break it unintentionally which is something I don't like.

valentinstoychev:
@hshristov @NathanaelA - you folks know each other very well, why not do a quick chat to resolve this :). Just saying :)

NathanaelA:
@hshristov I'm up for a chat.  Let me know what day & time...  I'm on the slack channels; so feel free to ping me there or via skype...  

I understood part of what you are saying, and I was afraid that was what you are suggesting; I can think of two issues with this idea

1. There are several delegate functions that only run when it is created and the delegate is attached; and in two of the cases that I'm discussing these are the ones I need and are accessed way before I can insert my open delegate using the extend method you are proposing.

2. if I capture the delegate in this fashion; doesn't that mean my delegate has to implement every single one of your delegate functions and pass it on to your delegate.  If you later go to add a new delegate function then my plugin will probably break the core modules.

NathanaelA:
@hshristov - Any more thoughts on this.  I would like to make sure we have this resolved one way or the other by 2.4's release.

hshristov:
@NathanaelA Well I still don't like the idea of opening all delegates. Too much exposure and while I'm sure you will patch the plugins you do, most people probably won't. So I don't like the idea to introduce breaking changes by exposing internal delegates.

If there are delegate methods that you cannot override because they executed too early we will expose events so that everyone could act on them.

The other suggestion to wrap our delegates (yes you will need to implement all delegate methods but in most of them you will call our delegate) will still work most of the time. When they don't we will review them case by case.

NathanaelA:
@hshristov - I have to say I totally disagree with this decision.  ;-(              The best method imho would be some sort of delegate registry like I mentioned in the comment above https://github.com/NativeScript/NativeScript/pull/2827#issuecomment-251747927  but that requires additional work to be built and implemented.    I was hoping in the meantime that exposing them would have been preferable so that we aren't blocking development, as the decision is now.

However, I have to say this decision wasn't unexpected.    I had already brainstormed and come up with an alternative plan of action a while back when I first started discussing this.   The `tns` tool chain has a cool `nativescript-hook` publish step, I can write a simple hook plugin `tns-core-patcher` that adds all these "exports.x" to all these files manually during the build and it would eliminate any issue for anyone using any of my plugins when I need to patch one of these delegates...   Any plugin I write that requires this functionality will just depend on the new tns-core-patcher` plugin as a dependency.   

Since it is hook plugin it will allow everything to continue working even when the core modules are updated and doesn't require install or any other work by users; and the end results acts just like you exposed them for me by accepting this patch would have done.  :grinning:

PeterStaev:
I tend to agree with this one with @NathanaelA and I know how he feels as I had to cope with this in a few situations as well. I ended up copy-paste the implementation from the core modules in my code and adding whatever additional I needed ([example](https://github.com/PeterStaev/NativeScript-Drop-Down/blob/master/drop-down.ios.ts#L147))... 

@hshristov, the reasoning that changing the internal implementation might break plugins/code is also valid, but even with your suggestion to create a wrapper delegate, if you make breaking changes to the original delegate, it will probably break this kind of implementation as well. Same goes for my workaround of copy-paste implementation. But with having delegates exported, it will make code a lot more readable and easy to change in case of a breaking change in the base implementation.

hshristov:
@NathanaelA And I totally disagree with your decision. Patching tns-core-modules with plugin makes them unsupportable so anyone using such plugin will have hard time to explain which version they are using, which version of the patch plugin and so on.
Delegate registration still won't work because we have some logic in the delegate that needs to call to delegate owner (like in `UITextFieldDelegateImpl`). We don't have a way to force the delegate implementation to call these methods and without calling them the control won't work.

@PeterStaev My suggestion for wrapping existing delegate is temporary until we know which cases are not possible right now. Please list the cases that you need to modify/extend current delegate implementation so that we plan the changes for the next version.

Just to make sure we are on the same side - I want to make modules as open as possible but still keeps them solid. Exposing private/internal parts of implementation is against all good coding policy. We have to still enhance and fix existing implementation, support iOS 8, 9, 10 and so on.

Lets discuss the delegates you guys need and make them cover your cases.

NathanaelA:
@hshristov - Oh, I bet you don't like my "final" solution.  I don't really like it either.  :grinning:  I totally realize the pitfalls it opens up; and unfortunately for you(Telerik); Telerik will totally get blamed if it breaks anything at any time...   (Especially since it only modifies the platform versions...)    I fully understand why it is the worst solution of the bunch; **but my point is; if that is the only solution to go forward**; then unfortunately I have very little choice other than to use it.   I get to choose between access or no access; I'm always going to choose access, so that I can get my job done.     I didn't toss it into the discussion earlier because it really is a method of last resort, and was not a route I wanted to take (because of the pitfalls), and so I had hoped to come to some sort of agreement that would work for everyone, before having to use my final failsafe method.  :grinning: 

As for a delegate registry; I don't see anything in UITextFieldDelegateImpl that would block the idea of a delegate registration system.  The idea of a delegate registration system would basically allow me to subclass the delegates BEFORE it is actually created so that any function I am overriding I would call super on.     This allows every single delegate function registered in the system for that delegate type to run.   It allows ALL of our delegate functions in the chain to work including your ability to call back to the owner to do your things.  Your delegate function would be just like mine, somewhere in the *super* chain...    Technically it isn't a very complicated issue to implement the logistics of that part of the piece.   However, the piece that is complicated which we do have to figure out in this idea is how to setup all the registrations during a very early point of bootstrap; before ANY delegates are actually created, since you have to have them registered to actually set them up.      Since we really don't want to scan every plugin folder during runtime at bootup.  We probably have to have an actual build step that combines all the plugins `delegate.registration.js` files into a single file or something that can make this a single file that the runtime can load from the main tns-core during startup.      I honestly thing this is the best way to implement this and solve all of our issues.   But it requires putting some engineering work into it from several points which means it will take probably a couple versions before it works well.   In the meantime I don't want to be blocked while we figure this out, so opening up access right now is my stopgap measure until a final solution is developed.   We can depreciate the direct access and re-make them all private when we have a final solution.

---

> Just to make sure we are on the same side - I want to make modules as open as possible but still keeps them solid. Exposing private/internal parts of implementation is against all good coding policy. We have to still enhance and fix existing implementation, support iOS 8, 9, 10 and so on.

I am fully aware of this issue from your point of view; but as a plugin developer I also have to deal with the exact same issues.  If my plugin breaks because you modified an internal delegate; then it is my responsibility to fix it, not yours.    I've had to fix my ns-orientation, and ns-dom at least once or twice over the TNS core upgrades because you fixed bugs and/or completely changed internal and even once  you eliminated/modified a public routines that I was depending on.  This is to be expected, if I use internal calls; I expect to have to fix them when they are broken/changed/removed.  

However, on the flip side I think every single one of my projects uses my NS-Dom library (& my liveedit), and half of them use my ns-orientation plugin; and a whole slew of my other plugins are in several of my projects...   I wouldn't ever want to live without most of my plugins; they provide me a lot more flexibility that the default platform.   So they will get fixed...

---

> Lets discuss the delegates you guys need and make them cover your cases.
This idea really really works very poorly for me...   The TNS releases are typically somewhere around 5+ weeks.   (We are already at 48 days for 2.4 and it looks like it will be >60 days now).   In addition if my request/fix comes in too late in the cycle it is pushed to the next release.   This is extremely horrible when we are trying to get things done, to have to wait forever before we can use it.    When I need it; it isn't that I need it in a couple months, it is I needed it yesterday because I'm creating a plugin to fix a clients need right now.   I can't tell my clients, oh yeah we "hope" it will be fixed in a couple months, that just doesn't fly.  ;-(       

Oh and if you are thinking of bringing up the @next versions -- the @next's are not really a useful solution for actual clients; they tend to be buggy and they are "not" really the `next` version, they are the `master` version meaning they frequently contain code that is not going to be in a real `release` version until (hopefully) some distant future version.   If you were to go to a weekly `next` version with primarily just bug fixes and smaller changes; rather than the current `master` version that would go a long ways to make this more of a non-issue for me.    

So there was a reason I exported all of these delegates.  I may NEVER use most of them, but I when I need them, I don't want to be begging to "maybe" see them in a couple months...

PeterStaev:
@hshristov , the ones that pop up in my mind are `UIPickerViewDelegateImpl` and `UIScrollViewDelegateImpl`

hshristov:
@PeterStaev I mean what is the case that you need to implement, which delegates and methods you need?

@NathanaelA As I said before there is another option - to create custom UI component. This way you will manage the native view, the delegate, everything. We won't break you when we change something. 

But again as I said I want to make modules as useful as possible so if you list all scenarios that need custom delegates we might be able to handle them without the need of any hack.

hamorphis:
@NathanaelA What about adding the missing functionality to the delegates directly in tns-core-modules. In this way, the default views in tns-core-modules will benefit from your enhancements and will improve NativeScript for all people that are using it.

PeterStaev:
@hshristov , for the picker one I needed it, because I wanted to format the text in the picker. Here is again the link to the code: https://github.com/PeterStaev/NativeScript-Drop-Down/blob/master/drop-down.ios.ts#L147. For the scrollview I needed to implement some custom logic for pageable scrolling in addition to the default event notification.

NathanaelA:
@hamorphis - I know I tend to write books in response; but I did answer this issue above in my other response to @hshristov .  ;-)

1. I'm actually all for adding missing functionality to the framework (when it needs it).   However, some of the plugins (like forcing the screen orientation programmatically), is questionable if it is actually needed in the core framework or better served as a plugin.   It isn't something used very much.   In fact I haven't used it in any of my apps; but a client needs it for his app.  This is a real example that I have had to hack around and the best solution is locked behind one of those stupid "private" delegates that I'm trying to free with this patch.      It cost me MANY MANY hours to bypass the chain in a way that worked reliably, where directly/manually modifying the delegate source prototype inside tns-core-modules worked immediately.   

2. The NS releases are slow; 5+ weeks between releases is a killer.   If the patch is too late in the cycle; make that over 2 months wait before it shows up...   This does not work for clients who are paying for a **plugin** to say; sorry 2.3 doesn't offer it; 2.5 might, see you next January!     I don't really want to start distributing customized tns-core-modules if I can avoid it...   To me if the framework stands in my way, then creating a plugin that re-writes the core-modules and fixes the framework is a **way better** option for me then me distributing customized tns-core-modules to a plugin clients.  At least rewriting the code should continue to work when you release 2.4, 2.5, etc...   

3. There is no guarantee a patch of mine will make it into core; take this one for example.  I have outlined several times WHY I need this ability (with actual REAL use cases); and I understand your reluctance (don't agree with it, but understand it) to implement it.    Yet, you haven't offered me any workable solution in this thread.    I can't wrap the app delegate; that is frequently too late in the component lifecycle -- the functions I need to wrap will have already been called by iOS when the component was initialized.   And waiting 3-8 weeks to **MAYBE** (again, no guarantees a patch will be accepted) to let my client have a plugin I wrote, doesn't work very well either for me to get paid.   

In summary; the framework needs to be enabling me; not standing in my way...  If it stands in my way, I will beat it down with a :hammer:    And yes, no one may like the solution, but at least it is a solution.  :grinning:

NathanaelA:
@hshristov - Your suggestion of creating my own component is also a non-starter in a some of these cases.  I really have NO issue creating my own components, I'm up to somewhere around 26+ public plugins and almost as many private...  ;-)

Lets go **BACK to my REAL** use case...   Which I have provided above, so we have something to talk about that is **REAL**.     I can't really create a new TNSApplication component for iOS, do you realize how deeply embedded this class is in the entire TNS core modules.   Like I mentioned *above* I need access to the app delegate before it is created.  

I guess **technically** I could copy the TNS application.ios.js code into my own file; make my 10 lines of code changes to it then require the client to call my `require ("ns-mt-application")` as the very first line of app.js/app.ts/main.ts file, which would then swap out the entire tns core modules application variable for mine, so that it is running my code and then my delegate.    BUT when you go to release 2.4 or 2.5, the odds of something breaking are pretty high.  This seems like a huge non-starter to me.   ;-)

It would be way more easy and safe to just create/fork my own tns-core-modules with all my patches and distribute that as a requirement for all my plugins.    I have so far attempted to NOT fork the core-modules; but I guess it is technically a valid option...   It would for sure simplify some things drastically for me, and might be valuable to a lot of people since simple bug fixes could be released quickly...    Hmm, I'll have to seriously think about that option, I already have all the infrastructure to do it...

Back to your original idea, my objective is not to be maintaining my own TNSApp class for each release.  My objective is to release a plugin that I **hopefully never** have to touch again because it is solid and will continue to work.  ;-)     That is why I patch my own prototype into the delegate.   

---

So I think these are all the potentially valid cases provided so far:
1. exporting the delegates (this patch)
2. a delegate registration system (a lot of infrastructure has to be done)
3. tns-core-patcher (potentially can mess up platform/tns-core files, Telerik blamed)
4. Fork tns-core-modules (Might split the community, or might be empowering like io.js)
5. Making my own component (A lot of work to just add 5-10 lines of code, but could be valid in some cases)
6. Wrapping the already created delegate.  (Timing failure, more memory, and changes to core delegate code will break the plugin)

Any other ideas?

hshristov:
@NathanaelA We already support adding custom application delegate:
```
    class AppDelegate extends NSObject implements UIApplicationDelegate {
        static ObjCProtocols = [UIApplicationDelegate];
        applicationHandleOpenURL(app, url): boolean {
            // Your code here.
        }
    }
    application.ios.delegate = AppDelegate;
```

The other cases if I understand correctly is ScrollChanged event on the scrollView.
If would be best if you submit PR directly with the changes you need to make in order to support your scenario or at least open an issue and describe them so we could think of the best way to implement it.

NathanaelA:
@hshristov 

Sorry my mistake I referred to the wrong delegate.  To many delegates.  :grinning:    NavControllerDelegate (ie. the UINavigationControllerImpl  in frame.ios.js) is the delegate I need access to.  However, based on the init cycle this gets created when the app is created; I need access to change the delegate BEFORE the app creates the delegate.    Also if I recall correctly, one of the ios TabBar delegates; I'm pretty sure I need before it is created also as it does some initial sizing of parts of the tab bar code when created...   

If you are opposed to this pull request, then just please close it  (& #2471)-- me arguing for it is pointless for both of us -- if there isn't a chance it will be accepted.   :grinning:  

You know my position, and what I believe needs to be done, if you decide in the future to implement the delegate registry system correctly, then that would solve the issue too.  Otherwise, we are just wasting each others time, and we both would be better off working on the other projects that we are working on right now...  :grinning:

NathanaelA:
I have decided to start down the [TNS-Core-Patcher plugin](https://github.com/NathanaelA/nativescript-dev-tnscorepatcher)  idea...    The more I think about it; the more I realized it would totally solve **all** my patch problems with the team.   :grinning:

I can build them, and they can be applied by TCP and then if you decide you want them built into the engine you will see the benefit because of the usage of the plugin that needs it, you will then have real data on how much usage it gets -- which will allow you to make a real decision if it is benefits the core modules, rather than arbitrarily dismissing patches because you don't see the benefit...  :grinning: 

Please feel free to comment on the repo, for any safety or issue you want me to try and take in account.  My objective isn't to break anything; and I've been brainstorming how to do this safely; and have a issue relating to this subject on my repo...

:grinning:

valentinstoychev:
Sometimes it is hard to come to a conclusion that suits everyone. This is one of these cases. I'm closing this request after I discussed with both Nathanael and the core team.

speigg:
@valentinstoychev I ran into this issue as well, trying to add `preferredScreenEdgesDeferringSystemGestures` to the Page's UIViewController. It seems like there isn't a good solution for these kinds of problems. 

Unfortunately, I understand that new javascript methods can't be added to a Obj-C subclass after `extends` is called... however, if this limitation were reversed with a new feature (`global.__updateNative`?), then I could simply do the following:

```ts
const viewController = <UIVIewController>page.ios;
viewController.preferredScreenEdgesDeferringSystemGestures = () => UIRectEdge.All;
global.__updateNative(viewController); // update native class to call into js for implemented methods
viewController.setNeedsUpdateOfScreenEdgesDeferringSystemGestures();
```

NathanaelA:
For anyone interested in a solution to this artificial door.  @speigg actually caused me to see a open window to allow you to be able to do this now.

http://fluentreports.com/blog/?p=558

NathanaelA:
Unfortunately a different bug in the iOS version of NativeScript prevents the ability I outlined in my blog post.   (Talk about seriously disappointing!)   Odds of getting that bug fixed; I'd wager are about as high as the Pope renouncing the Catholic church.   So we are **still** stuck with hand modifying the core modules to get access to any delegates.

PR #2848, opened 2016-10-05
remove gc call

hshristov:
Most of the OOM Exceptions were caused by holding Bitmap references inside image-source. Image component was using imageSource internally. Now Image component is loading images in Java so OOM should be rare. 
Removing the GC call improve the UX but there still could be a case were manual gc call is needed.


vchimev:
run ci

PR #2851, opened 2016-10-06
Make fps-meter callback execute in the same zone as they are created

vakrilov:
Wrap the callbacks in fps-meter module in `zonedCallback`.
The callbacks are called form native code and we need to make sure they are fired in the same zone as the `addCallback` method to work with angular (or any other zoned app)

cc @tsonevn

### DIFF
--- a/tns-core-modules/fps-meter/fps-meter.ts
+++ b/tns-core-modules/fps-meter/fps-meter.ts
@@ -1,16 +1,16 @@
 ﻿import fpsNative = require("fps-meter/fps-native");
 
-var callbacks = {},
-    idCounter = 0,
-    _minFps = 1000,
-    framesRendered = 0,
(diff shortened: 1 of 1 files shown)


tns-core-modules/fps-meter/fps-meter.ts
```diff
@@ -69,9 +69,10 @@ export function stop() {
 }
 
 export function addCallback(callback: (fps: number, minFps?: number) => void): number {
-    var id = idCounter;
+    const id = idCounter;
 
-    callbacks[id] = callback;
+    // Wrap all calback in zonedCallback so that they work with the current zone.
+    callbacks[id] = zonedCallback(callback);
```
vakrilov: The important bit - rest is just refactoring

PR #2855, opened 2016-10-06
Added npm run typedoc and npm run dev-typedoc scripts, read the Devel…

PanayotCankov:
@ErjanGavalji Can you check if this can be integrated in the docs build downstream?


PanayotCankov:
I fixed the Grunt too. I will merge on successful build.

PR #2862, opened 2016-10-07
Move imageCache initialization in the createUI method so that it is s…

hshristov:
…ure that we have valid context. application.android.nativeView could be null if image module is requested before application.start


dtopuzov:
run ci

PR #2909, opened 2016-10-17
ListView item template selector

### LINKED ISSUE #2820: ListView item template selector
To give more context (based on the slack chat we had):
We need to be able to display sets of data with different "designs" in a ListView without the need of using collapsed views (like a huge template which will be impossible to maintain and cost CPU cycles) or logic inside the view-model for `ContentView` which would have performance penalties.

A good  scenario could be a news app, that needs to display the news with different designs based on the importance of the article: 

Thank you 😄

hamorphis:
Resolves #2820


hamorphis:
uitests

hamorphis:
cuteness

hamorphis:
uitests android

tns-core-modules/ui/builder/component-builder.ts
```diff
@@ -181,7 +180,12 @@ export function setPropertyValue(instance: View, instanceModule: Object, exports
         if (isFunction(handler)) {
             instance.on(propertyName, handler);
         }
-    } else {
+    }
+    //TODO_ITS: This code may break some people if they have an export named like one of their attribue values.
+    else if (isFunction(exports && exports[propertyValue])) {
```
PanayotCankov: Using isKnownMultiTemplate should be an option.

tns-core-modules/ui/list-view/list-view-common.ts
```diff
@@ -67,6 +79,20 @@ export class ListView extends view.View implements definition.ListView {
     public static itemLoadingEvent = "itemLoading";
     public static itemTapEvent = "itemTap";
     public static loadMoreItemsEvent = "loadMoreItems";
+    
+    private _itemTemplateSelector: (index: number, item: any) => string;
```
PanayotCankov: I'd put the item first, the index second and probably the items collection third, like in Array.forEach.

hamorphis:
uitests

tns-core-modules/ui/list-view/list-view-common.ts
```diff
@@ -126,6 +162,40 @@ export class ListView extends view.View implements definition.ListView {
         this._setValue(ListView.itemTemplateProperty, value);
     }
 
+    get itemTemplates(): string | Array<view.KeyedTemplate> {
+        return this._getValue(ListView.itemTemplatesProperty);
+    }
+    set itemTemplates(value: string | Array<view.KeyedTemplate>) {
```
PanayotCankov: I'd avoid forcing the data structure here by specifying Array, can't we use a `string | (key: string) => View` factory function here? Similar to the Template function we already have.

tns-core-modules/ui/list-view/list-view-common.ts
```diff
@@ -126,6 +162,40 @@ export class ListView extends view.View implements definition.ListView {
...
+            newValue = builder.parseMultipleTemplates(<string>newValue, this);
+        }
+        this._setValue(ListView.itemTemplatesProperty, newValue);
+    }
+
+    get itemTemplateSelector(): string | ((index: number, item: any) => string) {
+        return this._itemTemplateSelector;
+    }
+
+    set itemTemplateSelector(value: string | ((index: number, item: any) => string)) {
```
PanayotCankov: Here too use the Array.forEach signature?

tns-core-modules/ui/builder/component-builder.ts
hamorphis: Can you please elaborate?

tns-core-modules/ui/list-view/list-view-common.ts
hamorphis: I agree. Will change this.

tns-core-modules/ui/list-view/list-view-common.ts
hamorphis: But this should accept multiple things -- not a single factory function? How should I type it?

tns-core-modules/ui/list-view/list-view-common.ts
hamorphis: Got it!

tns-core-modules/ui/list-view/list-view-common.ts
hamorphis: If I understand correctly, you want to change the KeyedTemplate interface from this:
```
    interface KeyedTemplate {
        /**
         * The unique key of the template.
         */
        key: string;

        /**
         * The function that creates the view.
         */
        createView: Template;
    }
```
to this:
```
    interface KeyedTemplate {
        (key: string): View;
    }
```
Am I correct?

tns-core-modules/ui/builder/component-builder.ts
PanayotCankov: Sorry, you can't directly use the isKnownMultiTemplate, but you can create isKnownFunction in a similar way. Or similar to the isEventOrGesture above.

hamorphis: I got it. I will add the isKnownFunction stuff so we will handle only the 'itemTemplateSelector' property name. 10x

tns-core-modules/ui/builder/builder.ts
```diff
@@ -34,11 +34,16 @@ export function parse(value: string | Template, context: any): View {
...
 
         return viewToReturn;
-    } else if (isFunction(value)) {
+    } else if ( isFunction(value)) {
         return (<Template>value)();
     }
 }
 
+export function parseMultipleTemplates(value: string, context: any): Array<KeyedTemplate> {
+    let dummyComponent = `<ListView><ListView.itemTemplates>${value}</ListView.itemTemplates></ListView>`;
```
hdeshev: Won't this create a full-blown list view just to throw it away after parsing the templates? What are the performance implications?

tns-core-modules/ui/list-view/list-view-common.ts
```diff
@@ -126,6 +162,40 @@ export class ListView extends view.View implements definition.ListView {
...
+                targetProperty: "templateKey",
+                expression: <string>value
+            });
+            this._itemTemplateSelector = (index: number, item: any) => {
+                item["$index"] = index;
+                this._itemTemplateSelectorBindable.bindingContext = item;
+                return this._itemTemplateSelectorBindable.get("templateKey"); 
+            };
+        }
+        else if (isFunction(value)) {
```
hdeshev: `} else {` (else on the same line as closing brace)

tns-core-modules/ui/core/view.d.ts
```diff
@@ -657,6 +657,21 @@ declare module "ui/core/view" {
     }
 
     /**
+     * Defines an interface for Template with a key.
+     */
+    interface KeyedTemplate {
+        /**
+         * The unique key of the template.
+         */
+        key: string;
```
hdeshev: Just a thought - does `id` sound better here?

tns-core-modules/ui/list-view/list-view-common.ts
PanayotCankov: No, KeyedTempalte is ok. Inside the control when you are using it you should simply use a `let itemView = this.itemTemplates(key)` and move the selection near 236:
https://github.com/NativeScript/NativeScript/pull/2909/files#diff-392843dff8d0afda16e33c7f1fced70cR236
in the ui builder just before you set the itemTemplates property on the view.
instead of setting it to the array you can create a function such as:
```
view.itemTemplates = key => {
    for (let i = 0, length = itemTemplatesArray.length; i < length; i++) {
        if (itemTemplatesArray[i].key === key){
            return itemTemplatesArray[i].template();
        }
    }
    return null;
}
```
This way Angular shouldn't need to care about the KeyedTemplate used under the hood by the builder when they are adopting the feature. It will also allow templates to be created in code behind a little bit easier:
```
view.itemTemplate = key => {
  switch(key) {
    case "title": return new Label();
    case "action": return new Button();
    default: throw "Unknown key";
  }
}
```

hamorphis:
@PanayotCankov I did refactor the code as per your suggestions. Code samples can be found here: https://github.com/NativeScript/NativeScript/tree/its/apps/app/ui-tests-app/list-view

hamorphis:
uitests

hamorphis:
@dtopuzov @vchimev Guys, we will need to rebuild the functional tests app, because I have two new overridden methods in the Android ListView adapter which are not being called and a test fails due to this. Rebuilding the app will generate the correct bindings. The Travis build passes since there everything is created from scratch.

hamorphis:
build

hamorphis:
build

hamorphis:
build

vakrilov:
Nace work @hamorphis

PR #2921, opened 2016-10-18
FlexboxLayout for iOS

PanayotCankov:
Translated the Android implementation of FlexboxLayout to {N} iOS TypeScript.
Migrated most of the unit tests.

There are some more unit tests required before this feature is completely done, mostly with our markup, special properties, and some integration examples with a repeater.


tns-core-modules/ui/layouts/flexbox-layout/flexbox-layout.android.ts
```diff
@@ -77,56 +77,63 @@ export class FlexboxLayout extends FlexboxLayoutBase {
 
     protected setNativeFlexDirection(flexDirection: FlexDirection) {
         let value = flexDirectionMap[flexDirection];
-        console.log("setNativeFlexDirection: " + flexDirection + " -> " + value);
         this.android.setFlexDirection(value);
```
hshristov: You could use the local `this._layout` filed for faster access.
But the bigger problem is your property won't be set to the native view if you use the propertyChangedHandler of property. You could use `onSetNativeValue` of PropertyMetadata from `proxy` module.

tsconfig.json
```diff
@@ -10,6 +10,8 @@
         "experimentalDecorators": true,
         "diagnostics": true,
         "sourceMap": true,
+        "jsx": "react",
```
hshristov: Why do we need this?

PanayotCankov: It is to enable .tsx within typescript: https://www.typescriptlang.org/docs/handbook/jsx.html
It allows View instances to be created using XML tags within TypeScript and appears native similar to writing JSON literals... The builder implementation is here:
https://github.com/PanayotCankov/nativescript-tsx
It is useful for UI declarations within the tests.
We can consider adding the code in the core at a later stage.

For now I moved to ui/builder.parse since we have to test that too.

tns-core-modules/ui/layouts/flexbox-layout/flexbox-layout.android.ts
PanayotCankov: The PropertyMetadata here is from proxy. Read the imports at the start. The "attached" properties however use propertyChange just like the GridLayout does, we should double check these.

PR #2924, opened 2016-10-19
Support of percentage margins for certain elements

zh-m:
ContentView, Page and all layouts will have their margins, width and height calculated based on a value set through percent. 

For example:
StackLayout {
    margin: 30% 15% 30% 15%;
}


zh-m:
#2924

PR #2934, opened 2016-10-20
npm scripts for typechecking public .d.ts-es and running tslint

PanayotCankov:
Moving one step closer to remove the grunt file.
The `tsconfig.public.es2016.json` now adds tsconfig that can be used to type check the public .d.ts-es exported by the tis-core-modules-es2016.d.ts using `npm run test-tsc-es2016`.
The tslint has been added as npm script too: `npm run tslint`.

### DIFF
--- a/tns-platform-declarations/test.ts
+++ b/tns-platform-declarations/test.ts
@@ -5,6 +5,7 @@ class IOSTest extends NSObject {
 }
 
 let pointer = new interop.Pointer();
+console.log(pointer);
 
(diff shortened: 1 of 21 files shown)


tns-platform-declarations/test.ts
```diff
@@ -5,6 +5,7 @@ class IOSTest extends NSObject {
 }
 
 let pointer = new interop.Pointer();
+console.log(pointer);
```
hshristov: Do we need this console.log here?

PR #2936, opened 2016-10-20
Fixed issue with image resizing.

nsndeck:
Fixes issue with resizing image for ios. For android this will also work but will return an image that is smaller sample size (based on requested size). For a complete solution we will update `tns-core-modules-widgets` to make such resizing on native side.

### DIFF
--- a/tns-core-modules/image-asset/image-asset-common.ts
+++ b/tns-core-modules/image-asset/image-asset-common.ts
@@ -56,23 +56,23 @@ export function getAspectSafeDimensions(sourceWidth, sourceHeight, reqWidth, req
     };
 }
 
-export function getRequestedImageSize(src: {width: number, height: number}): {width: number, height: number} {
+export function getRequestedImageSize(src: { width: number, height: number }, options: definition.ImageAssetOptions): { width: number, height: number } {
     let reqWidth = platform.screen.mainScreen.widthDIPs;
-    let reqHeight = platform.screen.mainScreen.heightDIPs
-    if (this.options && this.options.width) {
-        reqWidth = (this.options.width > 0 && this.options.width < reqWidth) ? this.options.width : reqWidth;
+    let reqHeight = platform.screen.mainScreen.heightDIPs;
+    if (options && options.width) {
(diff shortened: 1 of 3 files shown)


tns-core-modules/image-asset/image-asset-common.ts
```diff
@@ -56,23 +56,23 @@ export function getAspectSafeDimensions(sourceWidth, sourceHeight, reqWidth, req
     };
 }
 
-export function getRequestedImageSize(src: {width: number, height: number}): {width: number, height: number} {
+export function getRequestedImageSize(src: { width: number, height: number }, options: definition.ImageAssetOptions): { width: number, height: number } {
     let reqWidth = platform.screen.mainScreen.widthDIPs;
     let reqHeight = platform.screen.mainScreen.heightDIPs
```
hshristov: missing `;`

tns-core-modules/image-asset/image-asset-common.ts
```diff
@@ -56,23 +56,23 @@ export function getAspectSafeDimensions(sourceWidth, sourceHeight, reqWidth, req
     };
 }
 
-export function getRequestedImageSize(src: {width: number, height: number}): {width: number, height: number} {
+export function getRequestedImageSize(src: { width: number, height: number }, options: definition.ImageAssetOptions): { width: number, height: number } {
     let reqWidth = platform.screen.mainScreen.widthDIPs;
```
hshristov: Should we always decode to screen dimensions? Why not leave this option to the user?

tns-core-modules/image-asset/image-asset.ios.ts
```diff
@@ -35,12 +31,21 @@ export class ImageAsset extends common.ImageAsset {
...
                 }
                 else {
                     callback(null, imageResultInfo.valueForKey(PHImageErrorKey));
                 }
             }
         );
     }
+
+    private resizeImage(image, requestedSize): any {
+        UIGraphicsBeginImageContextWithOptions(requestedSize, false, 0.0);
```
hshristov: Does this code resize the image or just decode it in specified bounds?
If just decode lets change the name of the method.

PR #2948, opened 2016-10-24
Added pressed, active and disabled states for button.

nsndeck:
Disabled state is actually added to view, so it is available for all controls.


vchimev:
run ci

PR #2973, opened 2016-10-27
SelectedBackgroundColor property exposed via css.

nsndeck:
SegmentedBar selectedBackgroundColor property will be set via css like:

SegmentedBar {
    selected-background-color: red;
}


nsndeck:
run uitests android

PR #2989, opened 2016-10-30
Camera gallery as image source.

vbresults:
This also fixes a bug that would show video as an option when it shouldn't on iOS, since video is not yet implemented.


nsndeck:
Thank you for your pull request.

First of all we plan to deprecate build-in `tns-core-modules` camera module so we extract current functionality to a plugin named [nativescript-camera](https://github.com/NativeScript/nativescript-camera). For that plugin we received similar pull request (adding different sources for images). However we provide same functionality with another pluigin [nativescript-imagepicker](https://github.com/NativeScript/nativescript-imagepicker). Therefore our plans is to provide camera hardware only access via `nativescript-camera` plugin.

PR #2990, opened 2016-10-30
Set package.json 'main' to application/application.js for webpack.

vbresults:
When dividing the bundle into chunks, it's necessary to import `tns-core-modules`.  This change allows webpack to go down the import graph and manually register modules via `bundle-entry-points.js`.

### DIFF
--- a/tns-core-modules/package.json
+++ b/tns-core-modules/package.json
@@ -3,6 +3,7 @@
   "description": "Telerik NativeScript Core Modules",
   "version": "2.5.0",
   "homepage":"https://www.nativescript.org",
+  "main":"application/application.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/NativeScript/NativeScript"


hdeshev:
> When dividing the bundle into chunks, it's necessary to import tns-core-modules.

We never meant for users to import tns-core-modules directly. At runtime all submodules below tns-core-modules get moved one level up, so instead of importing `tns-core-modules/application`, you just import `application`

Could you post an example, or a more detailed description on what you are trying to achieve here? Does the approach in [nativescript-dev-webpack](https://github.com/NativeScript/nativescript-dev-webpack) not work for you? I am working on an improved (more customizable) webpack procedure that will be released with NativeScript 2.4 and I'd love to check if it will work in your case.

vbresults:
@hdeshev See NativeScript/app-protection#18

hdeshev:
> generate one bundle for nativescript and another for their app

Yes, you can do that without needing to import tns-core-modules. We will do it automatically starting with the 2.4 release -- you can see the [npm-scripts branch](https://github.com/NativeScript/nativescript-dev-webpack/tree/hdeshev/npm-scripts) that generates a 2-chunk bundle containing a vendor chunk and an app chunk.

It will generate a [`vendor` module](https://github.com/NativeScript/nativescript-dev-webpack/blob/hdeshev/npm-scripts/vendor.ts.nativescript.template) in the app folder that will let you fine tune what goes in that bundle and optimize your bundle size.

vbresults:
@hdeshev Rather than adding two new places to manually maintain modules to register, and adding two files to the webpack package, why not just add main as suggested which loads bundle-entry-points?

hdeshev:
I was thinking about this a lot. The way we modeled our solution(using a separate vendor entrypoint module that lets you configure what gets put there via simple requires) was inspired by configs used by   popular Angular seed projects. You raise a good point about duplicating imports though. I was already planning to change the default vendor.js config to something like:

```
require("./vendor-platform");                                                                                                                                                         
require("bundle-entry-points");
```

PR #2996, opened 2016-10-31
Fixed test about percent margin applied on a page

zh-m:
Fixed - test _test_percent_width_and_height_set_to_page_support_ failing as the height and width properties are calculated as (int) (availableHeight * percent) thus rounding the number to int.


zh-m:
uitests

vchimev:
run ci

Issue #2952, opened 2016-10-24
Error message given by `fetch` for xhr.onerror is not useful


vossad01:
### Did you verify this is a real problem by searching [Stack Overflow](http://stackoverflow.com/questions/tagged/nativescript) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?
Yes.  I did not see a direct match yet on StackOverflow for NativeScript though I did find this same issue for React Native on StackOverflow.  A related issue against the `fetch.js` NativeScript derived its implementation from is github/fetch#367.

### Tell us about the problem
The function assigned to `xhr.onerror` does not propagate helpful information concerning the error that occurred, instead only "Network request failed" is presented.

### Which platform(s) does your issue occur on?
Both (tested on Android)

### Please provide the following version numbers that your issue occurs with:
- CLI: 2.3.0
- Cross-platform modules: 2.3.0
- Runtime(s): tns-android (2.3.0)
- Plugin(s): none

### Please tell us how to recreate the issue in as much detail as possible.
Create a new project using the typescript template. Use the code below and observe the output on the console.

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.

app.ts
``` typescript
import * as app from 'application';

fetch("unknownProtocol://whatever", {
  method: "POST",
  body: JSON.stringify({}),
  headers: {
    "Content-Type": "application/json"
  }
})
  .then(resp => resp, error => {
    console.log(error);
    return error;
  });

app.start({ moduleName: 'main-page' });
```

### Proposed fix
I would like to submit a PR that changes `fetch.js` from 

``` javascript
xhr.onerror = function () {
    reject(new TypeError('Network request failed:'))
}
```
to 
``` javascript
xhr.onerror = function (error) {
    reject(new TypeError('Network request failed: ' + error.message))
}
```

In changing teh above you get an more helpful error such as,
> TypeError: Network request failed: java.net.MalformedURLException: unknown protocol: unknownprotocol

instead of,
> TypeError: Network request failed

NickIliev:
Hey, @vossad01 we are open source product and we are welcoming all contributors.
In order to create PR you should first sign the CLA (Contribution License Agreement) and once added to our contributors list you can post your PR. The process is few steps only and once ready create your PRs against the master branch - they will be revised from other member and when tested and upvoted they can be merged.

More about contributing can be found [here](https://www.nativescript.org/contribute).

NickIliev:
[...]

PR #3003, opened 2016-11-02
Improve error message of `fetch` when xhr.onerror occurs

vossad01:
Fixes/Implements #2952.

### You have [unit tests](https://github.com/NativeScript/NativeScript/blob/master/running-tests.md)
Added one test to cover this case.  Since I do not know what the error message will be on iOS, a pattern was used rather than checking for an exact message.

Additionally some cleanup of the existing tests was done to improve correctness of the tests.  Some would timeout, rather than fail, swallowing the failure message.


enchev:
run ci

Issue #2986, opened 2016-10-30
Enhancement: Page Imports


vbresults:
It would be great if we could have a new attribute for the `<Page>` element called `import` that lets us specify a module name to require/import, instead of it searching for a filename.

This will be very helpful in implementing a clean directory structure.

Also an extension on #1667; codeFile breaks in bundlers like webpack since bundles are by definition single-file.  `import` fixes this by importing with `global.loadModule`.

@enchev, IIRC you were talking about turning CSS into modules for webpack and this could lend itself to that too by doing the same with a new `stylesheet` attribute -- with module.exports being a string.

NickIliev:
Hi @vbresults thank you for your feature suggestion! 
Everyone who has an opinion on that matter can write it down in this thread. 

Ping @hshristov , @enchev

vbresults:
@NickIliev Ok, let me know and I will submit a PR: https://github.com/vbresults/NativeScript/commit/b26ed6b575758c04b2906a7bb0e7829d471bd0f6

hdeshev:
I like the idea. It looks like we can deprecate the `codeFile` feature as `import` seems to do the same thing in a more flexible way. For example, you could use it to "import" a platform-specific `my-page.android.js` file, if needed.

PR #3009, opened 2016-11-02
Added import attribute to page module.

vbresults:
#2986


hshristov:
@vbresults Why not using the existing properties:
```
<Page codeFile="~/your-code-file" cssFile="~/your-styles.css">
```

PR #3010, opened 2016-11-02
Image loading is done through native imageView

hshristov:
This is done so that we can clean up the bitmap if imageView is removed from window and then load it again if imageView is added to window.


hshristov:
build

PR #3020, opened 2016-11-04
Shrink will not damage the last item when it is shrink=0

PanayotCankov:
This will fix only the iOS version of the FlexboxLayout and may damage other functionality.


vchimev:
run ci

vchimev:
uitests

PR #3047, opened 2016-11-08
Add TabView.androidOffscreenTabLimit property

### LINKED ISSUE #2815: Tab View content reloads too often

It blown my head for 3 days, while I tried to find a solution and found that the issue comes because TabView recreates every tab after it becomes inactive which leads to huge freeze if you have some heavy components.

**I have 5 tabs. In the second one I have a list view with 100 rows.**

At first load, all tabs are showing very fast.

If i open tab 4 or 5, next time i open tab 2, i have a freeze of 2 seconds. I saw that the list view is recreated every time after the tab becomes inactive. And it becomes inactive only when i open a tab which is not the next or previous one. 

**If i switch from 1 or 3 to second tab(and I haven't switched before to 4 or 5) the change is instant. But if I switch to 4 or 5, the next time I open a tab with ListView the switch takes minimum 2 seconds.**

The switch to the other tabs(without a list views) doesn't need more than 300-400ms even if they were inactive before. Tested with both ListView and RadListView. And I use async pipes right now.

**If I remove the ListView component every tab switch takes maximum 300-400ms for reload.**

By the way, I use my Samsung Galaxy S7 for testing, and I didn't had any other application hanging like this since i bought it(6months ago)

Yes, i checked documentation, stackoverflow, all related nativescript issues and all the internet related blog posts.
### Tell us about the problem

I need to prevent recreation of TabView content after the tab becomes inactive, because the switch to a tab with a ListView freezes de UI for 2 seconds.

Android
- CLI: 2.3.0
- Cross-platform modules: 2.3.0 and tested also with @next
- Runtime(s): 2.3.0
- Plugin(s): 
  @angular/common: 2.0.1
  
  @angular/compiler: 2.0.1
  
  @angular/core: 2.0.1
  
  @angular/forms: 2.0.1
  
  @angular/http: 2.0.1
  
  @angular/platform-browser: 2.0.1
  
  @angular/platform-browser-dynamic: 2.0.1
  
  @angular/platform-server: 2.0.1
  
  @angular/router: 3.0.1
  
  nativescript-angular: 1.0.1
  
  nativescript-drop-down: ^1.3.2
  
  nativescript-intl: 0.0.4
  
  nativescript-statusbar: ^1.0.0
  
  nativescript-telerik-ui: ^1.4.1

angular app with 5 tabs(tabview), and insert one ListView/RadListView in one of the tabs, then switch the tabs. Everytime when you switch to the tab with the ListView(from a tab which is not next one or previous one) you get a freeze of 2 seconds.

I will try to make a small demo with the problem soon.

hamorphis:
Resolves #2815

* Gets or sets the number of tabs that should be retained to either side of the current tab in the view hierarchy in an idle state.
* Tabs beyond this limit will be recreated from the TabView when needed.

### DIFF
--- a/tests/app/ui/tab-view/tab-view-tests.ts
+++ b/tests/app/ui/tab-view/tab-view-tests.ts
@@ -275,6 +275,60 @@ export class TabViewTest extends testModule.UITest<tabViewModule.TabView> {
         TKUnit.assertEqual(actualOldIndex, expectedOldIndex, "expectedOldIndex");
         TKUnit.assertEqual(actualNewIndex, expectedNewIndex, "expectedNewIndex");
     }
+
+    public testAndroidOffscreenTabLimit_Default = function () {
+        let tabView = this.testView;
+        if (!tabView.android){
+            return;
+        }
+        
+        // The default setting is 1.
+        // tabView.androidOffscreenTabLimit = 1;
+        tabView.items = this._createItems(20);
+        this.waitUntilTestElementIsLoaded();
+        for (let index = 0, length = tabView.items.length; index < length; index++){
+            tabViewTestsNative.selectNativeTab(tabView, index);
+            TKUnit.waitUntilReady(function () {
+                return tabView.selectedIndex === index;
+            }, helper.ASYNC);
+        }
+
+        let viewsWithParent = 0;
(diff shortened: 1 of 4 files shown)


hamorphis:
uitests

PR #3105, opened 2016-11-16
Fix: connectivity module always returns undefined on android <= 4.4

### LINKED ISSUE #3098: Bug: connectivity module always return undefined on android <= 4.4
Android <= 4.4 (Tested on 4.2.2 and 4.4.2 with emulator)
Android >= 5.0 works as expected (at least all the versions I've tested on)

- CLI: 2.3.0
- Cross-platform modules: 2.3.0
- Runtime(s): tns-android@2.3.0
- Plugin(s): 

*Steps to reproduce: *
* Setup a project, enable android.permission.ACCESS_NETWORK_STATE in `app/App_Resources/Android/AndroidManifest.xml`
* Import the connectivity module;
* Setup and `connectivity.startMonitoring` or call `connectivity.getConnectionType()`
* On older androids (I tested on an emulator running 4.2.2 and 4.4.2, CPU: Google APIs Intel Atom x86) connectivity status is always `undefined`.

Expected result:
Connection status on both older and newer android.

### Is there code involved? If so, please share the minimal amount of code needed to recreate the 
problem.

I've made a simple example here:
https://github.com/m-abs/ns-debugging/tree/connectivity-issue

```typescript
    connectivity.startMonitoring((status) => {
      console.log(status);
      switch (status) {
        case connectivity.connectionType.mobile: {
          this.networkStatus = 'MOBILE';
          break;
        }
        case connectivity.connectionType.wifi: {
          this.networkStatus = 'WIFI';
          break;
        }
        case connectivity.connectionType.none: {
          this.networkStatus = 'OFFLINE';
          break;
        }
        default: {
          this.networkStatus = 'UNKNOWN';
          break;
        }
      }
    });
```

hamorphis:
Resolves #3098

### DIFF
--- a/tns-core-modules/connectivity/connectivity.android.ts
+++ b/tns-core-modules/connectivity/connectivity.android.ts
@@ -4,8 +4,8 @@ import utils = require("utils/utils");
 
 global.moduleMerge(common, exports);
 
-var WIFI = "WIFI";
-var MOBILE = "MOBILE";
+let wifi = "wifi";
+let mobile = "mobile";
 
 // Get Connection Type
 function getConnectivityManager(): android.net.ConnectivityManager {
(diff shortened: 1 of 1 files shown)


hamorphis:
build

PR #3132, opened 2016-11-21
Fix: Modal dialogs don't have background color

### LINKED ISSUE #3129: Modal dialogs don't have background color
In fact, no style properties are ever applied to the Page shown as a modal view with the `showModal` method.

https://github.com/NativeScript/NativeScript/issues/3113

hamorphis:
Resolves #3129


hamorphis:
uitests

SamuelGRwebfab:
when should we expect to be able to use this?

PR #3173, opened 2016-11-25
FlexboxLayout profiling and improvements

PanayotCankov:
Performance tests showed that the flex box in some cases fall up to 30% slower than grid and stack counterparts. After profiling the problems narrow down to intense use of inline styles and some redundant code. Measurement now shows times like:

iOS 
flexbox-grid: 865
gird: 790
-10%

flex-stack: 710
stack: 675
-5%

Android
flexbox-grid: 1115
gird: 1195
+7%

flexbox-stack: 1039
stack: 960
-10%

Which makes the flex box about 10% slower. This should be acceptable. We should also check if we can speed up inline styles.


vchimev:
uitests ios

PR #3189, opened 2016-11-28
Fix: TextView new line closes virtual keyboard on Android

### LINKED ISSUE #3111: TextView new line closes virtual keyboard on Android
I want to enter multiple lines in a TextView. As far as I understood this is what the TextView is for.
I use NativeScript with Angular 2.
I use a simple TextView like this in the xml:
`<TextView></TextView >`
When I tap on it the virtual keyboard opens as expected. Then I enter text. But when I enter return to start a new line, the keyboard closes. The new line is still inserted. But I have to click on the TextView again to reopen the keyboard.

This happens with version 2.3 and 2.4
I use an Android Device (LG Flex 2)

Do I have to configure the TextView somehow? How can I stop the keyboard from closing?

Thanks a lot,

Timo

hamorphis:
Resolves #3111


hamorphis:
uitests

Issue #3181, opened 2016-11-26
Cursor position resets on toggling value of secure property of TextField on Android


shripalsoni04:
### Did you verify this is a real problem by searching [Stack Overflow]
Yes.

### Tell us about the problem
In my nativescript-angular project, I have a `TextField` control as follows:
```
<TextField
    hint="Password"
    [secure]="!isPasswordVisible">
</TextField>
```
Now, when we change `isPasswordVisible` value programmatically, it works great on iOS. But on android it always resets the cursor position to the first character.

I tried to find the root cause of the issue and found that whenever `setInputType` is called, android will reset the cursor position to the first character. 

I checked the code of `text-field.android.js` and the issue is we are calling `setInputType` but we are not setting the cursor to its original position. This results in cursor resets to first character. So just getting the cursor position before calling `setInputType` and setting that cursor position  after calling `setInputType` will solve this issue as shown below.

`text-field.android.js`
```
function onSecurePropertyChanged(data) {
   var textField = data.object;
   ...
   var cursor = textField.android.getSelectionStart();
   ...
   textField.android.setInputType(newInputType);
   textField.android.setSelection(cursor);
}
```

#### Edit
I also noted that currently when `secure` is set to false, the input type is being set to `TYPE_TEXT_VARIATION_NORMAL`. This opens a different soft keyboard than that is visible when `secure` is true. So it will be great if we can have some provision so that when password is visible (`secure` is false), the inputType is set to `TYPE_TEXT_VARIATION_VISIBLE_PASSWORD` instead of `TYPE_TEXT_VARIATION_NORMAL`. This will keep the soft keyboard same irrespective of whether password is visible or not.

### Which platform(s) does your issue occur on?
Android

### Please provide the following version numbers that your issue occurs with:
- CLI: 2.4.0
- Cross-platform modules: 2.4.0
- Runtime(s): 2.4.1

Let me know if you want me to create a PR for this.

NickIliev:
Hey @shripalsoni04

Thanks for reporting this one - we are welcoming all PRs from the community that can improve and solve existing issues. If you want to contribute all you have to do is to sign the CLA and follow the [contributors guide](https://www.nativescript.org/contribute).

gurgenDP:
I have this issue with NativeScript 3.3.1.

AmitGurbani:
Same issue with version 5.4.2

PR #3197, opened 2016-11-29
Fix for issue #3181 of cursor position for secure text-field in Android

shripalsoni04:
In android if we toggle the `secure` property of `text-field`, the cursor position was getting reset to first character. 

This PR solves that issue by setting cursor position after setting `inputType`.

Find more details at #3181.


enchev:
run ci

enchev:
uitests

enchev:
uitests android

PR #3228, opened 2016-12-02
ListPicker respects color and bckgColor set from CSS

zh-m:
So far, ListPicker hasn't respected its color (both for IOS and Android) and background-color (for IOS) properties.

ListPicker {
    color: pink;
    background-color: lightskyblue;
}


zh-m:
uitests

PR #3233, opened 2016-12-02
separatorColor property of ListView can be set in CSS

zh-m:
separatorColor property can be set through CSS as follows:

ListView {
      separator-color: lightskyblue;
}

### DIFF
--- a/tns-core-modules/ui/list-view/list-view.android.ts
+++ b/tns-core-modules/ui/list-view/list-view.android.ts
@@ -8,6 +8,7 @@ import definition = require("ui/list-view");
 import {ProxyViewContainer} from "ui/proxy-view-container";
 import * as layoutBase from "ui/layouts/layout-base";
 import * as colorModule from "color";
+import { separatorColorProperty, registerHandler, Styler, StylePropertyChangedHandler } from "ui/styling/style";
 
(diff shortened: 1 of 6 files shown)


zh-m:
uitests

tns-core-modules/ui/list-view/list-view.android.ts
```diff
@@ -285,4 +286,27 @@ function ensureListViewAdapterClass() {
...
+    // separator-color
+    private static setSeparatorColorProperty(view: viewModule.View, newValue: any) {
+        let listView = <android.widget.ListView>view._nativeView;
+        listView.setDivider(new android.graphics.drawable.ColorDrawable(newValue));
+        listView.setDividerHeight(1);
+    }
+
+    private static resetSeparatorColorProperty(view: viewModule.View, nativeValue: any) {
+        let listView = <android.widget.ListView>view._nativeView;
+        listView.setDivider(new android.graphics.drawable.ColorDrawable(nativeValue));
```
EddyVerbruggen: I had to wrap these 3 lines with `if (nativeValue) {..}` because in my case it was `undefined` and threw an error in the console. Specified the color as `separator-color: #28b83c;` which works fine btw!

PR #3263, opened 2016-12-07
Add note about justify-content

kazemihabib:
Add note for nativescript FlexboxLayout docs.


hamorphis:
@kazemihabib This might be a bug. Can you please open a new issue describing the problem and send a sample layout that exhibits this behavior and we will investigate what is going on.

PR #3285, opened 2016-12-12
Fix: Custom components within other custom components: occasionally…

### LINKED ISSUE #1639: Custom components within other custom components: occasionally not loaded "in time" and ignoring css
I have a page where a custom component is included within another custom component as follows

> Page -> 1st level custom component (child of Page) -> 2nd level custom component (child of 1st level custom component)

With 1st-level only custom components I do **not** have the following issues, but with nested ones yes. Particularly:
- the **css class** applied to the **2nd level** custom component is **ignored**
- on the **1st level component** (pagePart), **loaded event** is sometimes triggered before the 2nd level custom component (switchComponent) is actually available and therefore I can't make any init code, resulting in various errors. It seems happening the 2nd time the <Page> containing the 1st level custom component is loaded. e.g. as in:

> Page with components -> (another) Page -> Page with components

**Note**: I tried to apply loaded=loaded event in <switchComponent:switchComponent id="first-switch" /> in **pagePart.xml** but it's not triggered (I guess because that applies only to <Page>)

Here is an overview of the code:

**main-page.xml**

```
<Page actionBarHidden="true"
 pagePart="includes/customComponents/pagePart"
 xmlns="http://www.nativescript.org/tns.xsd">
         <pagePart:pagePart id="first-part" />
</Page>
```

includes/customComponents/**pagePart.xml**

```
<StackLayout xmlns:switchComponent="includes/customComponents/switchComponent"
 xmlns="http://www.nativescript.org/tns.xsd">
      <switchComponent:switchComponent id="first-switch" />
</StackLayout>
```

includes/customComponents/**pagePart.ts**

```
exports.loaded = function() {
         // here goes init code which involves pagePart with id "first-part"
         // --> ISSUE: from time to time (usually when the view at **Page level** is loaded while another view was loading) 2nd level component (switchComponent) is not available resulting in various errors depending on the init code
}
```

includes/customComponents/**switchComponent.xml**

```
<StackLayout>
    <Image src="some_image" cssClass="myClass">
</Stacklayout>
```

includes/customComponents/**switchComponent.css**

```
.myClass {
 /* here go some styling for my custom component "switchComponent" */
}
```

hamorphis:
… not loaded "in time" and ignoring css

Resolves #1639

### DIFF
--- a/apps/app/ui-tests-app/issues/component/customView.css
+++ b/apps/app/ui-tests-app/issues/component/customView.css
@@ -1,0 +1,4 @@
+.local-green {
+    color: green;
+    margin-left: 10;
+}
--- a/apps/app/ui-tests-app/issues/component/customView.xml
+++ b/apps/app/ui-tests-app/issues/component/customView.xml
@@ -1,4 +1,5 @@
-<StackLayout xmlns:cC2="/ui-tests-app/issues/component2">
-    <Label class="ui-tests-app-issue-1639-green" text="customView green" textWrap="true" />
+<StackLayout xmlns:cC2="/ui-tests-app/issues/component2" style.borderWidth="1" style.borderColor="black">
+    <Label class="ui-tests-app-issue-1639-green" text="customView green from app.css" textWrap="true" />
+    <Label class="local-green" text="customView green from customView.css" textWrap="true" />
     <cC2:customView2/>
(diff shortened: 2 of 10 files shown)


hamorphis:
uitests

PR #3289, opened 2016-12-13
Fix: setInterval callback is not fired while touch events are being p…

### LINKED ISSUE #2116: setInterval callback is not fired while touch events are being processed
### Did you verify this is a real problem

Yes
### Tell us about the problem

setInterval callbacks are essentially paused during a touch interaction. This includes touches that happen via a scrollview, or gestures (like a pan gensture). 

iOS
- CLI: 1.7.0
- Cross-platform modules: 1.7.1
- Runtime(s): 1.7.0
- Plugin(s):  n/a

Create a callback that fires with setInterval, say, every 20 milliseconds (frequently enough for doing a dynamic animation, like with physicsjs). Create a scrollview or add a gesture recognizer to a view, and note that when the app is processing touch events, the setInterval callback does not fire. Once the touch interaction is done processing, the setInterval callback resumes. 

```
setInterval(()=>console.log(Date.now()));
// scrollview or something that handles touches. 
```

hamorphis:
…rocessed

Resolves #2116

### DIFF
--- a/tns-core-modules/timer/timer.ios.ts
+++ b/tns-core-modules/timer/timer.ios.ts
@@ -1,4 +1,12 @@
-﻿//iOS specific timer functions implementation.
+﻿import * as utilsModule from "utils/utils";
+let utils: typeof utilsModule;
+function ensureUtils() {
+    if (!utils) {
+        utils = require("utils/utils");
+    }
+}
+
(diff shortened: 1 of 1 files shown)


hamorphis:
uitests

PR #3395, opened 2017-01-03
fix TabView

hshristov:
1. TabViewItem is now added as a child to TabView
2. TabViewItem.view can now be set directly without specifying `.view` element:
```
<TabViewItem>
    <StackLayout>
        <Label text="no .view!!!" />
    </StackLayout>
</TabViewItem>
```


vchimev:
run ci

PR #3405, opened 2017-01-04
Enable evaluating in Chrome Console

KristinaKoeva:
Chrome calls compileScript method when an expression is evaluated in the console. So to enable the console evaluate add an empty compileScript implementation.
Also send an Runtime.executionContextCreated which is needed for Console evaluation.
Both of the command and the event do not exist in the Webkit protocol, so we cannot use InspectorBackendCommands definition file

### DIFF
--- a/tns-core-modules/debugger/webinspector.ios.ts
+++ b/tns-core-modules/debugger/webinspector.ios.ts
@@ -3,6 +3,7 @@ var inspectorCommands: typeof inspectorCommandTypes = require("./InspectorBacken
 
 import * as debuggerDomains from "./debugger";
 
+declare var __inspectorSendEvent;
 /**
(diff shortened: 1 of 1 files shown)


KristinaKoeva:
run ci

vchimev:
build

vchimev:
build

vchimev:
ci/jenkins/build — PASSed.

KristinaKoeva:
@vchimev So everything is green now, right ? Would somebody review this PR and merge it, because otherwise we would not be able to evaluate in the Chrome Dev Tools Console.

vchimev:
@KristinaKoeva, yes, unit tests passed. Ping @hshristov for a review.

PR #3432, opened 2017-01-06
Pin typescript dependency to 2.0.x

hdeshev:
Works around an incompatibility between TypeScript 2.1 and the android runtime tools.


hdeshev:
run ci

vchimev:
build

Issue #3492, opened 2017-01-15
Add optional callback parameter to utils.ios.openFile


jzgoda:
### Did you verify this is a real problem by searching [Stack Overflow](http://stackoverflow.com/questions/tagged/nativescript) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)? 
Yes

### Tell us about the problem
There is a lag between the time that `utils.ios.openFile` is called, and the file that is being opened is shown on screen. Adding an `ActivityIndicator` takes care of the user experience, but there is no way to know when to turn it off. `utils.ios.openFile` returns a boolean almost immediately, long before the file is displayed.  

### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- Cross-platform modules: suggesting a change to the [latest version of source](https://github.com/NativeScript/NativeScript/blob/84c469d61b148544bd1ef8b4f41ce689f9145754/tns-core-modules/utils/utils.ios.ts)

### Please tell us how to recreate the issue in as much detail as possible.
All uses of `utils.ios.openFile`

### Suggested Fix
I ended up creating my own version (at the suggestion of @NathanaelA), mimicking the functionality found in the current source. I added an optional callback parameter, which would allow this additional functionality, without breaking current uses.

Updated `utils.d.ts` definiton: 
```javascript
export function openFile(filePath: string, callback?: Function): boolean
```

Updated implementation of `openFile` (look at `if (callback) {...}`):
```javascript
export function openFile(filePath: string, callback?: Function): boolean {
	try {
		var fs: typeof fsModule = require("file-system");
		var path = filePath.replace("~", fs.knownFolders.currentApp().path);

		var controller = UIDocumentInteractionController.interactionControllerWithURL(NSURL.fileURLWithPath(path));
		controller.delegate = new UIDocumentInteractionControllerDelegateImpl();
		if (callback) {
			UIDocumentInteractionControllerDelegateImpl.didEndPreview.on('didEndPreviewEventName', (event) => {
				UIDocumentInteractionControllerDelegateImpl.didEndPreview.off('didEndPreviewEventName');
				callback();
			});
		}
		return controller.presentPreviewAnimated(true);
	}
	catch (e) {
		ensureTrace();
		trace.write("Error in openFile", trace.categories.Error, trace.messageType.error);
	}
	return false;
}
```

[...]

PR #3502, opened 2017-01-17
Add 'DidEndPreview' listener to utils.is.openFile

jzgoda:
Implements #3492.


tns-core-modules/utils/utils.ios.ts
```diff
@@ -185,6 +192,7 @@ export function openUrl(location: string): boolean {
 
 class UIDocumentInteractionControllerDelegateImpl extends NSObject implements UIDocumentInteractionControllerDelegate {
     public static ObjCProtocols = [UIDocumentInteractionControllerDelegate];
+    public static didEndPreview = new Observable();
```
hshristov: Could you try moving this `Observable` as instance field and attach it to the `controller`?
Then in `documentInteractionControllerDidEndPreview` you could take it from the `controller` parameter.

jzgoda: Can you provide me with some additional guidance as to what you mean by attaching the variable to the `controller`? (I'm not familiar with `UIDocumentInteractionController` or Obj-C.)

But, I can turn it into an instance variable like this:
```javascript
export function openFile(filePath: string, callback?: Function): boolean {
    ...

        var controller = UIDocumentInteractionController.interactionControllerWithURL(NSURL.fileURLWithPath(path));
        var delegate = new UIDocumentInteractionControllerDelegateImpl();
        controller.delegate = delegate;
        if (callback) {
            delegate.didEndPreview.on('didEndPreviewEventName', () => {
                delegate.didEndPreview.off('didEndPreviewEventName');
                callback();
            });
        }  

    ...
}
```
```javascript
class UIDocumentInteractionControllerDelegateImpl extends NSObject implements UIDocumentInteractionControllerDelegate {
    public static ObjCProtocols = [UIDocumentInteractionControllerDelegate];
    public didEndPreview = new Observable();

    ...

    public documentInteractionControllerDidEndPreview(controller: UIDocumentInteractionController) {
        let eventData: EventData = {
            eventName: "didEndPreviewEventName",
            object: this.didEndPreview
        };
        this.didEndPreview.notify(eventData);
    }
}
```

Would that be better?

jzgoda: @hshristov, I tried these changes that I just presented, and they don't seem to work...the `delegate` doesn't seem to be getting created properly. 

However, my original PR does work, and I'm still not completely sure what alternative route you want me to go down. Please let me know how I can help to get this accepted.  
Thanks.

vakrilov:
Closing due to inactivity and really really hard rebase.

@jzgoda If you are willing to give this one another try and base your changes on the latest code in master - it would be a blast.

You can also check our new [contributing guide](https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md)

jzgoda:
@vakrilov I'll be doing some work on the app that uses this feature in the next couple of months, so I will give it another shot with the latest version.

PR #3531, opened 2017-01-26
formatted text fix

hshristov:
Fix formattedText implementation for iOS
Fix typeof undefined checks
Add slow tests message
Improve tests speed
fix tslint
reduce transitions duration in test


tests/app/ui/list-view/list-view-tests.ts
```diff
@@ -100,42 +100,42 @@ export class ListViewTest extends testModule.UITest<listViewModule.ListView> {
...
-            } else if (platform.device.os === platform.platformNames.android) {
-                indexes[args.index] = args.android;
-            }
+            indexes[args.index] = isIOS ? args.ios : args.android;
         });
-        this.waitUntilListViewReady();
+        listView.items = colors;
+        TKUnit.waitUntilReady(() => indexes.length === 3);
 
         for (var item in indexes) {
```
vakrilov: Maybe add a check that `indexes.length` is actually `3` or the tests just timed-out

PR #3556, opened 2017-02-01
Segmented bar styles

vakrilov:


tns-core-modules/ui/segmented-bar/segmented-bar-common.ts
```diff
@@ -111,5 +113,5 @@ export const itemsProperty = new Property<SegmentedBarBase, SegmentedBarItemDefi
 });
 itemsProperty.register(SegmentedBarBase);
 
-export const selectedBackgroundColorProperty = new CssProperty<Style, Color>({ name: "selectedBackgroundColor", cssName: "selected-background-color", equalityComparer: Color.equals, valueConverter: (v) => new Color(v) })
+export const selectedBackgroundColorProperty = new InheritedCssProperty<Style, Color>({ name: "selectedBackgroundColor", cssName: "selected-background-color", equalityComparer: Color.equals, valueConverter: (v) => new Color(v) });
```
hshristov: Should we make this property Inheritable? Now it is possible to define style for SegmentedBarItem so you could style it without the need to make it inheritable.

vakrilov: I did it for back-compatibility. In NS 2.x the `selected-backgorund-color` was only available on `SegmentedBar` so everyone who uses it will be broken if it is not inheritable.

PR #3585, opened 2017-02-06
Ensure iOS navigation title view is removed if undefined

lfabreges:
This small PR is useful to ensure that the titleView is reset when using multiples ActionBar on iOS. Bellow an example with angular and tabs:

First tab:
```xml
<ActionBar *ngIf="selected" class="action-bar" title="{{ 'page.home.title' | L }}">
  <ActionItem (tap)="goToAdd()"
              ios.systemIcon="4" ios.position="right"
              android.systemIcon="ic_menu_add" android.position="actionBar">
  </ActionItem>
</ActionBar>
```

Second tab:
```xml
<ActionBar *ngIf="selected" class="action-bar" title="{{ 'page.search.title' | L }}">
  <SearchBar hint="Search..."></SearchBar>
</ActionBar>
```

If I navigate to the second tab, the search bar is shown, then when I navigate back to the first tab, it isn't removed, this PR ensure that it is.

### DIFF
--- a/tns-core-modules/ui/action-bar/action-bar.ios.ts
+++ b/tns-core-modules/ui/action-bar/action-bar.ios.ts
@@ -61,6 +61,9 @@ export class ActionBar extends common.ActionBar {
         if (this.titleView && this.titleView.ios) {
             navigationItem.titleView = this.titleView.ios;
         }
+        else {
+            navigationItem.titleView = null;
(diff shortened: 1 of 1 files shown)


vakrilov:
Closing in favour of #3953.
@lfabreges - thanks for the contribution

PR #3596, opened 2017-02-07
Animation properties and some backward compatability with the QSF

PanayotCankov:
Added CssAnimationProperty that supports multiple value sources. Accommodates for the CSS keyframe animation value source.


apps/app/package.json
```diff
@@ -1,4 +1,4 @@
 {
 	"name": "tns-samples-apps",
-	"main": "ui-tests-app/app.js"
```
vakrilov: Better keep the default app `ui-tests-app`

PR #3636, opened 2017-02-14
TypeScript and typings upgrade

hdeshev:
- Upgrade to TypeScript 2.1 for internal builds
- Kill grunt-ts and a bunch of related plugins. Use a plain `tsc` call to compile.
- Kill grunt-tslint. See above.
- Make typings compatible with the DOM TypeScript typings and get rid of name clashes.
- Make typings compatible with @types/node typings and get rid of name clashes.
- Use a shared tsconfig.shared.json file to avoid duplicating compiler options across projects (TS 2.1 feature)

Fixes NativeScript/nativescript-dev-typescript#19
Implements #3642


tns-core-modules/declarations.d.ts
```diff
@@ -99,21 +99,12 @@ declare type RequestInfo = Request|string;
...
 interface Console {
-    time(reportName: string): void;
-    timeEnd(reportName: string): void;
-    assert(test: boolean, message: string, ...formatParams: any[]): void;
-    info(message: any, ...formatParams: any[]): void;
-    warn(message: any, ...formatParams: any[]): void;
-    error(message: any, ...formatParams: any[]): void;
-    log(message: any, ...formatParams: any[]): void;
     trace(): void;
     dump(obj: any): void;
```
PanayotCankov: Copy the docs here.

tns-core-modules/timer/timer.ios.ts
```diff
@@ -63,19 +63,19 @@ function createTimerAndGetId(callback: Function, milliseconds: number, shouldRep
     return id;
 }
 
-export function setTimeout(callback: Function, milliseconds = 0): number {
-    return createTimerAndGetId(zonedCallback(callback), milliseconds, false);
+export function setTimeout(callback: Function, milliseconds = 0): NodeJS.Timer {
+    return <NodeJS.Timer><any>createTimerAndGetId(zonedCallback(callback), milliseconds, false);
```
PanayotCankov: Instead `<NodeJS.Timer><any>` use just `<any>`.

PR #3694, opened 2017-02-23
Fix the travis PR build

hdeshev:
- Use tns-core-modules-widgets@rc
- Pin to android runtime 2.5.0 since the `next` build currently has an issue that eats `console.log` messages.


hdeshev:
run ci

vchimev:
run ci

PR #3748, opened 2017-03-08
Pixels ios

PanayotCankov:
Length offers both dip and px as units but px so far behaved as dip in iOS. This PR aims to resolve 1px on retina display (@2, @3) to 1/2 dip or 1/3 dip respectively. All native APIs use dip, {N} layouts work in px, some APIs like the GridLayout that support rows and columns still support just dip and as such some actual sizes return dips. Effective values are supposed to work in px. Layout rounding will now round to px even when sizes are specified in px.

### DIFF
--- a/tns-core-modules/ui/utils.ios.ts
+++ b/tns-core-modules/ui/utils.ios.ts
@@ -18,7 +18,8 @@ export module ios {
         }
 
         var statusFrame = app.statusBarFrame;
-        return Math.min(statusFrame.size.width, statusFrame.size.height);
+        let min = Math.min(statusFrame.size.width, statusFrame.size.height);
+        return utils.layout.toDevicePixels(min);
     }
 
     export function _layoutRootView(rootView: View, parentBounds: CGRect) {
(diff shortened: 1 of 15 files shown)


tns-core-modules/ui/utils.ios.ts
```diff
@@ -38,8 +39,8 @@ export module ios {
 
         if (utils.ios.MajorVersion < 8 && utils.ios.isLandscape() && !superViewRotationRadians) {
             // in iOS 7 when in landscape we switch width with height because on device they don't change even when rotated.
-            width = size.height;
-            height = size.width;
+            width = utils.layout.toDevicePixels(size.width);
+            height = utils.layout.toDevicePixels(size.height);
```
vakrilov: I think `width` and `height` are switched on purpose in the original code. However, you can consider removing that at all as we don't claim to support `MajorVersion < 8` any more.

hshristov:
run ci

SvetoslavTsenov:
run ci

PR #3757, opened 2017-03-08
FIX: Android keyboard

vakrilov:
Fixes:
 - Android keyboard does not hide when focusing away from TextView
 -  The `nativeView` parameter of `dismissSoftInput` (from `utils`) is now optional.

### DIFF
--- a/tns-core-modules/utils/utils.android.ts
+++ b/tns-core-modules/utils/utils.android.ts
@@ -4,7 +4,7 @@
 
 export * from "./utils-common";
 
-import { getNativeApplication } from "application";
+import { getNativeApplication, android as androidApp } from "application";
 
 export module layout {
     let density: number;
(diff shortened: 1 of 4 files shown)


vakrilov:
run ci

tns-core-modules/utils/utils.android.ts
```diff
@@ -98,10 +98,19 @@ export module ad {
...
+    export function dismissSoftInput(nativeView?: android.view.View): void {
         const inputManager = getInputMethodManager();
-        if (inputManager && nativeView instanceof android.view.View && inputManager.isActive(nativeView)) {
-            inputManager.hideSoftInputFromWindow(nativeView.getWindowToken(), 0);
+        let windowToken: android.os.IIBinder;
+
+        if (nativeView instanceof android.view.View) {
+            windowToken = nativeView.getWindowToken()
+        } else if (androidApp.foregroundActivity instanceof android.app.Activity) {
+            const currentFocus = androidApp.foregroundActivity.getCurrentFocus();
```
hshristov: isn't getDecorView method better because it should always return android.view.View?

PR #3782, opened 2017-03-13
whiteSpace default value change to undefined

hshristov:
TKUnit default message change to empty string
isSet method is now instance method of Property classes
fix detaching from parent bindingContext - were using oldParent.parent instead of parent
editable-text-base.android - onTextChanged implementation commented. Does nothing.
frame - onCreateView wrapped in try/catch and shows label with exception message if any
text-base.android - should support reset of nativeView. TransformationMethod won’t be set if TextField is secure
Change some types to their string couterparts
TextField.android won’t support multilines anymore in order to work as iOS
In android when page is removed from native backstack we won’t call tearDownUI again a second time

### DIFF
--- a/tns-core-modules/ui/button/button.android.ts
+++ b/tns-core-modules/ui/button/button.android.ts
@@ -1,6 +1,7 @@
 ﻿import {
     ButtonBase, PseudoClassHandler,
-    paddingLeftProperty, paddingTopProperty, paddingRightProperty, paddingBottomProperty, Length, zIndexProperty
+    paddingLeftProperty, paddingTopProperty, paddingRightProperty, paddingBottomProperty, 
+    Length, zIndexProperty
 } from "./button-common";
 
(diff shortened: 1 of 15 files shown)


tns-core-modules/ui/button/button.android.ts
```diff
@@ -35,16 +36,13 @@ function initializeClickListener(): void {
 
 export class Button extends ButtonBase {
     _button: android.widget.Button;
```
vakrilov: `_button` is never used except for setting it - better remove it at all.

tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
```diff
@@ -39,11 +39,11 @@ function initializeEditTextListeners(): void {
         }
 
         public onTextChanged(text: string, start: number, before: number, count: number) {
```
vakrilov: Any ideas why this code was there at all? If not used - better delete it.

PR #3796, opened 2017-03-15
Network Agent implementation (Android)

Pip3r4o:
Add calls to runtime-exposed methods to communicate with the Network Agent of Chrome DevTools inside the android `http-request` implementation. Method calls to the Network Agent will only be made in debug and when the inspector has been fired up, meaning no overhead should occur when not debugging inside DevTools or VSCode

This PR depends on a small addition in the widgets -> https://github.com/NativeScript/tns-core-modules-widgets/pull/91

### DIFF
--- a/tns-core-modules/http/http-request/http-request.android.ts
+++ b/tns-core-modules/http/http-request/http-request.android.ts
@@ -8,6 +8,8 @@ import * as fsModule from "../../file-system";
 // this is imported for definition purposes only
 import * as http from "../../http";
 
+import { NetworkAgent } from "../../debugger/debugger";
+
 export const enum HttpResponseEncoding {
     UTF8,
     GBK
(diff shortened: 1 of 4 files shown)


tns-core-modules/http/http-request/http-request.android.ts
```diff
@@ -75,6 +94,56 @@ function onRequestComplete(requestId: number, result: org.nativescript.widgets.A
         }
     }
 
+    if (global.__inspector && global.__inspector.isConnected) {
```
vakrilov: Suggestion: move the contents of the if with all the inspector related logic in a separate method.

tns-core-modules/http/http-request/http-request.android.ts
```diff
@@ -194,6 +263,25 @@ export function request(options: http.HttpRequestOptions): Promise<http.HttpResp
             // initialize the options
             var javaOptions = buildJavaOptions(options);
 
+            if (global.__inspector && global.__inspector.isConnected) {
```
vakrilov: Suggestion: move the contents of the if with all the inspector related logic in a separate method.

Pip3r4o: Moved the calls to the network agent inside the debugger module where I should have put them initially. Thanks for the suggestions!

PR #3815, opened 2017-03-16
Always attach console implementation on android

Pip3r4o:
Recent fixes in the android debugger forced that the same v8 (javascript) context is used in both user code execution and devtools inspector. That meant that attaching a debugger (building in debug configuration) will apply predefined properties to the global object (`console` being one). This would cause the check in `globals` module to fail when checking whether to initialize the custom implementation. 

The proposed changes ensure that the modules' console implementation will always be used when running on Android devices, while still not polluting the iOS's global object.

### DIFF
--- a/tns-core-modules/globals/globals.ts
+++ b/tns-core-modules/globals/globals.ts
@@ -121,7 +121,11 @@ if ((<any>global).__snapshot) {
     registerOnGlobalContext("fetch", "fetch");
 }
 
-if (!(<any>global).console) {
+// check whether the 'android' namespace is exposed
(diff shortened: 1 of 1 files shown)


tns-core-modules/globals/globals.ts
```diff
@@ -121,7 +121,7 @@ if ((<any>global).__snapshot) {
     registerOnGlobalContext("fetch", "fetch");
 }
 
-if (!(<any>global).console) {
```
vakrilov: Can you leave a comment in the code why this is needed?

PR #3843, opened 2017-03-21
FIX: WebView not showing in gallery app

vakrilov:
Using `url` will throw exception now.

### DIFF
--- a/apps/app/gallery-app/content/web-view.xml
+++ b/apps/app/gallery-app/content/web-view.xml
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <Page>
-  <WebView margin="10" url="http://www.telerik.com/nativescript"/>
+  <WebView margin="10" src="http://www.telerik.com/nativescript"/>
 </Page>
--- a/tns-core-modules/ui/web-view/web-view-common.ts
+++ b/tns-core-modules/ui/web-view/web-view-common.ts
@@ -21,15 +21,9 @@ export abstract class WebViewBase extends View implements WebViewDefinition {
         "other"
     ];
 
-    public _suspendLoading: boolean;
-
     public src: string;
 
     public _onLoadFinished(url: string, error?: string) {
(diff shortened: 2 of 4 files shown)


SvetoslavTsenov:
👍

EddyVerbruggen:
Nooooooooooooo! I need that property. There is no other way to load a local html file with a querystring; passing `~/file.html` to `src` works, but `~/file.html?foo=bar` loads a blank page because it can't find the local file. `url` works perfectly for that case.

vakrilov:
@EddyVerbruggen We should be able to make this work with `src` property as well. 
Do you have specific html doc that we can test with? 
Also is this android/ios (or both) related issue?

EddyVerbruggen:
@vakrilov That would be awesome!

Theoretically (looking at the code, which is in the common file of the module) it's for both platforms, but I only verified this on iOS.

What I'm doing now is loading a map with natively determined lat/lng (to suppress a very nasty web-geolocation prompt on iOS):

```js
webView.url = `file:///${fs.knownFolders.currentApp().path}/mypath/map.html?lat=${location.lat}&lng=${location.lng}`;
```

With the `src` attribute that would be:
```js
webView.src = `~/mypath/map.html?lat=${location.lat}&lng=${location.lng}`;
```

The local HTML file queries the url and extracts the lat/lng params. But it could be anything of course. You can just do:

```js
webView.src = '~/mypath/myfile.html?foo=bar';
```

Currently it won't find `myfile.html` at all because of the query params, but if you can make that work again you can alert the URL in `myfile.html` - if that still contains `?foo=bar` then we can use `src` instead of `url`.

```html
<!DOCTYPE html>
<html>
  <script type="text/javascript" charset="utf-8">
    alert("URL: " + document.location.href);
  </script>
</html>
```

speigg:
IMO, I think we need to keep both:

* `src` should be the url that was explicitly set
* `url` should be the current url (following redirects, navigation, etc)

PR #3857, opened 2017-03-23
Fix several angular animation issues

hdeshev:
- Not crashing when we get a LinearInterpolator for android curves.
- Always starting animations and not doing that only for fully loaded elements (nothing to start them once the element loads)
- Catch rejections in keyframe child animations to avoid unhandled rejection warnings.

Fixes NativeScript/nativescript-angular#431, NativeScript/nativescript-angular#534

### DIFF
--- a/tns-core-modules/ui/animation/keyframe-animation.ts
+++ b/tns-core-modules/ui/animation/keyframe-animation.ts
@@ -212,9 +212,14 @@ export class KeyframeAnimation implements KeyframeAnimationDefinition {
             let animationDef = this.animations[index];
             (<any>animationDef).target = view;
             let animation = new Animation([animationDef]);
+            // Catch the animation cancel to prevent unhandled promise rejection warnings
             animation.play().then(() => {
                 this.animate(view, index + 1, iterations);
-            });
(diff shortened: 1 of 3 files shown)


irman:
Has this been lived? Can't seem to do normal navigation in ios. The `animation cancelled` error keeps popping up.

tns-core-modules/ui/animation/keyframe-animation.ts
```diff
@@ -212,9 +212,14 @@ export class KeyframeAnimation implements KeyframeAnimationDefinition {
             let animationDef = this.animations[index];
             (<any>animationDef).target = view;
             let animation = new Animation([animationDef]);
+            // Catch the animation cancel to prevent unhandled promise rejection warnings
             animation.play().then(() => {
                 this.animate(view, index + 1, iterations);
-            });
+            }).catch((error: any) => {
+                if (error.message.indexOf("Animation cancelled") < 0) {
+                    throw error;
```
NathanWalker: @hdeshev this `throw` defeats the solution and destabilizes apps.

Issue #3889, opened 2017-03-28
Add support for both 'focusLost' and 'textChanged' events for Text input (or) add a focusLost/blur Event


EddyVerbruggen:
Hi,

I encountered a case where I'd like to observe both `focusLost` and `textChanged` events on a `TextField`. There seems to be no way to currently observe a "blur" event when `textChanged` is required as well.

So can [this enum](https://github.com/NativeScript/NativeScript/blob/17610daa9a182fd999efb7c5b69a983015801b17/tns-core-modules/ui/enums/enums.ts#L90-L93) please be extended with an `all`?

I'm up for a PR if you think this is a good idea.

Thanks,
Eddy

EddyVerbruggen:
While we're on the subject.. it may be even more useful to have a dedicated `focusLost` event that binds to a function. So for Angular: `(focusLost)="myBlurHandler($event)"`.

vakrilov:
About the `UpdateTextTrigger` enum - generally `textChange` is more frequent than `focusLost` and it doesn't make sense to trigger text change twice with the same text - after the last change and on lost focus.
I think a dedicated `focusLost` or even `blur`would be more useful. 
PR would be great - thanks @EddyVerbruggen

EddyVerbruggen:
@vakrilov Yeah that makes more sense to me as well. Could be a bit harder to pull off but I'll give it a shot 👍

PR #3896, opened 2017-03-29
#3889 Add a 'blur' event to text input fields

EddyVerbruggen:
Implements #3889


EddyVerbruggen:
Locally tested with a {N} 3.0 non-Angular starter app with this added to the XML:

```xml
        <TextField text="blurrable textfield" blur="{{ onBlur }}"/>
        <TextField text="non-blurrable textfield"/>
```

And this added to `main-view-model.js`:

```js
    viewModel.onBlur = function() {
        console.log("Blurred!");
        this.set("message", "Blurred @ " + new Date().getTime());
    }
```

hshristov:
run ci

Issue #2740, opened 2016-09-16
Feature Request: Improved ActionBar CSS Support


firescript:
In order to take out the action bar shadow you must use js, here is the code to do it.

```
    if (page.ios) {
        navigationBar.shadowImage = new UIImage();
        navigationBar.setBackgroundImageForBarMetrics(new UIImage(), UIBarMetrics.UIBarMetricsDefault)
    }   
```

I think it would make sense to make use of the border css property to over ride that. Setting `border: 0;` on an ActionBar, should remove that naturally. 

The other thing is I cant seem to set the actionbar to transparent/transulcent? `background-color:transparent;` or `background-color: rgba(0,0,0,.8)` would make sense, in combination with the removal of the shadow on the action bar, it would give me a clean, transparent/transulcent actionbar.

sitefinitysteve:
+1 for "transparent"

wnas:
+1 for border 0 in css.

pmz:
+1. This is bugging me :)

EddyVerbruggen:
Please join the discussion at #3900, thanks!

vakrilov:
@EddyVerbruggen Thanks for the great PR - it is merged now.
@SvetoslavTsenov Can you please add this scenario in the e2e tests for the action-bar.

SvetoslavTsenov:
A test is already included but I found an issue on iOS. When we update the flat prop. from code behind it doesn't update the UI.
Here is the sample page that we use:
https://github.com/NativeScript/NativeScript/blob/master/apps/app/ui-tests-app/action-bar/flat.ts
Please have in mind that so far for Android is it working perfectly fine.

PR #3900, opened 2017-03-29
Implement an ActionBar 'flat' property to remove the bottom border/shadow and (on iOS) translucency

EddyVerbruggen:
Implements #2740:

I wanted to have an ActionBar that seamlessly blends with the page, like the screenshot at the bottom shows.

That was only possible by making the background colors of the `ActionBar` and the `StackLayout` below it match (setting `translucent` to `false` on the `ActionBar`, and removing the "1px" border at the bottom of the `ActionBar`.

That works fine in my project so I thought I'd share and that's when I encountered #2740 by @firescript.

This PR allows setting `border: 0` / `border-width: 0` / `border-bottom-width: 0` to get rid of the line/shadow at the bottom of the `ActionBar`. It is a bit opinionated though as it will also clear the translucency of the `ActionBar` when the border is removed. Which (as @firebase seems to agree with) makes most sense.

**There is one big problem with this PR though: it will change the default style of the `ActionBar` for iOS as there's seems to be no way to distinguish between an "unset" border width and one being 0 (as that's the default). That's why I've labeled it as [Do not merge] and seek advise. I personally don't mind this default and it can be "fixed" by using platform-specific css files, but still it's a breaking change UI-wise.**

simulator screen shot 29 mar 2017 20 59 18 2


EddyVerbruggen:
I'm thinking it may be better to add a new `flat` property (default `false`) to the `ActionBar`.

When set it removes the shadow for Android and the line/shadow for iOS, and also sets `translucent` to `false` on iOS.

Thoughts?

SvetoslavTsenov:
run uitests

tjvantoll:
I prefer the idea of the `flat` attribute that @EddyVerbruggen suggests as an alternative to the current border-based implementation. With the implementation currently in the PR adding `ActionBar { border-bottom: 0; }` would change your ActionBar’s background color on iOS, and I think that could be a bit of an unexpected side effect.

`<ActionBar flat="true">` makes more sense to me, and I think it’s a common enough problem that it’s worthy being included in the core modules.

SvetoslavTsenov:
run uitests ios

EddyVerbruggen:
Hi,

I've now finished the implementation of the `flat` property (default `false` so current behavior is not changed).

vchimev:
run ci

vakrilov:
run uitests

tns-core-modules/ui/action-bar/action-bar.android.ts
```diff
@@ -382,6 +383,18 @@ export class ActionBar extends ActionBarBase {
...
         this.nativeView.setTitleTextColor(color);
     }
+
+    [flatProperty.setNative](value: boolean) {
+        const compat = <any>android.support.v4.view.ViewCompat;
+        if (compat.setElevation) {
+            if (value) {
+                compat.setElevation(this.nativeView, 0);
+            } else {
+                const val = DEFAULT_ELEVATION * layout.getDisplayDensity();
```
vakrilov: Is there a way to get the default elevation directly form the view.
If the elevations is **4dips** it's OK be a constants as it is now. If it is something that is ApiLevel or theme specific - it would be better to use the `getDefault` callback and get it runtime.

EddyVerbruggen: It seems to change per API level a bit and I think you can determine the default level, but I didn't add that because of [this precedent](https://github.com/NativeScript/NativeScript/blob/23757e5dfcd6ea58153a0f724b3bb54f2489d9d4/tns-core-modules/ui/tab-view/tab-view.android.ts#L254) I found in the code.

vakrilov: Fair enough

vakrilov:
run uitests

vbresults:
Not working on tns-core-modules and nativescript @rc (3.2.0)

Issue #3614, opened 2017-02-08
Feature request: TextField `maxLength` property support.


tsonevn:
Including new TextField property, which allows to specifying the maximum length of the text

Workaround
Android 
```
textfieldloaded(args){
var textfield = args.object;
var legth = parseInt(this.maxvalue);
            var array = [];
            array[0] = new android.text.InputFilter.LengthFilter(legth);
            textfield .android.setFilters(array);
}
```

iOS
```
textfieldloaded(args){
var textfield = args.object;
var legth = parseInt(this.maxvalue);
           var uiTextView = textfield .ios;
           let tf = <any>this.textField;
           var newWeakRef = new WeakRef(uiTextView);
           let newDelegate = newUITextFieldDelegateImpl.initWithOriginalDelegate(tf._delegate, legth);
           uiTextView.delegate = newDelegate;
}

class newUITextFieldDelegateImpl extends NSObject implements UITextFieldDelegate {

    public static ObjCProtocols = [UITextFieldDelegate];

    private _originalDelegate: UITextFieldDelegate;
    private firstEdit: boolean;
    private maxlegth:number

    public static initWithOriginalDelegate(originalDelegate: UITextFieldDelegate, maxl:number): newUITextFieldDelegateImpl {
        console.log("initWithOwner")

        let delegate = <newUITextFieldDelegateImpl>newUITextFieldDelegateImpl.new();
        delegate._originalDelegate = originalDelegate;
        delegate.maxlegth=maxl;
        return delegate;
    }

    public textFieldShouldBeginEditing(textField: UITextField): boolean {
        return this._originalDelegate.textFieldShouldBeginEditing(textField);
    }

    public textFieldDidEndEditing(textField: UITextField) {
        return this._originalDelegate.textFieldDidEndEditing(textField);
    }

    public textFieldShouldClear(textField: UITextField) {
    

        return this._originalDelegate.textFieldShouldClear(textField);
    }

    public textFieldShouldReturn(textField: UITextField): boolean {
        
        return this._originalDelegate.textFieldShouldReturn(textField);
    }

    public textFieldShouldChangeCharactersInRangeReplacementString(textField: UITextField, range: NSRange, replacementString: string): boolean {
        if ((this.maxlegth <= textField.text.length)&&(replacementString.length > range.length)) {
            return false;
        }

        return this._originalDelegate.textFieldShouldChangeCharactersInRangeReplacementString(textField, range, replacementString);
    }
    
}
```

[...]

PR #3902, opened 2017-03-29
Implemented feature request: TextField `maxLength` property support.

EddyVerbruggen:
Implements #3614

Based on the code samples @tsonevn supplied. Works like a boss in both `TextField` and `TextView` on iOS and Android.

_I would have loved to add some tests, but for the life of me I can't seem to run them. I tried following the suggestions in the readme's but to no avail._

### DIFF
--- a/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
+++ b/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
@@ -2,7 +2,7 @@
     EditableTextBase as EditableTextBaseCommon, keyboardTypeProperty,
     returnKeyTypeProperty, editableProperty,
     autocapitalizationTypeProperty, autocorrectProperty, hintProperty,
-    textProperty, placeholderColorProperty, Color
+    textProperty, placeholderColorProperty, Color, maxLengthProperty
 } from "./editable-text-base-common";
 
 import { ad } from "../../utils/utils";
@@ -409,5 +409,9 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
         } else {
             this.nativeView.setHintTextColor(value);
         }
+    }
+
+    [maxLengthProperty.setNative](value: number) {
+        this.nativeView.setFilters([new android.text.InputFilter.LengthFilter(+value)]);
     }
 }
--- a/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
+++ b/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
@@ -412,6 +412,21 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
     }
 
     [maxLengthProperty.setNative](value: number) {
-        this.nativeView.setFilters([new android.text.InputFilter.LengthFilter(+value)]);
+        let lengthFilter = new android.text.InputFilter.LengthFilter(+value);
+        let filters = this.nativeView.getFilters();
+        let newFilters = [];
+
+        // retain existing filters
+        for (let i = 0; i < filters.length; i++) {
+            newFilters.push(filters[i]);
+            // update an existing length filter
+            if (filters[i] instanceof android.text.InputFilter.LengthFilter) {
+                filters[i] = lengthFilter;
+                return;
+            }
+        }
+        // no current length filter found, so add it
+        newFilters.push(lengthFilter);
+        this.nativeView.setFilters(newFilters);
     }
 }
--- a/tns-core-modules/ui/text-view/text-view.ios.ts
+++ b/tns-core-modules/ui/text-view/text-view.ios.ts
@@ -2,7 +2,7 @@
 import {
     EditableTextBase, editableProperty, hintProperty, textProperty, colorProperty, placeholderColorProperty,
     borderTopWidthProperty, borderRightWidthProperty, borderBottomWidthProperty, borderLeftWidthProperty,
-    paddingTopProperty, paddingRightProperty, paddingBottomProperty, paddingLeftProperty, maxLengthProperty,
+    paddingTopProperty, paddingRightProperty, paddingBottomProperty, paddingLeftProperty,
     Length, _updateCharactersInRangeReplacementString, Color, layout
 } from "../editable-text-base";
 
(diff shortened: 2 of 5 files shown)


tns-core-modules/ui/text-view/text-view.ios.ts
```diff
@@ -51,8 +51,14 @@ class UITextViewDelegateImpl extends NSObject implements UITextViewDelegate {
 
     public textViewShouldChangeTextInRangeReplacementText(textView: UITextView, range: NSRange, replacementString: string): boolean {
         const owner = this._owner.get();
-        if (owner && owner.formattedText) {
-            _updateCharactersInRangeReplacementString(owner.formattedText, range.location, range.length, replacementString);
+        if (owner) {
+            if (owner.maxLength > -1 && owner.maxLength <= textView.text.length && replacementString.length > range.length) {
```
PanayotCankov: What will happen if I have 3 characters selection and paste 5 characters and the new string is 1 character longer than the maxLength? It seems you've designed it so characters can always be deleted, that's great, but can it be something more verbose as:
```
const delta = replacement.length - range.length;
const isAddingChars = delta > 0;
const maxLength = owner.maxLength > -1 ? owner.maxLength : Number.positiveInfinity
const newLength = textView.text.length + delta;
if (isAddingChars &&  newLength > maxLength) {
  return false;
}
```
maxLength's default value may serve better here if it were Number.POSITIVE_INFINITY

tns-core-modules/ui/text-view/text-view.ios.ts
```diff
@@ -268,6 +275,13 @@ export class TextView extends EditableTextBase implements TextViewDefinition {
         let left = layout.toDeviceIndependentPixels(this.effectivePaddingLeft + this.effectiveBorderLeftWidth);
         this.nativeView.textContainerInset = { top: inset.top, left: left, bottom: inset.bottom, right: inset.right };
     }
+
+    [maxLengthProperty.getDefault](): number {
```
EddyVerbruggen: Thanks for the explanation! I was indeed wondering whether or not those were meant to be optional.

tns-core-modules/ui/text-view/text-view.ios.ts
```diff
@@ -63,6 +69,7 @@ export class TextView extends EditableTextBase implements TextViewDefinition {
     private _ios: UITextView;
     private _delegate: UITextViewDelegateImpl;
     private _isShowingHint: boolean;
+    private _maxLength: number = -1;
```
EddyVerbruggen: Let me try that!

tns-core-modules/ui/text-view/text-view.ios.ts
EddyVerbruggen: Will look into it, thanks a lot!

EddyVerbruggen:
@PanayotCankov I've implemented the changes you requested and added them to this PR.

I understand the new get/set structure much better now, thanks for taking the time to explain them as part of your review remarks.

Note that there is a slight difference between the iOS and Android implementations: if you paste something that's longer than allowed, Android will cut off the superfluous characters and inserts the characters that fit (this aligns with browser implementations). iOS just doesn't allow any of those characters. I don't think that should stop us from merging this PR though.

tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
```diff
@@ -410,4 +410,8 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
             this.nativeView.setHintTextColor(value);
         }
     }
+
+    [maxLengthProperty.setNative](value: number) {
+        this.nativeView.setFilters([new android.text.InputFilter.LengthFilter(+value)]);
```
PanayotCankov: The filter has to be cleared if the value is < 0. If the maxLength is no longer required the default -1 may be set by a developer. Also if the native view will be recycled (some fascinating new stuff @hshristov is working on) the maxLength will be reset to -1 calling this method in order to clean up the nativeView.

PanayotCankov:
We will have to add some unit tests before we merge this PR.

tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
EddyVerbruggen: Makes sense!

EddyVerbruggen:
I'd love to help adding tests but the problem for me was not being able to run the current set in the first place. I've looked at the docs on how to do it but to no avail. What are your steps to execute all (or the relevant) tests?

PanayotCankov:
@EddyVerbruggen 
```
git clone https://github.com/NativeScript/NativeScript.git
npm i
npm run setup
cd tests
npm link tns-core-modules-widgets
rm -rf node_modules/tns-core-modules/node_modules/tns-core-modules-widgets
tns run its
```
Where the `link tns-core-modules-widgets` are the widgets built in the `tns-core-modules-widgets` repo and can be pretty much installed using `npm i <filepath-to-tgz>` too.

Further maxLength will require automation tests since the tests will have to try to enter longer texts from the keyboard. For unit tests I can't figure API meaningful for tests besides setting maxLength when the text is bound with binding perhaps?! And setting maxLength back and forth to ensure the native setter will also reset the filter.

Our schedule is quite tight at the moment, perhaps later we will be able to spend some resources on this PR.

tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
EddyVerbruggen: Updated the PR to make sure current filters are retained on Android, and if a current length filter was found, replace that with the new filter.

EddyVerbruggen:
I've tried coming up with a working test for iOS but failed because the delegate code doesn't seem to be triggered. On Android the tests won't run at all.

I've extensively verified all changes with a test app, but I must ask others to add tests for this PR. I understand that may take some time of course.

jlooper:
@EddyVerbruggen This PR qualified for special treatment! Congratulations on your new badge!

contrib

PanayotCankov:
There are some conflicts, I've merged your changes on top of the current master and plan to merge the feature: #4190
Thanks!

PR #3927, opened 2017-04-03
Apply page css when using nav-entry with create method

vakrilov:

### DIFF
--- a/tns-core-modules/ui/frame/frame-common.ts
+++ b/tns-core-modules/ui/frame/frame-common.ts
@@ -83,14 +83,16 @@ export function resolvePageFromEntry(entry: NavigationEntry): Page {
         if (!page) {
             throw new Error("Failed to create Page with entry.create() function.");
         }
+
+        page._refreshCss();
     }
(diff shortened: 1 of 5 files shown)


tns-core-modules/ui/frame/frame-common.ts
```diff
@@ -83,14 +83,16 @@ export function resolvePageFromEntry(entry: NavigationEntry): Page {
...
     }
     else if (entry.moduleName) {
         // Current app full path.
         let currentAppPath = knownFolders.currentApp().path;
         //Full path of the module = current app full path + module name.
         let moduleNamePath = path.join(currentAppPath, entry.moduleName);
-        console.log("frame module path: " + moduleNamePath);
-        console.log("frame module module: " + entry.moduleName);
+        traceWrite("frame module path: " + moduleNamePath, traceCategories.Navigation);
+        traceWrite("frame module module: " + entry.moduleName, traceCategories.Navigation);
```
hshristov: Probably `frame module name`?

PR #3931, opened 2017-04-03
Fix iOS animations, transition was workin in DIP, when css class is change objects were flickering

PanayotCankov:


vchimev:
run ci

PR #3940, opened 2017-04-04
WebView NavigationType converted to type

vakrilov:


tns-core-modules/ui/web-view/web-view-common.ts
```diff
@@ -8,19 +8,12 @@ export * from "../core/view";
 
 export const srcProperty = new Property<WebViewBase, string>({ name: "src" });
 
+export type NavigationType = "linkClicked" | "formSubmitted" | "backForward" | "reload" | "formResubmitted" | "other" | undefined;
```
hshristov: Import it from definition file so that it is define only once

hshristov:
run ci

PR #3954, opened 2017-04-06
Fix animation's "Reset"

PanayotCankov:
Animations were always played as keyframes and kept overriding the local values. Android setAlpha has deprecated int overload, now it is guided to use the float one. Android background color was animated using css values instead of local or the keyframe one.


vakrilov:
AWESOMEAWESOMEAWESOME

PR #3980, opened 2017-04-12
Web view query params

### LINKED ISSUE #3872: Can't load a local HTML file with querystring via WebView.src (WebView.url works but is removed)
As noted [here](https://github.com/NativeScript/NativeScript/pull/3843#issuecomment-289091626) I'd like to request to not remove the `WebView.url` property as it's the only way to currently load a local HTML file with query parameters.

Passing `~/file.html` to `WebView.src` works, but `~/file.html?foo=bar` loads a blank page because it can't find the local file. `WebView.url` works perfectly for that case.

`WebView.src` can be altered to make it load the file successfully, but still I see no way in the current implementation to pass querystring params to the file. I really need a way to access those params in the loaded HTML file.

vakrilov:
Respect query params when setting `src` to `WebView`.

Added test page in `ui-test-app` -> `webview -> query`
It covers the following test cases:
1. `rel` - set `src` to a relative url (`"~/..."`)
2. `abs` - set `src` to a absolute url (`"/..."`)
3. `file` - set `src` to a file-prefixed url (`"file:///..."`)
4. `str` - set `src` to an html string - no query params passed in this case

CC: @SvetoslavTsenov @vchimev 

Resolves #3872.

### DIFF
--- a/apps/app/ui-tests-app/web-view/main-page.ts
+++ b/apps/app/ui-tests-app/web-view/main-page.ts
@@ -13,6 +13,7 @@ export function pageLoaded(args: EventData) {
 
     examples.set("webview", "web-view/web-view");
     examples.set("webtest", "web-view/web-view-test");
+    examples.set("query", "web-view/query-params");
 
     let viewModel = new SubMainPageViewModel(wrapLayout, examples);
     page.bindingContext = viewModel;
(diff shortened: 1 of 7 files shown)


EddyVerbruggen:
Thanks guys!

vchimev:
Add test: https://github.com/NativeScript/functional-tests/commit/7de59b7f0cad650957a255887ea35be78c0c217c#diff-6643555e18cf2449a28e16bf787c2785R30

PR #4056, opened 2017-04-24
Revert the translateX/Y to be in dip, Length prefer number to object type

PanayotCankov:
- Lengths will now prefer to result in dip number rather than { unit: "dip", value: <dip number> } to make general computations with the stored dip easier.
 - translateX/translateY will be no longer in Length, now it is in dip
 - Aliased number with px, dip and percent types to indicate the expected units in several APIs.


vakrilov:
run ci

PanayotCankov:
run-ci

PR #4002, opened 2017-04-16
Add sibling combinators for CSS rules

SBats:
Hi !
After asking for the opportunity to do a PR in this issue (https://github.com/NativeScript/NativeScript/issues/3727), I've worked a bit on it and, event if it's not complete, I would love to have a first feedback on the direction I'm taking.

I've implemented direct sibling combinator support, so the `+` combinator in web css.
To do so, I've modified the `createSelector` method and created a custom selector for handling this specific type of combination.

- Do you think it's a good approach?
- In this case, are extensive tests page in the `app` app enough, or do you want other kinds of automated unit tests on the methods?

Thank you a lot in advance for your help on this subject!

PR #4093, opened 2017-05-02
[In Progress] Add CSS positional selectors

SBats:
Hi !
As the PR https://github.com/NativeScript/NativeScript/pull/4002 add `getNodeIndex` method, I'm trying to add some of the positional selectors available in css.

Right now I'm targeting `:first-child`, `:last-child`, and `:nth-child`
I've got `:first-child` and `:nth-child` working basically and needs opinion on the way I implemented it (for example, wasn't sure if creating a new selector type was really necessary as it's a subtype of pseudo selectors)

But I'm not sure on how to implement `last-child` because of one thing.
Nodes are tested against selectors when they're added to the view, meaning technically at this moment they are always the last child of their parent. I'm guessing I should change the `trakChanges` method to take into account if the parent childCount change, but not sure about this approach. If you could confirm this or set me on the right track, I would very much appreciate it.

I've based this PR from the one with siblings selectors so it shouldn't be merged in master while the other one is not also merged. Not sure if you'd prefer me to target the other PR in that case. Please let me know what's better for your workflow.

Thank you a lot in advance for your help on this subject!

PS: I've also spread the regex for parsing selectors into multiple primitive ones, grouped into a Singleton which exports only one instance of the final dynamic regex (better performances wise).
I found it easier to maintain and read like this, but it can be reverted to a single primitive regex if you wish.


tns-core-modules/ui/styling/css-selector/css-selector.ts
```diff
@@ -194,11 +221,47 @@ export class AttributeSelector extends SimpleSelector {
 
 @SelectorProperties(Specificity.PseudoClass, Rarity.PseudoClass, Match.Dynamic)
 export class PseudoClassSelector extends SimpleSelector {
-    constructor(public cssPseudoClass: string) { super(); }
+    constructor(public cssPseudoClass: string, public value: string) { super(); }
     public toString(): string { return `:${this.cssPseudoClass}${wrap(this.combinator)}`; }
-    public match(node: Node): boolean { return node.cssPseudoClasses && node.cssPseudoClasses.has(this.cssPseudoClass); }
+    public match(node: Node): boolean {
+        if (this.value) {
```
PanayotCankov: In the CSS3 spec there is a reference to "The UI element states pseudo-classes" and "Structural pseudo-classes", I guess you can make two separate classes, rename PseudoClassSelector to ElementStatePseudoClassSelector and introduce a new base class StructuralPseudoClassSelector. Optionally these two can have a **new** common base class PseudoCLassSelector to keep the implementation that can be shared. The StructuralPseudoClassSelector can have multiple smaller derived classes that have their "match" implemented, so the branching (if-s) here won't happen at runtime, but rather while the CSS is parsed and our selector tree is generated. Multiple if-s are generally avoided by proper class chain and polymorphism.

tns-core-modules/ui/styling/css-selector/css-selector.ts
```diff
@@ -194,11 +221,47 @@ export class AttributeSelector extends SimpleSelector {
...
+        }
+        if (this.cssPseudoClass === "first-child") return isFirstChild(node);
+        if (this.cssPseudoClass === "last-child") return isLastChild(node);
+        return node.cssPseudoClasses && node.cssPseudoClasses.has(this.cssPseudoClass);
+    }
     public mayMatch(node: Node): boolean { return true; }
     public trackChanges(node: Node, map: ChangeAccumulator): void { map.addPseudoClass(node, this.cssPseudoClass); }
+
+    private _nthChildMatch(node: Node, eq: string): boolean {
+        if (eq.toLowerCase() === "even") eq = "2n";
```
PanayotCankov: This is a parser, it does not belong here.

tns-core-modules/ui/styling/css-selector-parser.ts
```diff
@@ -48,11 +48,84 @@ export function isAttribute(sel: SimpleSelector): sel is AttributeSelector {
     return sel.type === "[]";
 }
 
-var regex = /(\s*)(?:(\*)|(#|\.|:|\b)([_-\w][_-\w\d]*)|\[\s*([_-\w][_-\w\d]*)\s*(?:(=|\^=|\$=|\*=|\~=|\|=)\s*(?:([_-\w][_-\w\d]*)|"((?:[^\\"]|\\(?:"|n|r|f|\\|0-9a-f))*)"|'((?:[^\\']|\\(?:'|n|r|f|\\|0-9a-f))*)')\s*)?\])(?:\s*(\+|~|>|\s))?/g;
-// no lead ws     univ   type pref and ident          [    prop                   =                            ident    -or-    "string escapes \" \00aaff"    -or-   'string    escapes \' urf-8: \00aaff'       ]        combinator
+// Keep it here for simple test with tools
```
PanayotCankov: Legendary stuff :+1:, I just hope the original meaning of the RegEx is preserved and we had enough unit tests to catch potential breaking changes.

tns-core-modules/ui/styling/css-selector/css-selector.ts
PanayotCankov: Sad to say "do as we say, not as we do", If I have time I will refactor the AttributeSelector here:
https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/styling/css-selector/css-selector.ts#L142
And move the parsing logic in the css parser instead of the constructor of the AttributeSelector. Do not use this as reference as its code already smells a little.

PanayotCankov:
Overall it looks good. I think you are on the right track.

Structural changes at the moment reapply CSS to the added/removed child and all its descendants. So if you remove a child and add it on a different position all the selector matching logic will be re-triggered and it will be fine.

When adding or removing siblings though, the parent will have to notify children that are interested in siblings. It is not only the count. I am not sure what will happen if the CSS applied to a view depends on its previous sibling, and a new view is inserted before that view. The change map will have to carry flags that indicate whether changing the children of a view should notify the current view (the one for which the change map is build) that its siblings changed. You can introduce a "siblingsChanged" event (similar to countChanged), that may also be used in next sibling selectors.

As a side note if these selectors are not used they won't add overhead, but if they are, they may trigger a lot of change notifications while the visual tree is build. In the future we will have to find a way to defer CSS updates while building the visual tree, or add a CSS update cycle similar to the layout measure/layout cycle.

PanayotCankov:
P.S. Sorry for the delayed response I was on a trip :)

tns-core-modules/ui/styling/css-selector/css-selector.ts
SBats: That's some really good points! I'll go in that direction to continue this feature.
Thanks for the info about css3 spec, I didn't know that (should really look this spec in detail)

tns-core-modules/ui/styling/css-selector/css-selector.ts
SBats: Sure, make sense. Don't worry as a software evolve you can't keep everything at the right place at any time. If what I do can help refactoring some parsers, that's for the better!

tns-core-modules/ui/styling/css-selector-parser.ts
SBats: Thanks! I'm going to recheck it a bit more extensively. The source property of the final regex is useful to test and compare with the old one.

SBats:
Thanks for this feedback. It's really insightful! I am indeed a bit afraid of performances impacts with those selectors. Maybe it could be interesting to create a kind of stress test.
I'll keep going with your advice and let you know when I have something ready and more robust.

And no worries for the delay. I do it on my free time, there's no rush on my side and we all need to take time for ourself! I'm already really glad you take that much time to make extensive feedback!

SBats:
Sorry for the notification made by the bot. I'm trying to put my head back into this but had little time recently. I'll ping you if it goes further.
From what I've seen right now, the regex part did broke something so I need to fix this before anything.

PanayotCankov:
We may move to: https://github.com/NativeScript/css-parser
If it is properly done, there would be a stack of tokeniser and CSS parser, and for selectors and values we will get input tokens from the tokeniser that will be suitable for selectors parsing (e. g. things like (), [] blocks will be handled there, values parsing) I will close the PR for now.

SBats:
Okay. Thanks for the follow-up. I've been way over my head recently to be able to complete this work. Still have the code somewhere, so I might give it a shot again someday. Nice thing to have the parser separated! Good luck!

PR #4095, opened 2017-05-03
ImageView clear bitmap only when src is null/undefined

hshristov:
This fix image view losing bitmap when ImageSource was set.


hshristov:
run ci

PR #4167, opened 2017-05-12
Improvements after profiling the android flexbox example

PanayotCankov:


tns-core-modules/ui/core/bindable/bindable.ts
```diff
@@ -26,25 +26,32 @@ const paramsRegex = /\[\s*(['"])*(\w*)\1\s*\]/;
...
-        for (let i = 0, resultLength = result.length; i < resultLength; i++) {
-            if (result[i] === "parentsMatch") {
-                result[i] = parentsMatches[parentsMatchesCounter++];
-            }
-        }
+    if (!property) {
+        return emptyArray;
+    }
+
+    let result: Array<string> = propertiesCache[property];
```
PanayotCankov: Getting perent properties for binding expressions appear in the profiler. There won't be many bindings in an app so it should be save to cache all the result of the string parsing here. Reusing the cache seems to improve a little the bindings used in item templates.

tns-core-modules/ui/image/image.android.ts
```diff
@@ -9,6 +9,8 @@ export * from "./image-common";
 const FILE_PREFIX = "file:///";
 const ASYNC = "async";
 
+let AndroidImageView: typeof org.nativescript.widgets.ImageView;
```
PanayotCankov: Accessing android namespaces seems somewhat slow, simply caching the namespaces improves the performance but to allow android snapshot we can't cache them here in the module, instead they are cached in the createNativeView.

tns-core-modules/ui/layouts/flexbox-layout/flexbox-layout.android.ts
```diff
@@ -127,29 +134,32 @@ export class FlexboxLayout extends FlexboxLayoutBase {
         super._updateNativeLayoutParams(child);
 
         const lp = <org.nativescript.widgets.FlexboxLayout.LayoutParams>child.nativeView.getLayoutParams();
-        lp.order = child.order;
-        lp.flexGrow = child.flexGrow;
-        lp.flexShrink = child.flexShrink;
-        lp.wrapBefore = child.flexWrapBefore;
-        lp.alignSelf = alignSelfMap[child.alignSelf];
+        const style = child.style;
+        lp.order = style.order;
```
PanayotCankov: The shorthand properties each accesses this.style. internally, getting the values from style directly seems somewhat faster since this method is executed multiple times. We should check in future if we can provide a single point where the lp can be set just once.

Issue #4151, opened 2017-05-09
[TNS 3.0] iOS:Layout will not update on modal page


kssfilo:
### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.0.0
- Cross-platform modules: 3.0.0
- Runtime(s): 3.0.0
- Plugin(s): -

### Please tell us how to recreate the issue in as much detail as possible.

1. create default project with iOS platform(tns create..)
2. make a modal page like below and open it by tap event on 1st page

```
<StackLayout>
 <Label id=‘label1’ text=‘short’ textWrap=‘true’ />
 <Button button=‘label’ text=‘tap’ />
 <Label id=‘label2’ text=‘’ textWrap=‘true’ />
</StackLayout>
```

3. in modal pages tap handler, modify label1 or label2 text property to long text(like ‘foooooooooooooooooooo……’)

-> text will be not wrapped(label1) and text will not be shown (label2)

Besides this issue, I found that another UI has similar problem in modal page. such as Html or Image. when expanding height of these UI parts  programmatically, no layout change happens.

When text or src or html property is bound to model. The behavior is same.This problem is very critical when bounded properties initial value is null. (Nothing shown when setting value.)

I checked same test on non modal page and seems ok. so the problem looks like modal page specific issue.

tsonevn:
Hi @kssfilo,
If I understand you correctly after you opened a Modal page, you update text value for the Labels with a long string and the new text is not rendered properly. 
I tested this case on my side, however, it seems to work as expected. Please review the attached file and verify, whether I am missing something.
[Archive.zip](https://github.com/NativeScript/NativeScript/files/989107/Archive.zip)

kssfilo:
You have to write updating text code into Tap handler not loaded(). like this.

```
export function loaded(args:EventData){
}

export function onTap(args:EventData){
    var page:Page =<Page> args.object.parent;
     var label:Label = <Label> page.getViewById("label1");
     label.text="fooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo

     var label2:Label = <Label> page.getViewById("label2");
[...]

PR #4170, opened 2017-05-12
Fixes ios modal layout issue #4151

kssfilo:
it needs to be considered for merging.
=
Before 3.0.0, Page class calls setNeedsLayout() by it self. 

After commit:6b61335675587343f053b6f6607eae9e3f7907f1. Page class belongs parent Frame class.  Frame class also calls setNeedsLayout()  on this requestLayout(). 

But in condition a Page is Modal Page , Parent Frame class can't get "nativeView.window' , so the frame class fails to call setNeedsLayout().  And child Page class doesn't call setNeedsLayout() if they have parent property(before 3.0.0, modal page doesn't have parent)

I tried to get window by UIApplication.sharedApplication().delegate.window or UIApplica.....windows.first  in Frame class, but setNeedsLayout() doesn't update modal controls.

I am not expert for ios native application. so I don't know about this behavior.

This PR is just workaround to let Page class calls setNeedsLayout() when it is Modal like before 3.0.0.

I checked this by sample application for issue #4151 and some of my application and seems working fine.

### DIFF
--- a/tns-core-modules/ui/page/page.ios.ts
+++ b/tns-core-modules/ui/page/page.ios.ts
@@ -343,7 +343,7 @@ export class Page extends PageBase {
 
     public requestLayout(): void {
         super.requestLayout();
-        if (!this.parent && this.ios && this.nativeView) {
+        if ((!this.parent || this._modalParent) && this.ios && this.nativeView) {
             this.nativeView.setNeedsLayout();
         }
     }


SvetoslavTsenov:
run uitests

SvetoslavTsenov:
run uitests ios

hshristov:
run uitests ios

SvetoslavTsenov:
run uitests ios

PR #4171, opened 2017-05-12
[do not merge] Performance improvements after profiling the iOS flexbox example

PanayotCankov:
The native property updates are not suspended while the view is initialised and the views end up consuming too much time creating backgrounds for iOS. Background color is created 4 times when border width is set, 4 times when corner radius is set and 2 more times for border and background color setters. The PRs _nativeViewCreated flags is intended to suspend updates for a while (from the View constructor to the time createNativeView is called) but the implementation is weak as some Views set nativeView using means other than createNativeView/resetNativeView pair (e.g. TabView assigns TabViewItems nativeView, probably segmented bar does too).


SvetoslavTsenov:
run uitests

PanayotCankov:
Closing this in favour of #4227

PR #4172, opened 2017-05-12
Profiling module

vakrilov:
Added profiling module with some profiling methods.

### DIFF
--- a/tns-core-modules/profiling/profiling.ts
+++ b/tns-core-modules/profiling/profiling.ts
@@ -49,9 +49,8 @@ export function start(name: string): void {
         return;
     }
 
-    let info: TimerInfo;
-    if (timers.has(name)) {
-        info = timers.get(name);
+    let info = timers.get(name);
+    if (info) {
         if (info.isRunning) {
             throw new Error(`Timer already running: ${name}`);
         }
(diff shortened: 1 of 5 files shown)


tns-core-modules/profiling/profiling.ts
```diff
@@ -0,0 +1,176 @@
...
+        info.totalTime += info.lastTime;
+        info.count++;
+        info.currentStart = 0;
+        info.isRunning = false;
+    }
+
+    return info;
+}
+
+export function profile(name?: string): (target, key, descriptor) => void {
```
PanayotCankov:
```
export function profile(name?: string): MethodDecorator {
}
```
The returned function when profiling is enabled doesn't return void but TypedPropertyDecorator.

tns-core-modules/profiling/profiling.ts
```diff
@@ -0,0 +1,176 @@
...
+    return (target, key, descriptor) => {
+        if (!ENABLED) {
```
PanayotCankov: Can we move the if outside?
```
export const profile: (name?: string) => MethodDecorator = ENABLED ? ... no if ... : () => () => {};
```

tns-core-modules/profiling/profiling.ts
```diff
@@ -0,0 +1,176 @@
...
+            name = key;
+        }
+
+        profileNames.push(name);
+
+        //editing the descriptor/value parameter
+        descriptor.value = function () {
+            start(name);
+
+            var result = originalMethod.apply(this, arguments);
```
vakrilov: Will add in another PR

Issue #4135, opened 2017-05-07
TextField: if the secure property is true, the text property doesn't work properly.


Abdumama:
### Tell us about the problem
If you use the property **_secure_** in the widget **TextField** the property **_text_** doesn't work properly.

### Which platform(s) does your issue occur on?
Android

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.0.0
- Cross-platform modules: 3.0.0

### Please tell us how to recreate the issue in as much detail as possible.
`<TextField hint="Test" secure=true text="test"></TextField>`
The **TextField** value is empty, it should be displayed ****

`<TextField hint="Test" secure=false text="test"></TextField>`
The **TextField** value shows **test**

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.
Here a simple project to reproduce the issue.

[secureTextFieldBug.zip](https://github.com/NativeScript/NativeScript/files/981495/secureTextFieldBug.zip)

**EDIT:** 
The problem occurs only if you are using **Angular**.

tsonevn:
Thank you for reporting this issue.
I tested it on my side and confirm this is a real bug while setting the secure property for TextField.
I also found that this problem is reproduced on non-Angular project. For you convenience, I am attaching a sample project.
[Archive.zip](https://github.com/NativeScript/NativeScript/files/985931/Archive.zip)

Something that I suggest as a workaround is to setup `secure` property for the field on component loadedEvent, which will display the value as expected. For example:

login.component.html
```
<TextField hint="Test" (loaded)="textfieldloaded($event)" text="test"></TextField>
```

login,component.ts
```
textfieldloaded(args){
        var textF:TextField = <TextField>args.object;
        textF.secure=true;
     
    }
```

IJRS:
I have the same problem, any solution?

PR #4192, opened 2017-05-15
Fix setting isSecure, text and hint leaves TextField with visible hint instead of password chars

hshristov:
fix https://github.com/NativeScript/NativeScript/issues/4135


hshristov:
run ci

PR #4202, opened 2017-05-16
App "displayed" event

PanayotCankov:
This event is launched at the latest possible moment I've found that can be used to track app start time. This event takes about 300ms in iOS and about 1 sec. more in Android than the time to "launched".


PanayotCankov:
run-ci

PR #4215, opened 2017-05-17
fix(livesync): attach __onLiveSyncCore to global

sis0k0:
That method needs to be exposed because it's used in NativeScript Angular (https://github.com/NativeScript/nativescript-angular/blob/master/nativescript-angular/platform-common.ts#L91).


vakrilov:
run ci

PR #3999, opened 2017-04-15
Remove query params from target filename on getFile (fixes #3957)

vjoao:
### The commit message references a specific issue in this repo
Fixes #3957 

Updated tests to better reflect the getFile with a query string;

Issue #3957, opened 2017-04-07
http.getFile(url) should strip query string from saved file name


PeterStaev:
I'm trying to download files that have some long query string parameters and I'm using `http.getFile` w/o specifying a local file name. The default file name that is picked by the framework includes both the file name and the query string and thus the save fails. 

Expected:
The default file name that is generated should strip out any query string from the URL and just save with a file name that matches the url.

tsonevn:
Hi @PeterStaev,
Could you provide some more info about your case and the need of such a functionality?
It would also help if you could provide some info about the use-cases, when this will improve the functionality of the `getFile` method.

In that point, if you want to separate some parts of the URL you could simply use JavaScript `split` method.
For example:
```
var test = "https://httpbin.org/image?sample=test";
    var splstring = test.split("?");
    console.log(splstring[0]);
```

PeterStaev:
In the specific case I'm using this is i'm downloading files from AWS. Those files are in the following format:
`https://d3a8fsy4z5uwny.cloudfront.net/Thumbnails/1122a4b1-e9a2-463c-838e-c6d8dfc89f35.jpg?Expires=1491627540&Signature=kZ83lzc669Xg8jq-xMYPGx7....&Key-Pair-Id=APKAI7LA3QA3MJYCSYFQ`

Since I really don't need a specific file locally as I just need it downloaded locally so I can email the file, i decided to use the call `http.getFile(url)` (w/o specifying a file path). This tries to save the file as `....../1122a4b1-e9a2-463c-838e-c6d8dfc89f35.jpg?Expires=1491627540&Signature=kZ83lzc669Xg8jq-xMYPGx7....&Key-Pair-Id=APKAI7LA3QA3MJYCSYFQ` and this fails. 

I've workaround this by specifing some temp local file name for the save, but I think the method that accepts only URL should not fail to save from URL that has a query string.

tsonevn:
Hi @PeterStaev,
Thank you for the detailed information.
Indeed this is a valid scenario when you would like to downloaded file from URL with query string.
We will research, what are the possible ways to implement this functionality for the `getFile` method.

Regarding the topic, you could always create [pull request](https://github.com/NativeScript/NativeScript#contributing), which extends the needed functionality for the method.

SvetoslavTsenov:
The fix is already available in tns-core-modules@next and will be included in tns-core-modules@3.1.0

PR #4291, opened 2017-05-30
Support http.getFile for URLs with query params

vakrilov:
Rebased and test fixed version of #3999
Courtesy of @vjoao 
Related to #3957


vjoao:
Great, thank you!

Issue #3354, opened 2016-12-24
iOS: TextView hint becomes text


eleddie:
I have a text view which has a hint, when I start typing, the hint disappears (as it should) and I type like normal text. But when I send the message (is a chat app) and set the `ngModel` var back to `""` again, the hint now becomes editable text and I start typing in gray color like hint.

Here's the problem:

![tv-err](https://cloud.githubusercontent.com/assets/13477629/21467817/aeae6ccc-c9d0-11e6-8930-f569f6cb092f.gif)

NickIliev:
Hi @eleddie 

I've tried to reproduce your case but to no avail! I've tested it with [this sample](https://github.com/NativeScript/nativescript-sdk-examples-ng/tree/master/app/ui-category/text-view/basic-text-view) (with some modifications) but still was not able to reproduce so perhaps you can share with us a code snippet or a sample project which can recreate this scenario.

eleddie:
I would like this issue to be reopened since it isn't fixed yet. I was o a vacation trip and I couldn't reply. I'm going to create a sample project today so you could try it.
Thanks

Daxito:
I also used to have this issue, I fixed it by removing the hint!!!

NickIliev:
@eleddie can you post a snippet or a link to sample project reproducing the described case?

eleddie:
Hello @NickIliev, sorry for the delay, I've kinda busy.

This is my html:
`<TextView col="0" hint="Type a message..." [(ngModel)]="typing" returnKeyType="done" class="background-write"></TextView>
<Button col="1" text="&#xE163;" class="mi" (tap)="sendMessage(typing)"></Button>

and ts:
`
 private sendMessage(message: string) {
        if (message == "") return;
        this.typing = "";
        var newMessage = new Message(message.trim());
        this.messages.push(newMessage);
}
`

As @Daxito said, I can't just remove the hint...

NickIliev:
@eleddie I was able to reproduce your case and marked is as a bug. We will investigate further and post additional info when we have a solution.

Code to reproduce this issue [here](https://github.com/NickIliev/NativeScript-Issues-2017/tree/master/NativeScript/issue_3354_ng)

Meanwhile as a workaround, you can reset your hint after the initial text value is send. This way you will still have initial hint.
Example:

```
import { Component } from "@angular/core";

@Component({
    selector: "my-app",
    templateUrl: "app.component.html",
})

export class AppComponent {
    public typing: string = "";
[...]

PR #4292, opened 2017-05-30
Fix: IOS text view hint should never show while editing

vakrilov:
Related to #3354 
UI test added as this involves keyboard focus. Steps to reproduce in ui-tests-app:
1. Go to `issues->3354-ios`
2. Type some text in the text-view
3. Click the `clear` button. 

**Expected result:** The focus remains inside the text-view, but the text is cleared.
**Result before the fix:** The focus remains inside the text-view, however the hint text is shown instead.

cc: @SvetoslavTsenov @vchimev

### DIFF
--- a/tns-core-modules/ui/text-view/text-view.ios.ts
+++ b/tns-core-modules/ui/text-view/text-view.ios.ts
@@ -71,7 +71,7 @@ export class TextView extends EditableTextBase implements TextViewDefinition {
     private _ios: UITextView;
     private _delegate: UITextViewDelegateImpl;
     private _isShowingHint: boolean;
-    public _isEditing: boolean = false;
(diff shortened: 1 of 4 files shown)


tns-core-modules/ui/text-view/text-view.ios.ts
```diff
@@ -63,6 +71,7 @@ export class TextView extends EditableTextBase implements TextViewDefinition {
     private _ios: UITextView;
     private _delegate: UITextViewDelegateImpl;
     private _isShowingHint: boolean;
+    public _isEditing: boolean = false;
```
hshristov: It is possible to leave it without value (it will be falsy by default)

PR #4351, opened 2017-06-09
Chrome devtools elements tab support for Android

vakrilov:
Enable chrome devtools elements tab for android.
Supported features:
- Devtools browser is updated when views are added/deleted and upon navigation
- Devtools browser is updated on property changes
- Can see CSS computed styles
- Can update/remove attributes form devtools
- Can remove elements from devtools

Related to NativeScript/android-runtime#745 and NativeScript/android-runtime#746
Pair-codeded with @Pip3r4o

### DIFF
--- a/apps/app/package.json
+++ b/apps/app/package.json
@@ -1,4 +1,4 @@
 {
 	"name": "tns-samples-apps",
-	"main": "ui-test-app/app.js"
+	"main": "ui-tests-app/app.js"
 }
--- a/apps/app/package.json
+++ b/apps/app/package.json
@@ -1,4 +1,4 @@
 {
 	"name": "tns-samples-apps",
-	"main": "ui-tests-app/app.js"
+	"main": "devtools-app/app.js"
 }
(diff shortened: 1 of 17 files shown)


vakrilov:
run ci

apps/app/package.json
```diff
@@ -1,4 +1,4 @@
 {
 	"name": "tns-samples-apps",
-	"main": "ui-tests-app/app.js"
+	"main": "devtools-app/app.js"
```
PanayotCankov: undo this

tns-core-modules/ui/core/view-base/view-base.ts
```diff
@@ -259,7 +261,17 @@ export abstract class ViewBase extends Observable implements ViewBaseDefinition
         return null;
     }
 
-    // Overriden so we don't raise `poropertyChange`
+    get domNode(): DOMNode {
```
vakrilov: Make this a field for performance reasons

PR #4359, opened 2017-06-12
fix: add dom-node declarations

sis0k0:
The d.ts is needed because `dom-node` is imported in `text-base.d.ts`


vchimev:
`tns-core-modules/debugger/dom-node.d.ts[22, 13]: Unused method: 'toObject'`

PR #4361, opened 2017-06-12
fix: add css-agent declarations

vchimev:
The d.ts is needed because `css-agent` is imported in `dom-node.d.ts`.


vakrilov:
This file exports only interfaces - I think it's OK to leave **only** the `d.ts` file (remove the `css-agent.ts`)

Issue #4349, opened 2017-06-09
Feature: support accessibility escape gesture for going back or dismissing modal


m-abs:
To offer better support for accessibility for our iOS users, we need to be able to dismiss a modal or go back with the two-finger-Z gesture.

From https://developer.apple.com/library/content/featuredarticles/ViewControllerPGforiPhoneOS/SupportingAccessibility.html#//apple_ref/doc/uid/TP40007457-CH12-SW5

> Escape. A two-finger Z-shaped gesture that dismisses a modal dialog, or goes back one level in a navigation hierarchy.

To implement this, we need to override the native function "accessibiliityPerformEscape()".
This can be done on `UINavigationController` for going back one level in the history. I've still to figure out how to it for modals.

Adding these lines to the `UINavigationControllerImpl` in **ui/frame/frame.ios.ts** implements go-back:
```typescript
    public accessibiliityPerformEscape() {
      if (this.owner.canGoBack()) {
        this.owner.goBack();

        return true;
      }

      return false;
    }
```
It is only called if VioceOver is enabled.
If we can go back, goBack() is called and the function returns true, otherwise false.

Dismissing modals still needs to be worked out, but I'll be making a PR for this soon.

m-abs:
I've started my work on this branch: https://github.com/m-abs/NativeScript/tree/feature/4349

m-abs:
I can't run the nativescript tests on iOS, both master and my own branch crashes with this:

```bash
1   0x102d9527c -[TNSRuntime executeModule:referredBy:]
2   0x1028fbe00 main
3   0x106c4e92d start
4   0x1
[native code]: JS ERROR Error: Could not find module './'. Computed path '/Users/msn/Library/Developer/CoreSimulator/Devices/.../data/Containers/Bundle/Application/.../tests.app/app'.
Jun 14 15:40:28 wsnota313 com.apple.CoreSimulator.SimDevice.....launchd_sim[75276] (UIKitApplication:org.nativescript.tests[0x9553][76725]): Service exited due to signal: Segmentation fault: 11
```

m-abs:
As explained in the PR, I'm closing this issue. https://github.com/NativeScript/NativeScript/pull/4382#issuecomment-312227388

PR #4382, opened 2017-06-14
#4349: support accessibility escape gesture for going back or dismissing modal

m-abs:
Implements #4349 

This adds the function **accessibiliityPerformEscape()** to UINavigationControllerImpl on iOS.

When VoiceOver is enabled on an iOS device, this function is called by the platform when the escape gesture (two finger Z) is performed.
If the frame can go back, frame.goBack() is called.
If the current page have an modal, that modal is closed.


m-abs:
Question:
Android triggers the `activityBackPressedEvent` which can be cancelled, when the back button is pressed, should this do something similar?

Maybe `accessibilityPerformEscapeEvent`?

SvetoslavTsenov:
run ci

SvetoslavTsenov:
run uitests

SvetoslavTsenov:
run uitests ios

SvetoslavTsenov:
Hey @m-abs .Thanks for the contribution!
We are going to run some e2e test on this PR and we will review it in a few days

hshristov:
@m-abs 
Why do you need to implement this method 'accessibilityPerformEscape'?
According to documentation it will do the default action if no responder in the chain handles it. So we don't have any view that handle this method it will fallback to dismiss modal page or going back.

And in my opinion it should be first close modal page if any and if not then `frame.goBack()`

m-abs:
Hi @hshristov,

I'm closing this PR, as it is a result of a misunderstanding and misidentification of a problem in our app.

For some reason, which I've yet to identify, the escape gesture doesn't trigger the default behavior in our {N}-2.5 + ng project..
A locally applied patch like this in 2.5.5, solved the problem for us.

I should of cause have tested unpatched {N} in a demo project before making this PR.

I did after your reply and it works as expected without patching, if I just create a project with `tns create hello --ng`.

PR #4387, opened 2017-06-15
Globals in snapshot are overriden by the V8 on start, so they have to be installed later

PanayotCankov:


tns-core-modules/globals/globals.d.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Register globals such as setTimeout, setInterval, console etc.
+ */
+export function install();
```
vakrilov: Consider moving this to `module.d.ts` with the rest of the `Global` object interface

PR #4426, opened 2017-06-20
chore(travis): do not run unit tests app in Travis

vchimev:
The {N} team's having a hard time maintaining the unit tests app execution in Travis CI.
The main problem is that PROFILING, HTTP, WEB-VIEW tests fail randomly due to timeouts, caused by the unstable environment.
The unit tests app execution remains by default on PRs in {N} Jenkins CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,52 +4,16 @@ env:
   - PACKAGE_VERSION=$DATE-$TRAVIS_BUILD_NUMBER
   - PACKAGE_NAME=tns-core-modules
   - NODE_VERSION=6.9.4
-  - EMULATOR_API_VER=21
-  - RUNTIMEVERSION=next
-  - AVD_NAME=Arm$EMULATOR_API_VER
-addons:
-  artifacts:
-    paths:
-    - "$HOME/test-run-results$PACKAGE_VERSION.xml"
-sudo: false
-language: android
+language: node_js
 node_js:
-- 6.9.4
-jdk:
-- oraclejdk8
-android:
-  components:
-  - platform-tools
-  - tools
-  - build-tools-23.0.3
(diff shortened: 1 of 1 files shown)


NathanaelA:
Rather than disable the entire set of unit tests in Travis, why not just disable those three tests in Travis so the rest of the unit testing continues on the Android emulator?

vchimev:
run ci

vchimev:
Thanks for the comment, @NathanaelA! Yes, this is an option.

Let me step into more details why we decide to stop executing unit tests in Travis than my initial description.

The [Travis CI](https://travis-ci.org/NativeScript/NativeScript/branches) build for the [NativeScript/NativeScript](https://github.com/NativeScript/NativeScript) repository executes the unit test application only for Android, however, in our Jenkins CI we execute the unit test application for both Android and iOS.
- As seen, there is a duplication of what we execute (for Android) and we think it will be better to stick to Jenkins as it is our primarily CI system and as a single source of truth.
- This duplication causes us double time and efforts to maintain similar builds in two different CIs. We want to use these time and efforts for tasks bringing more value to the framework.
- The Travis CI build (only Android) takes ~25 (in the best case) mins and the Jenkins CI build (both Android and iOS) takes ~10 min. Basically, it does not bring any value to wait 25 min for something we already know in 10 min.

I would like to mention once again that Jenkins CI build triggers by default on each pull request and we trigger it explicitly for external contribution. Also that it is expected the tests to be executed prior any pull request following our [development workflow](https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#tests).

NathanaelA:
Awesome, thanks for filling me in!    Yeah in this case you are correct; no reason to maintain something that gives you no benefits and just eats developer time.

Issue #4342, opened 2017-06-08
Issue: Unable to change the fontWeight for both iOS and Android.


tsonevn:
When you try to setup fontWeight, the text style will no be changed. This problem is reproducible on both iOS and Android. Only `bold` font weight will be applied. Attaching sample project, where the issue could be debugged.

### Which platform(s) does your issue occur on?
Both

### Please provide the following version numbers that your issue occurs with:
- CLI: 2.5.5 and 3.0.3
- Cross-platform modules: 2.5.2, 3.0.1
[Archive.zip](https://github.com/NativeScript/NativeScript/files/1061310/Archive.zip)

t,1113363

ishitatsuyuki:
@vakrilov can you provide some progress on this? Has the root cause been determined?

vakrilov:
Hi, we are currently testing a fix proposed by #4436

vchimev:
A compatible fix with *2.5* version has been released as **tns-core-modules@2.5.3**.
For *3.\** version, the fix is already available in **tns-core-modules@next** and is going to be released officially in *3.1.1* or *3.2.0*.

PR #4436, opened 2017-06-21
Fonts refactored

vakrilov:
Reworked how fonts are retrieved in IOS. Related to #4342
I have also created [ns-fonts-tests application](https://github.com/vakrilov/ns-fonts-tests) to test the fix.


SvetoslavTsenov:
It seem that a text in a button will always be bold
expected:
transform_decoration_color_05_expected

actual
transform_decoration_color_05_actual

vakrilov:
run uitests

Issue #4464, opened 2017-06-28
Enable overlap for custom transitions


danielgek:
When using a custom transition for android [(example)](https://gist.github.com/danielgek/d3d7812bd3ed9793efadabdebbec6c4e), [these lines](https://github.com/NativeScript/NativeScript/blob/f7c9bd5d52106353f203901a2c6c7b4fa9d996a4/tns-core-modules/ui/transition/transition.android.ts#L158-L163) that make overlap possible for both fragments are never called.

Possible fix:

Adding the same logic before this [line](https://github.com/NativeScript/NativeScript/blob/f7c9bd5d52106353f203901a2c6c7b4fa9d996a4/tns-core-modules/ui/transition/transition.android.ts#L270) if the sdk version is greater than 21

i can summit a pr for this if it's ok with the team

### Which platform(s) does your issue occur on?
- Android
- CLI: 3.0.4
- Cross-platform modules: 3.0.1
- Runtime(s): 3.0.1

NickIliev:
@danielgek we are always happy to receive PRs from our community members! Please take a look at our [contributing guide ](https://www.nativescript.org/contribute)and let me know if you need any assistance on that matter.

hshristov:
Hi @danielgek,

I'm in the process of rewriting most of `transition.android.ts` so I will add your change in my PR. Otherwise it will be hard for me to rebase over your changes. Are you ok with this plan?

danielgek:
@hshristov yes, no problem at all ;)

hshristov:
@danielgek I forgot to mention that while these APIs are called for lollipop transitions (e.g. API21+) they should not affect custom animation like the one you are doing (e.g. extending our Transition class produce android.animation.Animator instance, not android.transition.Transition)

danielgek:
@hshristov i see, but there is way of enabling overlap using android.animation.Animator instance ? can you include that on your rework ?

hshristov:
@danielgek In my opinion they are overlapped by default.

danielgek:
Hi @hshristov sorry for the late response, only now i find a little time to read the android documentation more carefully and yes you are right, my bad!

PR #4470, opened 2017-06-28
Enable overlap for custom transitions on android, see #4464

danielgek:
This fixes the problem described on #4464, i don't really know how to write tests for this, but i can write them with help from the team i guess


vakrilov:
Hey @danielgek 
It looks like this PR will not be applicable after @hshristov's changes described in #4464, so I'm closing it. Let's continue the discussion in the issue

PR #4527, opened 2017-07-07
disable recycling on specific button

hshristov:
add more thorough test for view recycling
fix memory leak with android ActionBar
improve padding reset when view is recycled
improve reset of several controls


NathanaelA:
@hshristov - One question I have been meaning to ask; is there a way to disable the reuse of components easily (i.e. like if there isn't a resetNativeView function?)    I did a custom component that must be destroyed and re-initialized on each use; so I want to make sure there is a way that when all this recycling lands; there is an easy way to disable it per-component type...

hshristov:
@NathanaelA - Yes, there is `recycleNativeView` property on `ViewBase`.
But due to the native of this feature we don't enable it but default for all controls. As a plugin author you can choose to enable it if you know that this control will work correctly. But by default it is disabled.

hshristov:
uitests

hshristov:
animation

hshristov:
run animations

hshristov:
animations

hshristov:
run ci

hshristov:
untests

hshristov:
@NathanaelA Yes, there is a property `recycleNativeView: boolean` on `ViewBase` type.
But due to the nature of the feature we don't enable it for all controls. Every plugin author should test extensively and then enable it if it makes sense (meaning if it gives performance improvements).

In some synthetic tests `Label` doesn't bring much improvements when its nativeView is reused. But `Button` yields around 40% less time when nativeView is reused.

SvetoslavTsenov:
run animations

SvetoslavTsenov:
animations

SvetoslavTsenov:
uitests

SvetoslavTsenov:
uitests

SvetoslavTsenov:
run uitests android

SvetoslavTsenov:
run ci

SvetoslavTsenov:
run ci

Issue #1664, opened 2016-02-28
Feature request: css line-height


lawrencetaur:
I'd like the css **line-height** property implemented in {N}.
### Why:

The current line height seems to be 100% and in a lot of websites i use, the line-height for text is 1.5 to 1.8. 

I think this is how you do it on
Android
http://stackoverflow.com/questions/6863974/android-textview-padding-between-lines

and iOS
http://stackoverflow.com/questions/5494498/how-to-control-the-line-spacing-in-uilabel 

That seems a bit complicated. So why there are many css properties in {N} i feel 'line-height' would be a great addition 

Example:
This is how my app looks in Ionic 
![screenshot_20160228-174944](https://cloud.githubusercontent.com/assets/818425/13379155/3f5bfca4-de44-11e5-9b3a-0e57d1e5cbb0.png)

and in Nativescript
![screenshot_20160228-174930](https://cloud.githubusercontent.com/assets/818425/13379152/3b954f94-de44-11e5-8eab-e16e09a11bec.png)

vjoao:
If you can't wait for this to be implemented I suggest you to do it by calling the native APIs (`setLineSpacing` on Android or follow [this thread](http://stackoverflow.com/questions/5494498/how-to-control-the-line-spacing-in-uilabel) for iOS)

burkeholland:
:+1:

AntonioCuevaUrraco:
For people that end up here before it is implemented, this is the manual way to do it.

```ts
function labelLineHeight(nsLabel) {

	if(isIOS){
		var label= nsLabel.ios;
		var attributedString;

		if(label.atributedText){
			attributedString = label.atributedText;
		}
		else{
			attributedString=NSMutableAttributedString.alloc().initWithString(label.text);
		}
		var paragraphStyle = NSMutableParagraphStyle.alloc().init();
		paragraphStyle.lineSpacing = 5;
		var range= NSMakeRange(0, label.text.length);
		attributedString.addAttributeValueRange(NSParagraphStyleAttributeName, paragraphStyle, range);
		label.attributedText = attributedString;
	}
	if(isAndroid){
		var label = nsLabel.android;
		//Default spacing is 20% of text size
		//setLineSpacing(add,multiplyby);
		label.setLineSpacing(12, 1);
	}   
}

mikebranstein:
@AntonioCuevaUrraco I'm using this approach on several static pages, and it works well. 

[...]

PR #4534, opened 2017-07-09
Add CSS support for line-height

EddyVerbruggen:
Implements #1664

Btw, added a density multiplier for Android so spacing is better aligned with iOS, as shown in this screenshot:

screen shot 2017-07-09 at 21 34 23

### DIFF
--- a/tns-core-modules/ui/text-base/text-base-common.ts
+++ b/tns-core-modules/ui/text-base/text-base-common.ts
@@ -55,6 +55,13 @@ export abstract class TextBaseCommon extends View implements TextBaseDefinition 
         this.style.letterSpacing = value;
     }
 
+    get lineHeight(): number {
+        return this.style.lineHeight;
+    }
+    set lineHeight(value: number) {
+        this.style.lineHeight = value;
+    }
+
     get textAlignment(): TextAlignment {
         return this.style.textAlignment;
     }
@@ -203,3 +210,6 @@ textDecorationProperty.register(Style);
 
 export const letterSpacingProperty = new CssProperty<Style, number>({ name: "letterSpacing", cssName: "letter-spacing", defaultValue: 0, affectsLayout: isIOS, valueConverter: v => parseFloat(v) });
 letterSpacingProperty.register(Style);
+
+export const lineHeightProperty = new CssProperty<Style, number>({ name: "lineHeight", cssName: "line-height", defaultValue: 0, affectsLayout: isIOS, valueConverter: v => parseFloat(v) });
+lineHeightProperty.register(Style);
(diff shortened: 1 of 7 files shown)


vakrilov:
run ci

vakrilov:
run ci

tns-core-modules/ui/text-base/text-base-common.ts
```diff
@@ -202,4 +209,7 @@ export const textDecorationProperty = new CssProperty<Style, TextDecoration>({ n
 textDecorationProperty.register(Style);
 
 export const letterSpacingProperty = new CssProperty<Style, number>({ name: "letterSpacing", cssName: "letter-spacing", defaultValue: 0, affectsLayout: isIOS, valueConverter: v => parseFloat(v) });
-letterSpacingProperty.register(Style);
\ No newline at end of file
+letterSpacingProperty.register(Style);
+
+export const lineHeightProperty = new CssProperty<Style, number>({ name: "lineHeight", cssName: "line-height", defaultValue: 0, affectsLayout: isIOS, valueConverter: v => parseFloat(v) });
```
vakrilov: Consider removing the default value and implementing the `[lineHeightProperty.getDefault]()` callbacks instead. This assure you can reset the `lineHeight` property and also can sapre you some of the `style.lineHeight !== 0` checks.

vakrilov:
Hey @EddyVerbruggen 
Thanks for the PR it is looking great - I have proposed only one change related to getting the default value of the property.

tns-core-modules/ui/text-base/text-base-common.ts
EddyVerbruggen: I've tried that approach and you're right, it works fine. I was wondering though: if I remove the  `lineHeight !== 0` checks the code inside those branches will also run when the lineheight is the default, so it will always create the `NSMutableParagraphStyle`, even when not strictly required.

vakrilov: Makes sense - you can leave the check or even do `if(this.style.lineHeight) { ... }` if you remove the default value

EddyVerbruggen: Please check again, I've removed the default and added your `if (this.style.lineHeight)` suggestion. Behaves the same, so I'm happy ;)

vakrilov:
run ci

vakrilov:
run uitests

vakrilov:
run uitests

vchimev:
uitests ios

vakrilov:
Thanks for the great PR @EddyVerbruggen

airandfingers:
Thanks for adding this!

Note: this isn't line-height in the CSS sense; it's line *spacing*.

    [CSS line-height] - [font-size] = [line spacing or NativeScript line-height]

Would it make sense for this code to subtract fontSize from lineHeight before applying it as line spacing, or is fontSize not available in this context?

PR #4543, opened 2017-07-12
xhr: include cookies in the XHR response header

Dragomir-Ivanov:
Implements: Provides the app with  `set-cookie` and `set-cookie2` besides `Set-Cookie` and `Set-Cookie2`, so app can implement its own cookie jar. This is useful to retain session between app start/stop cycle. 

Tests: No unit tests, because there was no unit tests for XHR Cookies initially, and the patch is trivial.

### DIFF
--- a/tns-core-modules/xhr/xhr.ts
+++ b/tns-core-modules/xhr/xhr.ts
@@ -200,10 +200,7 @@ export class XMLHttpRequest {
         let result = "";
 
         for (let i in this._headers) {
-            // Cookie headers are excluded
-            if (i !== "set-cookie" && i !== "set-cookie2") {
-                result += i + ": " + this._headers[i] + "\r\n";
-            }
+            result += i + ": " + this._headers[i] + "\r\n";
         }
         return result.substr(0, result.length - 2);
     }


vakrilov:
Merged - thanks @Dragomir-Ivanov

bigradish:
Hi @Dragomir-Ivanov , could you tell me how to make the cookie.jar to maintain the session after app's exit? Thanks.

Dragomir-Ivanov:
Hi @bigradish 
You can use `nativescript-localstorage` to store your cookie across app restarts. Then set this cookie on each HTTP request. When server tells you that authentication failed for some reason, you delete this cookie from the localstorage.
However, I am strongly advising you not to use cookies for smart app. This is a browser thing and not suitable for REST API ( if you are doing this ). You can use OAuth, or even `Authorization` HTTP header that is not managed by the NativeScript underlying HTTP client.

bigradish:
Hi @Dragomir-Ivanov , Thank you so much for your detailed answers.
I use encrypted cookie to save session data on the client side to reduce the server's burden. For the details, you can refer to [RFC 6896](https://datatracker.ietf.org/doc/rfc6896/) . Maybe this feature can only be supported by cookie. Do you think so?

PR #4548, opened 2017-07-13
Disable reflection recycling tests for api lvls 17 and 19

vakrilov:
Disable reflection recycling tests for api levels <= 19. Too many methods throw exceptions for newly created widgets.


vakrilov:
run ci

PR #4586, opened 2017-07-23
fix(animations): check if target is present before removing its animation

sis0k0:
In NativeScript Angular, if the NativeScriptAnimationModule is imported in another
NgModule more than once (which shouldn't be done), the renderer will be instantiated twice. This
causes animations with empty targets to be created. If such animation is
removed, the app will crash. The additional check if the target is present
will prevent this.


sis0k0:
run ci

sis0k0:
uitests

vakrilov:
run ci

vakrilov:
run ci

vakrilov:
run ci

vakrilov:
run ci

PR #4624, opened 2017-07-31
Fix unitests on API <20

hshristov:
Some of these methods depends on the parent Layout. In the tests we are using `FlexboxLayout` as parent so items are positioned on different rectangle.


hshristov:
run ci

Issue #830, opened 2015-09-28
border-radius on Android 4.4.4 (Samsung Tab) not applied to <Button>


bradmartin:
The following markup does not apply a border-radius to the <Button> on Android 4.4.2. 

     .borderRads{
    border-radius: 30;
    }

         <Button text="SIGN UP" id="register" tap="register" 
              cssClass="button-accent-a200 margin-left-40 margin-right-40" row="17" verticalAlignment="bottom" />

I've ran the same .apk through appetize.io which I believe runs the app through Android 5.0 and the border-radius works correctly.

NathanaelA:
Uhm, you aren't using the borderRads class in your cssClass's.

bradmartin:
That was just an example. Excuse the typo. It doesn't get applied on this
device but does through appetizer io when I run it there so I know it's
working just not on the 4.4.2 samsung

enchev:
Just tried locally with our [test app](https://github.com/NativeScript/NativeScript/blob/master/apps/ui-tests-app/image-view/roundbtn.xml) and everything worked as expected.
![untitled](https://cloud.githubusercontent.com/assets/5804953/10129329/9782b9c4-65c5-11e5-91a3-fcdf02b02214.png)

x4080:
Maybe because of cache problem? Strange behavior did occur. I experienced button like before flat design. In a lollipop device :)

bradmartin:
It's device specific. The same code works fine in my emulator and when I run the .apk through appetize.io but locally on my Samsung Tab with 4.4.2 no border-radius is applied to the button. I haven't checked other elements.

tsvetan-ganev:
I also encountered the same issue - **border-radius** works on versions >= 4.4, but it does not work on devices (including emulators) running Android 4.2.2.

bradmartin:
I can confirm border-radius is not applied to any element on Samsung Tab (4.4.2) when using cssClass. I have not tested setting the styles in JS.

vchimev:
Hi @tsvetan-ganev,

I tested **border-radius** on emulator running Android 4.2.2 and it seemed to work as expected. Details:
```
    Name: Api17
    Path: /Users/vchimev/.android/avd/Api17.avd
  Target: Android 4.2.2 (API level 17)
 Tag/ABI: default/x86
    Skin: WVGA800
  Sdcard: 16M
```
[...]

PR #4625, opened 2017-07-31
Do not change the Android layer types (hardware/software)

PanayotCankov:
I don't know why we did this in the first place,
but changing the rendering mode can not be good.
Let's re-run all possible UI tests.

### DIFF
--- a/tns-core-modules/ui/animation/animation.android.ts
+++ b/tns-core-modules/ui/animation/animation.android.ts
@@ -8,7 +8,7 @@ import {
     translateXProperty, translateYProperty, scaleXProperty, scaleYProperty
 } from "../styling/style-properties";
 
-import { CacheLayerType, layout } from "../../utils/utils";
+import { layout } from "../../utils/utils";
 import lazy from "../../utils/lazy";
 
 export * from "./animation-common";
@@ -161,8 +161,6 @@ export class Animation extends AnimationBase {
             }
         }
 
-        this._enableHardwareAcceleration();
-
         if (traceEnabled()) {
             traceWrite("Starting " + this._nativeAnimatorsArray.length + " animations " + (this._playSequentially ? "sequentially." : "together."), traceCategories.Animation);
         }
@@ -190,7 +188,6 @@ export class Animation extends AnimationBase {
         }
 
         this._propertyUpdateCallbacks.forEach(v => v());
-        this._disableHardwareAcceleration();
         this._resolveAnimationFinishedPromise();
 
         if (this._target) {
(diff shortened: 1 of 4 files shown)


PanayotCankov:
run ci

NathanaelA:
The reason that we were using Hardware layers is for speed.  If you eliminate the hardware layers the app itself will be slower when drawing things.      However, the reason there is code to disable it is because some ops are not supported in hardware mode; so it is switched to software mode while those ops are going on.

Maybe you should read this: https://developer.android.com/guide/topics/graphics/hardware-accel.html 

I am firmly against this patch, unless you can show a really good reason why slowing down the android platform is justified.

PanayotCankov:
Ya red that,

A little while ago I removed some of the ops that were clipping images. Clipping with paths were not applying antialiasing on the images so they looked terrible anyway. Containers may not clip content when with rounded corners but that would be rarely a problem.

In general I think we should not mess internally with the rendering mode. "Hardware acceleration is enabled by default if your Target API level is >=14" and we will now never set it to software, so layers instead of forced to software every once in a while by {N}, should stay hardware all the time.

Now since we do not enable/disable hardware acceleration I think you can control it (e.g. on the windows) and it will be inherited by the views down the hierarchy. Or at least you can use the `<application android:hardwareAccelerated="true">` mentioned in the article you refer and the {N} won't mess with it.

Can you actually try the next versions of the widgets and the tns-core-modules in your app and verify that there is a performance hit?

hshristov:
run ci

SvetoslavTsenov:
run ci

SvetoslavTsenov:
run ci

SvetoslavTsenov:
run ci

SvetoslavTsenov:
run android

dtopuzov:
run ci

SvetoslavTsenov:
Hey, @PanayotCankov could try to rebase this pull request.

bradmartin:
At one time, I think Enchev and I went back and forth on fixing an Android issue and it required changing the layer type. I can't recall the specifics, it was probably in late 2015 when I was working on some initial plugins for android components but something didn't work and we started stumbling on the layer types. If I find it, I'll drop a link. This might be totally unrelated but just wanted to throw it out there 😄

bradmartin:
Ah so it was @vakrilov who helped, Enchev just cleaned up 👍 
https://github.com/NativeScript/NativeScript/issues/830#issuecomment-155374124

PanayotCankov:
@bradmartin I think before we relied on Canvas clipping to round the button, and clipping was not antialiased and was not working properly on some devices, unless software mode is set. I think I prefer to keep the defaults now since the geometry is drawn for the button background using different Canvas APIs. If the issue reoccur, I'd like to rather provide a property on the view to control the rendering mode.

SvetoslavTsenov:
uitests

SvetoslavTsenov:
animations

PR #4627, opened 2017-07-31
recycling now happens only if nativeView and android properties are...

hshristov:
…not accessed.

recycleNativeView filed now accepts: "always" | "never" | "auto". Always will recycle the nativeView no matter if its nativeView or android proprties are accessed. Never will disable recycling. Auto will recycle it only if nativeView and android properties are not accessed.

This is done because there are way too many places where nativeView is changed from code and we can't safely recycle these nativeViews.


hshristov:
run ci

hshristov:
run ci

hshristov:
run ci

PR #4628, opened 2017-07-31
Fix: refresh flat property on change in IOS

vakrilov:
Update UI when flat property changes in IOS.
Related to #2740


SvetoslavTsenov:
uitests

SvetoslavTsenov:
run ci

PR #4652, opened 2017-08-04
cache page on forward navigation

hshristov:
- Change the behavior when forward navigation is made. Before we destroy all nativeViews. Now we keep them alive, so navigating forward and backward will be faster. Also navigating back will be super fast and back pages will have their UI state preserved.
- Adds the option to wait until page transition ends (even for the default animation) so now navigation should be smoother.
- Major refactoring on navigation
- Major refactoring on page transitions/animations
- Fix module cycling reference
- Disabled nativeView recycling when native animation in running
- Improve the speed of unit-tests.


hshristov:
uitests

tns-core-modules/ui/frame/fragment.transitions.android.ts
```diff
@@ -0,0 +1,552 @@
...
+const popExitFakeResourceId = -40;
+
+const waitingQueue = new Set<android.app.Fragment>();
+
+let AnimationListener: android.animation.Animator.AnimatorListener;
+let loadAnimatorMethod: java.lang.reflect.Method;
+let reflectionDone: boolean;
+let defaultEnterAnimatorStatic: android.animation.Animator;
+let defaultExitAnimatorStatic: android.animation.Animator;
+let fragmentCompleted: any;
```
vakrilov: Add type - `ExpandedFragment`

tns-core-modules/ui/frame/fragment.transitions.android.ts
```diff
@@ -0,0 +1,552 @@
...
+}
+
+function transitionsCompleted(fragment: android.app.Fragment): boolean {
+    waitingQueue.delete(fragment);
+    return waitingQueue.size === 0;
+}
+
+function transitionOrAnimationCompleted(fragment: ExpandedFragment): void {
+    if (transitionsCompleted(fragment)) {
+        const entry: BackstackEntry = (<any>fragment)._callbacks.entry;
```
vakrilov: Can we get rid of the casts by extending the `ExpandedFragment` interface

tests/package.json
```diff
@@ -4,12 +4,12 @@
   "readme": "NativeScript Application",
   "repository": "<fill-your-repository-here>",
   "nativescript": {
-    "id": "org.nativescript.UnitTestApp",
+    "id": "org.nativescript.tests",
```
hshristov: Reverted

tns-core-modules/ui/core/view-base/view-base.ts
```diff
@@ -814,8 +814,10 @@ export abstract class ViewBase extends Observable implements ViewBaseDefinition
         if (nativeView && isAndroid) {
             const recycle = this.recycleNativeView;
             if (recycle === "always" || (recycle === "auto" && !this._disableNativeViewRecycling)) {
-                const nativeParent = isAndroid ? (<android.view.View>nativeView).getParent() : (<UIView>nativeView).superview;
-                if (!nativeParent) {
+                // const nativeParent = isAndroid ? (<android.view.View>nativeView).getParent() : (<UIView>nativeView).superview;
```
hshristov: There because I don't want to forget to do it for iOS

tns-core-modules/ui/frame/frame.android.ts
```diff
@@ -119,16 +55,65 @@ export class Frame extends FrameBase {
...
+        backstackEntry.fragmentTag = fragmentTag;
+        backstackEntry.navDepth = navDepth;
+
+        return newFragment;
+    }
+
+    public setCurrent(entry: BackstackEntry): void {
+        this.changeCurrentPage(entry);
+        this._currentEntry = entry;
+        this._isBack = true;
```
hshristov: `_isBack` should be always `true` unless you call `frame.navigate` so only there we make it false. All other navigations are considered `Back` :)

tns-core-modules/ui/frame/frame.android.ts
```diff
@@ -119,16 +55,65 @@ export class Frame extends FrameBase {
...
+        this._processNavigationQueue(entry.resolvedPage);
+    }
+
+    private changeCurrentPage(entry: BackstackEntry) {
+        const isBack = this._isBack;
+        let page: Page = this.currentPage;
+        if (page) {
+            if (page.frame === this) {
+                this._removeView(page);
```
hshristov: Added the ` _isBack === true`. Failed merge :(
`_isBack` is set to false in `navigateCore` so no point in passing it as parameter here.

tns-core-modules/ui/frame/fragment.transitions.android.ts
```diff
@@ -0,0 +1,552 @@
...
+const waitingQueue = new Set<android.app.Fragment>();
+
+let AnimationListener: android.animation.Animator.AnimatorListener;
+let loadAnimatorMethod: java.lang.reflect.Method;
+let reflectionDone: boolean;
+let defaultEnterAnimatorStatic: android.animation.Animator;
+let defaultExitAnimatorStatic: android.animation.Animator;
+let fragmentCompleted: any;
+
+export function _waitForAnimationEnd(newFragment, currentFragment): void {
```
hshristov: True but I couldn't find better name. Any suggestions?

tns-core-modules/ui/frame/fragment.transitions.android.ts
hshristov: changed with the latest commit.

hshristov:
uitests

hshristov:
uitests

PR #4606, opened 2017-07-26
Fixes TabView icons changes.

NathanaelA:
This Fixes/Implements #4138 .

This issue has two fixes; the first is that the item is now a couple levels deeper in the hierarchy; so we have to go up a couple parents to get to the actual TabView root controller in NS 3.    

The second issue; is even though we change the itemSource value, the item assigned to this tab view was never refreshed; so we call the function that recreates the tabview item with the new values.

Issue #4138, opened 2017-05-08
Cannot change iconSource TabViewItem


noetius:
### Did you verify this is a real problem by searching [Stack Overflow](http://stackoverflow.com/questions/tagged/nativescript) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?
Yes

### Tell us about the problem
In a TabView, when the user taps on a TabViewItem in order to see its content, I need to replace the unselected version of the TabViewItem icon with an highlighted version.
Before NativeScript 3.0.0, it was possible to bind a callback to TabView's selectedIndexChanged event, accessing TabViewItems by appropriate indexes, and replacing their iconSource without hassle.
But now happens this exception:

<pre>
An uncaught Exception occurred on "main" thread.
com.tns.NativeScriptException: 
Calling js method onPageSelected failed

TypeError: tabLayout.updateItemAt is not a function
"file:///data/data/org.nativescript.fleetbar/files/app/tns_modules/tns-core-modules/ui/tab-view/tab-view.js, line: 174, column: 22

StackTrace: 
	Frame: function:'TabViewItem._update', file:'file:///data/data/org.nativescript.fleetbar/files/app/tns_modules/tns-core-modules/ui/tab-view/tab-view.js', line: 174, column: 23
	Frame: function:'set', file:'file:///data/data/org.nativescript.fleetbar/files/app/tns_modules/tns-core-modules/ui/tab-view/tab-view-common.js', line: 66, column: 22
	Frame: function:'onSelectedIndexChanged', file:'file:///data/data/org.nativescript.fleetbar/files/app/views/home/home.js', line: 198, column: 31
	Frame: function:'Observable.notify', file:'file:///data/data/org.nativescript.fleetbar/files/app/tns_modules/tns-core-modules/data/observable/observable.js', line: 103, column: 23
	Frame: function:'valueChanged', file:'file:///data/data/org.nativescript.fleetbar/files/app/tns_modules/tns-core-modules/ui/tab-view/tab-view-common.js', line: 205, column: 16
	Frame: function:'CoercibleProperty._this.set', file:'file:///data/data/org.nativescript.fleetbar/files/app/tns_modules/tns-core-modules/ui/core/properties/properties.js', line: 208, column: 25
[...]

PR #4666, opened 2017-08-08
Fixes TabView icons changes.

vakrilov:
This is an extended version of #4606 by @NathanaelA 
 
Original description:
> This Fixes/Implements #4138 .
> 
> This issue has two fixes; the first is that the item is now a couple levels deeper in the hierarchy; so we have to go up a couple parents to get to the actual TabView root controller in NS 3.  
>  
> The second issue; is even though we change the itemSource value, the item assigned to this tab view was never refreshed; so we call the function that recreates the tabview item with the new values.


vakrilov:
run ci

vakrilov:
@SvetoslavTsenov -> I have added a test page in the `ui-tests-app`. It is inside `tab-view -> tab-view-icon-change`. It will change the icons of the tab view items when switching tabs.

PR #4687, opened 2017-08-10
fix: typo on android utils, getPalleteColor to getPaletteColor

### LINKED ISSUE #4647: utils.ad.getPalleteColor should be named getPaletteColor
The method name is a typo, and should probably be deprecated in favor of the correct spelling.

danielgek:
This pr Fixes #4647 .


danielgek:
@hshristov totally agree with letting the old method, but i don't know what you mean by "Also in the Obsolete decorator and API help specify the newer methods." can you point me an example ?

hshristov:
@danielgek We have `@Deprecated` TypeScript decorator but apparently you cannot pass arguments to it. So comment is also good enough :)

vakrilov:
@danielgek Thanks for the PR!

PR #4707, opened 2017-08-16
Fixed a bunch of typo's and minor errors in TypeScript definition files.

EddyVerbruggen:
This PR corrects (mainly) spelling errors / typo's in `.ts.d` files that caught my eye.

I was actually only going to correct this comment of the native element in `html-view.d.ts`:

https://github.com/NativeScript/NativeScript/blob/2701ea3c1e2b758674fbb344f74ffb39ec223288/tns-core-modules/ui/html-view/html-view.d.ts#L22

.. but then I got carried away. 😋 🙃

### DIFF
--- a/tns-core-modules/utils/utils.d.ts
+++ b/tns-core-modules/utils/utils.d.ts
@@ -86,9 +86,9 @@ export module layout {
 
     /**
      * Rounds value used in layout.
-     * @param value to round.
+     * @param px to round.
      */
-    export function round(value: px): px;
+    export function round(px: px): px;
 
     /**
      * Converts device pixels to device independent pixes and measure the nativeView.
--- a/tns-core-modules/utils/utils.d.ts
+++ b/tns-core-modules/utils/utils.d.ts
@@ -41,36 +41,43 @@ export module layout {
     export var UNSPECIFIED: number;
     export var EXACTLY: number;
     export var AT_MOST: number;
+
     /**
-     * Gets measure specification mode from a given specification as string.
+     * Gets layout mode from a given specification as string.
      * @param mode - The measure specification mode.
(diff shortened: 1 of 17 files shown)


tns-core-modules/utils/utils.d.ts
```diff
@@ -81,7 +88,7 @@ export module layout {
      * Rounds value used in layout.
      * @param value to round.
      */
-    export function round(px: px): px;
```
hshristov: I believe this was intentional

EddyVerbruggen: FYI I saw that the documentation `@param` name was different and the implementation of this definition also uses the param name `value` so that's why I changed it from `px` to `value`. The type and returntype are fine though.

https://github.com/NativeScript/NativeScript/blob/bba7a82bdfceb77c306b6dd699f494df68e334e0/tns-core-modules/utils/utils-common.ts#L92

hshristov: The name was intentionally specified as `px` so that the caller knows that this method expects device pixels and not `dip` device independent pixels. As for the implementation - it is so important (IMO) to match the definition file. The definition file is used for intellisense.

EddyVerbruggen: I understand - do you want me to update the PR?

hshristov: Yes please :)

EddyVerbruggen: No problemo :)

hshristov:
run ci

hshristov:
run ci

hshristov:
run ci

PR #4713, opened 2017-08-18
Enable evaluating in Chrome Console (#3405)

vakrilov:
This is cherrypick of the commits by #3405 which where lost in the 2.5 -> 3.0 migration.

@pkoleva Can you verify the Chrome console eval works with this fix?


pkoleva:
Related to https://github.com/NativeScript/ios-runtime/issues/787

bundyo:
This also fixes the previously non-working Watch expressions.

PR #4742, opened 2017-08-23
Fix crash where some android Drawables doesn't implement getConstantState

hshristov:
Changed all controls that were using getConstantState.


hshristov:
uitests

PR #4750, opened 2017-08-24
Fix breaking change - nativeView property should have setter

hshristov:
Exposed savedInstanceState on LaunchEventArgs so some controls that need Bundle to initialize could get it from there.


SvetoslavTsenov:
run ci

PR #4794, opened 2017-09-04
TypeScript 2.5 compatibility

vakrilov:

### DIFF
--- a/tns-core-modules/ui/styling/style-properties.ts
+++ b/tns-core-modules/ui/styling/style-properties.ts
@@ -1,5 +1,5 @@
 // Types
-import { 
+import {
     Transformation,
     TransformationValue,
     TransformFunctionsInfo,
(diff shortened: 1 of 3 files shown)


tns-core-modules/ui/styling/style-properties.ts
```diff
@@ -528,8 +528,8 @@ function parseTransformString(text: string): Transformation[] {
     return matches;
 }
 
-function normalizeTransformation({ property, value }: Transformation) { 
-    return STYLE_TRANSFORMATION_MAP[property](value);
+function normalizeTransformation({ property, value }: Transformation): Transformation {
+    return <any>STYLE_TRANSFORMATION_MAP[property](value);
```
sis0k0: why is that cast needed?

vakrilov: TypeScript cannot assure that all callbacks from `STYLE_TRANSFORMATION_MAP` actually return `Transformation`

PR #4832, opened 2017-09-12
iOS: setting borderRadius on Image with http/s source fails to render

### LINKED ISSUE #4804: Image with https src and borderRadius doesn't render on iOS
### Did you verify this is a real problem?

Yes, I verified that the issue occurs outside of my application, and created a minimal reproduction repo. I searched for similar issues on stack overflow and in the issue tracker and could not find any.

### Tell us about the problem

When applying `borderRadius` to a user's profile image that is fetched over `https://`, rather than `res://`, the image stops rendering entirely. Removing the `borderRadius` brings the image back.

img_3045

iOS

- CLI: 3.1.3
- Cross-platform modules: 3.1.1
- Runtime(s): tns-ios@3.1.0
- Plugin(s): 

1. Clone this repo: https://github.com/justindujardin/nativescript-image-border-radius-bug
2. `tns run ios` and run on either the simulator or a device.

The repo is just the default angular app (`--ng`), with the routing removed so as to be focused on reproducing the bug.

```html
<StackLayout>
    <Image src="https://cdn.rawgit.com/justindujardin/angular-rpg/gh-pages/assets/icon/icon.png"
           borderRadius="100" width="32"></Image>
</StackLayout>
```

or

```html
<GridLayout rows="auto">
    <Image src="https://cdn.rawgit.com/justindujardin/angular-rpg/gh-pages/assets/icon/icon.png"
           row="0" borderRadius="100" width="32"></Image>
</GridLayout>
```

justindujardin:
When loading an image from a remote resource, all the background properties are applied before the imageSource. When assigning borderRadius to an image, then later setting its imageSource would cause the image to be in a bad render state and completely disappear from the UI.

This patch clones the backgroundInternal property after async assignment of the imageSource, which causes the render state to be corrected.

Fixes #4804 

I'm not sure this is the very best fix and am open to suggestions.

### DIFF
--- a/tns-core-modules/ui/image/image-common.ts
+++ b/tns-core-modules/ui/image/image-common.ts
@@ -78,6 +78,8 @@ export abstract class ImageBase extends View implements ImageDefinition {
                 fromUrl(value).then((r) => {
                     if (this["_url"] === value) {
                         this.imageSource = r;
+                        // Clone the backgroundStyle to correct (now invalid?) background state.
+                        this.style.backgroundInternal = this.style.backgroundInternal.clone();
                         this.isLoading = false;
                     }
                 });
--- a/tns-core-modules/ui/styling/background-common.ts
+++ b/tns-core-modules/ui/styling/background-common.ts
@@ -27,7 +27,7 @@ export class Background implements BackgroundDefinition {
     public borderBottomRightRadius: number = 0;
     public clipPath: string;
 
-    private clone(): Background {
+    clone(): Background {
         const clone = new Background();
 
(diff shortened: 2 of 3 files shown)


justindujardin:
@ns-bot done

hshristov:
@justindujardin I can't reproduce this issue with tns-core-modules@3.2 but I've saw another issue - borderRadius was not applied on Image element. This PR https://github.com/NativeScript/NativeScript/pull/4963 fix the new issue so I'm closing your PR in flavor of the other.

Issue #4858, opened 2017-09-20
Android editable text base needs better defaults for controlling closing keyboard


evanfuture:
### Did you verify this is a real problem by searching [Stack Overflow]
Yes

### Tell us about the problem
Android editable text base needs better defaults for controlling closing keyboard

### Which platform(s) does your issue occur on?
Android

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.1.0
- Cross-platform modules: 3.1.0
- Runtime(s): tns-android 3.1.1

### Please tell us how to recreate the issue in as much detail as possible.
When a TextView uses the Send action, the keyboard will always close.  This is because there is nothing to handle this specific case in the onEditorAction method in https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts#L99

In that method, the default is to return `false`.  Based on the android documentation, we should return true if we have consumed the action, else false.

In other words, if we're done using the keyboard, return false, otherwise return true.

In this particular case then (pressing the Send button), we are forcing the keyboard to close instead of letting the developer handle this.

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.

To reproduce this, create a textfield with a returnKeyType of send...
`this._input.returnKeyType = 'send';`

And try something like focusing the input during its returnPress method.

tsonevn:
Hi @evanfuture ,
Thank you for writing us.

Indeed this is the valid scenario and would be useful to let the developer choose, whether the keyboard to be closed or not. 
Regarding that, I would suggest making your own pull request, where the needed changes in the code could be made that would allow the developer to handle the event. 

As a starting point, you could also review [our contribution guidelines.](https://www.nativescript.org/contribute)

evanfuture:
Hi @tsonevn, I made the PR just after (we had to patch it for our own project): https://github.com/NativeScript/NativeScript/pull/4859

PR #4859, opened 2017-09-20
TextField onEditorAction should return true if developer has consumed the action

evanfuture:
Based on this: https://developer.android.com/reference/android/widget/TextView.OnEditorActionListener.html
And assuming that the developer will run dismissSoftInput in the owner.dismissSoftInput() only if they need to.

Returning False will close the keyboard regardless.

To help the rest of the community review your change, please ensure:

Fixes/Implements #4858


dtopuzov:
run ci

PR #4909, opened 2017-09-30
Line-height property breaks text alignment (iOS)

### LINKED ISSUE #4906: Line-height property breaks text alignment (iOS)

Yes ✅ 

### Tell us about the problem
When using the attribute or CSS style for line-height combined with text-alignment, the line-height attribute will cause the contents to always be left-aligned. 

iOS

- CLI: 3.2.0-2017-08-18-9693
- Cross-platform modules: 3.1.1
- Runtime(s): 3.2.0

1. `tns create exampleApp --ng`
2. Create a Label, give it the following style
```
.example-lbl {
   line-height: 5;
   text-align: center;
}
```

Notice that the alignment does not apply. 

With line-height:

screen shot 2017-09-29 at 9 33 09 pm

Without line-height:

screen shot 2017-09-29 at 9 32 57 pm

EddyVerbruggen:
Fixes #4906

### DIFF
--- a/tns-core-modules/ui/text-base/text-base.ios.ts
+++ b/tns-core-modules/ui/text-base/text-base.ios.ts
@@ -129,6 +129,8 @@ export class TextBase extends TextBaseCommon {
         if (this.style.lineHeight) {
             const paragraphStyle = NSMutableParagraphStyle.alloc().init();
             paragraphStyle.lineSpacing = this.lineHeight;
+            // make sure a possible previously set text alignment setting is not lost when line height is specified
+            paragraphStyle.alignment = (<UITextField | UITextView | UILabel>this.nativeViewProtected).textAlignment;
             attrText.addAttributeValueRange(NSParagraphStyleAttributeName, paragraphStyle, { location: 0, length: attrText.length });
         }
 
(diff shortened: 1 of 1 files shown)


vakrilov:
run ci

EddyVerbruggen:
@vakrilov Any idea what's wrong with those unittests? (I can't access those servers)

vakrilov:
I think we had some connectivity issues on the CI server, so it might be something totally unrelated.. Will rerun them again.
run ci

vakrilov:
uitests

hshristov:
@EddyVerbruggen There are more values in the [NSTextAlignment](https://developer.apple.com/documentation/uikit/nstextalignment) enum than our [TextAlignment](https://github.com/NativeScript/NativeScript/blob/c26f1c916a313c6950aa34eef644d79b394a867e/tns-core-modules/ui/text-base/text-base.d.ts#L114).

Wouldn't it be safer if we apply only matching values?

EddyVerbruggen:
@hshristov That makes sense, but I wonder if that hasn't already been checked at this point by [this code](https://github.com/EddyVerbruggen/NativeScript/blob/a00753964a22e747a9d1ca11c70353bb65dee9b4/tns-core-modules/ui/text-base/text-base.ios.ts#L67-L81)?

hshristov:
@EddyVerbruggen Lol the whole time I thought that you take the value from the JS object ...
Apparently I didn't paid enough attention. Nice shot 👍

hshristov:
run ci

hshristov:
uitests

hshristov:
run ci

hshristov:
uitests

Issue #4916, opened 2017-10-04
Incorrect "top offset" when navigating to a Page that contains ListView or RadListView


VladimirAmiorkov:
### Did you verify this is a real problem by searching [Stack Overflow](http://stackoverflow.com/questions/tagged/nativescript) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?

Yes, I was not able to find anything related to it.

### Tell us about the problem
Please, ensure your title is less than 63 characters long and starts with a capital
letter.

When running the [tns-template-hello-world-ts](https://www.npmjs.com/package/tns-template-hello-world-ts) template on iOS 11.0, after navigation, when you go back the ListView component is vertically offset and immediately gets pushed back in place after the backwards navigation is completed. The same behavior is observed with the RadListView component from the [nativescript-pro-ui](https://www.npmjs.com/package/nativescript-pro-ui) plugin.

I was able to reproduce the same behavior in pure native iOS project and it looks to be coming from a change on how the [UITableView](https://developer.apple.com/documentation/uikit/uitableview) is behavior in backwards navigation on the latest iOS 11, the same is true for components backed on [UICollectionView](https://developer.apple.com/documentation/uikit/uicollectionview) like the RadListView element. Here is a screenshot, I have made the items background red to show the above white offset:
on navigation: ![simulator screen shot - iphone 7 - 2017-10-04 at 17 39 24](https://user-images.githubusercontent.com/4989411/31181954-474d3d84-a92b-11e7-8a78-bc14f3d9027c.png) after: ![simulator screen shot - iphone 7 - 2017-10-04 at 17 41 51](https://user-images.githubusercontent.com/4989411/31182002-6181df84-a92b-11e7-8e21-5af68efb93a4.png)

### Which platform(s) does your issue occur on?
iOS - only on 11, on 10 works correctly.

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.2.1
- Cross-platform modules: 3.1.1
- Runtime(s): 3.2.0
- Plugin(s): _nothing related_

### Please tell us how to recreate the issue in as much detail as possible.

- Execute `tns create my-app-name --template tns-template-hello-world-ts`, as of writing this the latest version is **3.2.0**
- Execute `cd my-app-name`
- Execute `tns run ios` (make sure that you are running agains iOS 11)

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.

All the code is in the template.

### Semi workaround
[...]

PR #4919, opened 2017-10-05
Fix for iOS 11 ListView top offset 

VladimirAmiorkov:
This PR addresses this issue https://github.com/NativeScript/NativeScript/issues/4916. In iOS 11 there is a new [contentInsetAdjustmentBehavior](https://developer.apple.com/documentation/uikit/uiscrollview/2902261-contentinsetadjustmentbehavior) behavior of the UIScrollView which is inherited by the ListView's UITableView. The default value of this new behavior is [automatic](https://developer.apple.com/documentation/uikit/uiscrollviewcontentinsetadjustmentbehavior/2902258-automatic) which adjusts the top offset as described by the API description: 

> Content is always adjusted vertically when the scroll view is the content view of a view controller that is currently displayed by a navigation or tab bar controller. If the scroll view is horizontally scrollable, the horizontal content offset is also adjusted when there are nonzero safe area insets. 

After this PR it may be a good feature to actually support the new different value of this new behavior, at the moment the animations provided by this behavior are not working inside the {N} layouts (GridLayout, StackLayout etc.)

### DIFF
--- a/tns-core-modules/ui/list-view/list-view.ios.ts
+++ b/tns-core-modules/ui/list-view/list-view.ios.ts
@@ -8,6 +8,7 @@ import { StackLayout } from "../layouts/stack-layout";
 import { ProxyViewContainer } from "../proxy-view-container";
 import { ios } from "../../utils/utils";
 import { profile } from "../../profiling";
+import { device } from "../../platform";
 
 export * from "./list-view-common";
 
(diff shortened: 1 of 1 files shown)


vchimev:
android

vchimev:
ci android

vchimev:
ci android

hshristov:
My PR will resolve ios 11 layout issues and will make this property working out-of-the box. So I think we 
shouldn't merge this one.

VladimirAmiorkov:
As noted in my PR comment, if we can take advantage of the new behavior without the current issues that would be great. Feel free to skip this PR if you have made the new behavior work out of the box.

PR #4977, opened 2017-10-24
fix: Require core modules used for inspector lazily

vakrilov:
This should fix crash when starting debugger in IOS


apps/app/ui-tests-app/app.ts
```diff
@@ -1,4 +1,6 @@
-﻿import * as application from "tns-core-modules/application";
+﻿console.log("####### ------ APP MODULES START ")
```
PanayotCankov: Remove? Or use something more console friendly.

PR #4987, opened 2017-10-26
Wrong definition of canEvaluatePolicyError

akiten-io:
IOS LAContext's canEvaluatePolicyError method always throw an error
 because of the wrong definition. (missing parameter)

this pr will fix the problem.


PanayotCankov:
The iOS runtime has been designed to throw JavaScript errors and handle the error pointer internally. The error second parameter is NSError** and in {N} it has to be passed as interop.Reference, then the value obtained from it. We'd like to promote using:
```
try {
  obj.canEvaluatePolicyError(policy);
} catch(e) {
 // Log the error?
}
```
So probably we will not be updating the .d.ts-es.

If you really want to fix this, you may consider contributing to https://github.com/NativeScript/ios-metadata-generator that is used to generate the .d.ts-es dynamically and the `error: Error` should be `error: interop.Reference`

PR #5008, opened 2017-10-30
refactor: set selectedIndex = 3 of list-picker

SvetoslavTsenov:


apps/app/ui-tests-app/list-picker/list-picker.ts
```diff
@@ -1,7 +1,14 @@
+import { ios } from "tns-core-modules/application/application";
```
vchimev: Is this used?

PR #5042, opened 2017-11-07
Add a simple polyfill for System.import so we can use some Angular services out-of-the-box

PanayotCankov:


SvetoslavTsenov:
Hey @PanayotCankov, this PR look good to be merged.

PR #5070, opened 2017-11-14
Fix crash on iOS where JS object reference was not hold.

hshristov:
The crash was affecting UITestsApp as well as Marketplace demo.


dtopuzov:
run ci

PR #5091, opened 2017-11-22
Application-Settings: Non-blocking changes

### LINKED ISSUE #4871: Application Settings blocks UI thread
* YES

### Tell us about the problem
Currently the `application-settings` module synchronize changes to disk every time a value is set. This could lead to main-thread blocking when setting large amounts of settings and/or large values. Especially at key moments, when 60fps is desired, like page transitions.

* Post iOS 7, one should **not** call `synchronize` after each change. ([source](https://stackoverflow.com/a/37680539))
  * > Because this method is automatically invoked at periodic intervals, use this method only if you cannot wait for the automatic synchronization.
   * However it might be wise to do so on `applicationDidEnterBackground`, see source for reason.

* On Android, one should **not** call `commit` after each change, but instead `apply`. ([source](https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#apply()))
   * Commit blocks until changes are written to disk. Apply does so asynchronously and has been available since API 9. It is suggested to use this, if you do not care about the success boolean from commit.

### Proposed Change
1. Do not `synchronize` on iOS and use `apply` on Android.
2. Add a `flush` call which synchronize/commit changes to disk, so that the user still has the option.
3. (Optional?) Add a call to `appSettings.flush()` on `application.suspendEvent`.

Side-note: Ideally I'd like to have something like [RN's AsyncStorage](https://facebook.github.io/react-native/docs/asyncstorage.html)

### Plugin
[@nota/nativescript-appsettings-async](https://github.com/ddfreiling/nativescript-appsettings-async)

To demo this I created a plugin with the changes proposed above. In the demo you can see the difference, which is quite big on Android. 4000ms /w commit vs. 100ms /w apply.

ddfreiling:
Fixes #4871

Changes to Application-Settings were blocking, meaning that changes would completely block the main (UI) thread until they were written to disk. This had some pretty nasty stuttering effects when saving large strings or stringified objects with Application-Settings.

## Android - SharedPreferences
On Android, calling `apply` instead of `commit`, saves changes to the in-memory cache, and starts an async save to disk. `commit` blocks and is only relevant when you actually need the return value indicating whether the disk write succeeded, which Nativescript does not care about. [source](https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#apply())

> Unlike commit(), which writes its preferences out to persistent storage synchronously, apply() commits its changes to the in-memory SharedPreferences immediately but starts an asynchronous commit to disk and you won't be notified of any failures. If another editor on this SharedPreferences does a regular commit() while a apply() is still outstanding, the commit() will block until all async commits are completed as well as the commit itself.

> You don't need to worry about Android component lifecycles and their interaction with apply() writing to disk. The framework makes sure in-flight disk writes from apply() complete before switching states.

## iOS - NSUserDefaults

On iOS, any changes affects the in-memory version of UserDefaults and explicitly calling `synchronize` writes the in-memory UserDefaults to disk before returning. According to Apple documentation this call is now unnecessary and should not be used. [source](https://developer.apple.com/documentation/foundation/userdefaults/1414005-synchronize)

> At runtime, you use UserDefaults objects to read the defaults that your app uses from a user’s defaults database. UserDefaults caches the information to avoid having to open the user’s defaults database each time you need a default value. When you set a default value, it’s changed synchronously within your process, and asynchronously to persistent storage and other processes.

## Explicit disk sync

I've added an explicit `flush` call which has the old behaviour of blocking the thread until changes has been sync'ed to disk, if for some reason it should be needed by anyone.

These changes should be covered by existing unit tests.

### DIFF
--- a/tests/app/application-settings/application-settings-tests.ts
+++ b/tests/app/application-settings/application-settings-tests.ts
@@ -102,6 +102,16 @@ export var testClear = function () {
     TKUnit.assert(!appSettings.hasKey(numberKey), "Failed to remove key: " + numberKey);
 };
 
+export var testFlush = function () {
+    appSettings.setString(stringKey, "String value");
+    // >> application-settings-flush
+    var flushed = appSettings.flush();
+    // will return boolean indicating whether flush to disk was successful
+    // << application-settings-flush
+    TKUnit.assert(flushed, "Flush failed: "+ flushed);
+    TKUnit.assert(appSettings.hasKey(stringKey), "There is no key: " + stringKey);
+};
+
 export var testInvalidKey = function () {
     try {
         appSettings.hasKey(undefined);
--- a/tns-core-modules/application-settings/application-settings.android.ts
+++ b/tns-core-modules/application-settings/application-settings.android.ts
@@ -1,7 +1,7 @@
 ﻿import * as common from "./application-settings-common";
 import { getNativeApplication } from "../application";
 
-var sharedPreferences;
+var sharedPreferences: android.content.ISharedPreferences;
 function ensureSharedPreferences() {
     if (!sharedPreferences) {
         sharedPreferences = (<android.app.Application>getNativeApplication()).getApplicationContext().getSharedPreferences("prefs.db", 0);
@@ -80,6 +80,6 @@ export function clear(): void {
     sharedPreferences.edit().clear().apply();
 }
 
-export var flush = function (): void {
-    sharedPreferences.edit().commit();
+export var flush = function (): boolean {
+    return sharedPreferences.edit().commit();
 }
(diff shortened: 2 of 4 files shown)


PanayotCankov:
It would be nice if this had some unit-tests.

ddfreiling:
@PanayotCankov I believe the tests in [tests/app/application-settings/application-settings-tests.ts](https://github.com/NativeScript/NativeScript/blob/master/tests/app/application-settings/application-settings-tests.ts) are still relevant and passing. I guess I could add a test for `flush`, but it has no actual changes and just blocks until disk sync is done.

ddfreiling:
@PanayotCankov updates with extra test

SvetoslavTsenov:
uitests

SvetoslavTsenov:
run ci

SvetoslavTsenov:
uitests

SvetoslavTsenov:
uitests

PR #5126, opened 2017-12-04
Add Enterkey-Processing for some IMEs to editable-text-base.android

### LINKED ISSUE #5121: Could not make New-Line in TextView when Using Japanese IME
Yes. looks like Issue #4858
onEditorAction (at editable-text-base.ts) return value
 
### Tell us about the problem
Using Japanese IME (or some other IME) , Could not make New-line  on TextView by new-line-key(on softkeyborad).

Android

- CLI: 3.3.1
- Cross-platform modules: 3.3.0
- Runtime(s): tns-android 3.3.1
- Plugin(s): nothing

The best approach would be to get your code running in the [NativeScript Playground](https://play.nativescript.org) and share the link with us, along with any additional details or steps to reproduce needed for examining the issue there. 

simply make TextView , Input some text with enterkey by Google日本語入力(Japanese Input?) or Google Korean Input (I can't read any korean letters,but enter key makes same action)

occures without any ts code.
```
<TextView></TextView>
```

and I Patched this problems(https://github.com/zrelyydereva/NativeScript/commit/a00f55e3bb48eb3c11d5f968ef8e3640f12eddcf).
but I'm wondering that it was the right way to fix this problems

zrelyydereva:
default or preinstalled IME is not raise onEditorAction, but Google's input method(e.g. Google Japanese Input,Google Korean Input.) raises this.
so we have to treat this as returning false .If not so, we cannot make new-lines.

Fixes #5121 

linted,compiled

### DIFF
--- a/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
+++ b/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
@@ -103,13 +103,10 @@ function initializeEditTextListeners(): void {
             }
 
             // https://github.com/NativeScript/NativeScript/issues/5121 
-            // default or preinstalled IME is not raise onEditorAction, but Google's input method(e.g. Google Japanese Input,Google Korean Input.) raises this.
+            // default or preinstalled IME is not raise onEditorAction, but Google's input method(e.g. Google Japanese Input,Google Korean Input.)
+            // and hardware keyboard raise this.
             // so we have to treat this as returning false .If not so, we cannot make new-lines.
             if (actionId === android.view.inputmethod.EditorInfo.IME_NULL && (event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
-                if (textView.getMaxLines() === 1) {
-                    owner.dismissSoftInput();
-                    return true;
-                }
                 owner._onReturnPress();
                 return false;
             }
--- a/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
+++ b/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
@@ -102,6 +102,17 @@ function initializeEditTextListeners(): void {
                 return;
             }
 
+            // https://github.com/NativeScript/NativeScript/issues/5121 
+            // default or preinstalled IME is not raise onEditorAction, but Google's input method(e.g. Google Japanese Input,Google Korean Input.) raises this.
+            // so we have to treat this as returning false .If not so, we cannot make new-lines.
+            if (actionId === 0 && (event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
+                if (textView.getMaxLines() === 1) {
+                    owner.dismissSoftInput();
+                }
+                owner._onReturnPress();
+                return false;
+            }
+            
             if (actionId === android.view.inputmethod.EditorInfo.IME_ACTION_DONE ||
                 actionId === android.view.inputmethod.EditorInfo.IME_ACTION_GO ||
                 actionId === android.view.inputmethod.EditorInfo.IME_ACTION_SEARCH ||
--- a/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
+++ b/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
@@ -105,9 +105,10 @@ function initializeEditTextListeners(): void {
             // https://github.com/NativeScript/NativeScript/issues/5121 
             // default or preinstalled IME is not raise onEditorAction, but Google's input method(e.g. Google Japanese Input,Google Korean Input.) raises this.
             // so we have to treat this as returning false .If not so, we cannot make new-lines.
-            if (actionId === 0 && (event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
+            if (actionId === android.view.inputmethod.EditorInfo.IME_NULL && (event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
                 if (textView.getMaxLines() === 1) {
                     owner.dismissSoftInput();
+                    return true;
                 }
                 owner._onReturnPress();
                 return false;


SvetoslavTsenov:
uitests

SvetoslavTsenov:
run ci

SvetoslavTsenov:
cuteness

SvetoslavTsenov:
testsappng

tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
```diff
@@ -102,6 +102,17 @@ function initializeEditTextListeners(): void {
...
 
+            // https://github.com/NativeScript/NativeScript/issues/5121 
+            // default or preinstalled IME is not raise onEditorAction, but Google's input method(e.g. Google Japanese Input,Google Korean Input.) raises this.
+            // so we have to treat this as returning false .If not so, we cannot make new-lines.
+            if (actionId === 0 && (event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
+                if (textView.getMaxLines() === 1) {
+                    owner.dismissSoftInput();
+                }
+                owner._onReturnPress();
+                return false;
```
hshristov: Why returning `false` here? This means that we didn't handled `KEYCODE_ENTER` and we leave the OS to handle it. Was this the intended behavior?

tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
```diff
@@ -102,6 +102,17 @@ function initializeEditTextListeners(): void {
                 return;
             }
 
+            // https://github.com/NativeScript/NativeScript/issues/5121 
+            // default or preinstalled IME is not raise onEditorAction, but Google's input method(e.g. Google Japanese Input,Google Korean Input.) raises this.
+            // so we have to treat this as returning false .If not so, we cannot make new-lines.
+            if (actionId === 0 && (event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
```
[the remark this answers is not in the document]
zrelyydereva: sorry for rough patching. I think too. I'll fix it

tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
zrelyydereva: I'm worried about _onReturnPress that already used  returnPressEvent of TextView or TextField.
behaviour looks like I thought. but at least , we've return false after dissmissSoftInput.
can I have your idea?

zrelyydereva: no sorry,return true after dismissSoftInput

zrelyydereva: I'm plan to fix patch as below
```
if (actionId === android.view.inputmethod.EditorInfo.IME_NULL && (event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
    if (textView.getMaxLines() === 1) {
        owner.dismissSoftInput();
        return true;
    }
    owner._onReturnPress();
    return false;
}
```
why returning false is to avoid running below without touching existen code ,and keeping current behavior when actionId !== IME_NULL && keycode === KEYCODE_ENTER.
And I think _onReturnPress should called as in case of actionId is IME_ACTION_NEXT or IME_ACTION_PREVIOUS.

if it ok, I'll fix this pull request.

zrelyydereva:
according from issue conversation, we've to treat hardware keyboard too.
So, this patch is not enogh.
(dismissSoftInput should not be called except TextField has no next control. really simple solution is omit calling dismissSoftInput of case in actionId === IME_NULL , because I think hardware keyboard users don't use and set visible software keyboards )
I'll make patch again.

zrelyydereva:
Could you please review it again?

hshristov:
@zrelyydereva I'm sorry for the late response. We are preparing 3.4 and we got delayed a bit.
I was thinking that this would be enough. Could you verify please because I don't have hardware keyboard (usb).
Here is the change:
```
if (actionId === android.view.inputmethod.EditorInfo.IME_NULL || 
    actionId === android.view.inputmethod.EditorInfo.IME_ACTION_DONE ||
    actionId === android.view.inputmethod.EditorInfo.IME_ACTION_GO ||
    actionId === android.view.inputmethod.EditorInfo.IME_ACTION_SEARCH ||
    actionId === android.view.inputmethod.EditorInfo.IME_ACTION_SEND ||
    (event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
                // If it is TextField, close the keyboard. If it is TextView, do not close it since the TextView is multiline
                // https://github.com/NativeScript/NativeScript/issues/3111
                if (textView.getMaxLines() === 1) {
                    owner.dismissSoftInput();
                    owner._onReturnPress();
                    return false;
                }
                owner._onReturnPress();
                return true;
            }
 // code removed for brevity 
}
```

zrelyydereva:
@hshristov Please don't worry , I'm looking forward to v3.4
But your changes do not work well , (could not make new-line when returning true)
and I think we should not call dismissSoftInput even if it is TextField , for next control.
as below scene
```
<TextView></TextView>
<TextField></TextField>
<TextView></TextView>
<TextView></TextView>
<TextField></TextField>
```
when showing soft-keyboard and pressing hard-enterkey on second element(TextField) ,caret should move to second element(TextView) and soft-keyboard should not be hidden (as like actionId === android.view.inputmethod.EditorInfo.IME_ACTION_NEXT)

manoldonev:
@zrelyydereva As far as I can tell @hshristov addressed the original issue with https://github.com/NativeScript/NativeScript/commit/8df8d754abd548f8c161c4326eb96f26c3a3604b#diff-01a10958d7fab58f4dd4216fe46638af. 

The fix will be released with the next version of the modules but you can `npm install tns-core-modules@rc --save` in a sample project to verify that everything works as expected with the release candidate so we can close this PR here.

PR #5134, opened 2017-12-05
refactor: set parentNode property to ViewBase

sis0k0:
Setter:
You can set a '_templateParent' by setting node.parentNode. This won't
affect the native parent of the node.
Getter:
Returns '_templateParent', if set, or the native parent otherwise.
Reason:
This will help us remove that abstraction from nativescript-angular as
it matches the DOM Nodes API name.

### DIFF
--- a/tns-core-modules/ui/core/view-base/view-base.ts
+++ b/tns-core-modules/ui/core/view-base/view-base.ts
@@ -149,6 +149,7 @@ export abstract class ViewBase extends Observable implements ViewBaseDefinition 
     private _style: Style;
     private _isLoaded: boolean;
     private _visualState: string;
+    private _templateParent: ViewBase;
     private __nativeView: any;
     // private _disableNativeViewRecycling: boolean;
     public domNode: dnm.DOMNode;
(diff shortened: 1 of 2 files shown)


PanayotCankov:
I like where this is going,
Instead of supporting Angular by implementing their services adapted for {N}, just browserify our API so that it works with stock Angular.

tns-core-modules/ui/core/view-base/view-base.ts
```diff
@@ -221,6 +222,14 @@ export abstract class ViewBase extends Observable implements ViewBaseDefinition
         this._style = new Style(this);
     }
 
+    get parentNode() {
+        return this._templateParent || this.parent;
+    }
+
+    set parentNode(node: ViewBase) {
+        this._templateParent = node;
```
hshristov: Why don't you set `this.parent` directly instead of introducing new field?

sis0k0:
ios-issues

sis0k0:
@hshristov, I thought that overriding the original parent will cause problems but I can't find problems when we do that. @vakrilov, do you know if there's a reason for keeping the original parent and not setting it directly (aka, using templateParent)?

vakrilov:
@sis0k0 There are elements in ng-renderer that are marked have metadata marked with `{ skipAddToDom: true }`. In this case we rely on the template to retain the original parent. Setting parent in this case might be wrong because the element will not be actually added to it's parent.

I see no harm in this PR and as @PanayotCankov it will even make the API more browser-like.

PR #4917, opened 2017-10-04
feat(animation): support animating width/height properties (WIP)

justindujardin:
Add support for animating width/height View properties.

Implements #1764

- [x] Supports animating PercentageLength properties
- [x] Works with Angular component animations
- [x] Works on Android
- [x] Works on iOS
- [x] Add example to examples app
  - [x] basic height animation
  - [x] height animation in stack layout
  - [x] animation curves tester
  - [x] details disclosure effect
- [x] Verify works as expected with "auto" height
- [x] Verify works with all built-in animation curves
- [ ] Verify keyframe animation support
- [x] Add unit tests
  - [x] PercentLength.parse tests to verify input/output expectations
  - [x] View.animate resolves percentage width/height relative to parent
  - [x] View.animate works with pixel value strings
  - [x] View.animate works with number dip inputs
  - [x] View.animate works with number dip inputs
  - [ ] Keyframe animation tests
  - [ ] Spring curve tests (iOS has separate codepaths for spring)

## Use with View.animate

```typescript
view.animate({
  width:'40%', 
  height: '400px',
  duration: 300
});
```

## Use with Angular

```typescript
@Component({
  moduleId: module.id,
  selector: 'cool-component',
  template: `
      <Button (tap)="details = !details" text="Toggle"></Button>
      <Image backgroundColor="blue" height="100%"
             [@sizingImage]="details == true ? 'details' : 'summary'"></Image>
  `,
  animations: [
    trigger('sizingImage', [
      state('details', style({height: '55%'})),
      state('summary', style({height: '100%'})),
      transition('summary => details', animate('200ms ease-out')),
      transition('details => summary', animate('200ms ease-in')),
    ]),
  ]
})
export class CoolComponent {
  details = true;
}
```

**Unfinished thoughts**

There are a few limitations that could be lifted, but need input:

- Layout values are not interpolated for Views on iOS. StackLayout height animations will "jump" a little. Would need to add some kind of periodic JS callback during the animation to do this.
- View originX/Y values are different on iOS/Android by default. This means that height/width animations will act differently by default if the user does not specify origins explicitly on the View being animated.

Issue #1764, opened 2016-03-15
Add the ability to animate height and width


tjvantoll:
I would like the ability to animate a UI element’s height and width. Here’s my use case: I have a container that I need to grow after a user interaction. If I use a scale animation on the container, everything in the container (text fields, buttons, labels, etc) scales as well, which I don’t want. I just want the container to grow and everything in the container to keep the same dimensions.

Now I _can_ workaround this by absolutely positioning an element behind my content and scaling that element instead, but that code bloats my XML file, and, it’s _really_ hard to get all the dimensions and positioning of these elements to line up perfectly across devices and screen sizes.

I realize this is not an easy thing to implement, as I’m sure there are layout considerations, but I thought I’d log this limitation to see if others in the community are hitting this problem as well.

NathanWalker:
I ran into this last week as well and would love to see this as well. Aside from my emoticon reaction above, nothing like words of affection :+1:

vjoao:
I've been looking into this for quite a long time. Much appreciated.

TheOriginalJosh:
👍

patuku-roy:
+1

mikkeldamm:
👍

honrust:
+1, yes please, sounds pretty essential for animating list item expanding/collapsing.

codeback:
:+1:

NickArgyle:
You can sort of hack this together using the timer and a loop. You won't get curve based animation, but it works in a pinch. 

Something like this:

```
function sideBarIn() {
    console.log("starting loop");
    sideOpen = 1;
    pageData.set("hamburgerVis", "collapsed");

    function nextFrame() {
        if (x < 25) {
            x = x + 5;
            xString = x.toString() + "%";
            console.log(xString);
            pageData.set("sideWidth", xString);
            setTimeout(nextFrame, 1);
        }
    }
    // Start the loop
    setTimeout(nextFrame, 0);}

function sideBarOut() {
    sideOpen = 0;
    console.log("starting loop");

    function nextFrame() {
        if (x > 0) {
            x = x - 5;
            if (x == 0) {
                xString = x.toString();
                console.log(xString);
                pageData.set("sideWidth", xString);
            } else {
                xString = x.toString() + "%";
                console.log(xString);
                pageData.set("sideWidth", xString);
            }
            setTimeout(nextFrame, 1);
        }
    }
    // Start the loop
    setTimeout(nextFrame, 0);
}
```

[...]

PR #5147, opened 2017-12-07
feat(animation): support animating width/height properties

vchimev:
Original PR https://github.com/NativeScript/NativeScript/pull/4917 by @justindujardin that implements https://github.com/NativeScript/NativeScript/issues/1764.

### DIFF
--- a/apps/app/ui-tests-app/animation/animation-army-100.ts
+++ b/apps/app/ui-tests-app/animation/animation-army-100.ts
@@ -1,0 +1,114 @@
+import * as view from 'tns-core-modules/ui/core/view';
+import {View} from 'tns-core-modules/ui/core/view';
+import * as pages from 'tns-core-modules/ui/page';
+import {Button} from 'tns-core-modules/ui/button';
+import {SegmentedBar, SegmentedBarItem} from 'tns-core-modules/ui/segmented-bar';
+import {Label} from 'tns-core-modules/ui/label';
+import {Animation, AnimationDefinition} from 'tns-core-modules/ui/animation';
+import * as fpsMeter from 'tns-core-modules/fps-meter';
+
+let fpsCallbackId;
+export function onLoaded(args) {
+    const page = args.object;
+    const fpsLabel = view.getViewById(page, 'fps') as Label;
+    fpsCallbackId = fpsMeter.addCallback((fps: number, minFps: number) => {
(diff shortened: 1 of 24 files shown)


vchimev:
Hello All,

Great work on this pull-request!

Our main concerns with this implementation are related to performance. I am afraid that we have not had the time to measure and investigate it due to higher priority tasks. However, we are aware of its importance and wide demand, that is why we would like to invest further efforts and improve this feature as well as the animations at all.

This has not been scheduled for release yet. Currently, we are working on the 4.1 roadmap, so I hope we could get it there. Please follow the [issue](https://github.com/NativeScript/NativeScript/issues/1764) and this [pull-request](https://github.com/NativeScript/NativeScript/pull/5147) for any comments and updates.

MisterBrownRSA:
When are you planning on giving this Pull Request some daylight?

farfromrefug:
@vchimev i have read through the PR and I dont really see anything that would impact performances if not using width or height in into animations. I understand you want to improve the whole animation system but in the meantime it would still be very useful to get that PR in.
And it would allow us not to use tricks like rxjs or tweenjs  which might have worst performance consequences.

Jonarod:
Hey !!
What is to be done for this to be released ???
I can see some UI tests were not successful, however seems like the page hosting details is down... so it is not possible to check where it fails.

@vchimev could you bring some details back ?

Also, I agree with @farfromrefug here, as of today we use funky workarounds like Tween.js to animate width and height, for what I understand from @justindujardin 's code is that it should not be worse than todays hacks... Maybe you could release that PR with some beta warnings like **to use with caution - still in beta** or whatever: the community will surely help to stabilize it.

I can totally understand concerns (plus I can't see why tests fail...) but the PR is over 1 year old now :sweat_smile: 

Thanks !!

elvticc:
Hello! Thank you very much for your improvements to NativeScript. This feature is essential for many developers, could you implement it please? Would you have a planned date?

jbarnabe:
Thanks @MartoYankov and @SvetoslavTsenov for this improvement! Looking forward to use it.

danielnitu:
Is this scheduled for future releases? I'm not able to use it in 5.4.3 and couldn't find more details elsewhere

MartoYankov:
@danielnitu Sorry for the late reply. This will be available in the 6.0 release.

PR #5156, opened 2017-12-11
feat: Add require.context typings

PanayotCankov:
Add WebpackRequire interface and append it with the require typing,
this will let us use `require.context("~/", true, /page\.(xml|js|css)/g)`


SvetoslavTsenov:
uitests

PR #5158, opened 2017-12-11
feat: Register ./app.css instead of app.css so it can be provided by webpack context

PanayotCankov:
This will let us register the app.css in webpack from a context, and potentially
have a configuration such as:
```
const appCssContext = require.context("~/", false, /^\.\/app\.(css|scss|less|sass)$/);
global.registerWebpackModules(appCssContext);
```
That will work with all of the app.css, app.scss, app.less etc. without further manual reconfiguration.


SvetoslavTsenov:
uitests

SvetoslavTsenov:
testsappng

SvetoslavTsenov:
@PanayotCankov, it looks like the builds are fine. Just ignore the single test failure, it is not related to this PR.

PR #5163, opened 2017-12-12
update(launch screen): Add iPhone X launch screens

tdermendjiev:


hshristov:
@tdermendjiev I think these changes need to go into templates

SvetoslavTsenov:
uitests

PR #5192, opened 2017-12-20
Exported Android WebView client implementation

### LINKED ISSUE #3998: Can't handle onReceivedSslError in Android WebView
Currently a user receives no feedback when attempting to load a webpage with a bad certificate. Developers need to handle the `onReceivedSslError` callback in WebViewClient, but there is no obvious way to override this.

I've tried adding the handler in web-view.android.js directly:

```
var WebViewClientClassInner = (function (_super) {
...
    WebViewClientClassInner.prototype.onReceivedSslError = function (view, handler, error) {
        console.log("WebViewClientClass.onReceivedSslError");
        _super.prototype.onReceivedSslError.call(this, view, handler, error);
    };
```

But this is never fired.

Ideally this handler would become a part of the nativescript code base, otherwise we need a clear way to override it.

To test:

Navigate to "https://badssl.com" in an Android WebView and click on any of the bad certificate links (expired, wrong.host, etc.) The `onPageFinished` event is fired without providing any feedback. There is no way to catch the `onReceivedSslError` event.

Platform: Android
CLI: 2.5.0
Cross-platform modules: 2.5.2
Runtime: 2.5.0

surdu:
Fixes #3998 

Having the NS WebView client class exported will allow developers to extend it without using future fixes and developments added to the client class.

### DIFF
--- a/tests/app/ui/page/page-tests.android.ts
+++ b/tests/app/ui/page/page-tests.android.ts
@@ -105,6 +105,6 @@ export var test_Resolve_Fragment_ForPage = function () {
     frame.navigate(pageFactory);
     TKUnit.waitUntilReady(() => frame.navigationQueueIsEmpty());
 
-    const fragment = frame.android.fragmentForPage(testPage);
+    const fragment = frame.android.fragmentForPage(frame._currentEntry);
     TKUnit.assertNotNull(fragment, "Failed to resolve native fragment for page");
 }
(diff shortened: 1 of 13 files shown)


surdu:
I'm having a small problem figuring out how to use this from JS. I'm trying this:

```js
WebViewClientImpl.prototype.onReceivedSslError = function (view, handler) {
	handler.proceed();
};

webview.android.setWebViewClient(new WebViewClientImpl(webview));
```

But the onReceivedSslError is not called when it should.

surdu:
Apparently, what I tried to do here [will not work due to a limitation in NS](https://discourse.nativescript.org/t/extend-nativescripts-default-webviewclient-implementation/4077/7?u=surdu).

PR #5211, opened 2017-12-27
example - how to use BackstackEntry with back navigation

tsonevn:
example - how to use BackstackEntry with back navigation

### DIFF
--- a/tests/app/ui/frame/frame.md
+++ b/tests/app/ui/frame/frame.md
@@ -20,3 +20,32 @@ To perform navigation, you will need a reference to the topmost frame of the app
 
 ### Navigating Back
 {%snippet frame-back%}
+
+### Back Navigation with BackstackEntry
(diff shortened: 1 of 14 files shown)


tests/app/ui/frame/frame.md
```diff
@@ -20,3 +20,32 @@ To perform navigation, you will need a reference to the topmost frame of the app
...
+        <Label text="Page 5" class="h1 text-center"/>
+        <Button text="back to the first page" tap="onTap" class="btn btn-primary btn-active"/>
+    </StackLayout>
+</Page>
+```
+```TypeScript
+import {topmost, BackstackEntry} from "ui/frame"
+
+export function backNavigation(args){
+    let backstackEntryFirstPage = topmost().backStack[0];
```
hshristov: Instead of 'topmost()' use 'args.object.page.frame'

Issue #3605, opened 2017-02-08
ImageSourceModule method toBase64String not working (both iOS and Android)


emmanuel128:
On line 140 of image-source-android.js `this.android.compress(targetFormat, quality, base64Stream);` I'm getting this error `this.android.compress is not a function`. I'm using the nativescript-camera to take a picture and then save it to base64 string. Also, the image is saving to my photos even when I set `saveToGallery: false`.

Here my code: 
```
cameraModule.takePicture({
        width: 1280, height: 720, keepAspectRatio: false, saveToGallery: false
    }).then(function (imageAsset) {
        console.log("Result is an image asset instance");
        viewModel.set("BoardingPassSource", imageAsset);
        var image = ImageSourceModule.fromNativeSource(imageAsset);
        var base64 = image.toBase64String("jpeg");
        // var image = new imageModule.Image();
        // image.imageSource = imageAsset;
        // image.toBase64String("jpeg");
        viewModel.set("base64String", base64);
    }).catch(function (err) {
        console.log("Error -> " + err.message);
    });
```

package.json
```
{
  "description": "NativeScript Application",
  "license": "SEE LICENSE IN <your-license-filename>",
  "readme": "NativeScript Application",
  "repository": "<fill-your-repository-here>",
  "nativescript": {
    "id": "org.nativescript.****",
    "tns-android": {
      "version": "2.5.0"
    }
  },
  "dependencies": {
    "nativescript-background-http": "^2.4.2",
    "nativescript-barcodescanner": "^2.3.3",
    "nativescript-camera": "0.0.8",
    "nativescript-drawingpad": "^1.1.2",
    "nativescript-drop-down": "^1.5.1",
    "nativescript-loading-indicator": "^2.2.2",
    "nativescript-telerik-ui": "^1.5.1",
    "nativescript-theme-core": "^1.0.2",
    "nativescript-timedatepicker": "^1.1.0",
    "tns-core-modules": "^2.5.0"
  },
  "devDependencies": {
    "babel-traverse": "6.21.0",
    "babel-types": "6.21.0",
    "babylon": "6.14.1",
    "lazy": "1.0.11"
  }
}
```

NickIliev:
Hi @emmanuel128 

I can confirm this issue is reproducible with tns-core-modules 2.5.0. We will get back at you in this thread once we have applicable solution and/or fix.

[...]

PR #5273, opened 2018-01-15
refactor(image-source): throw if source is not a correct native instance

vchimev:
https://github.com/NativeScript/NativeScript/issues/3605

### DIFF
--- a/tns-core-modules/image-source/image-source.android.ts
+++ b/tns-core-modules/image-source/image-source.android.ts
@@ -140,6 +140,9 @@ export class ImageSource implements ImageSourceDefinition {
     }
 
     public setNativeSource(source: any): boolean {
+        if (!(source instanceof android.graphics.Bitmap)) {
+            throw new Error("The method setNativeSource() expects android.graphics.Bitmap instance.");
+        }
         this.android = source;
         return source != null;
     }
--- a/tns-core-modules/image-source/image-source.android.ts
+++ b/tns-core-modules/image-source/image-source.android.ts
@@ -139,12 +139,11 @@ export class ImageSource implements ImageSourceDefinition {
         });
     }
 
-    public setNativeSource(source: any): boolean {
-        if (!(source instanceof android.graphics.Bitmap)) {
+    public setNativeSource(source: any): void {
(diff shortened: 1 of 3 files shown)


tns-core-modules/image-source/image-source.android.ts
```diff
@@ -140,6 +140,9 @@ export class ImageSource implements ImageSourceDefinition {
     }
 
     public setNativeSource(source: any): boolean {
+        if (!(source instanceof android.graphics.Bitmap)) {
```
DimitarTachev: In this way, the method will no longer return false, it will throw an exception instead. It could be something like:

    if (source && !(source instanceof android.graphics.Bitmap))

tns-core-modules/image-source/image-source.ios.ts
```diff
@@ -119,9 +119,10 @@ export class ImageSource implements ImageSourceDefinition {
     }
 
     public setNativeSource(source: any): boolean {
-        if (source instanceof UIImage) {
-            this.ios = source;
+        if (!(source instanceof UIImage)) {
```
DimitarTachev: In this way, the method will no longer return false, it will throw an exception instead. It could be something like:

    if (source && !(source instanceof UIImage))

vchimev:
Thank you, @DimitarTachev!

tns-core-modules/image-source/image-source.ios.ts
```diff
@@ -118,11 +118,11 @@ export class ImageSource implements ImageSourceDefinition {
         });
     }
 
-    public setNativeSource(source: any): boolean {
-        if (source instanceof UIImage) {
-            this.ios = source;
+    public setNativeSource(source: any): void {
```
vakrilov: This is a change in the public API -> mark this as breaking change if it is really needed.
Also the definitions (`image-source.d.ts`) should be updated

vchimev: Done.

vchimev:
Changes are done. Branch is rebased. Merge after green CI.

vchimev:
iOS

vchimev:
unit-tests-ios

PR #5303, opened 2018-01-19
chore(uitestsapp): replace start with run

SvetoslavTsenov:


SvetoslavTsenov:
uitests

SvetoslavTsenov:
Using run application with frame root instead start and page root (old way), causes two tests to fail which failures seems to be related to css scope e.t.c.

PR #5309, opened 2018-01-22
feat(observable): Implement observable .once

PanayotCankov:
Node has a handy one-liner method on its EventEmitter called "once".
It is similar to "on", but fires a single time and automatically unsubscribes.


vakrilov:
run ci

PR #5315, opened 2018-01-24
fix: layoutParent crash with ProxyViewContainer

vakrilov:
`layoutParent` method crashes with `ProxyViewContainer` as it does not have a native view.


vakrilov:
testsappng

PR #5320, opened 2018-01-24
feat(frame): introduce defaultPage property

ADjenkov:
- Add `defaultPage` string property to Frame 
- Forbid inline content for Frame (_Frame page should be only specified by `defaultPage` property_)


SvetoslavTsenov:
uitests

hshristov:
Maybe `firstPage` is more meaningful?

PR #5333, opened 2018-01-25
fix-next(android-tabview): handle tab fragments attach/detach properly

MartoYankov:
Force the android TabView pager adapter to detach tab fragments on TabView unloaded and to re-attach them on TabView loaded. Keeping the tab fragments attached after unload/navigate is problematic since they will try to recreate their native views on suspend/resume of the app.

This handles the scenario:
1. TabView is in a Frame.
2. Navigate to a new page.
3. Suspend/resume the app with "Don't keep activities".
4. Go back to the page with tabs.
Result: the tabs are empty.


SvetoslavTsenov:
uitests

SvetoslavTsenov:
cuteness

SvetoslavTsenov:
testsappng

PR #5404, opened 2018-02-08
hiding cookbook/ui/action-bar article in favor of ui/action-bar

tsonevn:
hiding cookbook/ui/action-bar article in favor of ui/action-bar


SvetoslavTsenov:
test

Issue #5407, opened 2018-02-09
Revise frame access API in the context of {N} 4.0


manoldonev:
### Which platform(s) does your issue occur on?
Both

### Please provide the following version numbers that your issue occurs with:
- Cross-platform modules: 4.0.0

frame.topmost() API is the most obvious "static" API to access navigation frame that is very popular among customers. The fact that it just works (until now) is related to {N} apps mostly having a single internally created frame up to {N} 4.0. With {N} 4.0 we expose the ability to the developer to create multiple frames in different scenarios and there are scenarios where frame.topmost() is not really applicable. Generally frame.topmost() returns the last frame that performed navigation but that is not entirely obvious either from its name, or from the documentation -- with {N} 4.0 the following scenario will be possible that renders frame.topmost() meaningless:

- Root tabview with two tab items
- Each tab item contains frame as a root element (with defaultPage property set to module name)

When you load the app:
- First tab item is processed thus its frame is processed and navigation occurs to the page set as frame.defaultPage (implicitly this adds frame1 to the top of the frame stack)
- Second tab item is processed (nevertheless it is not visible) thus its frame is processed and its navigation entry is added to the navigation queue (but at the same time this implicitly adds frame2 to the top of the frame stack)

So if while playing with the visible first tab, you try to reference its frame via frame.topmost() you will actually get a reference to the invisible frame from the second tab.

Consider:
- Updating frame.topmost() documentation to explain that it returns the frame that performed navigation last
- Extend frame access API besides frame.topmost() to support scenarios like the one above (possible use frame.id and provide API for the user to lookup frames by id)
- Do we need the frame stack at all? Is it a stack or something else?

MartoYankov:
@manoldonev 
Note: The frame.topmost() API returns the last navigated Frame instance that hasn't been already in the stack. This means that if you have 2 Frames and navigate in this order, Frame1 then Frame2 then again Frame1, the frame.topmost() API will return Frame2.

Issue #5340, opened 2018-01-26
Frame.topmost() unexpected behavior after closing modal dialog with frame inside


manoldonev:
### Tell us about the problem
Frame.topmost() generally returns the last frame that performed navigation. However, with ns-4.0 there is a scenario where you can open a modal dialog with frame inside (and presumably navigate within the modal). Then at some point you close the modal dialog -- the unexpected behavior is that if you call Frame.topmost() at this moment it will return the frame from the modal dialog (that is actually unloaded) -- it should return the host frame or undefined.

### Which platform(s) does your issue occur on?
Both

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.4.1
- Cross-platform modules: 4.0.0

PR #5408, opened 2018-02-09
fix(frame): root tabview with modal frame when suspend/resume app Fixes #5398

### LINKED ISSUE #5398: Error in a root tabview scenario with modal navigation when recreating app activities
Android

- CLI: 3.4.1
- Cross-platform modules: 4.0.0

`Error: Cannot pop a Frame which is not at the top of the navigation stack.` in the following scenario (can be reproduced in apps/modal-navigation-app):
[Set "Don't keep activities" flag in Android developer options]
1. Start app with root tabview visual.
2. Open modal dialog with root frame inside (frame element is crucial to reproduce the issue).
3. Minimize test app and then restore it.
4. [Modal dialog is restored open] Try to close the modal dialog via closeModal(...) call or press Android "Back" button twice -- you get the error above.

manoldonev:
- Fixes https://github.com/NativeScript/NativeScript/issues/5398
- Revises fix for https://github.com/NativeScript/NativeScript/issues/5340
- Logged new feature https://github.com/NativeScript/NativeScript/issues/5407

### DIFF
--- a/tests/app/testRunner.ts
+++ b/tests/app/testRunner.ts
@@ -226,7 +226,7 @@ import * as cssAnimationTests from "./ui/animation/css-animation-tests";
 allTests["CSS-ANIMATION"] = cssAnimationTests;
 
 import * as transitionTests from "./navigation/transition-tests";
-allTests["TANSITIONS"] = transitionTests;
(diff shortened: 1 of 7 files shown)


manoldonev:
test

PR #5411, opened 2018-02-12
 fix(ios-layouts): switch contentInsetAdjustmentBehavior on ScrollView 

ADjenkov:
Do not auto adjust the `ScrollView` insets, by disabling [contentInsetAdjustmentBehavior](https://developer.apple.com/documentation/uikit/uiscrollviewcontentinsetadjustmentbehavior) property, when `ScrollView` is used as a root `Page` element.

_Node:_ By default UIScrollView is auto adjusting its content inset depending on the constants indicating how safe area insets are added. This 'scrolls' the `ScrollView` under the ActionBar and the safe area when it's placed as a child of the root `Page` element:

```
<Page>
   <ActionBar/>
   <GridLayout>
     <ScrollView/>
   </GridLayout>
</Page>

### DIFF
--- a/tns-core-modules/ui/scroll-view/scroll-view.ios.ts
+++ b/tns-core-modules/ui/scroll-view/scroll-view.ios.ts
@@ -1,10 +1,13 @@
 ﻿import { ScrollEventData } from ".";
 import { View, layout, ScrollViewBase, scrollBarIndicatorVisibleProperty } from "./scroll-view-common";
+import { ios as iosUtils } from "../../utils/utils";
 // HACK: Webpack. Use a fully-qualified import to allow resolve.extensions(.ios.js) to
 // kick in. `../utils` doesn't seem to trigger the webpack extensions mechanism.
 import * as uiUtils from "tns-core-modules/ui/utils";
 
 export * from "./scroll-view-common";
+
+const majorVersion = iosUtils.MajorVersion;
 
 class UIScrollViewDelegateImpl extends NSObject implements UIScrollViewDelegate {
     private _owner: WeakRef<ScrollView>;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: tns-core-modules/ui/scroll-view/scroll-view.ios.ts [hunk 1]
```diff
@@ -120,6 +123,16 @@ export class ScrollView extends ScrollViewBase {
...
         this._contentMeasuredWidth = this.effectiveMinWidth;
         this._contentMeasuredHeight = this.effectiveMinHeight;
+
+        // `_automaticallyAdjustsScrollViewInsets` is set to true only if the first child 
+        // of UIViewController (Page, TabView e.g) is UIScrollView (ScrollView, ListView e.g).
+        // On iOS 11 by default UIScrollView automatically adjusts the scroll view insets, but they s
+        if (majorVersion > 10 && !this.parent._automaticallyAdjustsScrollViewInsets) {
+            // Disable automatic adjustment of scroll view insets when ScrollView 
+            // is not the first child of UIViewController.
+            this.nativeViewProtected.contentInsetAdjustmentBehavior = 2;
```
vakrilov: You can reset the property to automatic(`0`) if `_automaticallyAdjustsScrollViewInsets` is `true`. Also You can add definitions for the enum value. It is not generated, because it ios 11 only property:
- [`automatic = 0`](https://developer.apple.com/documentation/uikit/uiscrollviewcontentinsetadjustmentbehavior/2902258-automatic)
- [`never = 2`](https://developer.apple.com/documentation/uikit/uiscrollviewcontentinsetadjustmentbehavior/2902255-never)

PR #5426, opened 2018-02-15
fix(ios-frame): do not update backstack when navigating the same page

ADjenkov:
The problem:
Currently the entries backstack is updated on every page navigation (even if the entry is the same as the current one) which leads to unnecessary unload events.

The Solution: Prevent backstack update when navigating to the same entry ([already done for Android](https://github.com/NativeScript/NativeScript/blob/b8e0beccddbc538860ba72991e3c225fa470836f/tns-core-modules/ui/frame/frame.android.ts#L202))

### DIFF
--- a/tns-core-modules/ui/frame/frame.ios.ts
+++ b/tns-core-modules/ui/frame/frame.ios.ts
@@ -37,6 +37,14 @@ export class Frame extends FrameBase {
         return this._ios;
     }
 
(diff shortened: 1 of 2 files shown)


ADjenkov:
test uitests testsappng cuteness

dtopuzov:
run ci

PR #5435, opened 2018-02-19
fix-next(tabview): fix tab title layout in modal root tabview Fixes #5392

### LINKED ISSUE #5392: [TabView] Tabitem title not measured/arranged properly in a modal dialog with root tabview inside
Android

- CLI: 3.4.1
- Cross-platform modules: 4.0.0

Second tabItem title is not measured/arranged properly and gets wrapped in a scenario with opening a modal dialog with root tabview visual inside. Note that the problem is only present if the modal dialog is opened with fullscreen flag set to false.

See the attached screenshot (can be reproduced in e2e/modal-navigation):

screenshot_1517935730

manoldonev:
Fixes https://github.com/NativeScript/NativeScript/issues/5392

See comments below.

~~In a modal dialog scenario with root tabview inside, the DialogFragment content at some point is set via PhoneWindow.setContentView(view) call where `view` is the gridLayout root element of the tabview. This results in `setContentView(view, new ViewGroup.LayoutParams(MATCH_PARENT, MATCH_PARENT))` call that unconditionally overwrites the default instance of CommonLayoutParams attached to the gridLayout by us. As the default setter essentially created `new ViewGroup.LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT, Gravity.FILL)` we lose the Gravity.FILL setting and that prevents the gridLayout from stretching horizontally so tab item titles do not distribute the excess whitespace among themselves.~~

~~The proposed solution is to add a no-op outer native wrapper element to be laid out by the DialogFragment so we can perform the layout of the inner `gridLayout` element as expected.~~

~~Alternatively (as this problem is related only to modal dialog scenarios and I do not like the solution very much) we can make no changes to the tabview codebase and instruct those devs that hit the problem to add the additional GridLayout wrapper element by themselves in their own application code (the drawback is that it looks "broken" out-of-the-box).~~

Also, @MartoYankov pointed out something else that we might consider at some point -- generally the problem here is triggered by https://github.com/NativeScript/tns-core-modules-widgets/blob/a9ac20da20422ae57aa4cfa76bc03dc27d6bc885/android/widgets/src/main/java/org/nativescript/widgets/GridLayout.java#L259 in the GridLayout codebase that decides how the grid layout should behave based on a setting of its own LayoutParams while Android documentation states that "LayoutParams are used by views to tell their parents how they want to be laid out".  

BREAKING CHANGE: [Android] NativeScript no longer overwrites the horizontal/vertical alignment on the user-defined root visual element when opening it in modal dialog window (i.e. not fullscreen).

If your application logic relied on the previous behavior you need to manually set `horizontalAlignment="center"` and `verticalAlignment="middle"` on the root visual element you are showing modally.

### DIFF
--- a/apps/app/ui-tests-app/modal-view/login-page.xml
+++ b/apps/app/ui-tests-app/modal-view/login-page.xml
@@ -1,4 +1,6 @@
-﻿<Page xmlns="http://schemas.nativescript.org/tns.xsd" shownModally="onShownModally" loaded="onLoaded" unloaded="onUnloaded" backgroundColor="Red">
+﻿<Page xmlns="http://schemas.nativescript.org/tns.xsd" shownModally="onShownModally" 
+  loaded="onLoaded" unloaded="onUnloaded" backgroundColor="Red"
+  horizontalAlignment="center" verticalAlignment="middle">
   <StackLayout backgroundColor="PaleGreen" margin="10">
     <TextField hint="username" id="username" text="username"/>
     <TextField hint="password" id="password" text="password" secure="true"/>
--- a/tns-core-modules/ui/core/view/view.android.ts
+++ b/tns-core-modules/ui/core/view/view.android.ts
@@ -125,6 +125,7 @@ function initializeDialogFragment() {
         private _fullscreen: boolean;
         private _shownCallback: () => void;
         private _dismissCallback: () => void;
+        private _nativeView: android.widget.FrameLayout;
 
         constructor() {
             super();
@@ -142,10 +143,6 @@ function initializeDialogFragment() {
             this.setStyle(android.app.DialogFragment.STYLE_NO_TITLE, 0);
             
             const dialog = new DialogImpl(this, this.getActivity(), this.getTheme());
-
-            // adjust alignment based on fullscreen value.
-            this.owner.horizontalAlignment = this._fullscreen ? "stretch" : "center";
-            this.owner.verticalAlignment = this._fullscreen ? "stretch" : "middle";
           
             return dialog;
         }
@@ -154,8 +151,18 @@ function initializeDialogFragment() {
             const owner = this.owner;
             owner._setupAsRootView(this.getActivity());
             owner._isAddedToNativeVisualTree = true;
-            
-            return this.owner.nativeViewProtected;
+
+            // We need a native wrapper element to be laid out by the DialogFragment and 
+            // adjust its alignment for fullscreen so we can make sure that the actual 
+            // modal view will be laid out as expected on component level.
+            this._nativeView = new android.widget.FrameLayout(owner._context);
+            this._nativeView.addView(owner.nativeViewProtected);
+
+            // adjust alignment based on fullscreen value.
+            this._nativeView.horizontalAlignment = this._fullscreen ? "stretch" : "center";
+            this._nativeView.verticalAlignment = this._fullscreen ? "stretch" : "middle";
+
+            return this._nativeView;
         }
 
         public onStart(): void {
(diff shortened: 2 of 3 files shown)


vakrilov:
Another approach that might worth considering: Doing the wrap with the additional gird layout in the DialogFragment when showing the dialog (somewhere [here](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/core/view/view.android.ts#L153-L159)). This will probably prevent similar problems with other views hosted inside a dialog fragment.

manoldonev:
@vakrilov Tried this suggestion (moved the FrameLayout wrapper in DialogFragment.onCreateView(...)) as it would be a better solution to the problem but I cannot make it work. 

When we start opening the modal dialog everything looks good -- FrameLayout sets default layout params to its only child (the gridLayout that is tabview root) with height and width = MATCH_PARENT and gravity -1 (that we later coerce to Gravity.FILL). But at some point something happens and `GridLayout.setLayoutParams(...)` is called again with different layout params that change the gravity to 113 (Gravity.CENTER_HORIZONTAL | Gravity.FILL_VERTICAL). This again breaks the desired horizontal stretch behavior and unfortunately the stack trace is not helpful to get an idea what is going on as the call is raised by the infamous `com.tns.Runtime.callJSMethodNative(...)`:

screen shot 2018-02-23 at 7 12 44 pm

hshristov:
@manoldonev Check this:
https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/core/view/view.android.ts#L147

manoldonev:
test

manoldonev:
Thanks for the tip @hshristov 👍 

Indeed those explicit alignment setters are the root cause of the problem and not `PhoneWindow.setContentView(view) ` as I thought initially. I first moved those alignment setters to the new native wrapper element in the fragment and it was OK but then I tested with modal tabview root (https://github.com/NativeScript/NativeScript/commit/a416fa4fb61c105f3cce94b9c8496ea939ba9c27) without adding wrapper element and without alignment setters as well and it still visualizes in the same way so I removed the alignment setters altogether.

manoldonev:
test

manoldonev:
@vakrilov @SvetoslavTsenov There is actually a different behavior with the new implementation that fails one ui test here http://nsbuild01.telerik.com/build/job/integration-tests-stable-uitests-android-api19-emulator/1963/artifact/target/surefire-reports/html/index.html:

Actual:
modalview_01_popup_actual

Expected:
modalview_01_popup_expected


However, I think that the new behavior should be accepted as "correct" as we do not force alignment onto the user-defined content -- if he wishes so, he can achieve the old behavior by specifying the horizontal/vertical alignment in his own view (moreover, I think we started setting the alignment fairly recently so it hasn't actually been publicly released yet):
``` xml
<Page xmlns="http://schemas.nativescript.org/tns.xsd" shownModally="onShownModally" loaded="onLoaded" unloaded="onUnloaded" backgroundColor="Red"
horizontalAlignment="center" verticalAlignment="middle">
  <StackLayout backgroundColor="PaleGreen" margin="10">
    <TextField hint="username" id="username" text="username"/>
    <TextField hint="password" id="password" text="password" secure="true"/>
    <Button text="Login" tap="onLoginButtonTap"/>
  </StackLayout>
</Page>
```

manoldonev:
test

manoldonev:
test

apps/app/ui-tests-app/modal-view/login-page.xml
```diff
@@ -1,4 +1,6 @@
-﻿<Page xmlns="http://schemas.nativescript.org/tns.xsd" shownModally="onShownModally" loaded="onLoaded" unloaded="onUnloaded" backgroundColor="Red">
+﻿<Page xmlns="http://schemas.nativescript.org/tns.xsd" shownModally="onShownModally" 
+  loaded="onLoaded" unloaded="onUnloaded" backgroundColor="Red"
+  horizontalAlignment="center" verticalAlignment="middle">
```
manoldonev: This change is required as we no longer force alignment settings on user content unless it is absolutely necessary i.e. if fullscreen dialog is requested we would override these and stretch the user content.

manoldonev:
test

manoldonev:
test

Issue #5169, opened 2017-12-13
iOS: NavigationButton text 


tsonevn:
_From @Nurgunkalol on October 18, 2017 6:39_

Hello, im using navigation button in action bar with text="Go Back" (for example). When i am routing to page(component) first time navigation button always show default text "Back" and then i go to the next page and then return, navigation button text shows "Go Back". How to fix this bug ? or should i just use custom ActionItem ?
p.s I am using different language for navigation button text

_Copied from original issue: NativeScript/nativescript-angular#1045_

tsonevn:
_From @yamaoto on October 18, 2017 6:45_

+1

tsonevn:
Hi @Nurgunkalol 
I tested the scenario with setting up a custom title for the NavigationButton button on iOS, however, was unable to recreate an issue. For your convenience, I am attaching GIF file from the simulator. You should be able to use the NavigationButton element without the need to create a custom ActionItem.
I am attaching sample project, please review it and make the needed changes, which will allow us to recreate the problem. 
Also provide some more info about your case and about your environment(CLI, node, tns-core-modules, nativescript-angular versions).
![screencast 2017-10-18 at 1 14 41 pm](https://user-images.githubusercontent.com/17448734/31713600-6cac3bf2-b407-11e7-9941-ee7477a5c238.gif)
[Archive.zip](https://github.com/NativeScript/nativescript-angular/files/1394236/Archive.zip)

tsonevn:
_From @smarza on October 26, 2017 12:1_

@Nurgunkalol @yamaoto 

I had the same issue. Then I realised that the problem was performing because of the use of *ngIf in the NavigationButton. Removing it, the custom text works.

Maybe it is related to some king of Angular binding issue.

Let me know if it is your case too.

tsonevn:
Hi @Nurgunkalol  @yamaoto,
Were you able to resolve the issue or the problem still persists?

tsonevn:
closing due to inactivity

tsonevn:
_From @uzarsalan on December 9, 2017 14:43_

I have the same issue.
![-09-2017 22-42-27](https://user-images.githubusercontent.com/16972594/33796589-60506eec-dd32-11e7-88e6-aee0b1b69ee1.gif)

tsonevn:
Hi @uzarsalan,
[...]

PR #5458, opened 2018-02-23
fix(ios-action-bar): NavigationButton text change on first navigation

ADjenkov:
**The problem**: NavigationButton (a.k.a back button) won't change its title on first navigation when there 
is no transition animation - iOS 11+ only.

**The solution**: Set empty back button title on the parent view controller when navigating

For reference: [stackoverflow1](https://stackoverflow.com/questions/44815042/uinavigationcontroller-back-button-doesnt-hide-title-on-ios-11)  and  [stackoverflow2](https://stackoverflow.com/questions/1449339/how-do-i-change-the-title-of-the-back-button-on-a-navigation-bar) 

Fix https://github.com/NativeScript/NativeScript/issues/5169

### DIFF
--- a/tns-core-modules/ui/frame/frame.ios.ts
+++ b/tns-core-modules/ui/frame/frame.ios.ts
@@ -13,6 +13,8 @@ import * as utils from "../../utils/utils";
 
 export * from "./frame-common";
 
+const majorVersion = utils.ios.MajorVersion;
+
 const ENTRY = "_entry";
 const NAV_DEPTH = "_navDepth";
 const TRANSITION = "_transition";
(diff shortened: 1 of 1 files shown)


tns-core-modules/ui/frame/frame.ios.ts
```diff
@@ -86,6 +88,11 @@ export class Frame extends FrameBase {
         backstackEntry[NAV_DEPTH] = navDepth;
         viewController[ENTRY] = backstackEntry;
 
+        if (majorVersion > 10) {
```
vakrilov: Can you pls add a comment in the code explaining why this is necessary, as the fix is not obvious :)

hshristov:
@ADjenkov Won't this remove all back buttons?

tns-core-modules/ui/frame/frame.ios.ts
```diff
@@ -86,6 +88,13 @@ export class Frame extends FrameBase {
...
```
manoldonev: Maybe extend condition to `if (!animated && majorVersion > 10)` ?

SvetoslavTsenov:
test

PR #5468, opened 2018-02-26
fix-next: ensure proper events on tab change

MartoYankov:
* Add an executePendingTransactions() call to resetRootView, so that current transactions are executed properly. Add a test for this.
* [iOS] execute the Frame setCurrent() method only if the entry has changed, so that navigatedTo and navigatedFrom events are fired properly.
* [iOS] assign the willSelectViewController to tab pages on selectedIndexChanged, so that navigatingFrom isn't thrown if using the property directly.

### DIFF
--- a/tns-core-modules/ui/frame/frame.android.ts
+++ b/tns-core-modules/ui/frame/frame.android.ts
@@ -201,9 +201,7 @@ export class Frame extends FrameBase {
         const currentEntryChanged = current !== entry;
         if (currentEntryChanged) {
             this._updateBackstack(entry, isBack);
-        }
 
-        if (currentEntryChanged) {
             // If activity was destroyed we need to destroy fragment and UI
             // of current and new entries.
             if (this._tearDownPending) {
@@ -936,7 +934,9 @@ class ActivityCallbacksImplementation implements AndroidActivityCallbacks {
 
     public resetActivityContent(activity: android.app.Activity): void {
         if (this._rootView) {
-            // if we already have a root view, we reset it.
+            const manager = this._rootView._getFragmentManager();
+            manager.executePendingTransactions();
+
             this._rootView._onRootViewReset();
         }
         // Delete previously cached root view in order to recreate it.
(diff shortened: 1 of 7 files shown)


tns-core-modules/ui/frame/frame.android.ts
```diff
@@ -936,7 +934,9 @@ class ActivityCallbacksImplementation implements AndroidActivityCallbacks {
 
     public resetActivityContent(activity: android.app.Activity): void {
         if (this._rootView) {
-            // if we already have a root view, we reset it.
+            const manager = this._rootView._getFragmentManager();
```
hshristov: Are you sure that all `rootView`s will have fragmentManager?

MartoYankov: All instances of `View` should have fragment manager from the android context. Isn't this correct?

MartoYankov:
test branch_cuteness branch_testsappng branch_groceries

tns-core-modules/ui/tab-view/tab-view.ios.ts
```diff
@@ -399,6 +399,7 @@ export class TabView extends TabViewBase {
         }
 
         if (value > -1) {
+            (<any>this._ios)._willSelectViewController = this._ios.viewControllers[value];
```
manoldonev: If we try to select an invalid index (larger than the available tabs) this will probably crash.

tns-core-modules/ui/frame/frame.android.ts
hshristov: Actually yes 👍

tns-core-modules/ui/tab-view/tab-view.ios.ts
MartoYankov: selectedIndexProperty is a CoercibleProperty. CoerceValue is overridden to convert the passed value to max tabs length.

MartoYankov:
test branch_cuteness branch_testsappng branch_groceries

PR #5475, opened 2018-02-27
fix(animations): change throw -> trace to avoid unnecessary app crash

NathanWalker:
Fixes cause of crashes/bugs in production apps using animation.

## PR Checklist

Developers create apps using animation library and under normal circumstances and well coded scenarios the app will `throw` when navigating or other normal app usage causing bugginess.

## What is the new behavior?

Developers can now see the warning in the console when developing if tracing is enabled to let them know but keep their app stable.

## Cause of instability in production Portable North Pole

See screenshot of Crashlytics reports on this.

animation-bugs

/cc @atanasovg @vakrilov @sis0k0 @DickSmith

### DIFF
--- a/tns-core-modules/ui/animation/animation-common.ts
+++ b/tns-core-modules/ui/animation/animation-common.ts
@@ -10,9 +10,9 @@ import { View } from "../core/view";
 
 // Types.
 import { Color } from "../../color";
-import { isEnabled as traceEnabled, write as traceWrite, categories as traceCategories } from "../../trace";
+import { isEnabled as traceEnabled, write as traceWrite, categories as traceCategories, messageType as traceType } from "../../trace";
 
-export { Color, traceEnabled, traceWrite, traceCategories };
+export { Color, traceEnabled, traceWrite, traceCategories, traceType };
 export { AnimationPromise } from ".";
 
 export module Properties {
@@ -84,17 +84,16 @@ export abstract class AnimationBase implements AnimationBaseDefinition {
 
     abstract _resolveAnimationCurve(curve: any): any;
 
-    public play(): AnimationPromiseDefinition {
-        if (this.isPlaying) {
-            const reason = "Animation is already playing.";
-            if (traceEnabled()) {
-                traceWrite(reason, traceCategories.Animation, 2);
-            }
-            return <AnimationPromiseDefinition>new Promise<void>((resolve, reject) => {
-                reject(reason);
-            });
-        }
+    protected _rejectAlreadyPlaying(): AnimationPromiseDefinition{
+        const reason = "Animation is already playing.";
+        traceWrite(reason, traceCategories.Animation, traceType.warn);
+        
+        return <AnimationPromiseDefinition>new Promise<void>((resolve, reject) => {
+            reject(reason);
+        });
+    }
 
+    public play(): AnimationPromiseDefinition {
         // We have to actually create a "Promise" due to a bug in the v8 engine and decedent promises
         // We just cast it to a animationPromise so that all the rest of the code works fine
         var animationFinishedPromise = <AnimationPromiseDefinition>new Promise<void>((resolve, reject) => {
@@ -129,11 +128,7 @@ export abstract class AnimationBase implements AnimationBaseDefinition {
     }
 
     public cancel(): void {
-        if (!this.isPlaying) {
-            if (traceEnabled()) {
-                traceWrite("Animation is not currently playing.", traceCategories.Animation, 2);
-            }
-        }
+        // Implemented in platform specific files
     }
 
     public get isPlaying(): boolean {
--- a/tns-core-modules/ui/animation/animation-common.ts
+++ b/tns-core-modules/ui/animation/animation-common.ts
@@ -86,7 +86,13 @@ export abstract class AnimationBase implements AnimationBaseDefinition {
 
     public play(): AnimationPromiseDefinition {
         if (this.isPlaying) {
-            throw new Error("Animation is already playing.");
+            const reason = "Animation is already playing.";
+            if (traceEnabled()) {
+                traceWrite(reason, traceCategories.Animation, 2);
+            }
+            return <AnimationPromiseDefinition>new Promise<void>((resolve, reject) => {
+                reject(reason);
+            });
         }
 
         // We have to actually create a "Promise" due to a bug in the v8 engine and decedent promises
@@ -124,7 +130,9 @@ export abstract class AnimationBase implements AnimationBaseDefinition {
 
     public cancel(): void {
         if (!this.isPlaying) {
-            throw new Error("Animation is not currently playing.");
+            if (traceEnabled()) {
+                traceWrite("Animation is not currently playing.", traceCategories.Animation, 2);
+            }
         }
     }
 
--- a/tns-core-modules/ui/animation/keyframe-animation.ts
+++ b/tns-core-modules/ui/animation/keyframe-animation.ts
@@ -12,7 +12,7 @@ import { View, Color } from "../core/view";
 
 import { AnimationCurve } from "../enums";
 
-import { isEnabled as traceEnabled, write as traceWrite, categories as traceCategories } from "../../trace";
+import { isEnabled as traceEnabled, write as traceWrite, categories as traceCategories, messageType as traceType } from "../../trace";
 
 // Types.
 import { unsetValue } from "../core/properties";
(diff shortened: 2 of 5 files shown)


vakrilov:
Hey Nathan,
There is an `isPlaying` property of animations that you can check before calling `play()` or `cancel()`.

Trying to `cancel()` an animation that is not currently playing is probably caused by an error in the application code - thus we decided to throw in such cases. If it **is** intentional - you can always do the check with `isPlaying`?

NathanWalker:
Thanks @vakrilov - we always check `isPlaying` before attempting to call `cancel`. It's a side effect likely as a result of CSS processing keyframe animations in unforeseen ways (views getting destroyed when navigating or some other reasons, etc) - that neither you nor I can anticipate and really should not be considered *exceptional*.

A brief inspection of `tns-core-modules` reveals *198 matches across 67 files* where a`throw` is raised.
This breaks down to an average ~3 errors for every source file.
This looks over 538 individual files accounting for 2.5 Mb of source code in `tns-core-modules`.

In contrast React Native raises a `throw` *66 matches across 32 files* across it's core `js` libraries.
This breaks down to an average ~2 errors for every source file. 
This looks over 859 individual files accounting for 6 Mb of source code in React Native `Libraries`.

*Less than half the amount of throw's for more than double the amount of source code.*

https://stackoverflow.com/a/77164

> beware, throwing [...] exception can kill a senior user who is just a slow typer.

> +1 excellent answer. I am so frustrated by developers working on API's that I have to consume, and throw Exceptions for every little thing. VERY few cases really require exceptions.

> Exceptions should be reserved for what's truly exceptional.

@vakrilov  Is this animation condition *truly* exceptional? 

> Reasons not to use exceptions:
> Sometimes it's overkill if the error handling is simple.
> Throwing and catching exceptions is generally significantly more expensive...

The framework core currently makes developing bug free apps more difficult due to the unnecessary usage of `throw`.

What do you recommend?

/cc @jlooper @tjvantoll @sipacate

vakrilov:
Hey @NathanWalker -in this particular example - indeed trying to `play()` an animation that is already playing might not be a case for exception, although it might still be an indicator of faulty logic. If the general opinion is that this will make a better and more easy to use API - I'll be happy to change it. 

Looking at the crash analysis report you've attached - it is the exception thrown in [`keyframe-animation.ts` file](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/animation/keyframe-animation.ts#L164) that is causing trouble.

As for throwing exceptions in general - I absolutely agree that exceptions should not be used as a control flow mechanism. I can assure you that non of that APIs in the core framework is design in a way that requires you to catch exceptions in order achieve your goal. 

What we do use as a guiding principle when writing code is [fail early/fail early](https://stackoverflow.com/questions/2807241/what-does-the-expression-fail-early-mean-and-when-would-you-want-to-do-so/2807375#2807375). Most of the `throw` statements in the code are there to warn you about a problem in the code on the first possible place when it is visible. It is better to get a crash with a stack-trace than a program that is just not working and you don't know why. The exceptions are meant to point you to the exact place of a potential bug while you are developing and testing, not to crash your app in production.
That said, I don't agree that judging the quality and ergonomics of a framework, based on the number of throws in the code is fair and I don't see how comparing to React Native brings value to this discussion.

NathanWalker:
Thank you for the understanding @vakrilov and pointing out [keyframe-animation.ts](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/animation/keyframe-animation.ts#L164) file, likely the worst offender since CSS animation issues have been reported by others in several other outlets (forums,issues,etc.). I have updated the PR to include an adequate fix for that file as well.

Big believer in [fail fast/fail early](https://stackoverflow.com/questions/2807241/what-does-the-expression-fail-early-mean-and-when-would-you-want-to-do-so/2807375#2807375) however around animation handling doing the correct thing to help a sequence continue/succeed vs. fail is a far better route to take. Consider the comment in the SO post you mentioned:

> Consider this invalid HTML: <div><b>Bold Text!</div> It's missing the </b>, but web browsers choose to do the opposite of "fail early" by trying to recover from the problem and hoping it doesn't make things worse. In their case, it's a sensible route to take. 

The browser vendors do the right thing. Here we want to help an animation sequence/condition continue while still allowing the developer to dive deeper by turning tracing on to gain deeper insight *if they need to*. An appreciated priority would be on ensuring dev's can create animations via CSS keyframes or otherwise without coming to find out some 1% of users find a path that was impossible to find during testing and worse not even fixable by the developer themselves as is the case here since it is coming from the core framework in use.

1% of a 1.5 million user base is **15,000 users** that were adversely affected by the `throw` that this PR effectively solves.

Please let me know if there are other concerns to these changes and I can adjust per your advice.

vakrilov:
test

vakrilov:
test

tns-core-modules/ui/animation/animation-common.ts
```diff
@@ -86,7 +86,12 @@ export abstract class AnimationBase implements AnimationBaseDefinition {
 
     public play(): AnimationPromiseDefinition {
         if (this.isPlaying) {
-            throw new Error("Animation is already playing.");
+            if (traceEnabled()) {
+                traceWrite("Animation is already playing.", traceCategories.Animation, 2);
+            }
+            return <AnimationPromiseDefinition>new Promise<void>((resolve, reject) => {
+                reject();
```
vakrilov: Can you add `Animation is already playing.` as a reason for rejecting.

tns-core-modules/ui/animation/keyframe-animation.ts
```diff
@@ -161,7 +163,12 @@ export class KeyframeAnimation implements KeyframeAnimationDefinition {
 
     public play(view: View): Promise<void> {
         if (this._isPlaying) {
-            throw new Error("Animation is already playing.");
+            if (traceEnabled()) {
+                traceWrite("Keyframe animation is already playing.", traceCategories.Animation, 2);
+            }
+            return new Promise<void>((resolve, reject) => {
+                reject();
```
vakrilov: Can you add `Keyframe animation is already playing` as a reason for rejecting

vakrilov:
Hey thanks for your explanations.
Indeed you are right that for animations a more forgiving API is perhaps more useful. 

I have run the CI tests. Can I ask you to write a couple of unit tests (probably in [`animation-tests.ts`](https://github.com/NativeScript/NativeScript/blob/master/tests/app/ui/animation/animation-tests.ts) assuring this behavior:
 - Calling `play()` second time does not trow but just rejects.
 - Calling `cancel()` does nothing if animation is not playing).

NathanWalker:
test

@vakrilov Changes made and [tests](https://github.com/NativeScript/NativeScript/pull/5475/files#diff-254e142e83e931ee309eeb0742220168R57) added 👍

DickSmith:
Any way we could see this in a hotfix release? This appears to have been a regression in 3.4.x. I had none of these before 3.4, but 3.4 seems to have surfaced this and I'm seeing it at around a ~2% user rate in production. I'd be happy to take an `rc` tagged build for this to roll out to production for confirmation.
https://pasteboard.co/H9YhUDp.png

vakrilov:
test

vakrilov:
@NathanWalker the unit tests you have written seem to fail now on Android (that's what test do).

Indeed that was helpful because seems there is something we've missed. There is a platform-specific implementation for both `play()` and `cancel()` methods for [android](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/animation/animation.android.ts#L137-L178) and [ios](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/animation/animation.ios.ts#L212-L232). The code there will still execute even though the  `super.play()/super.cancel()` calls do not throw.

NathanWalker:
Ok excellent I’ll fix that up tomorrow since I’m going to bed now. If you knew the fix and wanted to pull my branch to finish it up would be fine, otherwise I’ll get in morning.

vakrilov:
https://github.com/NathanWalker/NativeScript/pull/1 ;)

NathanWalker:
Perfection thank you 👍

vakrilov:
test

NathanWalker:
@vakrilov I'm unable to see the jenkins test detail - just lemme know if any other modifications are needed to get those to clear 👍 - thanks again.

SvetoslavTsenov:
test

SvetoslavTsenov:
test branch_animations#master

PR #5489, opened 2018-03-02
chore(docs): Update writing-tests.md with note about code snippets comments

vakrilov:


WritingUnitTests.md
```diff
@@ -82,3 +82,7 @@ export function test_getJSON(done) {
         (error) => { done(error); }); // fail
 };
 ```
+
+# Misc
+
+When looking into the code of the existing tests, you migh encounter strange comments looking like this `// >> animation-chaining`. These are markers for code snippets generated in the docs documetation. They are not related to testing so you don't need to add any of those in your tests.
```
manoldonev: migh -> might typo

PR #5490, opened 2018-03-02
fix(android-images): set decodeHeight/decodeWidth default values to dip

ADjenkov:
Currently `decodeWidth` and `decodeHeight` properties accept values as pixels which can lead to confusion and poor picture quality in some scenarios. 

From now on `decodeWidth` and `decodeHeight` can be used in the following way:
- `decodeHeight="200"` - will be converted to DIP
- `decodeHeight="200dip"` - will be converted to DIP
- `decodeHeight="200px"` - used as exact pixels with no convertion.

BREAKING CHANGE: change decodeHeight/decodeWidth properties from accepting exact pixels to device independent pixels (DIP) by default.


dtopuzov:
test

PR #5495, opened 2018-03-06
fix-next(tabview): visually pre-load tab items on android

MartoYankov:


tns-core-modules/ui/tab-view/tab-view.android.ts
```diff
@@ -535,6 +583,7 @@ export class TabView extends TabViewBase {
         return this._viewPager.getOffscreenPageLimit();
     }
     [androidOffscreenTabLimitProperty.setNative](value: number) {
+        console.log("----> androidOffscreenTabLimitProperty: " + value);
```
ddfreiling: Just noticed when looking at new PRs, probably didn't mean to include this.

MartoYankov: (y) thanks ^^

tns-core-modules/ui/frame/frame.android.ts
```diff
@@ -176,6 +176,21 @@ export class Frame extends FrameBase {
...
 
+    onUnloaded() {
+        if (this._currentEntry && this._currentEntry.fragment) {
+            const manager: android.app.FragmentManager = this._getFragmentManager();
+
+            const transaction = manager.beginTransaction();
+            transaction.remove(this._currentEntry.fragment);
+            transaction.commitAllowingStateLoss();
+
+            this._currentEntry.fragment = null;
```
hshristov: Removing `currentEntry.fragment` while animation is not completed will trigger problems with application restore.

tns-core-modules/ui/frame/frame.android.ts
```diff
@@ -679,6 +692,15 @@ class FragmentCallbacksImplementation implements AndroidFragmentCallbacks {
         }
 
         const entry = this.entry;
+
+        if (!entry) {
+            // On app suspend the Frames that are unloaded destroy their fragments after
```
hshristov: You need to match restored fragment to Page or TabItem otherwise you will have a blank pages in some cases when application is restored.

tns-core-modules/ui/frame/frame.android.ts
```diff
@@ -703,7 +725,7 @@ class FragmentCallbacksImplementation implements AndroidFragmentCallbacks {
         // created before Tab loads its items.
         // TODO: addCheck if the fragment is visible so we don't load pages
         // that are not in the selectedIndex of the Tab!!!!!!
-        if (!page.isLoaded) {
+        if (frame.isLoaded && !page.isLoaded) {
```
hshristov: if `frame.isLoades` is `false` will only make page load later and it will break page transitions (pages will be empty)

tns-core-modules/ui/tab-view/tab-view-common.ts
```diff
@@ -71,9 +71,8 @@ export abstract class TabViewItemBase extends ViewBase implements TabViewItemDef
     public loadView(view: ViewBase): void {
         const tabView = this.parent as TabViewBase;
         if (tabView && tabView.items) {
-            const index = tabView.items.indexOf(this);
             // Don't load items until their fragments are instantiated.
-            if (index === tabView.selectedIndex && (<TabViewItemDefinition>this).canBeLoaded) {
+            if ((<TabViewItemDefinition>this).canBeLoaded) {
```
hshristov: This will load several views (before and after the selectedIndex). The whole point was an improvement so that we fire loaded only for the current selectedIndex.
Firing loaded may trigger another events like navigation, etc...

MartoYankov: Unfortunately, for Android we have to load as much views as the ViewPagerAdapter creates. Otherwise the swipe gesture of the TabView shows blank page before the navigation is completed. CanBeLoaded is assigned by the ViewPagerAdapter instantiate item method. We have also enabled androidOffscreenLimit to be equal to 0, which enables the scenario you talk about. The idea is to have it 0 when the TabView is on the bottom and there is no swipe to navigate.

tns-core-modules/ui/frame/frame.android.ts
MartoYankov: Actually, this fixes an issue with application restore. In the scenario - Tabs - Frames - Pages - all of the Frames fragments are restored before the Tabs fragments are recreated and they try to load their pages before they have been selected. The solution we came up with is to remove the Frame fragment on unload of the frame. A new problem came up - on suspend the Frame is unloaded after saveInstanceState, so the fragments are not really removed according to Android and it tries to recreate them on resume. That is why we set the currentEntry's fragment to null, so that on fragment create we can set it's contents to null, making it a non UI fragment. When the Frame is then loaded it will create a new fragment for the page.

Basically, we want the Android to re-create the Tab fragment, which loads its Frame which creates a new fragment for the Page. I'm open to other ideas.

tns-core-modules/ui/frame/frame.android.ts
MartoYankov: This is connected to the currentEntry fragment assigned to null. It handles this case.

tns-core-modules/ui/frame/frame.android.ts
MartoYankov: Can you share a test case, so that I can check the problem?

tns-core-modules/ui/frame/frame.android.ts
MartoYankov: I couldn't reproduce the problem that nullifying the fragment was supposed to fix, so I removed it :+1:

MartoYankov:
test branch_cuteness#tab-view-fragments

PR #5533, opened 2018-03-13
fix-next(frame): remove current entry fragment on reset root view

vchimev:
Problem:

When we reset root view and minimize/restore application with "Don't keep activities" turned ON, we try to restore the latest fragment. However, it is not available anymore, so could not be restored. In this case, remove it when changing root view in order to create a new one.

Steps to reproduce:

- run the [modal-navigation](https://github.com/NativeScript/NativeScript/tree/vchimev/reset-root-view/e2e/modal-navigation) app
- turn on  the "Don't keep activities" setting in the Developer options for Android
- tap on "Change root view" to set `TabView` as the application root
- minimize and restore application

Stack trace:

```
System.err: Error: Cannot find Frame for NO ENTRY, FragmentClass{f14f72e #0 id=0x1 fragment0[0]}
System.err: File: "file:///data/data/org.nativescript.modalnavigation/files/app/tns_modules/tns-core-modules/ui/frame/frame.js, line: 651, column: 18
System.err: StackTrace:
System.err: 	Frame: function:'FragmentCallbacksImplementation.onCreate', file:'file:///data/data/org.nativescript.modalnavigation/files/app/tns_modules/tns-core-modules/ui/frame/frame.js', line: 547, column: 23
System.err: 	Frame: function:'FragmentClass.onCreate', file:'file:///data/data/org.nativescript.modalnavigation/files/app/tns_modules/tns-core-modules/ui/frame/fragment.js', line: 24, column: 25
System.err: 	Frame: function:'ActivityCallbacksImplementation.onCreate', file:'file:///data/data/org.nativescript.modalnavigation/files/app/tns_modules/tns-core-modules/ui/frame/frame.js', line: 651, column: 19
System.err: 	Frame: function:'NativeScriptActivity.onCreate', file:'file:///data/data/org.nativescript.modalnavigation/files/app/tns_modules/tns-core-modules/ui/frame/activity.js', line: 20, column: 25
```

### DIFF
--- a/tns-core-modules/ui/frame/frame.android.ts
+++ b/tns-core-modules/ui/frame/frame.android.ts
@@ -177,27 +177,22 @@ export class Frame extends FrameBase {
     }
 
     _onRootViewReset(): void {
-        if (this._currentEntry && this._currentEntry.fragment) {
-            const manager: android.app.FragmentManager = this._getFragmentManager();
-
-            const transaction = manager.beginTransaction();
-            transaction.remove(this._currentEntry.fragment);
-            transaction.commitAllowingStateLoss();
-        }
-
+        this.disposeCurrentFragment();
         super._onRootViewReset();
     }
 
     onUnloaded() {
+        this.disposeCurrentFragment();
+        super.onUnloaded();
+    }
+
+    private disposeCurrentFragment(){
         if (this._currentEntry && this._currentEntry.fragment) {
             const manager: android.app.FragmentManager = this._getFragmentManager();
-
             const transaction = manager.beginTransaction();
             transaction.remove(this._currentEntry.fragment);
             transaction.commitAllowingStateLoss();
         }
-
-        super.onUnloaded();
     }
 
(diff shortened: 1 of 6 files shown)


vchimev:
test branch_modal_navigation#vchimev/reset-root-view --ignore

tns-core-modules/ui/frame/frame.android.ts
```diff
@@ -176,6 +176,18 @@ export class Frame extends FrameBase {
         }
     }
 
+    _onRootViewReset(): void {
+        if (this._currentEntry && this._currentEntry.fragment) {
```
manoldonev: Extract the common logic from _onRootViewReset() and onUnloaded() to avoid code duplication e.g. method cleanupCurrentFragment(...) / disposeCurrentFragment(...) or something like this.

manoldonev: Extract the common logic from _onRootViewReset() and onUnloaded() to avoid code duplication e.g. method cleanupCurrentFragment(...) / disposeCurrentFragment(...) or something like this.

Issue #5540, opened 2018-03-15
http service is not working in iOS when invoked from worker thread


slsumathi:
Hi,

   I iOS build the http service is not working, throwing the below error.

    **TypeError: global.moduleMerge is not a function. (In 'global.moduleMerge(httpRequest, exports)', 'global.moduleMerge' is undefined)**

   I noticed this problem when we invoke http service in a worker thread.
   I guess the problem exists even when we run in main thread.

   The fix I did for this problem is.. include the below line in 
   node_modules/tns-core-modules/http/http.js    at the beginning of the file

    require('globals');

vakrilov:
The `global.moduleMerge` method is not needed any more. We prefer to use typescript `export` syntax instead. So instead of 
```
import * as httpRequest from "./http-request";
global.moduleMerge(httpRequest, exports);
```
You can do:
```
import * as httpRequest from "./http-request";
export * "./http-request";
```
The result should be the same. However, the code looks cleaner and there is no dependency in the `global` module.

slsumathi:
Vakrilov,

   When we added tns-core-modules in node_modules we got http.js file not ts file.
   Thanks for the respective fix in typescript.
   BTW this fix needs to go into tns-core-modules.
   Who will be doing that ?

vakrilov:
The `http.js` is generated from the `http.ts`. So the typescript file *is* the correct place for fix.
Once merged, the fix will come out in the `@next` builds and in the next official release

PR #5541, opened 2018-03-15
HTTP service is not working in ios

slsumathi:
Hi,
   I raised the issue #5540 few minutes ago and provided the fix to this issue as well.
   The fix needs to go into http.js file. 

   If you think that this fix is valid then please accept my pull request.

Sumathi


vakrilov:
Hey @slsumathi thanks for the effort.
It looks like this PR has too many changes and it is not clear what is your fix.
Can you please make sure you target the `master` branch and you have rebased all your changes onto the current master.

Also can you check [my comment in the issue you opened.](https://github.com/NativeScript/NativeScript/issues/5540#issuecomment-373357280)

PR #5251, opened 2018-01-10
Fix: Line-height breaks line break mode (iOS)

hettiger:
When setting line-height using a stylesheet, line break mode will stop working.

__Working line break mode:__
https://play.nativescript.org/?template=play-ng&id=3HPeu8

__Now adding a style to home.component.css:__

```css
Label {
    line-height: 1;
}
```

__Results in broken line break mode:__
https://play.nativescript.org/?template=play-ng&id=3HPeu8&v=5

My Pull Request fixes this behavior.

PR #5544, opened 2018-03-15
fix(uilabel): line height setter should not break line break mode

manoldonev:
NOTE: Cherry-picked from https://github.com/NativeScript/NativeScript/pull/5251 due to CI infrastructure issues with the original PR.

When setting line-height using a stylesheet, line break mode will stop working.

__Working line break mode:__
https://play.nativescript.org/?template=play-ng&id=3HPeu8

__Now adding a style to home.component.css:__

```css
Label {
    line-height: 1;
}
```

__Results in broken line break mode:__
https://play.nativescript.org/?template=play-ng&id=3HPeu8&v=5

My Pull Request fixes this behavior.


manoldonev:
👍 for PR approval.

Issue #3818, opened 2017-03-17
[iOS] ListView on initial load not displaying last item


praven07:
I tried to use the ListView and it kinda displays weird it does not size properly.
You can see the problem when you scroll the view.

<img width="383" alt="screen shot 2017-03-17 at 1 39 02 am" src="https://cloud.githubusercontent.com/assets/2223191/24031794/8c6232b6-0ab2-11e7-8208-48c5a33b26ec.png">

<img width="379" alt="screen shot 2017-03-17 at 1 39 40 am" src="https://cloud.githubusercontent.com/assets/2223191/24031810/9ebc4df2-0ab2-11e7-9648-5ddf86b1d1c2.png">

NickIliev:
Hi @praven07, please provide more information regarding your issue
- on which platform this occurs (Android or iOS)
- code snippet used for creating ListView including the CSS styling (orr a sample project to reproduce the issue)

As a possible workaround keep in mind that if you are creating ListView in iOS, you need to wrap in a container with predefined sizes.(meaning that if your ListView is wrapped in a StackLayout with no sizes similar clipping might occur)

praven07:
It's on iOS. I just created a new angular project and I removed the some of the items and just left like 4 items and It creates this problem. If you see the 4th item is hidden so you would have to scroll to view the item. But when you click on an item and navigate to the details page and then come back to the main page it resizes correctly. As you can see in the 3rd screenshot. 

![screen shot 2017-03-17 at 1 41 20 pm](https://cloud.githubusercontent.com/assets/2223191/24058073/b4f97c4c-0b17-11e7-926c-bea47e427c1c.png)

<img width="487" alt="screen shot 2017-03-17 at 1 41 07 pm" src="https://cloud.githubusercontent.com/assets/2223191/24058063/b05dbfd6-0b17-11e7-8780-dbd3feefb597.png">

<img width="487" alt="screen shot 2017-03-17 at 1 46 20 pm" src="https://cloud.githubusercontent.com/assets/2223191/24058185/229177a0-0b18-11e7-919f-e21f2584b6e8.png">

NickIliev:
Hey @praven07 first of all please excuse me for my late answer!
I tested the described case and indeed it seems that on initial load the ListView is not displaying the last loaded item. However, if some navigation is initialized to the list view then the item is loaded.

[...]

PR #5568, opened 2018-03-20
feat(ios-list-view): introduce iosEstimatedRowHeight property.

ADjenkov:
`iosEstimatedRowHeight` property sets [estimatedRowHeight](https://developer.apple.com/documentation/uikit/uitableview/1614925-estimatedrowheight) of the UITableView which forces the ListView to auto-measure its items.

Setting `iosEstimatedRowHeight="0"` can easily resolve [issue](https://github.com/NativeScript/NativeScript/issues/3818)

### DIFF
--- a/tns-core-modules/ui/list-view/list-view.ios.ts
+++ b/tns-core-modules/ui/list-view/list-view.ios.ts
@@ -340,7 +340,7 @@ export class ListView extends ListViewBase {
             return height;
         }
 
-        return this.ios.estimatedRowHeight;
+        return this._ios.estimatedRowHeight;
     }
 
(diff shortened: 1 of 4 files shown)


tns-core-modules/ui/list-view/list-view.ios.ts
```diff
@@ -340,7 +340,7 @@ export class ListView extends ListViewBase {
             return height;
         }
 
-        return DEFAULT_HEIGHT;
+        return this.ios.estimatedRowHeight;
```
manoldonev: It seems within the class we are using this._ios vs this.ios.

PR #5573, opened 2018-03-22
fix(webpack): register wrap layout

radeva:
Currently webpack fails if app is using wrap layout since it is not registered here. Registering it manually in the app works as a workaround, but still I think it would be nice to have it here by default.


manoldonev:
test

PR #5608, opened 2018-04-02
feat(frame): handle back navigation when common layout is used as a root element

ADjenkov:
- Delegate back navigation handling to the topmost Frame, if it's a child of the current View, to prevent closing modal view or app termination.
- Introduce `FrameStack` module to handle `frameStack` access and avoid circular dependencies (`topmost(), _pushInFrameStack()` etc) 
- Add tests where the root modal element is common layout (`GridLayout/StackLayout` eg.)

### DIFF
--- a/tns-core-modules/ui/frame/frame-stack.ts
+++ b/tns-core-modules/ui/frame/frame-stack.ts
@@ -1,0 +1,49 @@
+import { FrameBase } from "./frame-common";
+
+export let frameStack: Array<FrameBase> = [];
+
+export function topmost(): FrameBase {
+    if (frameStack.length > 0) {
+        return frameStack[frameStack.length - 1];
+    }
+
+    return undefined;
+}
+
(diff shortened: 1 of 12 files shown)


ADjenkov:
test

tns-core-modules/ui/frame/frame-stack.ts
```diff
@@ -0,0 +1,49 @@
+import { FrameBase } from "./frame-common";
```
vakrilov: Can you add `// Types` comment on this import to make it clear it should be removed after transpile

SvetoslavTsenov:
test branch_modal_navigation --ignore default

SvetoslavTsenov:
test branch_modal_navigation --ignore default api23 api19

SvetoslavTsenov:
test branch_modal_navigation api19 api23

SvetoslavTsenov:
test branch_modal_navigation#layout-root api19 api23

PR #5612, opened 2018-04-03
fix(style): Styles are not applied to dialogs

vakrilov:
Fix: Global `Button`, `Label` and `Text` selectors with color styling, should be applied to dialogs


SvetoslavTsenov:
test --ignore default branch_uitests ios11

PR #5613, opened 2018-04-03
chore: set keepAspectRatio to true as default in image-asset

tbozhikov:
Currently, ImageAsset instances are created with options.keepAspectRatio = false as default. This PR makes the default true as this is the more expected value.


SvetoslavTsenov:
Hey @tbozhikov,

All test passed and I am ready to merge it in master. 

Just curious if it is a good idea to place those options as parameter in constructor which could be overridden respectivly in constructors of android and iOS assets.

` constructor (private _options: definition.ImageAssetOptions = { keepAspectRatio: true }) {
    super();
  }
 `

PR #5629, opened 2018-04-05
chore(image): expose decode image properties for the public API

ADjenkov:

### DIFF
--- a/tns-core-modules/ui/image/image.d.ts
+++ b/tns-core-modules/ui/image/image.d.ts
@@ -3,7 +3,7 @@
  * @module "ui/image"
  */ /** */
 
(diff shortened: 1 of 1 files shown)


tns-core-modules/ui/image/image.d.ts
```diff
@@ -52,6 +52,18 @@ export class Image extends View {
      * A color used to tint template images.
      */
     tintColor: Color;
+
+    /**
+     * Gets or sets the desired decode height of the image.
+     * This property is Android specific.
+     */
+    decodeHeight: Color;
```
ddfreiling: type Color :thinking:

PR #5668, opened 2018-04-12
feat(builder): enable reading xml from bundle

MartoYankov:
This should enable building pages from xml bundled with webpack. Currently, we copy the xml files even though they are already bundled.


MartoYankov:
test branch_webpack#myankov/xml-bundle branch_testsappng

vchimev:
What test could we have on this?

PR #5708, opened 2018-04-19
fix(android): returnPress invoked twice for some returnKeyTypes

### LINKED ISSUE #4461: ReturnPress is invoked twice on returnKey tap

The ReturnPressEvent is fired twice for text-fields with returnKeyType 'go', 'search' or 'send' on Android phones. 

I've created an [example project](https://github.com/jboeijenga/nativescript-double-return-press-event) to illustrate the problem

The issue is easily reproducible by starting up the app and pressing the different return keys. Every press should increase the count with 1 but for the return keys with type 'go', 'search' or 'send the count is increased with 2. 

There is a [closed issue](https://github.com/NativeScript/NativeScript/issues/1940) I've found because there was no ability to reproduce this issue.
Also it seems like there are [problem](https://github.com/facebook/react-native/issues/10443) for people using react native. So the issue might be more generic than the nativescript implementation

Android

Run the example project I've created.

EddyVerbruggen:
## PR Checklist

A `returnKeyType` like `IME_ACTION_GO` also fires an `IME_ACTION_UNSPECIFIED` actionId with `event.getKeyCode() === KEYCODE_ENTER` while navigating to the next field. This will trigger the `returnPress` event twice.

## What is the new behavior?
Check for the event not being `IME_ACTION_UNSPECIFIED` when evaluating the keycode.

Fixes #4461.


tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
```diff
@@ -121,7 +121,7 @@ function initializeEditTextListeners(): void {
                 actionId === android.view.inputmethod.EditorInfo.IME_ACTION_GO ||
                 actionId === android.view.inputmethod.EditorInfo.IME_ACTION_SEARCH ||
                 actionId === android.view.inputmethod.EditorInfo.IME_ACTION_SEND ||
-                (event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
+                (actionId !== android.view.inputmethod.EditorInfo.IME_ACTION_UNSPECIFIED && event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
```
manoldonev: hey @EddyVerbruggen -- I might be missing something but I don't think the proposed addition changes the logic of the conditional statement. If `actionId` was `IME_ACTION_UNSPECIFIED` the conditional would have been truthy because of the 2nd row in the statement so if we progressed till the end we already know it is not `IME_ACTION_UNSPECIFIED` and we do not need to check for it explicitly:
```
            if (actionId === android.view.inputmethod.EditorInfo.IME_NULL ||
                actionId === android.view.inputmethod.EditorInfo.IME_ACTION_UNSPECIFIED ||
                actionId === android.view.inputmethod.EditorInfo.IME_ACTION_DONE ||
                actionId === android.view.inputmethod.EditorInfo.IME_ACTION_GO ||
                actionId === android.view.inputmethod.EditorInfo.IME_ACTION_SEARCH ||
                actionId === android.view.inputmethod.EditorInfo.IME_ACTION_SEND ||
                (actionId !== android.view.inputmethod.EditorInfo.IME_ACTION_UNSPECIFIED && event && event.getKeyCode() === android.view.KeyEvent.KEYCODE_ENTER)) {
                // If it is TextField, close the keyboard. If it is TextView, do not close it since the TextView is multiline
                // https://github.com/NativeScript/NativeScript/issues/3111
                if (textView.getMaxLines() === 1) {
                    owner.dismissSoftInput();
                }

                owner._onReturnPress();
            }
```

EddyVerbruggen: @manoldonev That's a very valid point. I tested this change with the 3.4 version of code modules where the `actionId === android.view.inputmethod.EditorInfo.IME_NULL || actionId === android.view.inputmethod.EditorInfo.IME_ACTION_UNSPECIFIED ||` lines weren't present (both those static properties resolve to `0` btw) so that worked as expected. In the 4.0 branch this change doesn't make sense anymore as you pointed out.

EddyVerbruggen: @manoldonev I wonder why we need those first 2 lines anyway, because there doesn't seem to be a `returnPress` event being fired on a `TextView` anyway.

I mean, [this is the implementation on `TextField`](https://github.com/NativeScript/NativeScript/blob/6bdb5b553fb06e8a9c2d7b6617569990c5eed072/tns-core-modules/ui/text-field/text-field.android.ts#L13-L15), and there's [no implemenation on `TextView`](https://github.com/NativeScript/NativeScript/blob/6bdb5b553fb06e8a9c2d7b6617569990c5eed072/tns-core-modules/ui/text-view/text-view.android.ts), and the superclass implemenation [is empty](https://github.com/NativeScript/NativeScript/blob/6bdb5b553fb06e8a9c2d7b6617569990c5eed072/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts#L160-L162).

So I *think* we could remove those 2 lines from the `if` statement and this will work. Thoughts?

manoldonev: @EddyVerbruggen @hshristov added [those lines](https://github.com/NativeScript/NativeScript/commit/8df8d754abd548f8c161c4326eb96f26c3a3604b) to fix https://github.com/NativeScript/NativeScript/issues/5121 (new lines with Japanese IME).

I wonder now whether we can remove the IME_ACTION_GO, IME_ACTION_SEARCH, and IME_ACTION_SEND explicit checks as it seems each of them gets a complimentary IME_ACTION_UNSPECIFIED that we will handle -- have to check this.

manoldonev: New PR proposal for this issue here: https://github.com/NativeScript/NativeScript/pull/5727

manoldonev:
Closing this in favor of https://github.com/NativeScript/NativeScript/pull/5727

Issue #5735, opened 2018-04-24
Opening modal view from ViewBase instance does not work


manoldonev:
It should be possible to open a modal view from a ViewBase instance (e.g. TabViewItem) via the showModal(...) method if the correct set of arguments is passed and as long the ViewBase instance has a parent that is actual view (for TabViewItem -> TabView). Currently this only works for View instances.

If no arguments are passed, it is expected that ViewBase.showModal() or View.showModal will throw an exception.

PR #5737, opened 2018-04-24
fix(modal): exception when calling ViewBase.showModal(...)

manoldonev:
Fixes https://github.com/NativeScript/NativeScript/issues/5735


tests/app/ui/page/page-tests-common.ts
```diff
@@ -421,6 +421,110 @@ export function test_WhenPageIsNavigatedToFrameCurrentPageIsNowTheSameAsThePage(
...
+
+        return masterPage;
+    };
+
+    helper.navigate(masterPageFactory);
+
+    TKUnit.waitUntilReady(() => modalClosed);
+}
+
+export function test_WhenViewBaseCallsShowModal_WithArguments_ShouldThrow() {
```
MartoYankov: Should be without arguments.

PR #5747, opened 2018-04-26
chore(tslint): fix tslint config & errors

manoldonev:
Fix tslint.json syntax for some rules and the tslint errors uncovered by that fix.


build/tslint.json
```diff
@@ -3,56 +3,83 @@
...
+        "one-line": [
+            false,
             "check-open-brace",
             "check-catch",
             "check-else",
             "check-whitespace"
         ],
-        "quotemark": [false, "double"],
+        "quotemark": [
+            false,
```
sis0k0: Shouldn't this be `true` if we want to force `double` quotes?

manoldonev: @sis0k0 I did not disable this rule originally but it is now enabled and the errors fixed :).

manoldonev:
@sis0k0 enabled additionally `arrow-return-shorthand` and `whitespace` rules.

manoldonev:
test

Issue #5622, opened 2018-04-04
Removing all items from ObservableArray


pedromorgan:
I've got an app where I need to remove all items in an ObservableArray<any>, but using the same object.

I was looking for a `.clear()` or `.removeAll()` methods, but can only see `.pop()`.

My workaround is 
```ts
while(myarr.length > 0){
       myarr.pop();
}
```
- What is the correct way ?
- Can a ```.clear()``` method be added, similar to dozens of other api's ?

tsonevn:
Duplicate of #5515

Issue #5515, opened 2018-03-08
ObservableArray add .clear() or .removeAll()


LorenDorez:
I was surprised to find ObservableArry doesnt have a clear or removeAll method. 

Here is my usage senario
I am using OA to build a list view and add search functionality to it. So i carry and originalValues array and a values array. I the the originalValues array upon search and then load this into the values array. If i override the values array like values = filteredItems then this breaks the data-binding and the listview isnt updated. If i do values.push(filteredItems) then the listview is updated but the old values are not removed.

For now i call
while(values.length){
            values.pop();
        }

But would be nice to have a clear or remove all method to handle this

tsonevn:
Hi @LorenDorez,
Thank you for your suggestion. I logged this issue as a new feature request. For further info, please keep track on the issue.

MartoYankov:
@LorenDorez
The ObservableArray is designed to follow the API of the JS Array. The specs do not include a clear() or removeAll() method. If you want to empty the array you should use  `arrayInstance.length = 0` or `arrayInstance.splice(0)`. They do the same thing. You should also be able to use `splice()` to replace all or some items of an array with new items. See the example at [MDN](https://developer.mozilla.org/enUS/docs/Web/JavaScript/Reference/Global_Objects/Array/splice).

LorenDorez:
@MartoYankov 
Thanks the link is broken but i managed to find it. The splice(0) is much cleaner so I switched to that. I cant use it to replace items as they might not be in the same order. I understand with wanting to keep it to eh JS standard just thought of it as a suggestion,

VladsGG:
I'm trying to use the OservableArray.clear(); but it puts out an error that it is not a function

NickIliev:
@VladsGG there is no `clear` method for the reasons discussed above

VladsGG:
Is there any alternative code to clear all items in array?

NickIliev:
@VladsGG [the answer is posted in the very same thread two posts above your question...](https://github.com/NativeScript/NativeScript/issues/5515#issuecomment-372622094)

PR #5762, opened 2018-04-28
ObservableArray.clear() method

pedromorgan:
A convenience function (as xplatform dev, an obvious missing one

See https://github.com/NativeScript/NativeScript/issues/5515 and my dupe https://github.com/NativeScript/NativeScript/issues/5622

### DIFF
--- a/tns-core-modules/data/observable-array/observable-array.ts
+++ b/tns-core-modules/data/observable-array/observable-array.ts
@@ -188,6 +188,12 @@ export class ObservableArray<T> extends observable.Observable implements observa
     sort(compareFn?: (a: T, b: T) => number): T[] {
         return this._array.sort(compareFn);
     }
+    /**
+     * Remove all items in the array
+     */
+    clear() {
+        this.splice(0);
+    }
 
     /**
      * Removes elements from an array and, if necessary, inserts new elements in their place, returning the deleted elements.


manoldonev:
Hey @pedromorgan. We appreciate that you took the time to make the PR. However, our policy forbids us form merging it (and your previous PRs) before you sign the CLA at http://www.nativescript.org/cla

Note: You must sign the CLA with the e-mail used in your commits. As I see in your commit info [here](https://github.com/NativeScript/NativeScript/commit/37b33b40899beb4557d04c119ee652586ec8a2ba.patch) it is pedromorgan@users.noreply.github.com.

manoldonev:
Also, can you add relevant unit tests for the new functionality here: https://github.com/NativeScript/NativeScript/blob/master/tests/app/data/observable-array-tests.ts?

MartoYankov:
This looks like a syntactic sugar you can add in your project. Simply extend the ObservableArray class and add it. I don't think it should be in the framework. As I mentioned in https://github.com/NativeScript/NativeScript/issues/5515 this method is not in the JS Array spec and it's not needed. We have already implemented all the methods you need to clear an array in 3 different ways - splice, pop and length. The main problem I see is that the JS Array doesn't have it, so if you change an ObservableArray to the basic Array, your code won't work.

Issue #5785, opened 2018-05-05
[FR] Switch to AppCompatActivity


farfromrefug:
Hi,

I was trying to implement shared elements animations using this tuto:
https://android-developers.googleblog.com/2018/02/continuous-shared-element-transitions.html

To do that we need to use AppCompatActivity, compat Fragment ....
For that the whole Nativescript framework needs to switch to support v4 classes (frame, activity, fragment, fragment.transition).
I started to work on that but it's not easy because the FragmentManager does not have a loadAnimator but a loadAnimation. Which means it does not use animators but animations.

Also this will be a requirement as the android.app.FragmentManager is deprecated in P
:https://developer.android.com/reference/android/app/FragmentManager

If you are ok with that i can make a PR.

farfromrefug:
I have been looking deeper at this and started working on that.
The only real issue is that the support fragmentmanager uses animations and not animators whic makes the current code not compatible.
I al still tryingb to figure out why we need have those defaults animator which are cloned for every fragment.

farfromrefug:
I actually got it working. And shared elements got working with it too.
![may-06-2018 18-05-18](https://user-images.githubusercontent.com/655344/39675227-42e16f60-5158-11e8-8efc-8da2013c095a.gif)
I created a PR https://github.com/NativeScript/NativeScript/pull/5789

PR #4951, opened 2017-10-16
Fix clear history transition

hshristov:
Fix https://github.com/NativeScript/NativeScript/issues/4950
Implemented custom fragment save/restore state.
When navigating back we reverse manually transitions/animations because we no longer add them to navite backstack.
Fragment instance stored on entry.
Animation and Transition listeners now holds reference to entry instead of fragment for easier update of fragment.
Animation and Transition listeners removed when entry removed from backstack.
Animation and Transition removed from fragment when fragment activity is destroyed.

Issue #4908, opened 2017-09-30
Built-in Android CoordinatorLayout support for NativeScript | Hide actionbar while scrolling


shiv19:
Modern android apps, have the parallax scrolling behavior, where the action bar fades away when scrolling through the page, and it provides very good UX. As of now, there is no easy way to do this using NativeScript.

I have tried the nativescript-parallax plugin, but it seems to be broken as of now.

Please consider this feature request for the next release.

tsonevn:
Hi @shiv19,
Thank you for contacting us,
Indeed parallax scrolling behavior provides very good UX and this is the reason why the plugin was created.
I tested the NPM version of the plugin with inside a NativeScirpt application and indeed it will lead to a crash. In my further investigation, I found that the version in the master branch of `nativescript-parallax` works as expected on both iOS and Android. In this case, I would suggest building a local tgz file of the plugin and to use it in your project.
For your help, I also attach a zip file with an already built tgz file.
[nativescript-parallax-1.4.0.zip](https://github.com/NativeScript/NativeScript/files/1348812/nativescript-parallax-1.4.0.zip)

shiv19:
@tsonevn 
Hi Nikolay, I was able to try the plugin thanks to you. 
Hope to see better support for CoordinatorLayout in NativeScript.
Would like to see it as a Layout Type in NativeScript rather than a plugin.

Because internally, native-script parallax plugin is using scroll-view, grid-layout, absolute-layout and stack-layout to pull it off. Due to which the performance is not pure native. If you scroll fast enough, you can see glitches (white lines on the screen).

tsonevn:
Hi @shiv19,

Thank you for the additional info.
I have to agree that this would be a useful feature, regarding that I marked this as a feature request.
You could keep track on this issue for further info. If there is interest for similar functionality, this feature could be expected for some of the next versions of NativeScript.

ashisherc:
This is going to allow pure native looking apps in android. I hope to see the same effect in iOS as well.
:+1:

farfromrefug:
Actually this might be tough to achieve within the sdk itself I think. However material-components might be the way to go. Plus it will work on iOS !

tigrenok00:
+1 for the feature request

vishwa1937:
We should add this and other native feature of android and iOS into nativescript.

shiv19:
@vishwa1937 this will be possible after NativeScript 5.0 is released this month ^_^

[...]

PR #5789, opened 2018-05-06
AppCompatActivity and shared element

farfromrefug:
Follows issue: https://github.com/NativeScript/NativeScript/issues/5785
- switch to AppCompatActivity
- added “transitionName” property
- added prepareFragment optional function for navigate
- allow access to _fragment for page
- use ```manager.popBackStack();```

I should mention that switching to AppCompat actually changed to default fragment enter animation. With support it is a "push fade" animation (quite fast one). Without it was a fade animation.

For now i have let some commented code so that you can review it.
I am also not very good with tests. Let me know what tests you want me to add.

### DIFF
--- a/tns-core-modules/application/application.android.ts
+++ b/tns-core-modules/application/application.android.ts
@@ -42,8 +42,8 @@ export class AndroidApplication extends Observable implements AndroidApplication
     public paused: boolean;
     public nativeApp: android.app.Application;
     public context: android.content.Context;
-    public foregroundActivity: android.app.Activity;
-    public startActivity: android.app.Activity;
+    public foregroundActivity: android.support.v7.app.AppCompatActivity;
+    public startActivity: android.support.v7.app.AppCompatActivity;
     public packageName: string;
 
     public get currentContext(): android.content.Context {
@@ -218,7 +218,7 @@ global.__onLiveSync = function () {
 };
 
 function initLifecycleCallbacks() {
-    const setThemeOnLaunch = profile("setThemeOnLaunch", (activity: android.app.Activity) => {
+    const setThemeOnLaunch = profile("setThemeOnLaunch", (activity: android.support.v7.app.AppCompatActivity) => {
         // Set app theme after launch screen was used during startup
         const activityInfo = activity.getPackageManager().getActivityInfo(activity.getComponentName(), android.content.pm.PackageManager.GET_META_DATA);
         if (activityInfo.metaData) {
@@ -229,11 +229,11 @@ function initLifecycleCallbacks() {
         }
     });
 
-    const notifyActivityCreated = profile("notifyActivityCreated", function (activity: android.app.Activity, savedInstanceState: android.os.Bundle) {
+    const notifyActivityCreated = profile("notifyActivityCreated", function (activity: android.support.v7.app.AppCompatActivity, savedInstanceState: android.os.Bundle) {
         androidApp.notify(<AndroidActivityBundleEventData>{ eventName: ActivityCreated, object: androidApp, activity, bundle: savedInstanceState });
     });
 
-    const subscribeForGlobalLayout = profile("subscribeForGlobalLayout", function (activity: android.app.Activity) {
+    const subscribeForGlobalLayout = profile("subscribeForGlobalLayout", function (activity: android.support.v7.app.AppCompatActivity) {
         const rootView = activity.getWindow().getDecorView().getRootView();
         let onGlobalLayoutListener = new android.view.ViewTreeObserver.OnGlobalLayoutListener({
             onGlobalLayout() {
@@ -246,7 +246,7 @@ function initLifecycleCallbacks() {
     });
 
     const lifecycleCallbacks = new android.app.Application.ActivityLifecycleCallbacks({
-        onActivityCreated: profile("onActivityCreated", function (activity: android.app.Activity, savedInstanceState: android.os.Bundle) {
+        onActivityCreated: profile("onActivityCreated", function (activity: android.support.v7.app.AppCompatActivity, savedInstanceState: android.os.Bundle) {
             setThemeOnLaunch(activity);
 
             if (!androidApp.startActivity) {
@@ -260,7 +260,7 @@ function initLifecycleCallbacks() {
             }
         }),
 
-        onActivityDestroyed: profile("onActivityDestroyed", function (activity: android.app.Activity) {
+        onActivityDestroyed: profile("onActivityDestroyed", function (activity: android.support.v7.app.AppCompatActivity) {
             if (activity === androidApp.foregroundActivity) {
                 androidApp.foregroundActivity = undefined;
             }
@@ -274,7 +274,7 @@ function initLifecycleCallbacks() {
             gc();
         }),
 
-        onActivityPaused: profile("onActivityPaused", function (activity: android.app.Activity) {
+        onActivityPaused: profile("onActivityPaused", function (activity: android.support.v7.app.AppCompatActivity) {
             if ((<any>activity).isNativeScriptActivity) {
                 androidApp.paused = true;
                 notify(<ApplicationEventData>{ eventName: suspendEvent, object: androidApp, android: activity });
@@ -283,7 +283,7 @@ function initLifecycleCallbacks() {
             androidApp.notify(<AndroidActivityEventData>{ eventName: ActivityPaused, object: androidApp, activity: activity });
         }),
 
-        onActivityResumed: profile("onActivityResumed", function (activity: android.app.Activity) {
+        onActivityResumed: profile("onActivityResumed", function (activity: android.support.v7.app.AppCompatActivity) {
             androidApp.foregroundActivity = activity;
 
             if ((<any>activity).isNativeScriptActivity) {
@@ -294,15 +294,15 @@ function initLifecycleCallbacks() {
             androidApp.notify(<AndroidActivityEventData>{ eventName: ActivityResumed, object: androidApp, activity: activity });
         }),
 
-        onActivitySaveInstanceState: profile("onActivityResumed", function (activity: android.app.Activity, outState: android.os.Bundle) {
+        onActivitySaveInstanceState: profile("onActivityResumed", function (activity: android.support.v7.app.AppCompatActivity, outState: android.os.Bundle) {
             androidApp.notify(<AndroidActivityBundleEventData>{ eventName: SaveActivityState, object: androidApp, activity: activity, bundle: outState });
         }),
 
-        onActivityStarted: profile("onActivityStarted", function (activity: android.app.Activity) {
+        onActivityStarted: profile("onActivityStarted", function (activity: android.support.v7.app.AppCompatActivity) {
             androidApp.notify(<AndroidActivityEventData>{ eventName: ActivityStarted, object: androidApp, activity: activity });
         }),
 
-        onActivityStopped: profile("onActivityStopped", function (activity: android.app.Activity) {
+        onActivityStopped: profile("onActivityStopped", function (activity: android.support.v7.app.AppCompatActivity) {
             androidApp.notify(<AndroidActivityEventData>{ eventName: ActivityStopped, object: androidApp, activity: activity });
         })
     });
--- a/tns-core-modules/ui/core/view/view-common.ts
+++ b/tns-core-modules/ui/core/view/view-common.ts
@@ -568,16 +568,10 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
         this.style.scaleY = value;
     }
 
-    get transitionName(): string {
-        return this.style.transitionName;
-    }
-    set transitionName(value: string) {
-        this.style.transitionName = value;
-    }
-
     //END Style property shortcuts
 
     public automationText: string;
+    public transitionName: string;
     public originX: number;
     public originY: number;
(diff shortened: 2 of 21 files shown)


tns-core-modules/ui/core/view/view-common.ts
```diff
@@ -568,6 +568,13 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
         this.style.scaleY = value;
     }
 
+    get transitionName(): string {
+        return this.style.transitionName;
```
manoldonev: I don't think `transitionName` should be registered on Style but on ViewCommon. Check `automationText`: 
https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/core/view/view-common.ts#L1000

tns-core-modules/ui/frame/frame.android.ts
```diff
@@ -354,26 +357,27 @@ export class Frame extends FrameBase {
...
-        if (!backstackEntry.fragment) {
-            // Happens on newer API levels. On older all fragments
-            // are recreated once activity is created.
-            // This entry fragment was destroyed by app suspend.
-            // We need to recreate its animations and then reverse it.
-            backstackEntry.fragment = this.createFragment(backstackEntry, backstackEntry.fragmentTag);
-            _updateTransitions(backstackEntry);
-        }
+        const manager: android.support.v4.app.FragmentManager = this._getFragmentManager();
+        manager.popBackStack();
```
manoldonev: Currently we are not using the backstack functionality of FragmentManager but we are managing the fragments on our own. Could you elaborate a bit why are you proposing this change?

manoldonev:
@farfromrefug btw as far as I can see even with support libraries shared element transitions are supposed to work on Android API Levels 21 or higher, right?

Also, this is obviously Android only and ideally we would like to provide cross-platform support for iOS and Android. I did a quick search and there are iOS options ([here](https://www.cocoacontrols.com/controls/iossharedviewtransition) and [here](https://www.raywenderlich.com/170144/custom-uiviewcontroller-transitions-getting-started)) or cross-platform solutions like [this one](https://medium.freecodecamp.org/shared-element-transition-with-react-native-159f8bc37f50) that we might need to consider as alternatives as well.

tns-core-modules/ui/frame/frame.android.ts
farfromrefug: This one is quite important because otherwise shared element animation wont work. Or you need to do it manually. I think you did not use it to get listeners and be notified when transition finished. Which still works in this PR

tns-core-modules/ui/core/view/view-common.ts
farfromrefug: Great i am all new to nativescript. Will change that.

tns-core-modules/application/application.android.ts
```diff
@@ -42,8 +42,8 @@ export class AndroidApplication extends Observable implements AndroidApplication
     public paused: boolean;
     public nativeApp: android.app.Application;
     public context: android.content.Context;
-    public foregroundActivity: android.app.Activity;
-    public startActivity: android.app.Activity;
+    public foregroundActivity: android.support.v7.app.AppCompatActivity;
```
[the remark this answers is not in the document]
farfromrefug: Actually i made a all replace. I think you ate right in this case it should still be an activity as you can start another activity which would not be an AppCompatActivity

farfromrefug:
@manoldonev not in the support library. Was added in support 22.0.1 i think
Now about the iOS part that s a lot trickier
In my last framework i did noy manage to do it. So it was kind of an android thing. But now that you talk about it that would still be doable. For me the tricky thing is not technical as already do it, it is more with the expectation of the default transition. The default ios transition is the modern push in the navigation controller which would not work with shared elements (i have not tried with the lib you shared).
I will try and report

manoldonev:
test

manoldonev:
@farfromrefug btw we are actually working on a sample custom animation showcase that mimics shared element transitions and works on iOS and Android: https://play.nativescript.org/?template=play-ng&id=iDY4kF&v=10 (work in progress). Would this approach work for your scenario as well?

tns-core-modules/application/application.android.ts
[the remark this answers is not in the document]
manoldonev: I am still trying to get my head around why we need the AppCompat* API switch at all. Can you elaborate?

tns-core-modules/ui/frame/frame.android.ts
manoldonev: I guess we should explore doing this manually as I don't see us replacing our back stack management logic at this point (or adding the FragmentManager's one on top of it).

farfromrefug:
@manoldonev will look at the test error tomorrow morning!
Thanks for looking at this.

About your example it might work. Right now i see it does not handle different size from one view to the other? Also you don't animate the fragment transition itself right? Would it work with explode or fade transition? It would need to.
About the iOs counterpart, i really think i see a way to do it and i might have a look at it soon.

Thanks

tns-core-modules/application/application.android.ts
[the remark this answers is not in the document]
farfromrefug: @manoldonev You need to compat switch to get access to many methods, like ```startPostponedEnterTransition``` on fragments and activities.
You also need ```supportFinishAfterTransition``` to make sure transitions are applied when closing an activity.
Also i really want to point out that in P, they deprecate the default FragmentManager so you need to start using the support library anyway.
On a side note it wont make apps bigger as your already include it.

tns-core-modules/ui/frame/frame.android.ts
farfromrefug: @manoldonev i don't see how it breaks your back stack management? You still get the transition end event, and your stack is updated.

farfromrefug:
@manoldonev found a great library to do that on iOS. It's actually based on android transitions.
https://github.com/lkzhao/Hero
What's the rule about using native external lib within the nativescript core?

manoldonev:
test

manoldonev:
@farfromrefug about the iOS library -- we do not have a precedent for this and at this point we would prefer not to reference native external lib within NativeScript core. For one thing we would need to get much better understanding of this lib and its potential implications / effects, also adding it and not using it for anything else besides shared element transitions does not seem like the best strategy.

As for the Android part -- we had a discussion internally and we agreed that we would need to address Android P support first with highest priority due to its deprecations and breaking changes before even considering how to proceed with this PR (or alternative solution that adds support for cross-platform shared element transitions) as the proposed changes tamper with some core Android internals in NativeScript core (as discussed above one of the things we are concerned about is the backstack management changes and would not like to revert to a situation that https://github.com/NativeScript/NativeScript/pull/4951 addressed several months ago).

Btw, do you think it is feasible to extract the changes you are proposing in a NativeScript plugin? If we can agree on an approach and API to expose maybe that would be the best option for the time being (note that exposing the necessary API might not be that easy as currently this platform-specific implementation is intentionally hidden but at least we can try to discuss it).

farfromrefug:
@manoldonev i do understand. I don't think it can be made as a plugin because it requires to change the "extends" of the basic Nativescript Activity. At least i think it's not possible. 
Especially when we can't use generics in extends.

Now in my opinion and seeing the very limited amount of changes in the code, it could be part of the Android P support. Especially when you need to change the transition frame code because of the Android P security. If i am correct the way to get the default transition won't work anymore (same for my code). 
What i mean is that i think it could be part of the Android P support sprint.

But whatever you decide, i will understand.

Thanks

NathanaelA:
@manoldonev - Manol; can you tell me what the status of this is, I honestly can't believe this wasn't jumped on getting tested and added into the framework.  It fixes a lot of outstanding issues that we have had for a while now.    Even if this was a breaking change; it would be worth releasing NS 5.0 for this alone, imho.  :grinning:

Getting support for `AppCompatActivity` is actually EXTREMELY important for the ability to support Material Design on Android properly.   I've been trying to add several Material Design components to NativeScript and currently as it stands it is IMPOSSIBLE to add the controls to NativeScript because NativeScript Activity does not extend from the `AppCompatActivity`.   

MaterialDesign has been something that Progress/Telerik has wanted to implement in NativeScript for a while.   And this is **THE** key piece that is blocking it.  `AppCompatActivity` is the fundamental piece to making it all work.  

 In Addition, if you are planning on Android P support, Martin is correct the Framework "Fragment" as it stands is [depreciated in api 28](https://developer.android.com/reference/android/app/Fragment) because it is actually [buggy](https://plus.google.com/+IanLake/posts/WAGQiG7LaKs) across multiple versions of the OS and has not seen any fixes in a long time.     It has been the recommendation of Google devs for Fragments for a while now that you **NEED** to be using the Support based fragments as they are maintained and tied to [life cycle events](https://developer.android.com/topic/libraries/support-library/revisions#26-1-0) and are consistent across all versions of Android.    Also, the standard Activity does not implement many of the features of the [AppCompatActivity](https://developer.android.com/reference/android/support/v7/app/AppCompatActivity)

Btw, Even if iOS doesn't have the same animations, this is worth getting into play for what it brings to the table for Android and all the plugins it will then allow us to build once the AppCompatActivity is the main activity.    I also know I can create make my own apps and extend the AppCompatActivity, but this should be part of the core framework, that way standard Material design plugins can be created that work for everyone... 

CC: @vakrilov @vchimev @radeva

danielgek:
I agree with @NathanaelA, also consider this a major feat that nativescript needs!!

And there is [androidX](https://android-developers.googleblog.com/2018/05/hello-world-androidx.html) now, that brings among other things, support for the [material-components-android](https://github.com/material-components/material-components-android)(also available under the support.design lib), been using some of these components but it take to much to make it work.

This with android P would be great!

manoldonev:
@NathanaelA this PR is currently on-hold as we felt it required significant effort on our side to complete at the time. As for AppCompatActivity / deprecated native fragments -- we will revisit those building blocks separately from the shared element transition support in the coming weeks and we'll agree upon a course of action. Will update this thread when we have more info.

@danielgek are you referring to AppCompatActivity or the shared element transition support that this PR originally targets?

NathanaelA:
@manoldonev - I just was sent a email from a Google engineer (as follow up to the Android P issue we had a while ago) telling me that the framework fragment support (i.e. part of what broke in NS for Android P, that they had to Greylist) is highly recommended we ditch sooner than later.   He mentioned that the reason it has been deprecated is because it (Framework fragments) has not seen any fixes for a long while and won't see any fixes in the future, they are locked in their buggy states.   That for stable applications under recent versions of Android, we must switch to the support fragments, and eliminate any framework fragments.   Apparently the framework fragments are actually pretty brittle in several cases, especially with the newer OS's because some lifecycle events aren't triggered on framework fragments.  ;-(    

So I would recommend this becomes a higher priority item as the sooner we transition NS to the support fragments, the less likely released NS apps will break (or exhibit buggy behavior) in later versions of Android.   

Btw: Definition of Terms (just in case it isn't clear to people):
- Framework Fragments are those fragments that are defined in the base OS Framework; i.e. `android.app.Fragment` and descendants of that tree.   (Which is what NS is using)
- Support Fragments are those fragments that are defined by using `android.support.v4.app.Fragment` and descendants of that tree.  

 


`

farfromrefug:
@manoldonev What's the effort you are talking about? is that testing? Cause on the implementation side it seems like there is nothing else to do

manoldonev:
@farfromrefug testing is part of the effort too, yes (tests did not pass with this PR at any point in time) but as discussed above the proposed changes tamper with some core Android internals in NativeScript core and our main concern is the backstack management changes as we would not like to revert to a situation that #4951 addressed several months ago.

To summarize -- this feature is on our radar but we will handle breaking changes and deprecations like AppCompatActivity and native fragments first and then we will go back to the task of adding shared element transitions.

danielgek:
@manoldonev was talking about `AppCompatActivity ` !

shiv19:
Until we get the native shared elements animation support added to {N},
it is possible to implement something close to shared element transition,
that works on both Android and iOS, using existing {N} APIs
I've demonstrated it in this repo,
https://github.com/shiv19/nativescript-set-demo

shiv19:
@demetrio812
I totally agree with you!
I'm not happy with that implementation either, but that was a close as I could get to it back then. Maybe I can make a better solution now. But then, nothing beats native shared elements support. 

Another important support that is required is this:
https://github.com/NativeScript/NativeScript/issues/4908

Not sure if it is on their radar.

demetrio812:
Hello @shiv19 , thanks for pointing out the other issue, yes I think we really need to be able to create nice looking apps that follow the design guidelines from each OS to be really confident that ANY app can be built with NS, I'm really looking forward to it!

manoldonev:
Migration to support library APIs PR https://github.com/NativeScript/NativeScript/pull/6129 is now merged in the master branch of the NativeScript repo. Can be tested via tns-core-modules@next.

demetrio812:
@manoldonev @radeva @shiv19 now that version 6 is out, can you have another look at the shared elements? In general it would be cool to have a general solution that would work on iOS too as it's really needed to step up the UI capabilities of the apps written in Nativescript, especially after how much Flutter is focusing its efforts in UI quality and animations

Thanks a lot

demetrio812:
I'd like to add that in case you don't want to include them inside the NS Core, the new material components could be an opportunity to address it in a multiplatform way but keeping them in an external library

PR #5807, opened 2018-05-11
chore(e2e): clean up webpack files

vchimev:


.gitignore
```diff
@@ -52,7 +52,7 @@ apps/hooks/
 apps/node_modules/
 
 e2e/**/platforms
-!e2e/**/webpack.config.js
```
MartoYankov: Consider adding the vendor files to the ignore too. Also, might be good to unify the gitignores if possible. See the angular repo https://github.com/NativeScript/nativescript-angular/blob/master/.gitignore

.gitignore
```diff
@@ -1,15 +1,52 @@
...
+!.vscode/extensions.json
+
+# Appium files
+e2e/**/*.trace/
+e2e/**/test-results.xml
+
+# Webpack configuration files
+webpack.config.js
+vendor.ts
+vendor-platform.android.ts
```
sis0k0: the vendor* files are not needed anymore - the webpack plugin won't add them to the app

e2e/modal-navigation/package.json
```diff
@@ -17,29 +17,12 @@
...
-    "css-loader": "~0.28.7",
-    "extract-text-webpack-plugin": "~3.0.2",
-    "lazy": "1.0.11",
     "nativescript-dev-appium": "next",
     "nativescript-dev-typescript": "next",
     "nativescript-dev-webpack": "next",
-    "nativescript-worker-loader": "~0.8.1",
-    "raw-loader": "~0.5.1",
-    "resolve-url-loader": "~2.2.1",
-    "typescript": "~2.4.2",
```
sis0k0: I think we need that.

vchimev: These `devDependencies` will be added on the `npm install` execution, coming from the `nativescript-dev-*` plugins. I'd like to avoid executing `node_modules/.bin/update-ns-webpack --deps` in future if these present in the `package.json`.

vchimev:
test

PR #5823, opened 2018-05-16
keep some variable from early GC to speed up start time

vtrifonov:
To avoid early GC which will trigger MarkReachableObjects on startup I'm storing some callback locally which are cleared once they are not needed anymore.

### DIFF
--- a/tns-core-modules/application/application.android.ts
+++ b/tns-core-modules/application/application.android.ts
@@ -45,6 +45,8 @@ export class AndroidApplication extends Observable implements AndroidApplication
     public foregroundActivity: android.app.Activity;
     public startActivity: android.app.Activity;
     public packageName: string;
+    // we are using these property to store the callbacks to avoid early GC collection which would trigger MarkReachableObjects
+    private callbacks: any = {};
 
     public get currentContext(): android.content.Context {
         return this.foregroundActivity;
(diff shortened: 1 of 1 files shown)


tns-core-modules/application/application.android.ts
```diff
@@ -63,10 +65,11 @@ export class AndroidApplication extends Observable implements AndroidApplication
         this.packageName = nativeApp.getPackageName();
         this.context = nativeApp.getApplicationContext();
 
-        let lifecycleCallbacks = initLifecycleCallbacks();
-        let componentCallbacks = initComponentCallbacks();
-        this.nativeApp.registerActivityLifecycleCallbacks(lifecycleCallbacks);
-        this.nativeApp.registerComponentCallbacks(componentCallbacks);
+        // we store those callbacks and add a function for clearing them later so that the objects will be eligable for GC
```
pana-cc: Can we make the listeners static, create them once and keep them live forever? The listeners are small and I don't think they enclose something that would be collected otherwise?

tns-core-modules/application/application.android.ts
```diff
@@ -63,10 +65,11 @@ export class AndroidApplication extends Observable implements AndroidApplication
...
+        this.callbacks.lifecycleCallbacks = initLifecycleCallbacks(() => this.callbacks = {});
```
vakrilov: The clear callback will clear **all** callbacks, included those created in `initComponentCallbacks()` (next line). Is that intentional? Maybe the callback should look like:
``` 
() => delete this.callbacks.lifecycleCallbacks
```

vtrifonov: yep, basically we don't need the `callbacks`, we are linking them to avoid initial GC, but and we want to clean them once they are saved from the initial run. Cleaning this.callbacks shouldn't affect the callback execution however, we are just not keeping a reference to them anymore.

vtrifonov: Maybe actually it will be better not to clear them, so that we can avoid further big MarkReachableObjects triggering and they are not holding too much memory, so after a chat with @PanayotCankov we've decided not to clear them.

Issue #5687, opened 2018-04-16
Feature: Expose onLayout event on View


vakrilov:
Expose and event that will be fired when the View layout pass is completed.
Main use case is to be able to use `getActualSize()` and be sure the values are correct. It is a common thing when you want to trigger animations that are dependent on the size of the views.

A project with a custom implementation of such property can be found here: https://play.nativescript.org/?template=play-ng&id=Ximtcw&v=17&file=slides%2Fpager.component.ts

jogboms:
@vakrilov wow!!
I just ran the playground and i was amazed at what you did there. 👍

vakrilov:
Kudos go to @darind - he created the example ;)

PR #5825, opened 2018-05-16
 feat(view): introduce LayoutChanged event on every View component

ADjenkov:
`LayoutChanged` event should be invoked when the layout bounds of a view changes.

Fix https://github.com/NativeScript/NativeScript/issues/5687


tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -305,6 +307,17 @@ export class View extends ViewCommon {
     public initNativeView(): void {
         super.initNativeView();
         this._isClickable = this.nativeViewProtected.isClickable();
+
+        this.setOnLayoutChangeListener();
```
vakrilov: We should create a LayoutChangeListener only if someone has attached to this event. Currently listeners will be created for all views and will be triggered unnecessary.
We are doing something similar with the touch listener [here](https://github.com/NativeScript/NativeScript/blob/ebdabf21f5551c99cf3dcbf54ac78319db64f6a8/tns-core-modules/ui/core/view/view.android.ts#L241-L243). For the `onLayout` we should probably override another method as it is not attached trough `observe()` (observe is for gestures). Probably `on()/off()` events

ADjenkov: @vakrilov  You are right! 
I think we don't need to override another method though. How about:

```
if(this.hasListeners(ViewCommon.layoutChangedEvent)) {
    this.setOnLayoutChangeListener();
}
```
This way will listen for the event only if someone has attached to this event from markup or `.on('layoutChanged)'`

manoldonev:
https://github.com/NativeScript/NativeScript/issues/5687 says that the main use case is to be able to use `getActualSize()` and be sure values are correct -- shouldn't we add a test for this?

tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -241,6 +243,26 @@ export class View extends ViewCommon {
         }
     }
 
+    on(eventNames: string, callback: (data: EventData) => void, thisArg?: any) {
+        super.on(eventNames, callback, thisArg);
+        const isLayoutEvent = typeof eventNames === "string"  ? eventNames.indexOf(ViewCommon.layoutChangedEvent) !== -1 : false;
```
manoldonev: `"string"  ?` -- two spaces instead of one.

tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -305,6 +327,19 @@ export class View extends ViewCommon {
     public initNativeView(): void {
         super.initNativeView();
         this._isClickable = this.nativeViewProtected.isClickable();
+
+        if (this.hasListeners(ViewCommon.layoutChangedEvent)) {
```
manoldonev: I think we need to check for `!this.layoutChangeListenerIsSet` here?

ADjenkov: It should not be possible to `initNativeView` twice without `disposeNativeView`

ADjenkov:
test

PR #5834, opened 2018-05-18
fix(builder): check for custom components registered at build time

sis0k0:
related to https://github.com/NativeScript/nativescript-dev-webpack/pull/529


sis0k0:
related to https://github.com/NativeScript/nativescript-dev-webpack/pull/529

PR #5850, opened 2018-05-22
chore(layout-changed): update layoutChanged event tests delay

ADjenkov:
Extending time for layoutChanged event tests, since it could be fired later compared to other events


ADjenkov:
test

Issue #5731, opened 2018-04-24
Memory leaks with CSS animation 


NickIliev:
_From @RoyiNamir on April 20, 2018 6:33_

I've created a simple new project via `tns create bbb --ng`

![image](https://user-images.githubusercontent.com/2842123/39033892-edd54baa-447c-11e8-9a69-024e90c17b5c.png)

Added a CSS animation class via `animate.css` library : 

**app.css**
```
@import '~nativescript-theme-core/css/core.light.css';
@import "~animate.css/animate.min.css";

.breath {
    animation-name: pulse;
    animation-duration: 2s;
    animation-iteration-count: infinite;
    animation-fill-mode: forwards;
}

```

And now I've created 7 images with that class. Some with manual images and some are via `*ngFor` : 

![image](https://user-images.githubusercontent.com/2842123/39033944-247af402-447d-11e8-8b7f-a08f27626357.png)

This is the code for **items.component.html**

```
<ActionBar title="My App" class="action-bar"></ActionBar>
<GridLayout columns="*,*,*,*,*,*,*" rows="*">

    <StackLayout row="0" col="1"  >
        <Image class="breath " [src]="'~/images/hat00001.png'" stretch="aspectFit"></Image>
    </StackLayout>

    <StackLayout row="0" col="2"  >
        <Image class="breath " [src]="'~/images/hat00001.png'" stretch="aspectFit"></Image>
    </StackLayout>

    <Image *ngFor="let o of [1,2,3,4,5];let i = index;"
           [col]="i+3"
           [class]="'breath'"
          verticalAlignment="top"
           [src]="'~/images/hat0000'+(i+2)+'.png'"
           stretch="aspectFit"></Image>
</GridLayout>

```

**There's nothing else. No TS code ,  no nothing. Just this html and css.**

However  , while writing down this issue ^  , look at the memory consumption : 

![image](https://user-images.githubusercontent.com/2842123/39034024-777f0526-447d-11e8-8276-2bc5568797fa.png)

Started at : 

![image](https://user-images.githubusercontent.com/2842123/39034039-83f6e68e-447d-11e8-8c35-ecff4c1fa17f.png)

And now it's (keeps growing) : 

![image](https://user-images.githubusercontent.com/2842123/39034051-92f5d3f2-447d-11e8-9057-48f1a772524d.png)

**Attached project** : https://ufile.io/spicc

Video showing the ease growing leak : 

[...]

PR #5861, opened 2018-05-24
fix(animation): remove throw and fix promise handling

NathanWalker:
## PR Checklist

Usage of `throw` destabilizes apps during animation handling.

## What is the new behavior?

Promise rejection is now removed around keyframe handling since there was no valid way to handle. With tracing enabled developers can now have meaningful and productive debug sessions.

https://github.com/NativeScript/nativescript-angular/issues/431

### DIFF
--- a/tns-core-modules/ui/animation/keyframe-animation.ts
+++ b/tns-core-modules/ui/animation/keyframe-animation.ts
@@ -74,7 +74,6 @@ export class KeyframeAnimation implements KeyframeAnimationDefinition {
     public iterations: number = 1;
 
     private _resolve;
-    private _reject;
     private _isPlaying: boolean;
     private _isForwards: boolean;
     private _nativeAnimations: Array<Animation>;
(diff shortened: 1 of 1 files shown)


NathanWalker:
Excellent @RoyiNamir definitely agree. I’ll update this PR to include that fix as well later, thank you.

NathanWalker:
@RoyiNamir that spot is handled in this pr.

NathanWalker:
@RoyiNamir in your case above, anytime you use animation.play() it'd be best practice to wire up the success and error callbacks, ie:

```
animation.play(options).then(success => {
  //...
}, error => {
  //...
});
```

RoyiNamir:
@NathanWalker Not only. In my case ^ it was when I added a class ( which contains as single `animation-name` with 1 iteration) and then removed that class.
Error appeared also.

NathanWalker:
@RoyiNamir good to know. Nice one and valid use case. Will update PR to handle that as well, thank you for clarifying.

NathanWalker:
@RoyiNamir There really should never be a rejection in keyframe animation handling since there's no proper way to handle it therefore this removes rejections altogether. Will solve your case and many others that have been reported over time around this. You may try the changes seen in this PR in your project and report back the results for you.

RoyiNamir:
@NathanWalker  Having said that , there are also much more critical animation problems which is in progress of solving by the NS team.
I've already reported a memory leak which eventually stuck and shutdown the app. 
https://github.com/NativeScript/NativeScript/issues/5731

Just to clarify : I'm not saying that the current issue is not important  , as it is just an informative error.
You can't do `addAnimationClass().then(()=>{},error=>do somethng)`
And as you've said , it is feasible to do `view.animate()).then(()=>{},error=>do somethng)`

vakrilov:
test

tns-core-modules/ui/animation/keyframe-animation.ts
```diff
@@ -239,10 +236,10 @@ export class KeyframeAnimation implements KeyframeAnimationDefinition {
             // Catch the animation cancel to prevent unhandled promise rejection warnings
             animation.play().then(() => {
                 this.animate(view, index + 1, iterations);
+            }, (error: any) => {
+                traceWrite(typeof error === "string" ? error : error.message, traceCategories.Animation, traceType.warn);
```
vakrilov: Just wondering - do you need both error callback and `.catch()`?

vakrilov:
test

tns-core-modules/ui/animation/keyframe-animation.ts
NathanWalker: Although as @RoyiNamir points out I believe there are others cases that having both may provide utmost resiliency against.

Issue #5879, opened 2018-05-29
[iOS] Unable to show dialog from modal view without a page


ADjenkov:
**The original problem** occurred in an Angular app (Plugins team teamplate), that displays some dialogs inside modal view, bacause [we are no longer wrapping modal view inside a Page in nativescript-angular@next](https://github.com/NativeScript/nativescript-angular/pull/1298/commits/0b965a4b4cc9d6c198fb94977dba75f153f565f4).
This could also happen in a non Angular application (modal view without `<Page>`)

PR #5881, opened 2018-05-29
fix(ios-dialogs): unable to show dialog from modal view without a page

ADjenkov:
Fix https://github.com/NativeScript/NativeScript/issues/5879
**The original problem** occurred in an Angular app (Plugins team teamplate), that displays some dialogs inside modal view, bacause [we are no longer wrapping modal view inside a Page in nativescript-angular@next](https://github.com/NativeScript/nativescript-angular/pull/1298/commits/0b965a4b4cc9d6c198fb94977dba75f153f565f4).
This could also happen in a non Angular application (modal view without `<Page>`) 

**The solution** is to search up in the hierarchy for a parent with UIViewController on which we should present the dialog.

Added test in modal-navigation app that opens dialog from modal view (with and without wrapping modal view content in Page)

### DIFF
--- a/tns-core-modules/ui/dialogs/dialogs.ios.ts
+++ b/tns-core-modules/ui/dialogs/dialogs.ios.ts
@@ -207,7 +207,7 @@ function showUIAlertController(alertController: UIAlertController) {
             view = currentPage.modal;
 
             if (view.ios instanceof UIViewController) {
-                viewController = currentPage.modal.ios;
+                viewController = view.ios;
             } else {
                 viewController = iosView.getParentWithViewController(view).viewController;
             }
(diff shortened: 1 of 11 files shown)


tns-core-modules/ui/dialogs/dialogs.ios.ts
```diff
@@ -200,7 +200,19 @@ export function login(): Promise<LoginResult> {
...
     if (currentPage) {
-        let viewController: UIViewController = currentPage.modal ? currentPage.modal.ios : currentPage.ios;
+        let view: View = currentPage;
+        let viewController: UIViewController = currentPage.ios;
+
+        if (currentPage.modal) {
+            view = currentPage.modal;
+
+            if (view.ios instanceof UIViewController) {
+                viewController = currentPage.modal.ios;
```
MartoYankov: Shouldn't this be `viewController = view.ios`?

TomasKypry:
Can you please describe how to fix it? I have a <Page> in my modal and dialogs.action() for example doesn't work on iOS.

Issue #5914, opened 2018-06-05
[Feature Request] Flexible Error/Exception handling


vakrilov:
## The Problem
Currently, when an unhandled exception is thrown in NativeScript (inside app-code, core-modules or plugins) it will (in most) cases crash the application showing the error and stack trace. 

When in **development** - in many cases this is what you'd expect. You would want to have the stack-trace of the exact location the unexpected behavior has occurred, so that you can more easily understand what happened and fix it.

However, in **production**, an application crash can seriously hurt you application credibility and drive away customers. In many cases you might prefer anything else (app freeze, blank screen, failed navigation etc.) to an actual crash with an error log. There are even cases, in whitch the application will recover and will work OK even if something unexpected has happened - for exampled failed animation or a typo in CSS property.

## Proposal

The aim of this FR is to propose an API that provides a way to handle errors in different ways in development and in production. It should provide flexibility to configure how errors should be handled. For example:

1. _(development 1)_ Throw exceptions as soon as error occurs.
2. _(development 2)_ Show a scary `console.log` with `ERROR: Something bad happened` but continue execution of the app. You will see it in you terminal, but decide if it is critical based on what happens with the app after that.
3. _(production)_ Send an error report to your analytics/error-report server, but continue app execution. Maybe triggers some recover logic that will handle the app without a crash. 

### API Proposal
Extend the `trace` module with an `error()` method. 

Use the method in core-framework/plugins/app code:
```
function doSomething(arg) {
  // Instead of   
  if(!arg) throw new Error("Arg not provided in "doSomething");
  
  // Use trace.error()
  if(!arg) {
    trace.error("Arg not provided in "doSomething");
    return;
  }
  // ... implementation using arg
}
```

Developers can define custom error handler logic (probably on app startup):

```
const errorHandler: ErrorHandler = {
    handleError(err){
        //(development 1)
        throw err;

        //(development 2)
        trace.write(err, "unhandlede-error", type.error);

        //(production)
        reportToAnalytics(err)
    }
}

trace.setErrorHandler(errorHandler)
```

### Next Steps

- [x] Gather FB on this proposal and implement it in the `tns-core-modules`.
- [x] Create a comprehensive guide on how to use the error API in the [contributing.md](https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md). 
- [x] Require all PRs to follow the guide and handle errors in a similar fashion
[...]

PR #5929, opened 2018-06-08
feat: Flexible Error/Exception handling

vakrilov:
Implements #5914

### DIFF
--- a/HandlingErrors.md
+++ b/HandlingErrors.md
@@ -1,0 +1,36 @@
+# Handling Errors in NativeScript Core Modules
+
+One big difference between web-app and NativeScript app is that throwing and `Error` in NativeScript app cause the app to **crash**. Such crashes can be a serious problem for a production applications as they the application credibility and drive away customers.
+
+We want to provide the application developers with the flexibility to handle errors differently in *development* and *production* modes.
+
+## Using the Trace Module
+The `tns-core-modules/trace` utility module provides a good way to stream-line error logging and handling throughout the framework. It gives application developers a way to define custom `TraceWriter`s and `ErrorHandler`s for their apps and even specify different sets of those to be used in during development and in production.
+
+Here are the guidelines to using this when contributing to core-modules or creating your own plugins.
+
+### Use `trace.write()`
+Use trace.write() with the appropriate type to log non critical errors. 
+
+>Note: For the `error` message level all loggers will be notified unconditionally, for all other levels (`log`,`info`,`warn`), tracing should be enabled and the corresponding categories should be added.
+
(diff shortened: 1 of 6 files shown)


tests/app/trace/trace-error-tests.ts
```diff
@@ -0,0 +1,64 @@
...
+import { constants } from "zlib";
+import { trace } from "tns-core-modules/profiling/profiling";
+
+let cachedErrorHandler: ErrorHandler;
+export function setUpModule() {
+    cachedErrorHandler = getErrorHandler();
+}
+
+// before each
+export function tearDown() {
```
MartoYankov: Technically, this is executed after each test. There is a setUp() function for before each.

tests/app/trace/trace-error-tests.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    TKUnit.assert(called, "trace.error() should call handler")
+}
+
+export function test_trace_error_should_create_error_from_string() {
+    let called = false;
+    setErrorHandler({
+        handlerError(error) {
+            called = true;
+            TKUnit.assert(error instanceof Error, "trace.error() wrap string in error")
```
MartoYankov: Even if it's fine in this case, I think it's dangerous to use asserts in callbacks. It's better to cache the value and assert it after the assertion of the `called` variable. In cases where the callbacks are async asserting in them breaks the tests.

tns-core-modules/trace/trace.d.ts
```diff
@@ -63,6 +63,11 @@ export function isCategorySet(category: string): boolean;
  */
 export function write(message: any, category: string, type?: number);
 
+/**
+ * Passes an error ot the registered ErrorHandler
```
MartoYankov: typo - "ot" -> "to"

tests/app/trace/trace-error-tests.ts
vakrilov: yup - I want to cache the original `ErrorHandler` once and clean-up after each test to make sure the test does not leak side effects

HandlingErrors.md
```diff
@@ -0,0 +1,36 @@
+# Handling Errors in NativeScript Core Modules
+
+One big difference between web-app and NativeScript app is that throwing and `Error` in NativeScript app cause the app to **crash**. Such crashes can be a serious problem for a production applications as they the application credibility and drive away customers.
```
vchimev: Such crashes can be a serious problem for ~~a~~ production applications as they **hurt** the application credibility ...

PR #5953, opened 2018-06-14
test(): add more timeout to tab navigations tests

MartoYankov:
These tests fail on Android api19 only with waitUntilReady timeout. 5 seconds is the max timeout of the tests themselves.


MartoYankov:
test --ignore core_modules_unit_tests#myankov/fix-api19-tests api19

PR #5841, opened 2018-05-20
fix(modal): parent page invalid hierarchy handling

NathanWalker:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [ ] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#running-unit-tests.
- [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/WritingUnitTests.md.

## What is the current behavior?

Apps break with the current `throw` without any effective way to handle. Additionally a `while` loop was being used without safe guards to prevent `undefined` access.

## What is the new behavior?

With tracing on, developers can now properly debug their apps for various use cases without an untested (and most often not reproducible) ux flow/interaction hitting production resulting in this `throw` leaving the app unstable and unpredictable. Also proper safe guards have been added to prevent `undefined` access.

PR #5966, opened 2018-06-19
fix(modal): parent page invalid hierarchy handling [extended]

vakrilov:
Extension of #5841:
-  Added "report error and return immediately logic" `_showNativeModalView` in `_hideNativeModalView`
- `getParentWithViewController` refactor to be more concise
-  guard in `showUIAlertController` method

CC @NathanWalker
> 
> ## PR Checklist
> 
> - [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
> - [ ] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
> - [x] You have signed the [CLA](http://www.nativescript.org/cla).
> - [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#running-unit-tests.
> - [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/WritingUnitTests.md.
> 
> ## What is the current behavior?
> 
> Apps break with the current `throw` without any effective way to handle. Additionally a `while` loop was being used without safe guards to prevent `undefined` access.
> 
> ## What is the new behavior?
> 
> With tracing on, developers can now properly debug their apps for various use cases without an untested (and most often not reproducible) ux flow/interaction hitting production resulting in this `throw` leaving the app unstable and unpredictable. Also proper safe guards have been added to prevent `undefined` access.
>


tns-core-modules/ui/core/view/view.ios.ts
```diff
@@ -356,6 +364,11 @@ export class View extends ViewCommon {
     }
 
     protected _hideNativeModalView(parent: View) {
+        if (!parent || !parent.viewController) {
+            traceError("Trying to hide modal view but no parent with viewController specified.")
```
vakrilov: It's highly unlikely that the parent is undefined here as it's only called [_closeModalCallback](https://github.com/NativeScript/NativeScript/blob/0fc1547a19e29efba9e88d84dd24118262fdd18a/tns-core-modules/ui/core/view/view-common.ts#L263) where it is scoped.

vakrilov:
test

PR #5967, opened 2018-06-19
feat: Pass NS app to the native app instead of presenting it over the root VC

tdermendjiev:
When NativeScript embedded app is created from the native one we check for whether the topmost UIViewController has NativeScriptEmbedder protocol (implemented in the iOS Runtime) method 'presentNativeScriptApp:'. If yes, we call it with the NS app viewcontroller as a parameter so the embedder has control over the NS app (where and how to present it etc.) For backwards compatibility we present the NS app on top of the topmost UIViewController as a fallback.

## PR Checklist

Currently embedded app is always presented on top of the rootViewController.

## What is the new behavior?
The embedder is now able to choose where to  present the NS app (present, push, add as a child ViewController etc.) by implementing 'presentNativeScriptApp' method. Even when he does not implement it the NS app will present on top of the visible(topmost) viewcontroller instead of the root one (which may be hidden in some scenarios).


tns-core-modules/application/application.ios.ts
```diff
@@ -235,6 +235,20 @@ class IOSApplication implements IOSApplicationDefinition {
             this._window.makeKeyAndVisible();
         }
     }
+
+    public getTopmostViewController(rootViewController: UIViewController): UIViewController {
```
mbektchiev: @vakrilov Do we need to add documentation about this new public function? It can be quite useful when dealing with UIViewControllers and I think that it's a good idea.

vakrilov: Yup - we can expose it.
I think it better to make it static (it doesn't really need `this`). You can put it in the [`ios` namespace in the utils module](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/utils/utils.ios.ts#L47) and the documentation should go in [utils.d.ts[(https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/utils/utils.d.ts#L185)

tns-core-modules/utils/utils.d.ts
```diff
@@ -247,6 +247,8 @@ export module ios {
      * iOS - this folder is read-only and contains the app and all its resources.
      */
     export function getCurrentAppPath(): string;
+
+    export function getVisibleViewController(rootViewController: UIViewController): UIViewController;
```
vakrilov: Add JSDoc for `getVisibleViewController`

Issue #1677, opened 2016-03-01
ListPicker: enable working with an array of objects


sebawita:
Currently the **ListPicker** can only display values from an array of strings, however it is not usable for an array of objects, like: 
`var items = [ {name: "Pizza", price: "8"},{name: "Hot Dog", price: "2"},{name: "Burger", price: "6"} ];.`
It would be good if there were properties like:
- **textField** - to define which field should be used to display
- **valueField** - to define which field should be used to return
- **selectedValue** or **selectedItem** - bindable property, that would contain the selected value (as defined by **valueField**)

These could be used like this:
```
    <ListPicker
        items="{{ items }}"
        textField="name"
        valueField="price"
        selectedValue="{{ itemPrice }}"
    ></ListPicker>
```

enchev:
Hey @sebawita,

if you override toString() and use binding expressions you can something like this:

```XML
<Page loaded="pageLoaded">
  <StackLayout>
    <Label text="{{ items ? 'Selected item: ' + items[index].name : '' }}" />
    <ListPicker items="{{ items }}" selectedIndex="{{ index }}" />
  </StackLayout>
</Page>
```
```TypeScript
import observable = require("data/observable");
import pages = require("ui/page");

export function pageLoaded(args: observable.EventData) {
    let page = <pages.Page>args.object;

    let items = new Array<any>();
    for (let i = 0; i < 10; i++) {
        let name = `Name ${i}`;
        items.push({ id: i, name: name, toString: () => { return name; } });
    }

    page.bindingContext = { items: items, index: 2 };
}
```

Furthermore if your `bindingContext` is not anonymous but `Observable` the `Label` text will be updated when you raise change for `index` property.

sitefinitysteve:
@sebawita I feel like we should re-open this, it's more than a valid scenario, very much a common one that should need less code than this?  I feel like we need this to be more dropdownlist\comboboxy with object support.

x00:
he gave a perfectly good solution. 

You can define toString() to return a string.

jonnysamps:
Using `toString` is not valid if already being used for serialization.  Seems like this should be reopened.

sebawita:
To be honest this workaround feels a little bit hacky, as it requires extra code for something that should be rather simple.
[...]

PR #6033, opened 2018-07-05
feat(list-picker): add textField, valueField and selectedValue properties

sebawita:
## PR Checklist

Currently if we pass and array of json objects to a listview, there is no simple way to make it display a specific field from each item.
Also, we can only get the selectedIndex, but there is no direct way to get the selected object, or even specify which property should be returned.

## What is the new behavior?
Implements #1677

Adds new properties to the ListView component (`textField` and `valueField` - should be used with arrays of JSON objects):
 * `textField` - tells the listview which property should be used to display each item
 * `valueField` - tells the listview, which property should be used to update the selectedValue
 * `selectedValue` - is the property that will contain the selectedValue, if `valueField` is specified, then it will contain the value from that property, otherwise it will contain the whole selected item

## Breaking changes

No breaking changes.

## Demo app

I've added a sample in apps -> list-picker -> list-picker-json-array


sebawita:
An example of using the new properties:

```xml
<ListPicker items="{{ items }}"
  textField="name"
  valueField="role"
  selectedIndex="2"
  selectedValue="{{ selectedItem }}">
</ListPicker>
```

larssn:
It's unclear how `selectedValue` translates to Angular. How do you databind it there?

I've tried various ways:
(selectedValue)="selected = $event"
[selectedValue]="myObj?.field"

Doesnt seem to do anything.

Another thing. Binding the component via reactive forms will still bind just the index, and not the `valueField`:
`<ListPicker formControlName="printerUid" ...></ListPicker>` (This will bind 0,1,2 etc. to the field `printerUid`)

xxj2016:
When is Angular coming next?

PR #6067, opened 2018-07-13
refactor: streamline usage of submodules

manoldonev:
Related to https://github.com/NativeScript/nativescript-dev-webpack/issues/607

While researching the issue with webpack it became apparent that we were not strictly following package / module conventions for submodules in the tns-core-modules package:
- we had some folders like https://github.com/NativeScript/NativeScript/tree/master/tns-core-modules/utils that contained a bulk of d.ts files but empty package.json (with respect to module definition)
- we had some folders like  https://github.com/NativeScript/NativeScript/tree/master/tns-core-modules/debugger that contained a bulk of d.ts files but only single d.ts was actually referenced in the folder/module package.json

This PR streamlines the usage of submodules in the tns-core-modules package (i.e. introduces submodules where they are applicable but missing).

This PR is tested and works with the latest @ngtools/webpack 6.1.0-rc.2 as well.


manoldonev:
NOTE: it seems we cannot refactor ins-core-modules code that easily (especially moving files around) -- as on updating Android apps from the store they will contain both "old" and "new" file on the device. This would not be a problem if the app was published with a bundler (i.e. webpack) but we don't know how many {N} apps were published without webpack.

In summary -- cannot merge this right now.

PR #6079, opened 2018-07-17
fix: require devtools-elements.js with the extension mentioned explicitly

sis0k0:
This is a workaround for the issue with `@ngtools/webpack@6.1.0-rc.2`, described here: https://github.com/NativeScript/nativescript-dev-webpack/issues/607#issuecomment-405209925


manoldonev:
test

PR #6099, opened 2018-07-20
 feat(image-asset): add saveToFile method to ImageAsset

ADjenkov:
Expose `ImageAsset.saveToFile` method that:
- __[iOS]__ Can copy the original image file, without modifying it, inside Temporary (Cache) folder of the app.
- __[iOS and Android]__ Can create new resized image file (If width/height provided as `options` of the ImageAsset) inside Temporary (Cache) folder of the app. 

  _The mimeType/extension of the image should be automatically detected in order to use the proper compression (PNG vs JPEG)._
 
---
__Why `saveToFile` method for `ImageAsset`?__
Imagine a common scenario where you have to upload an image from the phone gallery, using `nativescript-imagepicker`, to the cloud.  Currently the lack of a method that saves ImageAsset as a file (inside the current application directory) imposes the use of `ImageSource.saveToFile()` in order to upload image.

Upload steps:
  1. Select an image from the gallery
  2. Convert the result returned from `nativescript-imagepicker` to `ImageSource` since it is an `ImageAsset` object.
 3. Create new image file inside Temporary (Cache) folder of the app using `ImageSource.saveToFile` method.

    _`ImageSource.saveToFile` expects `format` param which means you have to manually find the correct image extension/mimeType_ of the image.

    _`ImageSource.saveToFile` will always perform image compression while saving file, which can affect the file size, even if no resizing is required (no `width` and `height` options set for `ImageAsset`)_

 4. Use the newly created image file path for upload.
---

__Less code for our templates__:
[template-master-detail](https://github.com/NativeScript/template-master-detail/blob/b5f90da3e325bfb180d909fd64974c9d9bf5bafd/cars/car-detail-edit-page/car-detail-edit-view-model.js#L83) template could switch to `saveToFile` instead of using the above approach.
[template-construction-sites](https://github.com/NativeScript/template-construction-sites/blob/371966e8433d5cfe57871c5fb33077acde0daa59/app/construction-sites/construction-site-edit/construction-site-edit.component.ts#L125) template manually creates a copy of the selected image file for iOS.

### DIFF
--- a/tns-core-modules/image-asset/image-asset.android.ts
+++ b/tns-core-modules/image-asset/image-asset.android.ts
@@ -42,7 +42,7 @@ export class ImageAsset extends common.ImageAsset {
             let error = null;
             // read as minimum bitmap as possible (slightly bigger than the requested size)
             bitmap = android.graphics.BitmapFactory.decodeFile(this.android, finalBitmapOptions);
-            
+
             if (bitmap) {
                 if (requestedSize.width !== bitmap.getWidth() || requestedSize.height !== bitmap.getHeight()) {
                     // scale to exact size
@@ -66,6 +66,10 @@ export class ImageAsset extends common.ImageAsset {
         catch (ex) {
             callback(null, ex);
         }
+    }
+
+    public saveToFile(fileName: string, callback: (imagePath: string, error: any) => void) {
+        callback(this.android, null);
     }
 }
 
--- a/tns-core-modules/image-asset/image-asset.android.ts
+++ b/tns-core-modules/image-asset/image-asset.android.ts
@@ -69,7 +69,43 @@ export class ImageAsset extends common.ImageAsset {
     }
 
     public saveToFile(fileName: string, callback: (imagePath: string, error: any) => void) {
-        callback(this.android, null);
+        const shouldResize = this.options && (this.options.width || this.options.height);
+
+        if (shouldResize) {
+            this.getImageAsync((image, err) => {
+                if (image) {
+                    const extension = this.android.split(".").pop();
+                    const targetFormat = this.getTargetFormat(extension);
+                    const tempPath = knownFolders.temp().path;
+                    const localFullPath = fsPath.join(tempPath, `${fileName}.${extension}`);
+                    const quality = this.options.quality || 100;
+                    const outputStream = new java.io.BufferedOutputStream(new java.io.FileOutputStream(localFullPath));
+
+                    const result = image.compress(targetFormat, quality, outputStream);
+                    outputStream.close();
+
+                    if (result) {
+                        callback(localFullPath, null);
+                    } else {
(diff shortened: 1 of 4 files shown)


MartoYankov:
We should probably change the default quality to 90-95.

tns-core-modules/image-asset/image-asset.android.ts
```diff
@@ -67,6 +67,46 @@ export class ImageAsset extends common.ImageAsset {
...
+            this.getImageAsync((image, err) => {
+                if (image) {
+                    const extension = this.android.split(".").pop();
+                    const targetFormat = this.getTargetFormat(extension);
+                    const tempPath = knownFolders.temp().path;
+                    const localFullPath = fsPath.join(tempPath, `${fileName}.${extension}`);
+                    const quality = this.options.quality || 100;
+                    const outputStream = new java.io.BufferedOutputStream(new java.io.FileOutputStream(localFullPath));
+
+                    const result = image.compress(targetFormat, quality, outputStream);
```
manoldonev: Can the native image.compress(...) call throw exception? Should we wrap this in try/catch block and clase the stream at all times? Not sure how do we handle such scenarios.

tns-core-modules/image-asset/image-asset.ios.ts
```diff
@@ -42,19 +42,19 @@ export class ImageAsset extends common.ImageAsset {
 
         if (this.nativeImage) {
             let newSize = CGSizeMake(requestedSize.width, requestedSize.height);
-            let resizedImage = this.scaleImage(this.nativeImage, newSize);
+            let resizedImage = this.scaleImage(this.nativeImage, newSize, ignoreScaleFactor);
```
manoldonev: `ignoreScaleFactor` would be undefined if parameter is not specified. It is not a big deal here but still `!!ignoreScaleFactor` would make sure we are passing an actual boolean (or we can specify a default value for the parameter).

ADjenkov: We cannot add default value for optional param (we want it to be optional since it's public API)

tns-core-modules/image-asset/image-asset.ios.ts
```diff
@@ -64,8 +64,83 @@ export class ImageAsset extends common.ImageAsset {
...
+                    quality = (quality - 0) / (100 - 0);  // Normalize quality on a scale of 0 to 1
+                }
+
+                data = UIImageJPEGRepresentation(instance, quality);
+                break;
+        }
+        return data;
+    }
+
+    private getImageExtension(UTIType: string): "png" | "jpg" {
```
manoldonev:
``` ts
+    private getImageExtension(UTIType: string): "png" | "jpg" {
+        switch (UTIType) {
+            case kUTTypePNG:
+                return "png";
+            default:
+                return "jpg";
+        }
+    }
```

PR #6102, opened 2018-07-21
refactor(core-modules): implement createNativeView and initNativeView for all components

farfromrefug:
This PR propose a simple refactoring to make life easier for plugin developers.
I am writing a [plugin](https://github.com/Akylas/nativescript-material-components) which brings material components for iOS and Android.

in [material-components](https://github.com/material-components), Google simply inherited native components to create material ones (UIButton, UITextField, ...)
So creating a nativescript component was pretty straight forward as I simply had to subclass the corresponding tns components and replace the ```createNativeView```

Except that It was not working for some reasons:
- some native views were created in the tns constructor component. So if I was subclassing some native components were unnecessarily created  ([iOS label](https://github.com/Nativescript/NativeScript/blob/master/tns-core-modules/ui/label/label.ios.ts#L29) is an example)
- delegate/listener were created in ```createNativeView```. So I subclassed ```createNativeView``` I was loosing the delegate/listener setup, I then had do it myself and also re-define those delegate/listener classes exactly the same way it was done in Nativescript (unnecessary)
- in the case of the android textfield the material component was actually made of 2 components, the layout and text widget. Which means that, in nativescript, any layout operation is to be applied to the native layout widget. All text operations should be applied to the text widget. For now I had to use a dirty, not fully working working, overriding of the ```nativeViewProtected``` getter to either return the layout or the text widget. That PR actually defines a ```nativeTextViewProtected``` getter that is used for all text related operations. By default it returns ```nativeViewProtected``` That way a plugin can simply override it to use it's own widget. This has no performance consequences.

All those are really just refactoring which does not impact Nativescript performances. However they do make plugins more performant because the prevent redefining already defined Nativescript delegate/listener classes. But also because they prevent unused native widget creations for plugins subclassing some Nativescript widget classes like Label.

The only thing I am not sure about if is if ```createDelegate```  which I call in ```createNativeView``` could be called in ```initNativeView``` instead. I am waiting for some feedback from you on this.

## PR Checklist

normal

## What is the new behavior?
unchanged


SvetoslavTsenov:
test

farfromrefug:
@SvetoslavTsenov what does that mean? Also I can't see the details log of ci/jenkins/core-modules-tests — FAILed. So I can't see what to fix

SvetoslavTsenov:
Hey @farfromrefug, I have executed some integration tests on our CI (which is internal). The result so far is that unit tests are failing for both platform and uitests app crashes on start up(you can find it in app folder). Could you please rebase your branch again and run unit tests  on you side. To do that you can simple do:
`npm run setup
  npm run tsc
  cd tests
  tns run ios/ android
`

Let me know if you need some further information.

farfromrefug:
@SvetoslavTsenov I realised the issue now. On iOS ```createNativeView``` is not used.
I also realise that there is a comment about this in the code saying it should be done the same way as in android and use it.

DO you want me to expand that PR to do that? I am willing to but it would be more changes for iOS.
Could make that PR harder to merge. 
I
I still defined ```createNativeView``` for iOS Label and Button but call It in the constructor. Also I only create if not already created.
This does not follow the same rule as Android. But otherwise tests will fail.

farfromrefug:
@SvetoslavTsenov still can't see what tests are failing with ```ci/jenkins/core-modules-tests — FAILed.```. Details page brings fail to open page

As for the tests app. It succeeds now on iOS but I keep getting ```waitUntilReady Timeout., Stack: Error: waitUntilReady Timeout.``` errors on android. Any idea?
Finally I have a lot of tests failing on android not related to what I have done. Like :
```Test: --- [ACTION-BAR.test_ActionBar_set_title_as_number_doesnt_thrown] FAILED: Expected: false, Actual: true, Stack: Error: Expected: false, Actual: true```

So not sure how to know what makes the cli fail

**Comment by MartoYankov** [comment]
Hi @farfromrefug, thanks for the PR and thumbs up for the material components plugin. Just a heads up - the repo you linked to the plugin appears to be private, so we can't take a look.

I reviewed the PR, discussed it with the team and we agree with the changes.
 * Just for context and reference, I think you should take a look at `ViewBase`'s [setupUI](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/core/view-base/view-base.ts#L682) method. This is the place where `createNativeView()` and `initNativeView()` are called and where the backwards compatibility for native view create in constructor is handled.
 * Also, if you haven't already, take a look at this article in the docs - https://docs.nativescript.org/plugins/ui-plugin-custom#building-ui-plugins-using-custom-components - largely explaining what you are trying to do.
 * We consider the iOS components native view create in constructor as technical debt and we should move to `createNativeView()` on all components eventually. The goal is to have a unified api with the Android components.

That being said, here are some comments on how to improve the PR:
 * On all the iOS components where you added `createNativeView()` remove the constructor altogether. The `createNativeView()` method will be called from the `setupUI()` method.
 * There is no need to couple the new `createDelegate(nativeView)` method with `createNativeView()`. I think you even forgot to call it in `editable-text-base.android.ts`. My suggestion is to call this method in `setupUI()` right after `createNativeView()`. Of course this would mean to add it in `view-base.d.ts` and `view-base.ts` with an empty comment for backwards compatibility.
 * Also, please add a comment to the `nativeTextViewProtected` getter on why we are doing this and why you would want to override it.

**Comment by farfromrefug** [comment]
@MartoYankov Thanks.
Sorry about the link. I fixed it:
https://github.com/Akylas/nativescript-material-components

**Comment by farfromrefug** [comment]
@MartoYankov great comments.

I will look at all that !

Thanks

**Comment by MartoYankov** [comment]
@farfromrefug I think the two weak references to `this` are fine and also, updating the other iOS and Android classes can happen in separate PRs, so that they are more manageable. 

Now, seeing that implemented I have some doubts about the new `initNativeViewDelegates` method that has a parameter any. It's a bit confusing having both `initNativeViewDelegates` and `initNativeView`. What do you think about moving the contents of the `initNativeViewDelegates` to `initNativeView`? Will it work for your case? It will have the added benefit that during `initNativeView` we already have `nativeViewProtected`, so you won't have to pass parameters.

I'm thinking about making a PR to your fork if you are okay with the change. Let me know what you think.

**Comment by farfromrefug** [comment]
@MartoYankov a PR is fine !
Now about moving the ```initNativeViewDelegates``` inside ```initNativeView``` That s actually what I try to get around.
The issue is that if you component uses  a subclass of the nativeView Nativescript is using you want to have a custom ```initNativeView``` but keep the delegate from Nativescript core.
The real issue is that otherwise you have to declare again all the "listener" classes  like the ones define [here](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/editable-text-base/editable-text-base.android.ts)

**Comment by MartoYankov** [comment]
@farfromrefug That way you would avoid repeating the listeners, but there is still some custom code in `initNativeView` like assigning the listener owner. Wouldn't it be better if in your inheritor class you call the `super.initNativeView()` which would initialize the listeners of the base NativeScript core class and then put your custom `initNativeView` logic there?

**Comment by farfromrefug** [comment]
@MartoYankov sorry I think I misplaced ```initNativeView``` and ```createNativeView``` ...
So I am ok with ```initNativeView``` being the place where all listener/delegates are created. 
One thing: it seems ```initNativeView``` is never used on iOS. No issue there?
Cause I would move all delegate creation there even on iOS.

Also isn't ```initNativeView``` called multiple times? I mean it's called from ```setNativeView``` which is called from ```nativeView``` setter which could be called often? And there is no check to see if new ```nativeView``` is equal to current one. So if called again with same view, initNativeView will be called again and thus all listener created again.

BTW I am thinking about upgrading all ui components from that PR. Any issue with that?

**Comment by MartoYankov** [comment]
@farfromrefug Regarding your first question - yes, the current state of affairs is that iOS components are using their constructors to create and init, while Android components are using `createNativeView` and `initNativeView`. As I previously commented, we consider this technical debt and should move the code from all iOS constructors to `createNativeView` and `initNativeView`.

Regarding your second question - the `nativeView` setter shouldn't be called often. I'm not sure about the scenario where it was needed, but there is a check if the new value is equal to the current one. The first check in `setNativeView` should account for that.

I'm okay with making the changes to all components in this PR. We have to change the title though.

**Comment by farfromrefug** [comment]
@MartoYankov I updated all the UI components

File: tns-core-modules/ui/image/image.ios.ts [hunk 1]
```diff
@@ -10,15 +10,18 @@ export class Image extends ImageBase {
...
 
-    constructor() {
-        super();
-
-        //TODO: Think of unified way of setting all the default values.
+    public createNativeView() {
         const imageView = UIImageView.new();
         imageView.contentMode = UIViewContentMode.ScaleAspectFit;
         imageView.userInteractionEnabled = true;
         this.nativeViewProtected = imageView;
```
MartoYankov: I think this line should be removed since this is going to happen with the `createNativeView()` where it's called.

tns-core-modules/ui/image/image.ios.ts
```diff
@@ -10,15 +10,18 @@ export class Image extends ImageBase {
...
         this._setNativeClipToBounds();
```
MartoYankov: This is moved to `initNativeView()`. Is there a reason not to remove it from `createNativeView()` ?

tns-core-modules/ui/label/label.ios.ts
```diff
@@ -23,15 +23,14 @@ export class Label extends TextBase implements LabelDefinition {
     nativeViewProtected: TNSLabel;
     private _fixedSize: FixedSize;
 
-    constructor() {
-        super();
-
-        this.nativeViewProtected = TNSLabel.new();
-        this.nativeViewProtected.userInteractionEnabled = true;
+    public createNativeView() {
+        const view = this.nativeViewProtected = TNSLabel.new();
```
MartoYankov: The `this.nativeViewProtected` assignment should be removed here.

tns-core-modules/ui/switch/switch.ios.ts
```diff
@@ -33,14 +33,26 @@ export class Switch extends SwitchBase {
 
     constructor() {
         super();
-        const nativeView = UISwitch.new();
-        this._handler = SwitchChangeHandlerImpl.initWithOwner(new WeakRef(this));
-        nativeView.addTargetActionForControlEvents(this._handler, "valueChanged", UIControlEvents.ValueChanged);
-        this.nativeViewProtected = nativeView;
         this.width = 51;
         this.height = 31;
```
MartoYankov: Shouldn't these be moved to `createNativeView()`?

tns-core-modules/ui/tab-view/tab-view.ios.ts
```diff
@@ -205,9 +205,18 @@ export class TabView extends TabViewBase {
 
         this.viewController = this._ios = UITabBarControllerImpl.initWithOwner(new WeakRef(this));
         this.nativeViewProtected = this._ios.view;
```
MartoYankov: Is there a problem with implementing `createNativeView()` here?

tns-core-modules/ui/time-picker/time-picker.ios.ts
```diff
@@ -21,27 +21,35 @@ function getComponents(date: Date | NSDate): NSDateComponents {
...
-    private _ios: UIDatePicker;
+    nativeViewProtected: UIDatePicker;
     private _changeHandler: NSObject;
-    public nativeViewProtected: UIDatePicker;
 
     constructor() {
         super();
+        let components = getComponents(NSDate.date());
+        this.hour = components.hour;
+        this.minute = components.minute;
```
MartoYankov: Shouldn't these be moved to `createNativeView()` or `initNativeView()` ?

tns-core-modules/ui/web-view/web-view.ios.ts
```diff
@@ -83,25 +83,33 @@ export class WebView extends WebViewBase {
     private _ios: WKWebView;
```
MartoYankov: Since you removed this property, you should change it to `nativeViewProtected` and change the `ios` getter like in the other components.

tns-core-modules/ui/web-view/web-view.android.ts
```diff
@@ -116,6 +115,7 @@ export class WebView extends WebViewBase {
         }
 
         (<any>nativeView).client.owner = null;
+        (<any>nativeView).client = null;
```
MartoYankov: I think this is not necessary since we already removed the JS reference from it - the owner - and there is no reference to the native object from the JS object.

tns-core-modules/ui/switch/switch.android.ts
```diff
@@ -36,23 +36,22 @@ export class Switch extends SwitchBase {
...
+        initializeCheckedChangeListener();
+        const listener = new CheckedChangeListener(this);
+        nativeView.setOnCheckedChangeListener(listener);
+        (<any>nativeView).listener = listener;
     }
 
     public disposeNativeView() {
         const nativeView: any = this.nativeViewProtected;
         nativeView.listener.owner = null;
+        nativeView.listener = null;
```
MartoYankov: I think this is not necessary since we already removed the JS reference from it - the owner - and there is no reference to the native object from the JS object.

tns-core-modules/ui/segmented-bar/segmented-bar.android.ts
```diff
@@ -213,24 +213,23 @@ export class SegmentedBar extends SegmentedBarBase {
...
+        nativeView.setOnTabChangedListener(listener);
+        (<any>nativeView).listener = listener;
+        nativeView.setup();
         this._tabContentFactory = this._tabContentFactory || new TabContentFactory(this);
     }
 
     public disposeNativeView() {
         const nativeView: any = this.nativeViewProtected;
         nativeView.listener.owner = null;
+        nativeView.listener = null;
```
MartoYankov: I think this is not necessary since we already removed the JS reference from it - the owner - and there is no reference to the native object from the JS object.

tns-core-modules/ui/search-bar/search-bar.android.ts
```diff
@@ -95,32 +95,30 @@ export class SearchBar extends SearchBarBase {
...
-        nativeView.closeListener.owner = this;
-        nativeView.queryTextListener.owner = this;
     }
 
     public disposeNativeView() {
         const nativeView: any = this.nativeViewProtected;
         nativeView.closeListener.owner = null;
+        nativeView.closeListener = null;
         nativeView.queryTextListener.owner = null;
+        nativeView.queryTextListener = null;
```
MartoYankov: I think these are not necessary since we already removed the JS reference from it - the owner - and there is no reference to the native object from the JS object.

tns-core-modules/ui/list-view/list-view.android.ts
```diff
@@ -90,7 +85,9 @@ export class ListView extends ListViewBase {
         const nativeView = this.nativeViewProtected;
         nativeView.setAdapter(null);
         (<any>nativeView).itemClickListener.owner = null;
+        (<any>nativeView).itemClickListener = null;
         (<any>nativeView).adapter.owner = null;
+        (<any>nativeView).adapter = null;
```
MartoYankov: I think these is not necessary since we already removed the JS reference from them - the owner - and there is no reference to the native object from the JS object.

tns-core-modules/ui/list-picker/list-picker.android.ts
```diff
@@ -120,7 +111,9 @@ export class ListPicker extends ListPickerBase {
     public disposeNativeView() {
         const nativeView = this.nativeViewProtected;
         (<any>nativeView).formatter.owner = null;
+        (<any>nativeView).formatter = null;
         (<any>nativeView).valueChangedListener.owner = null;
+        (<any>nativeView).valueChangedListener = null;
```
MartoYankov: I think this is not necessary since we already removed the JS reference from it - the owner - and there is no reference to the native object from the JS object.

tns-core-modules/ui/image/image.android.ts
```diff
@@ -49,23 +49,22 @@ export class Image extends ImageBase {
...
+        initializeImageLoadedListener();
+        const nativeView = this.nativeViewProtected;
+        const listener = new ImageLoadedListener(this);
+        nativeView.setImageLoadedListener(listener);
+        (<any>nativeView).listener = listener;
     }
 
     public disposeNativeView() {
         (<any>this.nativeViewProtected).listener.owner = null;
+        (<any>this.nativeViewProtected).listener = null;
```
MartoYankov: I think this is not necessary since we already removed the JS reference from it - the owner - and there is no reference to the native object from the JS object.

tns-core-modules/ui/editable-text-base/editable-text-base.android.ts
```diff
@@ -156,34 +156,32 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
...
-        (<any>nativeView).listener.owner = this;
-        this._inputType = nativeView.getInputType();
+        this._inputType = editText.getInputType();
     }
 
     public disposeNativeView(): void {
-        super.disposeNativeView();
-        (<any>this.nativeViewProtected).listener.owner = null;
+        (<any>this.nativeTextViewProtected).listener.owner = null;
+        (<any>this.nativeTextViewProtected).listener = null;
```
MartoYankov: I think this is not necessary since we already removed the JS reference from it - the owner - and there is no reference to the native object from the JS object.

tns-core-modules/ui/date-picker/date-picker.android.ts
```diff
@@ -54,23 +54,23 @@ export class DatePicker extends DatePickerBase {
...
+        initializeDateChangedListener();
+        const nativeView = this.nativeViewProtected;
+        const listener = new DateChangedListener(this);
+        nativeView.init(this.year, this.month - 1, this.day, listener);
+        (<any>nativeView).listener = listener;
     }
 
     public disposeNativeView() {
         (<any>this.nativeViewProtected).listener.owner = null;
+        (<any>this.nativeViewProtected).listener = null;
```
MartoYankov: I think this is not necessary since we already removed the JS reference from it - the owner - and there is no reference to the native object from the JS object.

tns-core-modules/ui/button/button.android.ts
```diff
@@ -49,22 +49,23 @@ export class Button extends ButtonBase {
...
+        const clickListener = new ClickListener(this);
+        nativeView.setOnClickListener(clickListener);
+        (<any>nativeView).clickListener = clickListener;
     }
 
     public disposeNativeView() {
-        (<any>this.nativeViewProtected).clickListener.owner = null;
+        if (this.nativeViewProtected) {
+            (<any>this.nativeViewProtected).clickListener.owner = null;
+            (<any>this.nativeViewProtected).clickListener = null;
```
MartoYankov: I think this is not necessary since we already removed the JS reference from it - the owner - and there is no reference to the native object from the JS object.

tns-core-modules/ui/action-bar/action-bar.android.ts
```diff
@@ -131,21 +131,21 @@ export class ActionBar extends ActionBarBase {
...
+        const nativeView = this.nativeViewProtected;
+        initializeMenuItemClickListener();
+        const menuItemClickListener = new MenuItemClickListener(this);
+        nativeView.setOnMenuItemClickListener(menuItemClickListener);
+        (<any>toolbar).menuItemClickListener = menuItemClickListener;
     }
 
     public disposeNativeView() {
         (<any>this.nativeViewProtected).menuItemClickListener.owner = null;
+        (<any>this.nativeViewProtected).menuItemClickListener = null;
```
MartoYankov: I think this is not necessary since we already removed the JS reference from it - the owner - and there is no reference to the native object from the JS object.

farfromrefug: ok will clean up all of those then.
Thanks

tns-core-modules/ui/switch/switch.ios.ts
farfromrefug: I am not sure what lines you are talking about here. The comments show the deleted lines.
You are talking about the handler creation?

tns-core-modules/ui/tab-view/tab-view.ios.ts
farfromrefug: Not really but I kind of wanted to respect the way it is done for now in Page.
The question is do we need to have access to the controller even if the "View" (Page or TabView) is not yet added to the visual tree?

tns-core-modules/ui/time-picker/time-picker.ios.ts
farfromrefug: I didn't because you don't need to recreate it if you "unload" the "load" the view again.
Seemed to me more like a class variable.

tns-core-modules/ui/switch/switch.ios.ts
MartoYankov: I'm talking about the `this.width` and `this.height` assignments. They are not connected to the native view though. I'm not sure where they should be initialized. Probably it's good in the constructor for now.

tns-core-modules/ui/tab-view/tab-view.ios.ts
MartoYankov: Fair enough, I will take a look into it after this PR is merged.

tns-core-modules/ui/switch/switch.ios.ts
farfromrefug: Yeah I think it is better In the constructor. It s default width and height.
If you put it in the init or create it will override any width /height set in CSS

MartoYankov:
test

MartoYankov:
test

MartoYankov:
@farfromrefug It appears there are various errors due to undeclared variables and such. These are caught by the unit tests app. You can run the app following these instructions - https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md . Currently, many tests are failing so it will be better to run them by category. You can do that by commenting out categories in `/tests/app/testRunner.ts`. The `allTests` const is holding all the test categories. Don't forget to run `tsc` after you comment the categories.

I will make comments about a couple of issues I found by running the tests. Let me know if you have trouble looking into the others.

tns-core-modules/ui/action-bar/action-bar.android.ts
```diff
@@ -131,17 +131,16 @@ export class ActionBar extends ActionBarBase {
...
     }
 
     public initNativeView(): void {
         super.initNativeView();
-        (<any>this.nativeViewProtected).menuItemClickListener.owner = this;
+        const nativeView = this.nativeViewProtected;
+        initializeMenuItemClickListener();
+        const menuItemClickListener = new MenuItemClickListener(this);
+        nativeView.setOnMenuItemClickListener(menuItemClickListener);
+        (<any>toolbar).menuItemClickListener = menuItemClickListener;
```
MartoYankov: `toolbar` is not declared. Should be `nativeView`.

farfromrefug:
@MartoYankov thanks don't spend too much time on this. It s mine to handle ! ;)

tns-core-modules/ui/button/button.android.ts
```diff
@@ -49,22 +49,22 @@ export class Button extends ButtonBase {
 
     @profile
     public createNativeView() {
-        initializeClickListener();
-        const button = new AndroidButton(this._context);
-        const clickListener = new ClickListener(this);
-        button.setOnClickListener(clickListener);
-        (<any>button).clickListener = clickListener;
-        return button;
+        return new AndroidButton(this._context);
```
MartoYankov: The `AndroidButton` type is assigned inside `initializeClickListener()`, so after we moved it to `initNativeView()`, this is not declared. I think it will be safe to use the `android.widget.Button` constructor directly.

farfromrefug: I will actually define AndroidButton. I think the point is not to always have to go through ```android.widget.Button```?
Might be faster?

farfromrefug:
@MartoYankov I fixed the issues. Thanks! There is still a test error with the slider. Something I did not change. The fix is easy. Do you want to fix it? (test for owner existence).

MartoYankov:
@farfromrefug I fixed the slider android native view owner assignment and indeed all unit tests now pass. However, there are still a lot failing on iOS. Can you take a look at them?

farfromrefug:
@MartoYankov dumb me! I forgot to run the iOS one :s sorry ....

farfromrefug:
@MartoYankov I fixed most of the issues.
Now some tests are still failing which are due to the fact that now the nativeViewProtected is not created in the constructor.
Some are actually a bit tricky to "understand"

For example the test 'test_scrollToHorizontalOffset_with_animation" is failing because we don't get the expected scroll.
But I am struggling to understand why we are actually expecting such a value. Like [here](https://github.com/Akylas/NativeScript/blob/9908b4b60a87d2241439cec1108aeaf54a9f861b/tests/app/ui/scroll-view/scroll-view-tests.ts#L139) for example.

Still working on those!

farfromrefug:
@MartoYankov I should now have fixed all the errors! Sorry for the delay

tns-core-modules/ui/frame/frame.ios.ts
```diff
@@ -30,7 +30,11 @@ export class Frame extends FrameBase {
         super();
         this._ios = new iOSFrame(this);
         this.viewController = this._ios.controller;
-        this.nativeViewProtected = this._ios.controller.view;
+        this.nativeViewProtected = this.viewController.view;
```
MartoYankov: I think you should remove this line. If it stays, there is no need for `createNativeView`.

tns-core-modules/ui/scroll-view/scroll-view.ios.ts
```diff
@@ -38,10 +38,19 @@ export class ScrollView extends ScrollViewBase {
     private _contentMeasuredWidth: number = 0;
     private _contentMeasuredHeight: number = 0;
     private _delegate: UIScrollViewDelegateImpl;
+    
+    public createNativeView() {
+        const view = UIScrollView.new();
+        if (this.orientation === "horizontal") {
```
MartoYankov: I think this should be called in `initNativeView()`. The bonus is we can call `updateScrollBarVisibility()` method there.

tns-core-modules/ui/frame/frame.ios.ts
farfromrefug: Actually I am not sure. The reason I did that is for ```nativeViewProtected``` to be available as soon as the constructor is called. I think it's important from the frame.
Will test this

MartoYankov: I ran the tests and they worked. I'm also conflicted. I think for non UI components (Page, Frame), it might be better to keep the `nativeViewProtected` in constructor and omit `createNativeView`.

farfromrefug: Well I have both because if ```_tearDownUI``` you loose it without being able to get it back

tns-core-modules/ui/scroll-view/scroll-view.ios.ts
```diff
@@ -38,10 +38,19 @@ export class ScrollView extends ScrollViewBase {
...
+        const view = UIScrollView.new();
+        return view;
+    }
 
-    constructor() {
-        super();
-        this.nativeViewProtected = UIScrollView.new();
+    initNativeView() {
+        super.initNativeView();
+        if (this.orientation === "horizontal") {
```
MartoYankov: We can now remove this ìf` and call `this.updateScrollBarVisibility(this.scrollBarIndicatorVisibile)`.

tns-core-modules/ui/frame/frame.ios.ts
MartoYankov: Yes, the idea of `_tearDownUI` is to destroy the components, so that the garbage collection can release them. We are currently not recycling native views. I think for now you can remove this line and keep the one in `createNativeView()`.

MartoYankov:
test

MartoYankov:
test

MartoYankov:
test

MartoYankov:
@farfromrefug Some dialogs and modals UI tests are failing. I know you can't see them, so I'll take care of them.

farfromrefug:
@MartoYankov ok let me know If I can help in any way

MartoYankov:
test --ignore branch_ns_ui_sidedrawer_demo ios11 api23

MartoYankov:
@farfromrefug Thanks for the awesome PR. I hope the material components will come out nicely now.

farfromrefug:
@MartoYankov Thanks! Yes will update my plugin right now with this. Need some changes on the runtime though (might already be in right now) for that plugin to be published.

PR #6124, opened 2018-07-27
feat(styling): Added 2 functions to control applicationAdditionalSelectors

### LINKED ISSUE #5912: Application CSS no longer exposed
### Did you verify this is a real problem by searching the [NativeScript Forum]
yes

### Tell us about the problem
In NS 2.0  the application wide css variables where exposed so that my NS-Themes plugin could easily allow **dynmically** re-theming the application.  The user could chose any number of themes while running the application and it would apply them in real time.  Key factor, without messing up any other loaded css.  

In NS 4 (possibly in a 3.x version???) this value was moved was put into local variables of the Style-Scope.ts file https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/styling/style-scope.ts#L66

Without that ability anymore; I investigated the ability to change is the `localCssSelectors` 
https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/styling/style-scope.ts#L517 of the frame object.   However, these are NOT propagated to the children from the Frame (see: #5911) .  So basically with the new version of 4.x; I can't actually dynamically theme the entire app at runtime properly using either the frame or application value..     If I apply it to the page; then the theme is only applied to a single page.    If I use the `cssChanged` event to attempt to modify the application wide css; 
https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/styling/style-scope.ts#L239 then I will overwrite any "app.css" which is NOT the desired behavior.

I need either some method to access the `applicationAdditionalSelectors` or I need the Frame to propagate the CSS (see: #5911) to its child properly.    

Confirmed Android, iOS probably -- code is in common code files...

4.1 - Current latest release on everything.

https://github.com/NathanaelA/NativeScript-Themes

NathanaelA:
## PR Checklist

- [X] There is an issue for the bug/feature this PR is for.  #5912

Somewhere between version 2.x of NativeScript and 4.x of NativeScript, the Styling system was changed.  The application wide CSS selectors are no longer accessible from anything external to the style-scope file.  ([css selectors](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/styling/style-scope.ts#L69) are now a local variable)

## What is the new behavior?
I made a generic interface to be able to modify the `applicationAdditionalSelectors` basically I exposed two new functions off the StyleScope library;

1. `addTaggedAdditionalCSS` this function accepts any CSS text and can set an optional Tag on the imported rules.
2. `removeTaggedAdditonalCSS` this function will remove any CSS rules that were Tagged. 

This is a much better solution to solve #5912 (just exposing the `applicationAdditionalSelectors`) -- this will allow anyone to easily add more CSS text "globally" to the app that can be optionally removed at some point in the future -- because the rules are tagged; you can add any number of different rule sets, and remove any set of them on demand at a later time.

Fixes/Implements/Closes #5912

Please note to test the new functionality; you can use the new version of NativeScript-Themes and the demo it has.  (You'll have to run from master).

 This does not create any breaking changes; and unless the functions are actually utilized they do not have any impact on the existing code base.  If they are utilized then they will impact just the long term global selectors.

### DIFF
--- a/tests/app/ui/styling/style-tests.ts
+++ b/tests/app/ui/styling/style-tests.ts
@@ -9,7 +9,7 @@ import * as tabViewModule from "tns-core-modules/ui/tab-view";
 import * as helper from "../../ui/helper";
 import * as types from "tns-core-modules/utils/types";
 import * as viewModule from "tns-core-modules/ui/core/view";
-import { resolveFileNameFromUrl, removeTaggedAdditonalCSS, addTaggedAdditionalCSS } from "tns-core-modules/ui/styling/style-scope";
+import { resolveFileNameFromUrl, removeTaggedAdditionalCSS, addTaggedAdditionalCSS } from "tns-core-modules/ui/styling/style-scope";
 import { unsetValue } from "tns-core-modules/ui/core/view";
 import * as color from "tns-core-modules/color";
 
@@ -1480,10 +1480,11 @@ export function test_resolveFileNameFromUrl_unexisting_file() {
 }
 
 export function test_checkAddRemoveCSS() {
-    const css1 = "Label { color: #FF0000; }";
-    const css2 = "Label { color: #00FF00; }";
+    const css1 = "#test_checkAddRemoveCSS_label { color: #FF0000; }";
+    const css2 = "#test_checkAddRemoveCSS_label { color: #00FF00; }";
     const label = new labelModule.Label();
     label.text = "color coming from updated rules";
+    label.id = "test_checkAddRemoveCSS_label";
 
     helper.buildUIAndRunTest(label, function (views: Array<viewModule.View>) {
         const page = <pageModule.Page>views[1];
@@ -1499,11 +1500,13 @@ export function test_checkAddRemoveCSS() {
         helper.assertViewColor(label, "#00FF00");
 
         // Remove Green (Should revert to red, since we removed the green rule)
-        removeTaggedAdditonalCSS("green");
+        removeTaggedAdditionalCSS("green");
         page._onCssStateChange();
         helper.assertViewColor(label, "#FF0000");
-    });
 
+        //Cleanup
+        removeTaggedAdditionalCSS("red");
+    });
 }
 
 // <snippet module="ui/styling" title="styling">
--- a/tests/app/ui/styling/style-tests.ts
+++ b/tests/app/ui/styling/style-tests.ts
@@ -9,7 +9,7 @@ import * as tabViewModule from "tns-core-modules/ui/tab-view";
 import * as helper from "../../ui/helper";
 import * as types from "tns-core-modules/utils/types";
 import * as viewModule from "tns-core-modules/ui/core/view";
-import { resolveFileNameFromUrl } from "tns-core-modules/ui/styling/style-scope";
+import { resolveFileNameFromUrl, removeTaggedAdditonalCSS, addTaggedAdditionalCSS } from "tns-core-modules/ui/styling/style-scope";
 import { unsetValue } from "tns-core-modules/ui/core/view";
 import * as color from "tns-core-modules/color";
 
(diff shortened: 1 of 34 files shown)


dtopuzov:
test

NathanaelA:
@dtopuzov - Any idea why it failed the Jenkins core-modules-tests; there is nothing in this PR that failed any tests on my side.    I double checked the PR's after the initial one to make sure I didn't add a typo; and I don't see any reason why the tests should have failed.

NathanaelA:
Just a FYI, the https://github.com/NativeScript/theme/pull/148 PR that @shiv19 did (for the actual NativeScript / Theme repo is waiting on this PR.   So the sooner this PR is critiqued and accepted the sooner your own repo https://github.com/NativeScript/theme/ will work again in NS 4.  :grinning:

dtopuzov:
Hi @NathanaelA 

Failing unit test:
```
Test: VIEW-LAYOUT-EVENT.test_event_LayoutChanged_IsNotRaised_TransformChanged FAILED:  Actual: <0>(number). Expected: <1>(number)
```
It might be some flakiness in unittest, re-running it now.

NathanaelA:
Looks like it worked this time.   :grinning: 

---

Not that it is related to this PR but in
https://github.com/NativeScript/NativeScript/blob/master/tests/app/ui/view/view-tests-layout-event.ts#L119

Wouldn't the `TKUnit.waitUntilReady(() => button.height === 200, 5);` be better to be:
`TKUnit.waitUntilReady(() => stackLayoutChangedCount+buttonLayoutChagedCount > 1, 5);` because technically that is what you are waiting for; after 5 seconds then it will fail.   Checking if  button.height = 200 should be true almost instantly since you just assigned it, but the callback events might not have been called do to any number of timing factors.

NathanWalker:
@NathanaelA could you rebase latest master onto this PR?

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -262,6 +262,34 @@ class CSSSource {
...
+export function removeTaggedAdditonalCSS(tag: String | Number): Boolean {
+    let changed = false;
+    for (let i = 0; i < applicationAdditionalSelectors.length; i++) {
+        if (applicationAdditionalSelectors[i].tag === tag) {
+            applicationAdditionalSelectors.splice(i, 1);
+            i--;
+            changed = true;
+        }
+    }
+    return changed;
```
vakrilov: Probably you should call `mergeCssSelectors();` here again if there are changes.
It will regenerate the `applicationCssSelectors` and will increment `applicationCssSelectorVersion`.

vakrilov:
test

tns-core-modules/ui/styling/style-scope.ts
vakrilov: Awesome!
For to operations (remove and after that add) the perf penalty won't be that big. If there are cases that you want to have multiple operations with tagged selectors - we can probably create an API that can batch these and call `merge` only once at the end.

vakrilov:
test

vakrilov:
test

vakrilov:
test

SvetoslavTsenov:
test

tests/app/ui/styling/style-tests.ts
```diff
@@ -1480,10 +1480,11 @@ export function test_resolveFileNameFromUrl_unexisting_file() {
 }
 
 export function test_checkAddRemoveCSS() {
-    const css1 = "Label { color: #FF0000; }";
-    const css2 = "Label { color: #00FF00; }";
+    const css1 = "#test_checkAddRemoveCSS_label { color: #FF0000; }";
+    const css2 = "#test_checkAddRemoveCSS_label { color: #00FF00; }";
```
NathanaelA: The two tests above my test seemed to be flakey on my machine (they frequently failed while I was writing my test).
1. Doesn't `buildUIAndRunTest` reset everything each test?     So if so, what is the point of switching from `Label` to #test_checkAddRemoveCSS_Label?   (So I know for any future tests).

vakrilov: `buildUIAndRunTest` will clear everything on the page (page-level CSS too), but will not clear the application-wide CSS selectors.
I noticed that the registered tag selectors remain which was affection other tests especially when running the tests with livesync. Thats why I made them specific id-selectors ;)

nikolasdas:
Sorry if this is a dumb question... but do the two added functions need to be added to `style-scope.d.ts` as well?

PR #6139, opened 2018-08-01
chore(proxy-view-container): add property to store the root PVC

ADjenkov:
Required by [this nativescript-angular PR](https://github.com/NativeScript/nativescript-angular/compare/djenkov/nested-modals?expand=1)


SvetoslavTsenov:
test

SvetoslavTsenov:
test

vakrilov:
I think this one should go into `nativescript-angular`. We can just dynamically extend the ProxyViewContainer there. 
Having this property in the tns-core-modules does not add any value.

PR #6141, opened 2018-08-01
fix(android): suppress reflection for default animations

manoldonev:
Fixes `Error: java.lang.CloneNotSupportedException: Class android.support.v4.app.FragmentManagerImpl$AnimationOrAnimator doesn't implement Cloneable` in specific projects.

I was able to reproduce the problem when added the firebase plugin to a working hello-world app (cli next, android runtime next, tns-core-modules next). The root of the problem is the fact that different versions of the support library expose different API for android.support.v4.app.FragmentManagerImpl.loadAnimations(...) -- one returns Animation, the other AnimationOrAnimatior (that is not cloneable). Might be related to 
https://github.com/NativeScript/nativescript-cli/issues/3773 as we are really unsure which support library version is used (and why).

Related to #5785
Related to #6129 

BREAKING CHANGE

Before:
Default fragment enter animation was Android version specific

After:
Default fragment enter animation is now fade animation for all Android versions

You can customise the transition per navigation entry or globally via the [navigation transitions API](
https://docs.nativescript.org/core-concepts/navigation#navigation-transitions)

### DIFF
--- a/tns-core-modules/ui/frame/fragment.transitions.android.ts
+++ b/tns-core-modules/ui/frame/fragment.transitions.android.ts
@@ -50,17 +50,12 @@ interface ExpandedEntry extends BackstackEntry {
 
 const sdkVersion = lazy(() => parseInt(device.sdkVersion));
 const defaultInterpolator = lazy(() => new android.view.animation.AccelerateDecelerateInterpolator());
-const isAndroidP = lazy(() => sdkVersion() > 27);
 
 export const waitingQueue = new Map<number, Set<ExpandedEntry>>();
 export const completedEntries = new Map<number, ExpandedEntry>();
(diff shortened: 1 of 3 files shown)


SvetoslavTsenov:
test branch_ns_ng_animation_examples  branch_qsf

shiv19:
@manoldonev Just tried it out 😄 
This works like a charm 👍

dtopuzov:
test branch_ns_ng_animation_examples branch_qsf

dtopuzov:
test branch_ns_ng_animation_examples branch_qsf

dtopuzov:
test branch_ns_ng_animation_examples branch_qsf

Issue #6152, opened 2018-08-06
XHR -> Event not supported: progress


wendt88:
- tns version: 4.1.3
- @angular version: 6.0.6

adding property `reportProgress: true` to angular's `HttpClient` crashes with error `Event not supported: progress`

Found in `tns-core-modules/xhr/xhr.js` from line 131:
```
XMLHttpRequest.prototype.addEventListener = function (eventName, handler) {
        if (eventName !== "load" && eventName !== "error") {
            throw new Error("Event not supported: " + eventName);
        }
        var handlers = this._listeners.get(eventName) || [];
        handlers.push(handler);
        this._listeners.set(eventName, handlers);
};
```

By removing 
```
if (eventName !== "load" && eventName !== "error") {
    throw new Error("Event not supported: " + eventName);
}
```
it works!

By setting `reportProgress: true` angular adds a `progress` event to XHR, so it does not pass your check and throws that error.

tsonevn:
HI @wendt88,
This seems to be related to a missing feature. I will mark this as a feature request. For now, I would suggest keeping track of the issue for further info.

PR #6154, opened 2018-08-06
enabling reportProgress property for NativeScirpt Angular's HTTPClient

tsonevn:
enabling `reportProgress` property for NativeScirpt Angular's HTTPClient


tsonevn:
Related to issue https://github.com/NativeScript/NativeScript/issues/6152

PR #6164, opened 2018-08-08
fix(android/platform): reinitialise screen metrics on orientation change

### LINKED ISSUE #4588: Screen dimensions platform.android & platform.ios different
Dear friend, we, the rest of the NativeScript community really
appreciate your feedback! While we are doing all we can to take care of every
issue, sometimes we get overwhelmed. Because of that, we will consider issues
that are not constructive or problems that cannot be reproduced "dead".
Additionally, we will treat feature requests or bug reports with unanswered
questions regarding the behavior/reproduction for more than 20 days "dead". All
"dead" issues will get closed.

### Tell us about the problem
Please, ensure your title is less than 63 characters long and starts with a capital
letter.

Different behavior on iOS/Android so I just wanted to check if this was planned.

- CLI: 3.1.2
- Cross-platform modules: 3.1.0
- Runtime(s): ios 3.1.0 & android 3.1.1
- Plugin(s): see bottom for entire package.json

I was using:
`import { screen, device } from "platform";`
and in ngOnInit method:
`this.isLandscape = (screen.mainScreen.widthPixels > screen.mainScreen.heightPixels);`
to determine if phone is in landscape or portrait mode and noticed different behavior between iOS and Android. This calculation is always correct on the initial page of my app, but the difference appears when navigating to the second page and using the same code to determine landscape or portrait.

1) Starting on page 1 in portrait mode and then rotating phone to landscape.
2) Then navigate to page 2:
    a) iOS will correctly compute this.isLandscape
    b) Android will not

I believe the reason for this is [here](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/platform/platform.android.ts#L118-L124).

Another solution has been implemented in the app to avoid this, but I wanted to check with NS team if this is planned/expected behavior. Thanks for your time and support!

My package.json:
```
{
  "name": "",
  "version": "0.0.2",
  "description": "",
  "license": "SEE LICENSE IN <your-license-filename>",
  "readme": "NativeScript Application",
  "repository": "<fill-your-repository-here>",
  "nativescript": {
    "id": "",
    "tns-android": {
      "version": "3.1.1"
    },
    "tns-ios": {
      "version": "3.1.0"
    }
  },
  "dependencies": {
    "@angular/common": "4.1.3",
    "@angular/compiler": "4.1.3",
    "@angular/core": "4.1.3",
    "@angular/forms": "4.1.3",
    "@angular/http": "4.1.3",
    "@angular/platform-browser": "4.1.3",
    "@angular/platform-browser-dynamic": "4.1.3",
    "@angular/router": "4.1.3",
    "@ngrx/core": "1.2.0",
    "@ngrx/store": "2.2.2",
    "nativescript-angular": "3.1.1",
    "nativescript-camera": "3.0.0",
    "nativescript-google-analytics": "^0.3.7",
    "nativescript-imagepicker": "3.0.1",
    "nativescript-iqkeyboardmanager": "1.0.1",
    "nativescript-social-share": "1.3.2",
    "nativescript-telerik-ui": "3.0.1",
    "reflect-metadata": "^0.1.10",
    "rxjs": "5.4.2",
    "tns-core-modules": "3.1.0",
    "zone.js": "0.8.12"
  },
  "devDependencies": {
    "@angular/compiler-cli": "4.1.3",
    "@ngtools/webpack": "^1.3.0",
    "babel-traverse": "6.23.1",
    "babel-types": "6.23.0",
    "babylon": "6.16.1",
    "copy-webpack-plugin": "~3.0.1",
    "extract-text-webpack-plugin": "~2.0.0-beta.4",
    "lazy": "1.0.11",
    "nativescript-css-loader": "~0.26.0",
    "nativescript-dev-typescript": "^0.4.5",
    "nativescript-dev-webpack": "^0.3.6",
    "raw-loader": "~0.5.1",
    "resolve-url-loader": "~1.6.0",
    "tns-platform-declarations": "3.1.0",
    "typescript": "^2.3.4",
    "webpack": "2.2.0",
    "webpack-sources": "~0.1.3"
  },
  "scripts": {
    "ns-bundle": "ns-bundle",
    "start-android-bundle": "npm run ns-bundle --android --start-app",
    "start-ios-bundle": "npm run ns-bundle --ios --start-app",
    "build-android-bundle": "npm run ns-bundle --android --build-app",
    "build-ios-bundle": "npm run ns-bundle --ios --build-app"
  }
}
```

PeterStaev:
## PR Checklist

When getting the screen width/height on android those are the same, no matter if you rotate the device or not. Unlike iOS where the properties always reflect the actual device width/height. 

## What is the new behavior?
`platform.screen.mainScreen` width and height properties will always show the actual values like on iOS. 

Fixes #4588.

### DIFF
--- a/tns-core-modules/platform/platform.android.ts
+++ b/tns-core-modules/platform/platform.android.ts
@@ -101,16 +101,8 @@ class Device implements DeviceDefinition {
 
 class MainScreen implements ScreenMetricsDefinition {
     private _metrics: android.util.DisplayMetrics;
-    
-    constructor() {
-        appModule.on(appModule.orientationChangedEvent, () => {
-            if (this._metrics) {
-                this.initMetrics();
-            }
-        });
-    }
 
-    private cssChanged(args: appModule.CssChangedEventData): void {
+    private reinitMetrics(): void {
         if (!this._metrics) {
             this._metrics = new android.util.DisplayMetrics();
         }
@@ -125,7 +117,8 @@ class MainScreen implements ScreenMetricsDefinition {
     private get metrics(): android.util.DisplayMetrics {
         if (!this._metrics) {
             // NOTE: This will be memory leak but we MainScreen is singleton
-            appModule.on("cssChanged", this.cssChanged, this);
+            appModule.on("cssChanged", this.reinitMetrics, this);
+            appModule.on(appModule.orientationChangedEvent, this.reinitMetrics, this);
 
             this._metrics = new android.util.DisplayMetrics();
             this.initMetrics();
--- a/tns-core-modules/platform/platform.android.ts
+++ b/tns-core-modules/platform/platform.android.ts
@@ -102,6 +102,14 @@ class Device implements DeviceDefinition {
 class MainScreen implements ScreenMetricsDefinition {
     private _metrics: android.util.DisplayMetrics;
     
+    constructor() {
+        appModule.on(appModule.orientationChangedEvent, () => {
+            if (this._metrics) {
+                this.initMetrics();
+            }
+        });
+    }
+
     private cssChanged(args: appModule.CssChangedEventData): void {
         if (!this._metrics) {
             this._metrics = new android.util.DisplayMetrics();


PeterStaev:
@vakrilov , just to confirm, you mean to move the `cssChanged` sub in the constructor? I was thinking about this, but since I wasn't sure if there might be a reason for the delayed sub, I decided not to touch the existing code 😉

vakrilov:
Can you move the attaching the orientation changed event into the `private get metrics()` method. The idea here is that if nobody accesses the metrics filed - we won't listen to css/orientation events. As soon as someone does access them - we need to listen to those events so that we invalidate the cached metrics in case of an event.

PeterStaev:
Personally don't like subscriptions to happen inside a property since someone from outside could potentially nullify `this._metrics` and cause a double sub, but will see to change it a bit later.

vakrilov:
test

vakrilov:
@PeterStaev I can see your point and frankly - the constructor seems more logical place to put the subscriptions from a code-readability standpoint too. I think it is a kind of micro optimization to avoid subscribing if not necessary. The ` metrics()` is private so it should be enough from external tinkering (as far as anything can be safe in JS 😄 ).

vakrilov:
I checked the commits and it was moved to the dedicated property, because it was causing problems with generating android-snapshots. The snapshot should be OK now, even if we do the subscriptions are done in the constructor, because with webpack@4 we changed how the vendor-chunk is generated, but I don't think its worth it to move it again.

PR #6171, opened 2018-08-09
fix(gestures)propagate touch to parent so that gestures can work

farfromrefug:
## PR Checklist

Right now If a view has listening for swipe as a child listening for tap, you can't recognise the swipe on the child

## What is the new behavior?
swipe can be recognised on the child


farfromrefug:
Is something expected from me here?

vakrilov:
test

vakrilov:
Hi @farfromrefug - thanks for the awesome PR.
The code changes look good. 

Do you think there is a good way to test that to assure it works?
I'm not sure that you can simulate this in an unit test, but you can create a test page with the paren-child scenario in the test-app and our QAs can write an e2e(with appium) test against it. We test other gestures in a similar fashion. For example pan-gesture: [xml](https://github.com/NativeScript/NativeScript/blob/master/apps/app/ui-tests-app/events/pan-event.xml) and [ts](https://github.com/NativeScript/NativeScript/blob/master/apps/app/ui-tests-app/events/pan-event.ts).

Also, have you if the gestures are propagated correctly in IOS too?

farfromrefug:
@vakrilov thanks for the awesome framework !!!

Yes there is a easy test (which was the failure here) 

<view swipeEvent>
<button touchEvents/>
</view>

Because of the touchEvents on the button the ```onTouch``` method is called on the button. 
Without the PR, the view would not get the onTouch and thus the view would not handle the swipe.

The PR should fix that.

Let me know if you need more

vakrilov:
Hey @farfromrefug 
I have added a test page for your PR.

It consists of a parent grid with `swipe` handler and a child stack with `tap` handler. Without your fix, swiping inside the child stack does not trigger the `swipe` event of the parent.

I also remove the `view` param of the `handleGestureTouch` method as it was not really used.

farfromrefug:
@vakrilov perfect! Indeed the view arg is not needed .
All seems perfect to me!

Issue #6179, opened 2018-08-11
suspend/resume View: Crash - JS ERROR TypeError: undefined is not an object (evaluating 'view.isLoaded')


NathanWalker:
Put app in background then resume and this error can occur at anytime. Likely related to an `*ngIf` on a root component view which is attempting to retrigger change detection on app re-activation. The fix is very likely quite simple in the need to simply guard a fragile area of core modules which is doing this check.

### Which platform(s) does your issue occur on?

Both

### Please provide the following version numbers that your issue occurs with:
- CLI: 4.2.0
- Cross-platform modules: 4.2.0
- Runtime(s): 4.2.0

### Please tell us how to recreate the issue in as much detail as possible.

I will explore a way to create an exact repro of this issue however this is a race condition and not known to happen everytime and therefore will be dificult to repro 100%. Simple guard check around `if (view) view.isLoaded` will likely be all that's needed to fix this issue instead of simply assuming `view` would be defined at the expected moment in time.

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.

In a root app.component you could have something like the following:

```
<GridLayout>
  <page-router-outlet></page-router-outlet>
  <some-custom-component *ngIf="someCondition"></some-custom-component>
</GridLayout>
```

Here's the stack that can occur:
```
CONSOLE LOG file:///app/bundle.js:6882:24: suspendEvent
CONSOLE LOG file:///app/bundle.js:6882:24: resumeEvent
***** Fatal JavaScript exception - application has been terminated. *****
Native stack trace:
1   0x109b4cc1f NativeScript::reportFatalErrorBeforeShutdown(JSC::ExecState*, JSC::Exception*, bool, bool)
2   0x109b768ea NativeScript::FFICallback<NativeScript::ObjCMethodCallback>::ffiClosureCallback(ffi_cif*, void*, void**, void*)
3   0x10a543ced ffi_closure_unix64_inner
4   0x10a544642 ffi_closure_unix64
5   0x10f225a65 _CFXRegistrationPost
6   0x10f2257a1 ___CFXNotificationPost_block_invoke
7   0x10f1e7422 -[_CFXNotificationRegistrar find:object:observer:enumerator:]
8   0x10f1e65a1 _CFXNotificationPost
9   0x10ac20e57 -[NSNotificationCenter postNotificationName:object:userInfo:]
10  0x10b25a675 -[UIApplication _stopDeactivatingForReason:]
[...]

PR #6187, opened 2018-08-15
fix(view): isLoaded stability improvements

NathanWalker:
## PR Checklist

App's can randomly crash when various bindings update upon resuming an app dependent upon a wide array of possible view setups via developers.

## What is the new behavior?

Guards have been added to account for various troubling race conditions.

closes https://github.com/NativeScript/NativeScript/issues/6179

### DIFF
--- a/tns-core-modules/ui/frame/frame.android.ts
+++ b/tns-core-modules/ui/frame/frame.android.ts
@@ -9,7 +9,7 @@ import { Page } from "../page";
 import * as application from "../../application";
 import {
     FrameBase, stack, goBack, View, Observable,
-    traceEnabled, traceWrite, traceCategories
+    traceEnabled, traceWrite, traceCategories, traceError
 } from "./frame-common";
 
 import {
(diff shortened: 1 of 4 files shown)


MartoYankov:
test

tns-core-modules/ui/frame/frame.android.ts
```diff
@@ -718,7 +718,7 @@ class FragmentCallbacksImplementation implements AndroidFragmentCallbacks {
             this.frame._addView(page);
         }
 
-        if (frame.isLoaded && !page.isLoaded) {
+        if ((frame && frame.isLoaded) && (page && !page.isLoaded)) {
```
vakrilov: `page` is used couple of lines above - so no point of guarding it here. If it was `undefined` the code would have crashed [here](https://github.com/NativeScript/NativeScript/pull/6187/files#diff-c1ad77120477acf9bacdaabe6af2518dR705)

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -357,7 +357,7 @@ export class CssState {
      * As a result, at some point in time, the selectors matched have to be requerried from the style scope and applied to the view.
      */
     public onChange(): void {
-        if (this.view.isLoaded) {
+        if (this.view && this.view.isLoaded) {
```
vakrilov: What are the cases when you get error here?
The `view` is passed in the constructor and never changed. 
Maybe we should add a check there?

tns-core-modules/ui/frame/frame.android.ts
NathanaelA: I suspect that because frame can be null here; that he just safe guarded both at the same time.   Probably should also have the "page" test on the lines above it too, in case that is ever null.

tns-core-modules/ui/styling/style-scope.ts
NathanaelA: Well, onChange is called because of css changes; if css is in the process of changing and the view is being de-allocated it will crash because this.view = null.

NathanWalker: That’s exactly right @nathanaela

tns-core-modules/ui/frame/frame.android.ts
NathanWalker: @vakrilov @NathanaelA I updated the handling to ensure of number of possible errors that could occur due to timing and null/undefined - also added trace handling to ensure the log is notified in event of such conditions.

NathanWalker:
Worth noting I am using these changes in a custom pack of core modules in a current project to solve the issues reported and is working 100% without error with these in place.

MartoYankov:
We should consider merging this in the `release` branch.

tns-core-modules/ui/styling/style-scope.ts
vakrilov: I'm still struggling to get this one. The `view` is private and is never set (except from the constructor).
If there is a problem with trying to apply styles on disposed view - this check will do nothing more. The view will be disposed, but its `CssState` will still hold the reference and so `this.view` will not be null.

vakrilov:
Hey @NathanWalker and @NathanaelA. I will cherry pick the commit and open a PR in release, so that we can include it in a patch release. 

There are still some changes that don't really make sense to me and I'm trying to understand if you have crash reports related to them (the stack trace form the issue points only to the `loadView` method in the `view-base.ts`)?

vakrilov:
closing in favor of @6255

PR #6190, opened 2018-08-16
added support to clipToBounds for android

farfromrefug:
## PR Checklist

no ```clipToBounds``` support on android
## What is the new behavior?
```clipToBounds``` support on android

If you run the ui test as it is on android the view will still be clipped. This is because I set a padding property on the non clipping view. 
I you remove the padding it will not be clipped.

I added this on purpose cause it shows an issue with the padding on android. It seems like padding actually cause children to be limited in size to the parent size.
I don't think this should happen and it does not on iOS.

### CODE REVIEW

File: tns-core-modules/ui/core/view/view.android.ts [hunk 1]
```diff
@@ -822,6 +822,12 @@ export class CustomLayoutView extends View implements CustomLayoutViewDefinition
             if (child instanceof View) {
                 this._updateNativeLayoutParams(child);
             }
+
+            //if a child view asks to not be clipped we must set it up for us too
+            if (child.style.clipToBounds === false) {
+                console.log("adding child not clipping");
```
shiv19: is this console log necessary?

manoldonev:
test

tns-core-modules/ui/styling/style-properties.ts
```diff
@@ -1098,4 +1098,7 @@ export const visibilityProperty = new CssProperty<Style, Visibility>({
         target.view.isCollapsed = (newValue === Visibility.COLLAPSE);
     }
 });
-visibilityProperty.register(Style);
\ No newline at end of file
+visibilityProperty.register(Style);
+
+export const clipToBoundsProperty = new CssProperty<Style, boolean>({ name: "clipToBounds", cssName: "clip", valueConverter: booleanConverter });
```
manoldonev: I find it a bit confusing to introduce such `clip` css property and don't think it is a good idea:
- Web css `clip` lets you specify rectangle to clip an absolutely positioned element in the form `rect (top, right, bottom, left)`
- Web css `clip` property is deprecated in favor of `clip-path` that we already have.

If clipToBounds is to be exposed as a css property I think a better candidate would be the [css overflow property](https://www.w3schools.com/cssref/pr_pos_overflow.asp) with supported values "visible" and "hidden".

tns-core-modules/ui/layouts/layout-base-common.ts
```diff
@@ -105,7 +105,16 @@ export class LayoutBaseCommon extends CustomLayoutView implements LayoutBaseDefi
         this.style.paddingLeft = value;
     }
 
-    public clipToBounds: boolean;
+    get clipToBounds(): boolean {
+        return this.style.clipToBounds;
+    }
+    set clipToBounds(value: boolean) {
+        //we should never get a string as the booleanConverter is used in the property. Though it happens ...
```
manoldonev: This comment is unnecessary -- I would suggest to remove it.

tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -836,6 +836,11 @@ export class CustomLayoutView extends View implements CustomLayoutViewDefinition
             if (child instanceof View) {
                 this._updateNativeLayoutParams(child);
             }
+
+            //if a child view asks to not be clipped we must set it up for us too
```
manoldonev: Add white space after the //

tns-core-modules/ui/layouts/layout-base.d.ts
```diff
@@ -92,10 +92,4 @@ export class LayoutBase extends CustomLayoutView {
      */
     paddingTop: Length;
 
-    /**
-     * Gets or sets a value indicating whether to clip the content of this layout.
-     */
-    clipToBounds: boolean;
```
manoldonev: This shouldn't be removed from the d.ts file.

tns-core-modules/ui/layouts/layout-base.ios.ts
```diff
@@ -21,8 +21,8 @@ export class LayoutBase extends LayoutBaseCommon {
     }
 
     _setNativeClipToBounds() {
-        if (this.clipToBounds) {
-            this.nativeViewProtected.clipsToBounds = true;
+        if (this.style.clipToBounds !== undefined) {
+            this.nativeViewProtected.clipsToBounds = !!!this.style.clipToBounds;
```
manoldonev: Triple negation breaks this -- should be `!!this.style.clipToBounds`

tns-core-modules/ui/layouts/layout-base.android.ts
```diff
@@ -52,4 +36,16 @@ export class LayoutBase extends LayoutBaseCommon {
     [paddingLeftProperty.setNative](value: Length) {
         org.nativescript.widgets.ViewHelper.setPaddingLeft(this.nativeViewProtected, Length.toDevicePixels(value, 0) + Length.toDevicePixels(this.style.borderLeftWidth, 0));
     }
+
+    [clipToBoundsProperty.setNative](value: boolean) {
+        if (this.nativeViewProtected instanceof android.view.ViewGroup) {
+            const viewGroup = this.nativeViewProtected as android.view.ViewGroup;
```
manoldonev: Generally we use <> casting instead of `as` in our codebase. Please update to:
``` ts
const viewGroup = <android.view.ViewGroup>this.nativeViewProtected;
```

tns-core-modules/ui/layouts/layout-base.android.ts
```diff
@@ -52,4 +36,16 @@ export class LayoutBase extends LayoutBaseCommon {
...
+
+            viewGroup.setClipChildren(value);
+            if (!value && viewGroup.getParent() instanceof android.view.ViewGroup) {
```
manoldonev: I think `!value` should not be part of the if-statement -- if this property was first set to false, and then to true we would like to call `setClipChildren(true)` as well.

tns-core-modules/ui/layouts/layout-base.android.ts
```diff
@@ -52,4 +36,16 @@ export class LayoutBase extends LayoutBaseCommon {
...
+                (viewGroup.getParent() as android.view.ViewGroup).setClipChildren(value);
```
manoldonev: (<android.view.ViewGroup>viewGroup.getParent()).setClipChildren(value);

tns-core-modules/ui/layouts/layout-base.android.ts
```diff
@@ -52,4 +36,16 @@ export class LayoutBase extends LayoutBaseCommon {
...
+            }
+        }
+    }
+
```
manoldonev: No blank line here.

apps/app/ui-tests-app/layouts/clip.ts
```diff
@@ -0,0 +1,7 @@
+﻿import * as model from "~/ui-tests-app/layouts/myview";
```
manoldonev: Should be `import * as model from "./myview";`

manoldonev:
As for the issue with the padding -- I think you need to set both `setClipChildren(...)` and `setClipToPadding(...)` properties for Android to work as you would expect i.e. set them both to false if you would like content to overflow. See https://developer.android.com/reference/android/view/ViewGroup#attr_android:clipToPadding for more info.

manoldonev:
test

manoldonev:
Can be reopened once comments are addressed.

PR #6228, opened 2018-08-29
refactor: remove ununsed Layout class

manoldonev:
I came upon the Layout class by accident and I think it should be removed as it is not used anywhere in NativeScript besides in the unit tests app (where it can and should be replaced by `LayoutBase`). I could not find any clues or user guidelines that would describe how to use this class so I think it is some kind of legacy that we do not need any more. Code comments note that it is "Base class for all views that supports children positioning in cross platform manner" but that is not (no longer?) true as all built-in layouts inherit from LayoutBase/CustomLayoutView directly.

BREAKING CHANGE:  

Layout class (`tns-core-modules/ui/layouts/layout`) is now removed as it is not used in {N} framework any more.


MartoYankov:
Just to note that the class that is "Base class for all views that supports children positioning in cross platform manner" is now `CustomLayoutView`.

manoldonev:
@MartoYankov LayoutBase is CustomLayoutView btw... we have some complex hierarchy.

manoldonev:
test

PR #6266, opened 2018-09-14
docs: gather snippets from apps app folder only

MartoYankov:
Right now the docs build fails if you have already built either the /apps or /tests apps.


MartoYankov:
test --ignore

Issue #6292, opened 2018-09-21
interact with nested fragments only thru child fragment manager


manoldonev:
When interacting with nested fragments instead of using getSupportFragmentManager we must always use getChildFragmentManager instead. Failure to do so causes obscure issues on app suspend / resume (inner fragment unloads on app resume) due to the fact that inner fragments cannot be removed properly as their parent fragment has already been removed. Moreover, it seems that when child fragment manager is used Android is smart enough to remove child fragments automatically when parent fragment is removed.

PR #6293, opened 2018-09-21
fix: nested fragments interact thru child fragment manager

manoldonev:
Fixes https://github.com/NativeScript/NativeScript/issues/6292

~~TODO: review tab fragments' logic with respect to this change (and how did it work so far in the first place).~~

### DIFF
--- a/tns-core-modules/ui/core/view/view.android.ts
+++ b/tns-core-modules/ui/core/view/view.android.ts
@@ -3,10 +3,10 @@ import { Point, CustomLayoutView as CustomLayoutViewDefinition, dip } from ".";
 import { GestureTypes, GestureEventData } from "../../gestures";
 // Types.
 import {
-    ViewCommon, layout, isEnabledProperty, originXProperty, originYProperty, automationTextProperty, isUserInteractionEnabledProperty,
-    traceEnabled, traceWrite, traceCategories, traceNotifyEvent,
+    Color, EventData, ViewCommon, layout, getAncestor, automationTextProperty,
+    isEnabledProperty, isUserInteractionEnabledProperty, originXProperty, originYProperty,
     paddingLeftProperty, paddingTopProperty, paddingRightProperty, paddingBottomProperty,
-    Color, EventData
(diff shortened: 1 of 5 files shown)


tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -20,7 +20,7 @@ import {
 
 import { Background, ad as androidBackground } from "../../styling/background";
 import { profile } from "../../../profiling";
-import { topmost } from "../../frame/frame-stack";
+import { topmost } from "../../frame";
```
MartoYankov: If I remember correctly, the reason for the separate `frame-stack` file and import was to avoid circular dependencies. Is this problem fixed now?

tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -269,16 +269,36 @@ export class View extends ViewCommon {
...
-                    // the case is needed because _dialogFragment is on View
-                    // but parent may be ViewBase.
-                    view = view.parent as View;
                 }
+
+                // - frame1 -> frame2 (frame2 uses frame1 CHILD fm)
+                // - tabview -> frame1 (frame1 uses tabview item CHILD fm)
+                // - frame1 -> tabview (tabview uses frame1 CHILD fm)
+                // - frame1 -> tabview -> frame2 (tabview uses frame1 CHILD fm; frame2 uses tabview item CHILD fm)
+                if (view.typeName === "Frame" || view.typeName === "TabView") {
```
MartoYankov: I think the `typeName` will be wrong (and won't work) with uglify. The other property `cssType` that is set via the decorator might work, but it doesn't look right to use it.

What do you think about having a method `_hasFragments()` that returns true for `TabView` and `Frame` ?

tns-core-modules/ui/core/view/view.android.ts
manoldonev: Not sure but I'll revert that.

manoldonev:
test

manoldonev:
test

manoldonev:
test

PR #6336, opened 2018-10-02
feat: Add __assign ts-helper to support using spread operator

### LINKED ISSUE #4434: Execution of JS code stops when using Destructuring assignment(NativeScript + Angular)

I am using **TypeScript**
When using the destructuring assignment in JS, any code below that line will not be executed.
```typescript
  console.log('About to destructure an object')
  const user = { ...response, token: undefined }
  console.log('hi')
```

Compiled JS:
```
  var user = __assign({}, response, { token: undefined }); // Where is __assign defined?
```

Output:
```bash
JS: About to destructure an object
```

It's not just with `console.log`, simply any code is not executed, but no errors are thrown(on Android).

Fix with:
```typescript
  console.log('About to destructure an object')
  const user = response
  user.token = undefined
  console.log('hi')
```

Compiled JS:
```
  var user = response;
  user.token = undefined;
```

Output:
```bash
JS: About to destructure an object
JS: hi // works
```

Yeah, nothing there

### Tell us about the problem
Cannot use destructuring assignment in code. Although there is an easy workaround, it is not very obvious. It took me a couple of hours to go back and forth to the faulty code attempting to fix my app.

On Android, the app **DOES NOT** throw any errors, but code does not get executed below that line
On iOS, the app will crash of segmentation fault(not sure if this is really caused by the destructuring assignment, might be an emulator bug) **<- I'll get back to this**.

- CLI: 3.0.3
- Cross-platform modules: 3.0.1
- Runtime(s): 3.0.1 (Android + iOS)
- Plugin(s): The issue is not related to a plugin

To replicate my situation, paste in the code below into a button's `(tap)` event.

Simply copy this code, preferably in a button's `(tap)` event:

`response` is a response object from server. It contains some user data and the token:
```
const response = {
  name: 'Jane',
  last_name: 'Doe',
  token: 'ewqrewru0r7eyr87y87y43287r832yr8uye8uh82en8rxh8'
}
```

```typescript
  console.log('About to destructure an object')
  const user = { ...response, token: undefined }
  console.log('hi')
```

vakrilov:
Add `__assign` typescript helper function to support using spread operator.
Resolves #4434

### DIFF
--- a/tests/app/testRunner.ts
+++ b/tests/app/testRunner.ts
@@ -33,6 +33,9 @@ export function isRunningOnEmulator(): boolean {
 
 export const allTests = {};
 
+import * as tsTests from "./typescript/ts-tests";
+allTests["TYPESCRIPT"] = tsTests;
+
 import * as domNodeTest from "./debugger/dom-node-tests";
 allTests["DOM-NODE"] = domNodeTest;
 
--- a/tests/app/typescript/ts-tests.ts
+++ b/tests/app/typescript/ts-tests.ts
@@ -1,0 +1,9 @@
+import * as TKUnit from "../TKUnit";
+
+export function test_typescript_spread_operator() {
+    const a = { first: "test1", second: "test2" };
+    const b = { ...a, second: "newValue" };
+
+    TKUnit.assertEqual(b.first, "test1");
+    TKUnit.assertEqual(b.second, "newValue");
+}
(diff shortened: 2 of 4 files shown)


vakrilov:
test

vakrilov:
Closing in favor of #6351

PR #6337, opened 2018-10-02
fix-next(flexbox): flex end for justify content

vchimev:
Fix justify content calculations for flex end value of flexbox layout.


SvetoslavTsenov:
test

vchimev:
test

PR #6338, opened 2018-10-02
fix: CLAuthorizationStatus order and values

zbranzov:


vakrilov:
for reference: https://developer.apple.com/documentation/corelocation/clauthorizationstatus?language=objc

zbranzov:
run ci

PR #6353, opened 2018-10-04
fix: Clear the "view" to avoid left over referenced once the View is detached and should be collected by GC

VladimirAmiorkov:
## PR Checklist

- [ ] Tests for the changes are included - Due to having to profile the JS heap I am not sure how we can efficiently make a unit test for that

When using the Angular `createEmbeddeedView` from the `ViewContainerRef`, after calling `clear()` the created NativeScript Views are left in the memory

## What is the new behavior?
When using the Angular `createEmbeddeedView` from the `ViewContainerRef`, after calling `clear()` the created NativeScript Views are collected by the GC and a no longer in the memory

https://github.com/telerik/nativescript-ui-feedback/issues/825

### DIFF
--- a/tns-core-modules/ui/styling/style-scope.ts
+++ b/tns-core-modules/ui/styling/style-scope.ts
@@ -377,6 +377,9 @@ export class CssState {
 
     public onUnloaded(): void {
         this.unsubscribeFromDynamicUpdates();
+        this.view = undefined;
+        this._matchInvalid = false;
+        this._match = undefined;
     }
 
     @profile


tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -377,6 +377,9 @@ export class CssState {
 
     public onUnloaded(): void {
         this.unsubscribeFromDynamicUpdates();
+        this.view = undefined;
+        this._matchInvalid = false;
```
facetious: This variable is poorly named. "Match invalid = false" implies that the match is valid, but you're about to set the match to `undefined`, which strikes me as ultimately invalid.

NathanaelA: @facetious - The_matchInvalid is just another value used by the styling class.   Probably used as an optimization during css parses for rule checks.  Since Styling is actually one of the frequently ran classes --  He is clearing it because this is the "onUnloaded" (meaning the item is being removed from the view).

SvetoslavTsenov:
@VladimirAmiorkov, could you please rebase your branch on master, since there are a lot of important fixes.

vakrilov:
Hey @VladimirAmiorkov 
1. Can you elaborate a bit on what is holding a reference to the `CssState` in this scenario?
2. Any ideas on how this fix can be tested so that we ensure that there are no regressions in the future.
3. The `view` field is set in the constructor. It is possible that the view is loaded/unloaded multiple times. If you clear the reference in the `unloaded` -> that means that next time the view is `loaded` again it the `CSSState` associated to it will not  work as expected. You should either set the view again in `loaded` or clear the `view` field when you are sure the view will no longer be loaded again.

VladimirAmiorkov:
Hi @vakrilov ,

After researching this issue today I am unable to observe it even without any fixes in the `tns-core-modules`. Currently the only fixes that are required for [this issue](https://github.com/telerik/nativescript-ui-feedback/issues/825) are [this](https://github.com/NativeScript/nativescript-angular/pull/1542) PR in `nativescript-angular` and an fix in the `nativescript-ui-listview`.

I suggest we close this PR for now since I can no longer observe an memory leak when working with the latest version (**^4.2.1**) of _tns-core-modules_ and/or with the next version.

VladimirAmiorkov:
On a side note:

It may be nothing but I noticed that in the CssState class we are subscribing to one event [here](https://github.com/NativeScript/NativeScript/blob/e53bdd4f473b6c0b9dcf9b1838dc8a840cc9f09f/tns-core-modules/ui/styling/style-scope.ts#L489) and unsubscribing from a different one [here](https://github.com/NativeScript/NativeScript/blob/e53bdd4f473b6c0b9dcf9b1838dc8a840cc9f09f/tns-core-modules/ui/styling/style-scope.ts#L509), meaning we subscribe to "(attribute)Change" but unsubscribe from "onPropertyChanged:(attribute). This looks strange but I was not able to test it to see if it could lead to some sort of leak. If you know how I could test this part of the code let me know.

Issue #5744, opened 2018-04-25
iOS: issue when trying to chain modal pages within promise


felix-idf:
### Did you verify this is a real problem by searching the [NativeScript Forum]
_YES_

### Which platform(s) does your issue occur on?
iOS

### Tell us about the problem
I try to show multiple independent modal pages one after the other using a promise:
```
new Promise(function(resolve, reject) {
        application.getRootView().showModal(
            "modal-page",
            null,
            function() {
                resolve(true);
            },
            false
        );
    }).then(function(response) {
        if(application.getRootView().modal) {
            application.getRootView().closeModal();
        } else {
            console.log("closing modal page is not neccessary, as there is no open modal page...");
        }

        try {
            application.getRootView().showModal(
                "modal-page",
                null,
                function() {
                    var topmost = frameModule.topmost();
                    if(topmost) {
                        topmost.navigate({
                            moduleName: "second-page",
                            animated: true
                        });
                    }
                },
                false
            );
        } catch(e) {
            console.log(e);
        }
    });
```
It opens the first modal page and when I close it to see the next, I always get the following error mesage:
`Error: Parent page is not part of the window hierarchy. Close the current modal page before showing another one!`

I am not sure if this is the expected behaviour, so I need to know what I am doing wrong. Please note, that in our real app there are some conditions for showing the modal pages, so it might be possible to skip one of the modal pages or even all of them.

### Please provide the following version numbers that your issue occurs with:
- CLI: 4.0.0
- Cross-platform modules: next (4.0.0-2018-04-25-02)
- Runtime(s): 4.0.2

### Please tell us how to recreate the issue in as much detail as possible.
1. Start the app
2. Tap the button to open the first modal
3. tap the button to close this modal, so the second modal should be opened, but this does not happen on iOS.

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.
[NSIssues4.zip](https://github.com/NativeScript/NativeScript/files/1947861/NSIssues4.zip)

PR #6370, opened 2018-10-08
fix(iOS): viewhierarchy stability with modals

NathanWalker:
## PR Checklist

Apps can get a "white screen of death" due to invalid and improper viewController handling on iOS related to modals.

## What is the new behavior?

Ensures that a viewController is properly closed in the case one needs to be before opening to avoid app error.

The issue here relates to {N} + Angular however it's tied to this problem in core modules ultimately.
closes https://github.com/NativeScript/nativescript-angular/issues/1546

### DIFF
--- a/tns-core-modules/ui/core/view/view.ios.ts
+++ b/tns-core-modules/ui/core/view/view.ios.ts
@@ -1,4 +1,4 @@
-// Definitions.
+﻿// Definitions.
 import { Point, View as ViewDefinition, dip } from ".";
 import { ViewBase } from "../view-base";
 import { booleanConverter, Property } from "../view";
@@ -8,8 +8,9 @@ import {
     traceEnabled, traceWrite, traceCategories, traceError, traceMessageType, getAncestor
 } from "./view-common";
 
-import { ios as iosBackground, Background } from "../../styling/background";
+import { topmost } from "../../frame/frame-stack";
 import { ios as iosUtils } from "../../../utils/utils";
+import { ios as iosBackground, Background } from "../../styling/background";
 import {
     Visibility,
     visibilityProperty, opacityProperty,
@@ -362,6 +363,8 @@ export class View extends ViewCommon {
     }
 
     protected _showNativeModalView(parent: View, context: any, closeCallback: Function, fullscreen?: boolean, animated?: boolean, stretched?: boolean) {
+      const that = this;
+      const superCall = super._showNativeModalView;
         const parentWithController = ios.getParentWithViewController(parent);
         if (!parentWithController) {
             traceWrite(`Could not find parent with viewController for ${parent} while showing modal view.`,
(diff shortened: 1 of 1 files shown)


NathanWalker:
This PR is not cleanest solution however intended to provide suggestion on fix which core team may can address in best way possible. This modification has been tested pretty rigorously and has been found to be the type of fix needed to solve the issue mentioned.

NathanWalker:
Ideally we could use https://developer.apple.com/documentation/uikit/uiviewcontroller/1621505-dismissviewcontrolleranimated with the completion handler to ensure it's fully closed in that particular condition. Not sure if worth the trouble since usually this particular condition just needs to fully close (with **no** animation) before opening again but perhaps the completion handler's could be used to tighten all this up and avoid race conditions here.

MartoYankov:
@NathanWalker Indeed this fix doesn't look clean. It's not clear how the problem of a `viewController` not having a `view` is fixed by closing a modal view on the same `viewController`. Perhaps this fixes a certain scenario, but it's not immediately clear what it is.

My guess is that you are trying to open a modal view on a target that already opened a modal view. The iOS framework doesn't allow this by default and we are currently not supporting this. If indeed this is the case, we can discuss further if this is indeed a desirable behavior.

vakrilov:
Hey @NathanWalker 

I have done some tests and managed to reproduce the `"Close the current modal page before showing another one!"` error consistently in 2 scenarios:

### 1. Open a modal page in the `closeCallback` of another modal page
The issue here is that we fire the `closeCallback` *before* the closing animation is completed. As you suggested - I tried using the [`dismissViewControllerAnimatedCompletion`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621505-dismissviewcontrolleranimated) and fire the callback in the completion handler. This seems to resolve the issue. 🎉 
Also related to #5744

### 2. Open a modal page while another one is opened
The issue here is the same as before - there is already modally presented view controller, so we cannot present another one. The solution might be to have a parameter `forceCloseOpenedModal` in the `showModal()` method. The implementation should trigger a close on the current modal and use the completion callback (as in the previous case) to show the new modal. The value of this "force" param can also be true by default, which will make it behave much like what you are aiming with this PR.

[Here](https://github.com/NativeScript/NativeScript/tree/async-modals) is the branch that I have been experimenting with:
- Problem 1 is resolved thanks to [`dismissViewControllerAnimatedCompletion`](https://github.com/NativeScript/NativeScript/commit/b28837cde73897e2347e1b6d3aec385230c2351a#diff-2775a71872c1630de8108bf058a1628eR438)
- There are repro of both problems 1. and 2. in [the test-app](https://github.com/NativeScript/NativeScript/commit/b28837cde73897e2347e1b6d3aec385230c2351a#diff-930d331ce60f8ce090318a4b2f1fb5efR26).

## Next Steps
This PR triggered a good discussion. One thing is sure - `setTimeout(openNow(), 300);` is not the right approach here - that's why I'm closing this PR.
What's next:
1. Work on a PR based on [`dismissViewControllerAnimatedCompletion`](https://github.com/NativeScript/NativeScript/tree/async-modals) to resolve #5744 (problem 1)
2. Work on on the `forceClose` option (problem 2). We can set it to `true` by default or make angular always pass true when showing modals. 
3. Once the above are done. Let's see if there are more "view-hierarchy stability" issues that developers can easily hit.

Your help is much appreciated if you want to contribute on this plan.

NathanWalker:
@vakrilov hooray! this is great news thank you so much for digging into this one 🤗 
Love your proposal 1 & 2 in `next steps` - I'll definitely try to help on those when I can.

PR #6379, opened 2018-10-10
fix-next: drawer plugin viewcontroller init

MartoYankov:
Fixes the RadSideDrawer appearing black in `next`. The problem was caused by https://github.com/NativeScript/NativeScript/pull/6102 . The issue was that we waited for the native views to be instantiated before we create the wrapper UILayoutViewController. RadSideDrawer checks its modes when its native view is created. It requires the view controller to be created before the native view.

### DIFF
--- a/tns-core-modules/application/application.ios.ts
+++ b/tns-core-modules/application/application.ios.ts
@@ -227,6 +227,8 @@ class IOSApplication implements IOSApplicationDefinition {
             this._rootView._onRootViewReset();
         }
         const rootView = createRootView(view);
+        const controller = getViewController(rootView);
+
         this._rootView = rootView;
 
         if (createRootFrame.value) {
@@ -236,7 +238,7 @@ class IOSApplication implements IOSApplicationDefinition {
             // setup view as styleScopeHost
             rootView._setupAsRootView({});
         }
-        const controller = getViewController(rootView);
+        setViewControllerView(rootView);
         const haveController = this._window.rootViewController !== null;
         this._window.rootViewController = controller;
         if (!haveController) {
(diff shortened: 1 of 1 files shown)


MartoYankov:
test pro_templates

tns-core-modules/application/application.ios.ts
```diff
@@ -339,13 +341,21 @@ function getViewController(view: View): UIViewController {
...
-        }
+        // We set UILayoutViewController dynamically to the root view if it doesn't have a view controller
+        // At the moment the root view doesn't have its native view created. We set it in the setViewControllerView func
+        viewController = iosView.UILayoutViewController.initWithOwner(new WeakRef(view)) as UIViewController;
+        view.viewController = viewController;
+        return viewController;
+    }
+}
+
+function setViewControllerView(view: View): void {
```
manoldonev: Not sure what is the idea of the `setViewControllerView` naming, can you elaborate a bit?

tns-core-modules/application/application.ios.ts
```diff
@@ -339,13 +341,21 @@ function getViewController(view: View): UIViewController {
...
+    const viewController: UIViewController = view.viewController || view.ios;
+    const nativeView = view.ios || view.nativeViewProtected;
+
+    if (nativeView && viewController && viewController instanceof iosView.UILayoutViewController) {
+        viewController.view.addSubview(nativeView);
+        return;
     }
 
     throw new Error("Root should be either UIViewController or UIView");
```
manoldonev: Looks like we have a scenario where viewController can be UIViewController but not UILayoutViewController -- previously we did not throw exception in this case but with the current flow I think we will.

tns-core-modules/application/application.ios.ts
MartoYankov: We add a subview to the view of the view controller. `viewController.view = ...`. The correct `setViewControllerViewSubView` didn't sound good to me.

MartoYankov:
test pro_templates

PR #6402, opened 2018-10-12
fix(observable-array): reduce no longer ignores zero as initial value

### LINKED ISSUE #6385: ObservableArray.reduce from tns-core-modules doesn't handle initial value correctly
**Environment**
From `tns info`:
✔ Component nativescript has 4.2.4 version and is up to date.
✔ Component tns-core-modules has 4.2.1 version and is up to date.
✔ Component tns-ios has 4.2.0 version and is up to date.

**The bug**
The `reduce` function form `ObservableArray` doesn't handle initial value correctly. The first value will always be the first item in the list and not the initial value provided as a parameter.

**To Reproduce**
```javascript
let products = new ObservableArray([{price:10},{price:20},{price:30}]);

const sum = products.reduce((sum, p, index) => {
  console.log('Iteration', index, 'Sum:', sum, 'Type of sum:', typeof(sum), 'Price:', p.price);
  return sum + p.price;
}, 0);

console.log('Final sum', sum, typeof(sum));

// Output
// Iteration 1 Sum: { "price": 10 } Type of sum: object Price: 20
// Iteration 2 Sum: "[object Object]20" Type of sum: string Price: 30
// Final sum "[object Object]2030" string
```
Run the code and see that  `sum` will be `{price:10}` on the first iteration. This is wrong. It should be `0`. However, if we didn't pass an initial value as a parameter this would be the expected behaviour.

**Expected behavior**
On the first iteration the `sum` should be `0` as it is provided as the initial value.

surdu:
## PR Checklist

When using passing an initial value of zero to `.reduce()` from ObservableArray, the initial value is ignored

## What is the new behavior?
Initial value is used

Fixes #6385


vakrilov:
Hey @surdu 
The code looks right. Do you think you can also take a look at the `reduceRight` method. It seems that it's pretty identical and probably requires the same fix.

surdu:
@vakrilov Now I realize that what I was wondering is exactly what you said 😆. Sorry, I was tired.

I've just pushed the tests and the fix for reduceRight as well 😉

vakrilov:
Hey @surdu  - thanks again for the PR! LGTM!
We are currently in a code-freeze phase for the NativeScript 5.0 release. This means, it might take more time to merge PRs into master.

We will notify you if the is some action needed form your side. 
Thanks for understanding and being awesome part of the NS community!

PR #6405, opened 2018-10-12
fix-next: restrict to safe area when transform

MartoYankov:
Applying the safe area adjustments while there is a transform on the UIView leads to unexpected/undefined results. See warning in https://developer.apple.com/documentation/uikit/uiview/1622459-transform. This PR restricts the safe area logic when there is a transform set on the view. This should lead to the same behavior as 4.0.

An alternative would be to remove the transform, adjust the safe area, then re-apply the transform. Will check how this behaves.

Test with Groceries and NS-ng animations demo


MartoYankov:
test groceries_ios groceries_ios10 groceries_ios11 animations_ios animations_ios10 animations_ios11

MartoYankov:
test --ignore ng_animation_examples_ios ng_animation_examples_ios10 ng_animation_examples_ios11

Issue #6387, opened 2018-10-11
[iOS] providing an option to specify the modal view's UIModalPresentationStyle 


tsonevn:
Providing a way to specify the modal view's UIModalPresentationStyle by choosing some of the available [optional values from iOS docs](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle).
```
case fullScreen
A presentation style in which the presented view covers the screen.

case pageSheet
A presentation style that partially covers the underlying content.

case formSheet
A presentation style that displays the content centered in the screen.

case currentContext
A presentation style where the content is displayed over another view controller’s content.

case custom
A custom view presentation style that is managed by a custom presentation controller and one or more custom animator objects.

case overFullScreen
A view presentation style in which the presented view covers the screen.

case overCurrentContext
A presentation style where the content is displayed over another view controller’s content.

case blurOverFullScreen
A presentation style that blurs the underlying content before displaying new content in a full-screen presentation.

case popover
A presentation style where the content is displayed in a popover view.

case none
A presentation style that indicates no adaptations should be made.
```

surdu:
I could take this one, if it's ok.

@tsonevn How do you envision this be implemented: pass an argument to showModal that will only be used on iOS ?

demetrio812:
Hello guys, any news on this feature? Is there a workaround in the meantime that works across multiple iOS versions?

Cheers

tsonevn:
HI @demetrio812,
Currently, the workaround is to change the UIModalPresentationStyle directly in the [tns-core-modules](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/core/view/view.ios.ts#L405).
Keep in mind that if you delete the **node_modules** folder, you have to make this change again.

cerireyhan:
@tsonevn Can this also be used with nativescript-angular? Its not implemented in ModalDialogService.

dottodot:
It seems to have been implemented but no way to use in Angular.

PR #6409, opened 2018-10-13
feat(view): added iOS parameter for modal presentation style

surdu:
## PR Checklist

Other than fullscreen or not, there is no way to set the rest of modal presentation styles for a modal on iOS

## What is the new behavior?
You can now pass an extra parameter called `iOSPresentationStyle` to `.showModal()`

Implements #6387


surdu:
@tsonevn It turns out this is not so easy as just passing a parameter to `.showModal()`. Some of the presentation styles require additional logic in order to make them work.

I've already implemented [some default logic](https://github.com/NativeScript/NativeScript/pull/6409/files#diff-2775a71872c1630de8108bf058a1628eR403) for `UIModalPresentationPopover` and while I believe some default logic can also be implemented for most of the otehrs, `UIModalPresentationCustom` might be a little bit complicated withouth returning the view controller to the user.

I will investigate further, but wanted to open the discussion here for further ideas.

vakrilov:
Hey there @surdu - thanks for you effort. Here are a couple of thoughts:

### Definitions
First of all - all `d.ts` files should be clear of IOS or Android Specific classes(like `UIModalPresentationStyle`). The reason is that we don't want people to be forced to use `tna-platform-declarations` package in their apps.

### Options
The `showModal` API already has 3 optional parameters and this is making it little hard to use. Adding one more optional param will amplify the problem (what if you want to pass only the last one). 

My suggestion is to define a `ShowModalOptions` interface that you pass to the method. This way you can fill in only the options you care about. Furthermore you can have an `ios` and `android` portion of the interface that only deals with plat-spceific options (like `UIModalPresentationStyle`). 

You can also have a callback in the options - which will be called if they are using the an option, that needs to give the control back to the user for additional setup of a view-controller for example.

### PR
Seems that this PR is not as easy as originally thought. What do you think of moving the discussion back to the issue (#6387). We might pick only the `UIModalPresentationPopover` values that are most useful - no point of supporting all of them if they are not going to be used.

surdu:
@vakrilov I've made the changes that you requested.

Also wrote a test for my changes. I took another test for the modal and just changed the way the options are passed. The test is somewhat overkill, but I guess there was a good reason to implement the test I copied like this, so I went with it.

There is one thing that I want to do but I didn't manage to figure out (probably can't be done at this stage of execution): I want to get styling informations about the Page that is about to be presented [at this point](https://github.com/NativeScript/NativeScript/pull/6409/files#diff-2775a71872c1630de8108bf058a1628eR405). It would be nice to get the Page width and height expressed by the user is css and resize the pop-up accordingly. Also needed the background color few lines bellow.

For the old version of `showModal` that receives the options as arguments, do you want to deprecate it or will they live alongside each other?

vakrilov:
test

dtopuzov:
test

tns-core-modules/ui/core/view/view-common.ts
```diff
@@ -243,10 +243,10 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
           stretched = arguments[5];
         }
 
-        const view: ViewDefinition = firstAgrument instanceof ViewCommon
-            ? firstAgrument : createViewFromEntry({ moduleName: firstAgrument });
+        const view = firstAgrument instanceof ViewCommon
+            ? firstAgrument : <ViewCommon>createViewFromEntry({ moduleName: firstAgrument });
 
-        (<ViewCommon>view)._showNativeModalView(this, context, closeCallback, fullscreen, animated, stretched, iosOpts);
```
surdu: Note to self: removed `iosOpts` by mistake

vakrilov:
test

vakrilov:
test

vakrilov:
test

PR #6418, opened 2018-10-16
test(safe-area): add ui tests

MartoYankov:
New tests:
 * scroll-view -> safe-area-root-element
 * scroll-view -> safe-area-sub-element
 * scroll-view -> safe-area-images
 * scroll-view -> safe-area-images-overflow
 * list-view -> safe-area
 * layouts -> safe-area

Just compare images. If possible - Check in both Portrait and Landscape.


MartoYankov:
test --ignore

PR #6428, opened 2018-10-18
fix(ios delegate) Eliminate crash on selector looking for "Window"

NathanaelA:
## PR Checklist

NS Will crash on some plugins when they attempt to get the "window" from the main delegate.   With the error message Application terminated which can't be caught.

## What is the new behavior?
NS won't crash anymore

Fixes/Implements/Closes # https://github.com/NativeScript/ios-runtime/issues/430

Notes:
I am not 100% sure if I need to return a valid window, or if undefined is enough.   If you believe I should pre-set the window, I'm willing to update the PR if you can point me at which of our "windows" we should return.        However, in the plugin I was converting -- I just had to have a valid selector; once I had a property the selector could access;   all the random crashes about the missing selector stopped, even though the plugin was always getting "undefined" as the result back.     Having the selector was all that is needed to fix this issue for the plugin I was working on making work in NS.


vakrilov:
Hi and thanks for the PR!

The team have been quite busy baking the upcoming NativeScript 5.0 release. We hope you love it as much as we do! The major release is the reason we taking longer with PR reviews. We will resume active reviewing in the following weeks. 

Thanks for your patience and for being awesome contributor!

tns-core-modules/application/application.ios.ts
```diff
@@ -21,9 +21,18 @@ import { Frame, NavigationEntry } from "../ui/frame";
 import * as utils from "../utils/utils";
 import { profile, level as profilingLevel, Level } from "../profiling";
 
-class Responder extends UIResponder {
-    //
-}
+// Must use .extend() to make the "window" be a valid iOS selector
+var Responder = UIResponder.extend({
```
vakrilov: Can you refactor this code to use TypeScript extend. It will be more like the rest of the codebase. Should look something like:
```
class Responder extends UIResponder implements UIApplicationDelegate {
    public static ObjCProtocols = [UIApplicationDelegate];

    private _window: UIWindow;

    get window(): UIWindow {
        return this._window;
    }
    set window(setWindow: UIWindow) {
        this._window = setWindow;
    }
}
```

vakrilov:
Thanks for the PR!
About the `window` question. I would suggest doing:
```
    get window(): UIWindow {
        return iosApp ? iosApp.window : undefined;
    }
```
Not sure if the setter should do anything - probably not.

tns-core-modules/application/application.ios.ts
NathanaelA: @vakrilov - I wish that worked...  Unfortunately, I'm not 100% sure why; but I could NOT get the Typescrpt "extends" to work at all in this case.   My original code left that alone and just added the window to it.

For some reason using the typescript extends the "window" was not setup as a proper function that could be seen by a `selector`.    Hence the comment preceding it about having to use `.extend()`.    ;-)

Just in case I made a mistake; I'll try dropping your class into the code base and verify that maybe I didn't make a mistake originally...

NathanaelA: @vakrilov - I believe the issue is caused by this: https://github.com/NativeScript/ios-runtime/issues/818

NathanWalker: @vakrilov It's true - the standard TypeScript version does cause issues due to what @mbektchiev points out in that issue - recently ran into that and was surprised as well. His suggestion there to use `.extends` works great though.

mbektchiev: I guess that the reason is related to another issue we recently discovered in the runtime: https://github.com/NativeScript/ios-runtime/issues/1012. My guess is that if you implement the methods with 2 functions named `window` and `setWindow` they should be correctly callable from native.

vakrilov:
test

PR #6443, opened 2018-10-22
fix-next: apply ios safe area before transform

MartoYankov:
This ensures the same behavior between container views with transform and the ones without a transform.


MartoYankov:
test ng_animation_examples_ios ng_animation_examples_ios10 ng_animation_examples_ios11 animations_ios animations_ios10 animations_ios11

PR #6472, opened 2018-10-29
fix-next: handle view controller nesting in ng

MartoYankov:
This should fix failing tests in angular e2e/renderer test app


SvetoslavTsenov:
test ngapps

SvetoslavTsenov:
test ngapps

Issue #6490, opened 2018-11-01
IllegalStateException in a tabview scenario with nested frames


manoldonev:
1. Create app with root frame:
``` xml
<Frame defaultPage="home/tab-page"></Frame>
```
2. Default page of the frame should contain a navigation button to a different page and a tabview component whose tab items contain Frames:
``` xml
<Page class="page"
    xmlns="http://schemas.nativescript.org/tns.xsd">

    <GridLayout rows="auto, *">
        <Button text="navigate to second page" tap="onTap" />
        <TabView androidTabsPosition="bottom" row="1">
            <TabViewItem title="Home">
                <Frame defaultPage="home/home-items-page"></Frame>
            </TabViewItem>

            <TabViewItem title="Browse">
                <Frame defaultPage="browse/browse-page"></Frame>
            </TabViewItem>
        </TabView>
    </GridLayout>
</Page>
```
3. Navigate to second page (via the button) and navigate back to the tabview page (e.g. via hardware button, doesn't matter really)

The following exception occurs:
```
System.err: Calling js method finishUpdate failed
System.err:
System.err: Error: java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first.
```

PR #6489, opened 2018-11-01
fix(android): back navigation on app suspend/resume

manoldonev:
Fixes: https://github.com/NativeScript/NativeScript/issues/6488

PR #6421, opened 2018-10-16
fix-next(android): popExit fragment animation

manoldonev:
Fix issues with "popExit" fragment animation when FragmentTransaction.setCustomAnimations(...) are used.
Also reworked fix from https://github.com/NativeScript/NativeScript/pull/6339 

Kudos to @MartoYankov for the assistance! 👍

PR #6339, opened 2018-10-02
fix(android): nested frames on app suspend/resume

manoldonev:
Fixes https://github.com/NativeScript/NativeScript/issues/6312
Fixes https://github.com/NativeScript/NativeScript/issues/6201

PR #6495, opened 2018-11-02
fix(android): IllegalStateException with tabview&nested frames

manoldonev:
Fixes: https://github.com/NativeScript/NativeScript/issues/6490
Related to: https://github.com/NativeScript/NativeScript/pull/6339
Related to: https://github.com/NativeScript/NativeScript/pull/6421

This fix was originally applied with https://github.com/NativeScript/NativeScript/pull/6339 but was later reworked with https://github.com/NativeScript/NativeScript/pull/6421 (basically the part in frame.android.ts where disposing the fragments is performed after super.unloaded(...) logic).

The original fix did not work properly when called in onDestroyView(...) as it essentially removed the fragment view too early hence breaking exit transition. I actually tested onDestroy(...) then as well but due to a different bug addressed with https://github.com/NativeScript/NativeScript/pull/6421 and https://github.com/NativeScript/NativeScript/pull/6489 (essentially we need to cache and restore animators because of the simulated first navigation) onDestroyView(...) and onDestroy(...) were called at the same time (now onDestroy(...) is called after the actual exit transition is finished which means we can use it to detach the native view from its parent).

WIP: Add unit test

### DIFF
--- a/tns-core-modules/ui/frame/frame.android.ts
+++ b/tns-core-modules/ui/frame/frame.android.ts
@@ -859,11 +859,9 @@ class FragmentCallbacksImplementation implements AndroidFragmentCallbacks {
         if (traceEnabled()) {
             traceWrite(`${fragment}.onDestroy()`, traceCategories.NativeLifecycle);
         }
-        
+
         superFunc.call(fragment);
 
-        // fixes 'java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first'.	
-        // on app resume in nested frame scenarios with support library version greater than 26.0.0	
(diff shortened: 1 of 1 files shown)


manoldonev:
test --ignore uitests_android ngapps_android api19 api23 api28

manoldonev:
test --ignore uitests_android ngapps_android api19 api23 api28

SvetoslavTsenov:
test ngapps#svetoslavtsenov/enable-naviagtion-tests-android build_nativescript-angular#svetoslavtsenov/enable-naviagtion-tests-android

PR #6589, opened 2018-11-20
fix(android-fragment): child already has a parent

manoldonev:
Related to https://github.com/NativeScript/NativeScript/pull/6495


manoldonev:
test --ignore tests_app_transitions_android animations_android uitests_android ngapps_android api19 api23 api27 api28

Issue #6621, opened 2018-11-26
Can't control color of large titles in iOS


alfrednerstu:
**Environment**
✔ Component nativescript has 5.0.1 version and is up to date.
✔ Component tns-core-modules has 5.0.3 version and is up to date.
✔ Component tns-android has 5.0.0 version and is up to date.
✔ Component tns-ios has 5.0.0 version and is up to date.

**Describe the bug**
If you use `navigationBar.prefersLargeTitles = true;` you can't control the color of the title with css like you can with a normal title like this `ActionBar { background-color: #fff; color: #f00; }`

**To Reproduce**
Scroll the page and see the title change color between its large and normal versions.

**Expected behavior**
The large title should have the defined color too just like the normal title.

**Sample project**
https://play.nativescript.org/?template=play-vue&id=vLJE4t

**Additional context**
iOS only.

NickIliev:
@alfrednerstu you can change the text color using the native API

For example:
```
let actionBar = <ActionBar>args.object;

actionBar.ios.prefersLargeTitles = true;
actionBar.ios.largeTitleTextAttributes =  NSDictionary.dictionaryWithDictionary({
    [NSForegroundColorAttributeName]: UIColor.blueColor
});
```

NickIliev:
@alfrednerstu agreed - I've created a PR for this one [here](https://github.com/NativeScript/NativeScript/pull/6631) - once the PR is reviewed and merged it will be included in the `master` branch and from there in the next official version of `tns-core-modules`.

PR #6631, opened 2018-11-27
feat(ActionBar): apply text color when largeTitletextAttribute is set on iOS

NickIliev:
Related to [https://github.com/NativeScript/NativeScript/issues/6621](https://github.com/NativeScript/NativeScript/issues/6621)

Covers cases where large titles attribute is used on iOS
```
actionBar.ios.prefersLargeTitles = true;
```

As a result, the color set via CSS or `color` property will be applied for this specific attribute/


patrikengborg:
Would it be possible to add this for font weight as well?

PR #6637, opened 2018-11-28
feat(modals): Enable modal dialog chaining in IOS

### LINKED ISSUE #5744: iOS: issue when trying to chain modal pages within promise

vakrilov:
The `closeCallback` for modals in IOS is fired before the dialog is closed. If you try to open another dialog in the callback - it fails because the previous dialog is still showing.

## What is the new behavior?
The `closeCallback` is fired *after* the dialog is closed - in the completion callback of `dismissViewControllerAnimatedCompletion`.

**e2e test needed**: There is a `Login (second modal in cb)` button in the `modal-view` page in the `ui-tests-app`. It will open a modal and on its closeCallback it will open a second one.

Fixes #5744
This PR is followup of #6370


dtopuzov:
test

two-bridges:
This bug (#5744) is occurring for me in tns 5.3.3, nativescript-angular 6.2.  

Upon calling `(<ModalDialogParams>this.params).closeCallback()`, the return promise is resolved too soon such that a subsequent modal popup fails with the same error as #5744:

> ViewHierarchy: Parent is already presenting view controller. Close the current modal page before showing another one!

Issue #6511, opened 2018-11-05
Android: isUserInteractionEnabled stopped controlling the ScrollView's behaviour 


manojdcoder:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 5.0.0
 - Cross-platform modules: 5.0.1
 - Android Runtime: 5.0.0
 - iOS Runtime: 5.0.0
 - Plugin(s): N/A

**Describe the bug**
Until v4.2.0, setting `isUserInteractionEnabled` to `false` not only prevented click events but also the scrolling incase of ScrollView. Since 5.0.0, it's not the case anymore with Android, user is allowed to scroll anyways. 

**To Reproduce**
https://stackoverflow.com/questions/53158670/nativescript-scrollview-bug-on-android-with-runtime-version-5-0-0

**Expected behavior**
Setting `isUserInteractionEnabled` to `false` should prevent scroll on Android.

**Sample project**
https://play.nativescript.org/?template=play-js&id=hKrNlK

**Additional context**
If anyone looking for a workaround, here it is - https://play.nativescript.org/?template=play-js&id=hKrNlK&v=10

tsonevn:
Hi @manojdcoder,
Thank you for the provided sample projects. 
I tested the problem on my side and was able to recreate the issue. I will mark this as a bug.

Issue #3215, opened 2016-11-30
StackLayout isUserInteractionEnabled="false" value is not working


MadhuSudhanBhaskar:
StackLayout isUserInteractionEnabled="false" value is not working

https://docs.nativescript.org/api-reference/classes/_ui_core_view_.view.html#isuserinteractionenabled
Yes I did verify on stackover flow and other channels

I am trying to disable use interaction when I have a activity indicator. Below code is stack layout example. 
<StackLayout isUserInteractionEnabled="false">
    <Button text="Sign in"  (tap) = "test()"></Button>
    <ActivityIndicator  style="color:red" [busy] = "isLoading" ></ActivityIndicator>
</StackLayout>

Which platform(s) does your issue occur on?
Android [For now I tested this on Android alone]

### Please provide the following version numbers that your issue occurs with:
- CLI: (run `tns --version` to fetch it) 2.4.0
- Cross-platform modules: (check the 'version' attribute in the
`node_modules/tns-core-modules/package.json` file in your project) 2.4.0
- Runtime(s): (look for the `"tns-android"` and `"tns-ios"` properties in the
`package.json` file of your project) 2.4.1 (Android)
- Plugin(s): (look for the version number in the `package.json` file of your
project)

### Please tell us how to recreate the issue in as much detail as possible.

I have a button which enables the activity indicator, from my function I am trying to set isUserInteractionEnabled to false(Boolean) and trying to disable the interaction on the screen for certain time. But I still can click different buttons on the screen.

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.

Some pieces of code.
<StackLayout isUserInteractionEnabled="disStatus">
    <TextField hint="Password" secure="true" [(ngModel)]="user.password"></TextField>
    <Button text="Sign in"  (tap) = "test()"></Button>
    <ActivityIndicator  style="color:red" [busy] = "isLoading" ></ActivityIndicator>
</StackLayout>

Component function 
    test() {
        this.isLoading = !this.isLoading;
        this.defaultTest = false;
    }

tsonevn:
Hi @MadhuSudhanBhaskar,
Thank you for reporting this issue.
[...]

PR #6640, opened 2018-11-30
feat(scrollview): add isScrollEnabled property

manoldonev:
Add ScrollView.isScrollEnabled property that should cover the majority of the scenarios that users try to address with isUserInteractionEnabled / isEnabled properties. This is a straightforward and consistent solution that does one thing and one thing only -- enable/disable scroll.

We have a [known inconsistency](https://github.com/NativeScript/NativeScript/issues/3215) in Android / iOS with respect to the `isUserInteractionEnabled` property and container components (e.g. scrollview, various layouts, etc.)-- in iOS when property is set to false on the container, it disables user interaction recursively on all subviews, while on Android it doesn't. As this is the standard behavior on each OS we decided not to streamline this:
- on Android `isUserInteractionEnabled` false is basically equivalent to `isScrollEnabled` false (if we were to streamline the behavior for scrollview to match iOS counterpart, we would have to do the same for all layout containers too so for the time being we are "consistently inconsistent" -- consistent among Android components and inconsistent betwen Android and iOS)
- on iOS `isUserInteractionEnabled` false disables scroll and also disables user interaction for all subviews.

Also, ScrollView.isEnabled property currently has no effect neither on Android, nor on iOS.

Fixes https://github.com/NativeScript/NativeScript/issues/6511
Depends on https://github.com/NativeScript/tns-core-modules-widgets/pull/150 (merge after the widgets PR)


manoldonev:
test branch_tns_core_modules_widgets#mdonev/disable-scroll

PR #6258, opened 2018-09-11
feat(tabview:android/application): enable/disable android swipe for tabview

Rjgoolsby:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#running-unit-tests.
- [x] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/WritingUnitTests.md.

## What is the current behavior?

Current behavior: on Android if the tabs are at the "top" there is no way to disable the swipe gesture changing tabs. This is seriously inconvenient when putting things like maps, listviews or anything that require horizontal swipe gestures for functionality. Currently if you make a horizontal swipe it changes the tab. Segment bar does not suffice as it doesn't function the same as Tabview across all devices. iOs does not have this issue as the swipe gesture doesn't work anyway.

## What is the new behavior?

Now, the dev have the the option of setting swipe gesture to false (true by default for backwards compatibility). When set to false, swipe gesture will be disabled on Android like iOs.

Also added example to test app. 

Fixes/Implements/Closes #[Issue Number].
https://github.com/NativeScript/NativeScript/issues/2449

No breaking changes as the default is true which is the same as before.

PR #6652, opened 2018-12-03
feat(tabview): add androidSwipeEnabled property

### LINKED ISSUE #2449: Disable swipe left/right in TabView on Android
Hi,

i need to disable the native swipe ability in TabViews on Android. The solution mentioned [here](https://github.com/NativeScript/NativeScript/issues/958#issuecomment-153367667) is not working anymore - the swipe event seems to be already in motion when onTouch is called, so it moves a few pixels and gets stuck then.

The proper native solution seems to be, to override the ViewPage class like discussed [here](http://stackoverflow.com/questions/9650265/how-do-disable-paging-by-swiping-with-finger-in-viewpager-but-still-be-able-to-s).

Is it possible to achieve this with nativescript some other way?
Would it be possbile to add a TabView Attribute swipeEnabled="true/false"?

manoldonev:
Copied from https://github.com/NativeScript/NativeScript/pull/6258 with comments applied as it seems @Rjgoolsby deleted the forked repo on his side.

Fixes/Implements/Closes #2449

### DIFF
--- a/apps/app/ui-tests-app/tab-view/main-page.ts
+++ b/apps/app/ui-tests-app/tab-view/main-page.ts
@@ -24,6 +24,7 @@ export function loadExamples() {
     examples.set("tab-view-bottom-position", "tab-view/tab-view-bottom-position");
     examples.set("issue-5470", "tab-view/issue-5470");
     examples.set("tab-view-tab-text-font-size", "tab-view/tab-view-tab-text-font-size");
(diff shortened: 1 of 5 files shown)


Rjgoolsby:
@manoldonev Thank you. I was having problems keeping my fork in sync with the main repo so I just deleted it and made the changes locally. Sorry for the inconvenience and thank you for doing this. This will help us a tremendous without listviews in tabview

PR #6655, opened 2018-12-04
fix(modals): regression with modals in angular

vakrilov:
Fix regression in modals in angular caused by #6637


vakrilov:
test ngapps

Issue #6628, opened 2018-11-27
Overlapping elements in ListView on ios


lcini17:
**Environment**
 - CLI: 5.0.1
 - Cross-platform modules: 5.2.0-2018-11-27-104751-03
 - Android Runtime: 5.0.0
 - iOS Runtime: 5.0.0

**Describe the bug**
Refer to https://github.com/NativeScript/nativescript-angular/issues/1599

On iOS, the ListView layout is rendered incorrectly, text is overlapping. This is happening with the elements that are not initially displayed on screen. This works fine with tns-core-modules version 4.2.1. 

On Android, the ListView works fine.

**To Reproduce**
Create a ListView with a StackLayout large enough to go off the screen and then scroll to the bottom of the list.

**Expected behavior**
Correctly display ListView without elements overlapping.

melineric:
Adding iosOverflowSafeArea="false" to every StackLayout within the ListView template fixed it for me.

lcini17:
@melineric  Thank you :)

NickIliev:
@lcini17 can you create a Playground demo of your scenario (I've used [this one ](https://play.nativescript.org/?template=play-tsc&id=aEvc6O)but was not able to reproduce the case)

For anyone interested in what `iosOverflowSafeArea` property is doing, [here is the documentation reference.](https://docs.nativescript.org/ui/layouts/layouts#iosoverflowsafearea-property)

NickIliev:
Closing this one as the issue is not reproducible with the latest `tns-core-modules` and [the test Playground.](https://play.nativescript.org/?template=play-tsc&id=aEvc6O) 
@lcini17 if you are still experiencing this one please do not hesitate to contact me and if possible provide more information (or demo) on how to reproduce the issue.

PR #6656, opened 2018-12-04
fix(listview-ios): incorrect layout when scroll

MartoYankov:
Fixes https://github.com/NativeScript/nativescript-angular/issues/1626
Fixes https://github.com/NativeScript/NativeScript/issues/6628

During the first layout of the list view items, they have no window. On scroll when laying out the next items, they suddenly have a window. Maybe because of recycling. This is not the actual issue however. The problem is that for some reason the coordinates of these views in this window are negative, i.e. wrong. Not sure what is causing that.

The easiest and cleanest way to fix this would be to make the `iosOverflowSafeArea` property inherited and assign it as `false` to the list view cell. However, we can't do this by design - containers and components must have different default values. So the next best thing is add a new property which can be inherited - `iosOverflowSafeAreaEnabled`. Setting this to `false` would disable the setting of the other property to the View and all its children.

### DIFF
--- a/apps/app/ui-tests-app/issues/issue-ng-repo-1626.ts
+++ b/apps/app/ui-tests-app/issues/issue-ng-repo-1626.ts
@@ -1,0 +1,31 @@
+import { fromObject } from "tns-core-modules/data/observable";
+
+export function onLoaded(args) {
+    const page = args.object;
+    page.bindingContext = fromObject({
+        items: [
+            { id: 1, name: "Ter Stegen", role: "Goalkeeper" },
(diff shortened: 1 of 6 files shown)


marcocasadio:
In a RadListView if I scroll up/down I get this issue:
before
after

NickIliev:
@marcocasadio a PR is not the place to report an issue - please post your issue in the respective repository (in your case with RadListView that would be [https://github.com/telerik/nativescript-ui-feedback/issues)](https://github.com/telerik/nativescript-ui-feedback/issues)

Issue #6690, opened 2018-12-10
ModalView StatusBar color issue


tsonevn:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 5.0.3
 - Cross-platform modules:  5.0.5
 - Android Runtime: 5.0

Setting up [StatusBar color](https://docs.nativescript.org/ui/theme#status-bar-considerations), while changing the colors in `App_Resources/Android/colors.xml`, will not be respected when the ModalView is opened in fullscreen mode. The issue can be recreated with [nativescript-sdk-examples-js](https://github.com/NativeScript/nativescript-sdk-examples-js/tree/master/app/ns-ui-category/modal-view/basics).
![7aad22fc-9ce2-49cb-966d-08cd61f62307_image001](https://user-images.githubusercontent.com/17448734/49733241-a728fb00-fc89-11e8-939e-06d3854fdbe4.gif)

t.1364644

PR #6691, opened 2018-12-10
set the correct application theme while creating the dialog fragment

tsonevn:
Set the correct application theme while creating the dialog fragment. 
The current code `this.getTheme` always returns  `0` and does not set the correct theme for the fragment. 

 Fixes https://github.com/NativeScript/NativeScript/issues/6690


tsonevn:
test

PR #6729, opened 2018-12-19
fix-next: undefined root view when reapplying styles

vchimev:
Error:
```
JS ERROR TypeError: undefined is not an object (evaluating
'application_1.getRootView()._onCssStateChange')
```

Steps:
- `tns run <platform> --hmr`
- make a change in application styles
- restart the application

### DIFF
--- a/tns-core-modules/application/application-common.ts
+++ b/tns-core-modules/application/application-common.ts
@@ -89,7 +89,7 @@ export function livesync(context?: HmrContext) {
         reapplyAppCss = extensions.some(ext => context.module === fileName.concat(ext));
     }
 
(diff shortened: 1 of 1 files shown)


tns-core-modules/application/application-common.ts
```diff
@@ -89,7 +89,7 @@ export function livesync(context?: HmrContext) {
         reapplyAppCss = extensions.some(ext => context.module === fileName.concat(ext));
     }
 
-    if (reapplyAppCss) {
```
manoldonev: No need to call the method twice -- store the result of the first execution in a variable and use it instead. :)

vchimev:
test

PR #6755, opened 2019-01-04
fix(tab-view): remove onBackPressed override

ADjenkov:
Android back button will exit the app, instead going back to previous list page, when using lazy loaded nested named <page-router-outlets> inside TabViewItems. 

_This happens only if using <TabView> as a root component (template-tab-navigation-ng template)._
 
Remove onBackPressed override, since the topmost() frame is now properly assigned when switching between tab view items.

Fix https://github.com/NativeScript/nativescript-angular/issues/1680


ADjenkov:
test ngapps

Issue #6863, opened 2019-02-05
TextView content shifts up on every other newline character input


manoldonev:
**Environment**
 - CLI: 5.1.1
 - Cross-platform modules: 5.1.2
 - iOS Runtime: 5.1.1

**Describe the bug**
iOS TextView content shifts up and gets partially clipped on every other newline character input 

**To Reproduce**
1. Input some text
2. Input newline -- textview resizes properly but content shifts up and gets partially clipped
3. Input newline -- textview resizes properly and content offset is OK -- content visible

**Expected behavior**
Textview should not exhibit alternating behavior on inputing newline -- content should always be visible

**Sample project**
Playground project: https://play.nativescript.org/?template=play-tsc&id=c8e1hS&v=2

**Additional context**
The cause of the problem is described [here](https://stackoverflow.com/a/42480403/9332825):

> The problem is that UITextView's scroll animation and your frame setting action were happened at the same time. UITextView internally scrolls the texts you currently typing to visible when typed one more character at the end of the line or typed the new line character. But the scroll animation does not need because you are expanding the textview. Unfortunately we can't control textview's internal scroll action so the text scrolls to the top of the expanded textview weirdly. And that weird scroll makes unnecessary bottom padding too.

PR #6864, opened 2019-02-05
fix(ios): textview content clipped on every other newline

manoldonev:
Fixes: https://github.com/NativeScript/NativeScript/issues/6863

### DIFF
--- a/tns-core-modules/ui/text-view/text-view.ios.ts
+++ b/tns-core-modules/ui/text-view/text-view.ios.ts
@@ -95,6 +95,16 @@ class UITextViewDelegateImpl extends NSObject implements UITextViewDelegate {
     }
 }
 
+class NoScrollAnimationUITextView extends UITextView {
+    // see https://github.com/NativeScript/NativeScript/issues/6863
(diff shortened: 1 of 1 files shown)


tns-core-modules/ui/text-view/text-view.ios.ts
```diff
@@ -95,6 +95,16 @@ class UITextViewDelegateImpl extends NSObject implements UITextViewDelegate {
     }
 }
 
+class NoScrollAnimationUITextView extends UITextView {
+    // see https://github.com/NativeScript/NativeScript/issues/6863
+    // UITextView internally scrolls the text you are currently typing to visible when newline character
+    // is typed but the scroll animation is not needed because at the same time we are expanding
+    // the textview (setting its frame)
+    public setContentOffsetAnimated(contentOffset: CGPoint, animated: boolean): void {
```
vakrilov: Are there cases that we actually want the default scrolling behavior - for example when you have set the width/height properties explicitly?

SvetoslavTsenov:
test ngapps tsapps

SvetoslavTsenov:
test groceries

Issue #6348, opened 2018-10-04
Accessing NativeScript component properties with CSS


pushqrdx:
### Tell us about the problem
According to the docs you can set a component property using CSS as longs as it's of a primitive type; However, trying to set `androidTabsPosition` property of the TabView doesn't seem to work.

### Which platform(s) does your issue occur on?
Android

### Please provide the following version numbers that your issue occurs with:
- CLI: latest

### Please tell us how to recreate the issue in as much detail as possible.
[Playground Reproduction](https://play.nativescript.org/?template=play-vue&id=kyxr9o)

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.

```
    TabView {
        tab-background-color: #53ba82;
        selected-tab-text-color : white;
        android-selected-tab-highlight-color: #71f7ad;
        android-tabs-position: "bottom";
    }
```

pushqrdx:
@tsonevn Why marked as a feature. It's an unexpected behavior! because the docs say otherwise.

Issue #5372, opened 2018-02-02
[SearchBar] Allow setting textFieldHintColor and textFieldBackgroundColor via CSS


PeterStaev:
### Tell us about the problem
Setting `text-field-hint-color` and `text-field-background-color` properties in CSS has no effect. This is because the properties are not defined as `CssProperty` https://github.com/NativeScript/NativeScript/blob/0f14101238fa44b7b58f2c382b0c31a4aae93777/tns-core-modules/ui/search-bar/search-bar-common.ts#L25-L29
I don't see a reason why those cannot be set up as `CssProperties` with the respective names so we can easily set those through css. 

### Which platform(s) does your issue occur on?
Both

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.4.1
- Cross-platform modules: 3.4
- Runtime(s): 3.4.1

Issue #5355, opened 2018-01-30
textWrap NativeScript property is not accessible via CSS


NickIliev:
As stated in [this documentation section](https://docs.nativescript.org/ui/styling#accessing-nativescript-component-properties-with-css) the NativeScript specific properties should be accessible via CSS
with syntax like
```CSS
Label {
    text-wrap: true;
}

StackLayout {
    orientation: horizontal;
}
```

However, the `text-wrap` can't be set via CSS. Playground demo demonstrating the issue can be found [here](https://play.nativescript.org/?template=play-tsc&id=IUOq0x) - reported via the feedback form.

NickIliev:
Marking this one as a feature as we can use `white-space` CSS property to set the wrapping.
For example:
```
Label {
    white-space: normal;
}
```

PR #6889, opened 2019-02-11
fix(css): widget properties in css didn't work

MartoYankov:
Based on the docs - https://docs.nativescript.org/ui/styling#accessing-nativescript-component-properties-with-css - A developer can declare non-CSS widgets properties in the CSS and this would work if the properties accept text based value.

This works only for lowercase only properties, i.e. single word camel case, and it seems this never worked for multi-word camel case properties. The example in the docs works, but properties like `textFieldBackgroundColor` of the `SearchBar` didn't work. The solution to this bug is to allow the developer to declare such properties as kebab case like this `text-field-background-color`.

This actually implements a number of issues marked as features:
Fixes https://github.com/NativeScript/NativeScript/issues/5355
Fixes https://github.com/NativeScript/NativeScript/issues/5372
Fixes https://github.com/NativeScript/NativeScript/issues/6348

Docs needed:
https://docs.nativescript.org/angular/ui/styling#accessing-nativescript-component-properties-with-css

### DIFF
--- a/tests/app/ui/search-bar/search-bar-tests-native.android.ts
+++ b/tests/app/ui/search-bar/search-bar-tests-native.android.ts
@@ -23,6 +23,15 @@ export function getNativeHintColor(searchBar: searchBarModule.SearchBar): colorM
     return undefined;
 }
 
+export function getNativeTextFieldBackgroundColor(searchBar: searchBarModule.SearchBar): colorModule.Color {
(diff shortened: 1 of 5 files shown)


SvetoslavTsenov:
test package_version#rc

dtopuzov:
test package_version#rc

PR #6892, opened 2019-02-12
fix(ios-frame): tearDownUI and reset UINavigationController

ADjenkov:
**The problem**: The memory allocated for the modal view, with Frame inside, is not released.

**Solution**: [iOS] Reset reference to UINavigationController in Frame when UINavigationController view disappeared. _`_teadDownUI` method will call `disposeNativeView` where the clean up is done_


ADjenkov:
test package_version#latest

Issue #6901, opened 2019-02-14
[iOS] SearchBar hint color cannot be set before hint is set


MartoYankov:
**Environment**
 - CLI: 5.2.0
 - Cross-platform modules: 5.2.0
 - iOS Runtime: 5.2.0

**Describe the bug**
If you set a `SearchBar` `textFieldHintColor` before setting a `hint`, the color will not be assigned. This is very important for CSS styling. You usually want to set the color in CSS and the hint in the code behind and the CSS is always applied first.

**To Reproduce**
1. Have a `SearchBar`.
2. Set `textFieldHintColor` first
3. Set `hint` after this

**Expected behavior**
The color should be applied to the hint.

**Sample project**
https://play.nativescript.org/?template=play-tsc&id=20b7Dg

MartoYankov:
fixed by https://github.com/NativeScript/NativeScript/pull/6902

PR #6902, opened 2019-02-14
fix(ios): searcbar hint color before hint property

MartoYankov:
fixes https://github.com/NativeScript/NativeScript/issues/6901

### DIFF
--- a/tns-core-modules/ui/search-bar/search-bar.ios.ts
+++ b/tns-core-modules/ui/search-bar/search-bar.ios.ts
@@ -71,7 +71,6 @@ export class SearchBar extends SearchBarBase {
     nativeViewProtected: UISearchBar;
     private _delegate;
     private __textField: UITextField;
-    private __placeholderLabel: UILabel;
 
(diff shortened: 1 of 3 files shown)


MartoYankov:
package_version#latest

MartoYankov:
test package_version#latest

tns-core-modules/ui/search-bar/search-bar.ios.ts
```diff
@@ -210,18 +198,30 @@ export class SearchBar extends SearchBarBase {
     }
 
     [textFieldHintColorProperty.getDefault](): UIColor {
-        const placeholderLabel = this._placeholderLabel;
```
manoldonev: Theoretically this might qualify as a breaking change but as the old API did not behave consistently in all scenarios I guess we can let this pass.

MartoYankov: Also, this is now more consistent with TextField placeholderColor default value.

Issue #6891, opened 2019-02-12
[iOS] Translucent tab bar breaks safe area


MartoYankov:
Reported by @blizmo in https://github.com/NativeScript/NativeScript/issues/6788

When setting a TabView tab bar translucent to false, there is a space between tab bar and any layout inside the TabView.

```
const tabView = this.tabViewRef.nativeElement as TabView;
tabView.ios.tabBar.translucent = false;
```

https://play.nativescript.org/?template=play-ng&id=ZyeGeF&v=4

Issue #6684, opened 2018-12-07
Navigating to a page on iOS with a flat action bar breaks safe area overflowing


interrobrian:
**Environment**
N/A (running in NativeScript Playground)

**Describe the bug**
In iOS, if you setup a page with a flat action bar and then navigate to it, then layouts won't overflow into the safe area correctly.

**To Reproduce**
Create a page with an action bar set to flat and a layout which should span the safe area. Set a background on the layout so that you can see it. Run on an iPhone with a safe area (I'm using an iPhone XS). Navigate to this page from another page. Note that the layout does not overflow into the safe area. Also note that if you repeat these steps and don't set the action bar to flat, the layout will overflow correctly.

**Expected behavior**
As noted in the [safe area support documentation](https://www.nativescript.org/blog/nativescript-5-ios-safe-area-support), any layout which touches the safe area should overflow it, unless specified otherwise. Furthermore, the flat property of the action bar shouldn't effect the safe area overflow behavior of other widgets on the page.

**Sample project**
I don't know how to use NativeScript Core, so here is a NativeScript Angular project which demonstrates this bug: https://play.nativescript.org/?template=play-ng&id=WcOmao&v=12
Using an iPhone X*, from the start page, tap "Correct Component" and you'll get a component which overflows the safe area correctly. Go back and tap "Broken Component" and it will not overflow the safe area. Note that the only difference between the two components is that the action bar is set to flat in BrokenComponent.

**Additional context**
This bug *only* happens when you navigate to the page. It does not happen on the starting page of an app.

yoat:
I was originally watching #1638, which was closed in favor of this bug.  However the issue I'm experiencing happens on all pages, not just after navigation.

From this bug: "This bug only happens when you navigate to the page. It does not happen on the starting page of an app."
If that is true there may be two separate issues here. I believe #1638 should be reopened to address the issue of spacing under the ActionBar on all iOS devices (not just notched X devices) and for all pages (including the first page).

b4rtt:
Any progress here? Thats really bad and I need to use flat="true".

salvadorcoronel:
I’m working with Angular like Framework, I have the same problem, but It only happens when I do a navigation to other view using "clearHistory:true" as a parameter of  next code this.router.navigate(["page2"], `{clearHistory:true});`
When I do a navigation without "clearHistory:true" parameter I don't have the issue

yoat:
This issue appears to be resolved for me in NS 5.2.

[...]

PR #6918, opened 2019-02-18
fix(ios): opaque bars break ui layout

MartoYankov:
The `extendedLayoutIncludesOpaqueBars` property set to `true` unifies layout behavior between translucent and opaque bars in iOS. Remove additional special handling for flat action bars.

fixes https://github.com/NativeScript/NativeScript/issues/6891
fixes https://github.com/NativeScript/NativeScript/issues/6684


MartoYankov:
test package_version#latest

PR #6942, opened 2019-02-20
fix(android-bottom-tabs): transition type on programmatic selectedIndex change

ycherniavskyi:
Currently on programmatically change selectedIndex property always applied smoothly scroll to the new item (smoothScroll = true), even for TabView with bottom tabs position. Which lead to unnecessary tabs layout render, because if TabView has for example 3 tabs, and from tab3 you programmatically move to tab1 then tab2 also render.

## What is the new behavior?
For TabView with bottom tabs position use immediately transition (smoothScroll = false) on programmatically change selectedIndex property.
It makes perfect sense with already disable the swipe navigation and the items preloading functionality.

## Difference visualisation (with animated gif)

* smoothScroll = true
smoothscroll-true
* smoothScroll = false
smoothscroll-false


ycherniavskyi:
Unfortunately, it is not clear for me how to implement tests which will test different transition scheme for `top` and `bottom` `androidTabsPosition` property.

manoldonev:
test

Issue #6810, opened 2019-01-21
Using default setting for all transitions crashes with TabView and Side Drawer


stefan-kolb:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
nativescript. Your current version is 5.1.0
tns-core-modules. Your current version is 5.1.1
tns-android has 5.1.0 version 
component tns-ios. Your current version is 5.1.0

**Describe the bug**

I'm using the Telerik Sidebar component `nativescript-ui-sidedrawer` together with another view that has a TabView Component.
Also I'm setting the default transition to `Frame.defaultTransition = { name: "fade" };`
Without the TabView navigating back to a home component works as expected.
When I add the TabView inside the page and try to navigate back to the home view it crashes with:

```
An uncaught Exception occurred on "main" thread.
android.content.res.Resources$NotFoundException: Unable to find resource ID #0xffffffec
	at android.content.res.ResourcesImpl.getResourceTypeName(ResourcesImpl.java:249)
	at android.content.res.Resources.getResourceTypeName(Resources.java:1964)
	at android.support.v4.app.FragmentManagerImpl.loadAnimation(FragmentManager.java:1172)
	at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1543)
	at android.support.v4.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManager.java:1784)
	at android.support.v4.app.BackStackRecord.executeOps(BackStackRecord.java:797)
	at android.support.v4.app.FragmentManagerImpl.executeOps(FragmentManager.java:2625)
	at android.support.v4.app.FragmentManagerImpl.executeOpsTogether(FragmentManager.java:2411)
	at android.support.v4.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManager.java:2366)
	at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:2273)
	at android.support.v4.app.FragmentManagerImpl$1.run(FragmentManager.java:733)
	at android.os.Handler.handleCallback(Handler.java:789)
	at android.os.Handler.dispatchMessage(Handler.java:98)
	at android.os.Looper.loop(Looper.java:164)
	at android.app.ActivityThread.main(ActivityThread.java:6541)
	at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #6677, opened 2018-12-06
fix(android): nested fragment disappears on parent fragment removal

manoldonev:
~~TODO: Known crash with Flip navigation due to not handling AnimatorSet animator properly.~~
~~TODO: consider implementing base fragment class in widgets and place the shared code between frame and tab fragment there.~~
TBD: consider do we need a property to enable / disable lollipop transitions (currently implicitly disabled for nested frame scenarios). This property could be useful in other cases as well e.g. this bug https://github.com/NativeScript/NativeScript/issues/6510 is only present for the "lollipop" Slide transition but if we use our own implementation, it is OK (if user *could* disable "lollipop" explicitly, it would work).

Fixes: https://github.com/NativeScript/NativeScript/issues/6641
Fixes: https://github.com/NativeScript/NativeScript/issues/6659
Fixes: https://github.com/NativeScript/NativeScript/issues/6698

Merge after https://github.com/NativeScript/tns-core-modules-widgets/pull/154

BREAKING CHANGE:  

[Android] Explode transition cannot be used when navigating in nested frame scenarios.

[Android] Custom nested fragments should inherit from `org.nativescript.widgets.FragmentBase` class instead of `android.support.v4.app.Fragment` directly to benefit from the applied workaround for https://code.google.com/p/android/issues/detail?id=55228

PR #6955, opened 2019-02-21
fix(android): resource ID #0xffffffec not found on nav

manoldonev:
Fixes: https://github.com/NativeScript/NativeScript/issues/6810

The problem is not related to the sidedrawer and can be reproduced with plain navigation between pages as well:
1. Default home page (no nested frames/fragments)
2. Navigate to a second page that contains tabview (implicitly this means nested frames/fragments)
3. Navigate to a third page (no nested frames/fragments) --> exception occurs

This problem is caused by https://github.com/NativeScript/NativeScript/pull/6677 and basically is triggered by not properly setting up transitions/animators logic upon handling fragments with and without nested fragments upon navigation.


manoldonev:
test package_version#latest tsapps_android ngapps_android api21 api23 api25 api28 --ignore

manoldonev:
test package_version#latest tsapps_android ngapps_android api21 api23 api25 api28 --ignore

PR #6968, opened 2019-02-27
fix(ios): disable default tab reselect behavior

MartoYankov:
## PR Checklist

On iOS by default the TabView resets the navigation in a single tab when the tab is re-selected. This breaks navigation as it's not handled by NS.

## What is the new behavior?
This behavior is disabled.

Fixes https://github.com/NativeScript/nativescript-angular/issues/1466

### DIFF
--- a/tns-core-modules/ui/tab-view/tab-view.ios.ts
+++ b/tns-core-modules/ui/tab-view/tab-view.ios.ts
@@ -90,6 +90,10 @@ class UITabBarControllerDelegateImpl extends NSObject implements UITabBarControl
             owner._handleTwoNavigationBars(backToMoreWillBeVisible);
         }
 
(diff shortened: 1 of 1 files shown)


MartoYankov:
test package_version#latest

vakrilov:
AFAIU, this is a problem only with Angular as the native navigation gets out of sync with the Angular navigation.
Is it possible to make a property disabling the re-select tab feature and use it for Angular?

MartoYankov:
There are several points to be made here:
1. This actually breaks the Core framework too, but it's not that obvious. If you navigate several times and reselect the tab, it will reset the `Frame` to the root view controller and it's UI will work, but we haven't cleared the `Frame` history stack, so if you call `goBack()` it will break.
2. Making this a property and integrating it in Angular would make this more a feature than a patch. We should probably do this separately.

PR #6693, opened 2018-12-11
feat(tns-platform-declarations): Generate iOS typings from iOS 12.1 SDK

mbektchiev:
* Instructions for generating iOS typescript definitions
* Refactor `typings-gen.sh` to use {N} CLI and generate fuller 
metadata for the system libraries
* Add interop pointers and references as alternatives
to `string` for `char*` args
(implemented in https://github.com/NativeScript/ios-runtime/pull/1022)

## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#running-unit-tests.
- [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/WritingUnitTests.md.

## What is the current behavior?
Typings have not been updated with the latest iOS SDK

## What is the new behavior?
iOS typings have been regenerated with iOS 12.1

PR #7019, opened 2019-03-11
tns-platform-declarations(ios): add extend to NSObject

mudlabs:
In TypeScript, method calls like `AVCaptureMetadataOutput.extend(...)` throw the error `Property 'extend' does not exist on type 'typeof AVCaptureMetadataOutput'.`

## What is the new behavior?
Adding `extend` to the `NSObject` class resolves this compiler error for all classes that are extended from `NSObject`.

Partially fixes [issue 376](https://github.com/NativeScript/docs/issues/376)


manoldonev:
@mudlabs those iOS typings are automatically generated and any manual change is overwritten with the next version of the iOS SDK (see https://github.com/NativeScript/NativeScript/pull/6693).

Moreover, https://github.com/NativeScript/docs/issues/376 is related to Android and not iOS and seems to be outdated as the linked [documentation article](https://docs.nativescript.org/core-concepts/android-runtime/binding-generator/extend-class-interface) actually contains examples how to extend native Android classes in TypeScript.

You can refer to this article [here](https://docs.nativescript.org/core-concepts/ios-runtime/how-to/ObjC-Subclassing#typescript-support) that demonstrates how to extend native iOS classes in TypeScript (using the `extends` keyword and not extend(...) method).

Issue #550, opened 2015-08-04
Implement CSS Box Shadow


lscown:
Would be great to see a CSS3 style box shadow for UI elements.

Possible implementation:
`box-shadow: none|h-shadow v-shadow blur spread color |inset|initial|inherit;`

mischah:
Would love to have this too. 
Is this on the roadmap? Since #191 is closed.

jlooper:
I'd like to see this too, to make nice looking card interfaces. Pretty!

lyonzy:
If anyone's interested, I've discovered this sort-of-hack:
`view.android.setBackgroundResource(android.R.drawable.dialog_holo_light_frame)`
which gives an alright drop shadow on Android.

However add my +1 to this as well, it would be great to have a customisable, cross platform solution!

aliatsis:
+1

tiagordc:
+1

andreyluiz:
+1

shantamg:
+1

gabrielbiga:
+1

arnolduis:
+1

ljx0517:
+1

treio:
+1

burkeholland:
+1

Premitium:
+1

k3yb0ardn1nja:
+1

TheOriginalJosh:
+1

kamilmac:
+1

vjoao:
+1

hekigan:
+1

theBEERd:
+1

Alphmega:
+1

yezarela:
+1

abinici:
I've found out how to achieve the box shadow effect on iOS, for example on a TextField:

```
    import {Color} from "color";
    import * as view from "ui/core/view";
    ...
    let myTextField = view.getViewById(this.page, 'myTextField');
    myTextField.ios.layer.masksToBounds = false;
    myTextField.ios.layer.shadowOpacity = 1.0;
    myTextField.ios.layer.shadowRadius = 0.0;
    myTextField.ios.layer.shadowColor = new Color('#000000').ios.CGColor;
    myTextField.ios.layer.shadowOffset = CGSizeMake(2.0, 2.0);
```

In order to reference this field from code you need to give the element an ID:

```
    <TextField id="myTextField"></TextField>
```

If you use TypeScript then you'll also need to declare some iOS-specific functions:

[...]

PR #7055, opened 2019-03-21
feat(android): implement BorderDrawable outline

edusperoni:
## PR Checklist

BorderDrawable does not have an outline

## What is the new behavior?
BorderDrawable has an outline

Provides initial support for #550 

This is a port from https://github.com/NativeScript/tns-core-modules-widgets/pull/158


edusperoni:
@manoldonev It seems there was a failing test (https://github.com/NativeScript/tns-core-modules-widgets/pull/158#issuecomment-473948011), but allow me to elaborate:

Currently, on Android, Button has an elevation. This elevation is not respected by BorderDrawable because it doesn't provide an outline so no shadow is cast.

This is the exact reason this PR was created, so any BorderDrawable can cast shadow. Once @EddyVerbruggen and I properly finish the elevation property for all views, the developer will be able to specify elevations for views.

Currently, adding borders do buttons gets rid of Ripple (because it resets the background) and Shadow (because the outline is not provided).

I believe we might have to discuss how a button **should** behave. If the current implementation is an issue, I suggest we change the Button's elevation to 0 to keep things consistent

manoldonev:
test

manoldonev:
@edusperoni thank you for elaborating on this. I think we are good with this change and will update our ui tests next week to accept it.

edusperoni:
@manoldonev Awesome! Thanks!

We'll open a PR to support elevation (via CSS too) some time in the future

manoldonev:
test

manoldonev:
test

SephReed:
oh man, I hope this comes out soon

manoldonev:
@edusperoni do you have an estimate when you'll be able to add the elevation support?

We have a specific scenario with outline buttons that clearly doesn't benefit from the BorderDrawable outline support:
Actual:
healthsurvey-welcome_actual

Expected:
healthsurvey-welcome_expected

You can restore the old behavior (no shadow in this scenario) by setting `button.nativeViewProtected.setStateListAnimator(null);` but that's not a good solution for public API and I don't think we'd like to release a minor / major NativeScript version like this.

My understanding is that with elevation support (your future PR), users will be able to set `button.elevation = 0;` or `<Button elevation="0" ... />to achieve the outline button scenario above -- is this correct?

With this in mind I see the following options if elevation support will not be added in the immediate future:
- Revert this PR and merge it again together with the elevation support PR when ready (ensuring that outline button scenario can be achieved)
- Add property that disables the BorderDrawable outline functionality by default

What do you think?

edusperoni:
I could make a quick patch for elevation which would be android only, iOS support coming after. You can also switch to a `Label` for that case, since it has no elevation by default and also no statelistanimator.

So we can better determine the elevation properties, I invite you to check this repo https://github.com/JoshDSommer/nativescript-ngx-shadow#list-of-attributes

And this PR: https://github.com/JoshDSommer/nativescript-ngx-shadow/pull/19/files

Currently, all that code is done natively. My PR in that repo is a mimic of this feature (and also compatible with it).

While we can set the elevation, the statelistanimator is a whole other issue. Should we always override the statelistanimator? Just for views that have an animator? Or should we expose extra android-only properties for this?

Or maybe we just set the elevation for the view and set statelistanimator to `null`, allowing the user to use external plugins if they want to extend the default behavior?

If we can't find a good solution now, we can just set the `Outline` alpha to 0 to disable it (bound to a property) for the time being

PR #7067, opened 2019-03-24
chore: fix app.start deprecation

manoldonev:
Currently app.run relies on app.start so deprecation warning was shown all the time.


manoldonev:
test package_version#rc

Issue #3671, opened 2017-02-20
Feature request:  ClickableSpan support for iOS and Android 


tsonevn:
Ability to use clickable Spans in a NativeScirpt project.

At the moment as workaround it could be used `ClickableSpan` for `Android` and `NSLinkAttribute` for `iOS`. For `NativeScript TypeScirpt` or `NativeScript Angular 2` project we should also install `tns-platform-declarations` with `npm install tns-platform-declarations --save` and to follow the setup instructions [here](https://www.npmjs.com/package/tns-platform-declarations).

For `iOS` also should be implemented `textViewShouldInteractWithURLInRange` delegate method, which will be fired, when some of the clickable text parts have been selected.

 An example could be found [here](https://github.com/tsonevn/clickableSpanIOSAndroid).

BMwanza:
@tsonevn Is there anyway to get this working as a Label and not a Text View? Having trouble styling the font of the TextView, the font size won't increase

tsonevn:
Hi @BMwanza,
For Android, the approach for setting up a clickable span will work in the same way while using a Label. For iOS, you can set up the attributed text to the Label, however, you should implement different event listener for handling the tap event. Check out [this thread](https://stackoverflow.com/a/28519273/6133327) in StackOverflow

haingdc:
Does anyone does it work in the 2019? I have to admit the sample is quite challenge to get into, "nativescript" and "angular" is new to me.

Arystosedes:
Any news?

luis-gmonkey:
> Any news?

For those that land here. Using _linkTap_ works:

Example:

```
<StackLayout class="helpdesk">
        <Label textWrap="true">
          <FormattedString>
            <Span text="{{ L('login_page.helpdesk') + ' ' }} " class="text" />
            <Span text="{{ entitySettingsObj.supportEmail }}" linkTap="{{ openEmail }}" class="content--link"  />
          </FormattedString>
        </Label>
</StackLayout>
```

CatchABus:
Closed by #8256

PR #7076, opened 2019-03-26
(WIP) feat: TappableSpan support

edusperoni:
## PR Checklist

There is no way to add links to Labels/TextViews.

## What is the new behavior?
Span now has a `linkClick` event which, when bound, makes it possible for it to be interacted with via tap.

Implements #3671.


edusperoni:
If possible, I'd like that the core team test this implementation on Android and validate it's styling, so I can replicate it on iOS.

Challenges:

1. On iOS, making a Label clickable requires a [dirty hack](https://stackoverflow.com/a/28519273/6133327).
2. I have no idea if this works when the parent is a button (further testing needed).

There's an option to use the usual link highlighting (some adaptations are needed, or we could depend on https://github.com/saket/Better-Link-Movement-Method) to get this effect:

image

We could accomplish it by setting something like `android.highlightColor`, but when you override the span's background color, the highlight color never appears. This means we would not be able to create a span with a background that highlights.

This is also not possible on iOS (at least not in a trivial way). If our goal is to have the same style on both platforms, we can skip this step.

shiv19:
Hi @edusperoni!
Can we please rename this to TappableSpan,
and change the name of the event to `linktap` since "click" is not a term that is used on mobile. :)

edusperoni:
@shiv19 Absolutely, I used `CliackableSpan` because Android actually defines it like that (https://developer.android.com/reference/android/text/style/ClickableSpan). I'll update it to match `TappableSpan`

elvticc:
Hello ! Great improvement, look forward to it being implemented :) Thank you very much for your work.

edusperoni:
Closing in favor of #8256

Issue #7065, opened 2019-03-22
White space below action bar on IOS


brandon-sigao:
**Environment**
 - CLI: 5.2.4
 - Cross-platform modules: 5.2.2
 - Android Runtime: 5.2.1
 - iOS Runtime: 5.2.0
 - Nativescript Angular

**Describe the bug**
On IOS, navigating to a page results in the page content not extending to the action bar and leaving a large gap of white space.  User input such as a tap event will cause the page the redraw into the correct format without the white space.

**To Reproduce**
I honestly don't have a clue what specifically makes this bug occur.  I suspect its related to recently upgrading to "tns-core-modules: 5.2.2" as we weren't having these issues a few weeks ago, but I could be wrong. The best I can give you is that it happens on initial page load and persists until you interact with the page, at which point the page will be formatted correctly.

If it helps, we are also using flat="true" on all of our action bars.

**Additional context**
An example can be found [here](https://imgur.com/2KjBov9).  

**EDIT 1**
I'm in the process of switching back to 5.2.1 to confirm my suspicions.  As a work around, adding "changeDetectorRef.detectChanges()" at the end of the ngOnInit function seems to fix the problem, but I'm hoping to find a solution that doesn't involve injecting that into every single page.

**EDIT 2**
We confirmed that the issue exists only on tns-core-modules v5.2.2, reverting back to 5.2.1 solves this.

tsonevn:
Hi @brandon-sigao,
I was able to recreate the case with  `flat="true"` in a non-Angular project and will mark the issue as a bug that we will investigate further. I also checked the workaround with `detectChanges()` suggested in your comment in an NS Angular project, however, it does not seem to fix the issue on my side. 
Regarding that, as a temporary solution, I would suggest removing the `flat` property from the ActionBar.
[Archive.zip](https://github.com/NativeScript/NativeScript/files/3001984/Archive.zip)

brandon-sigao:
@tsonevn Thanks for the update.  We ended up sticking with 5.2.1 as there wasn't any real need to jump to 5.2.2 right now and 'flat="true"' is necessary for our app design to work.

djmgh:
I'm seeing the same behavior in NS Vue and not using flat="true"

lohiarahul:
> @tsonevn Thanks for the update. We ended up sticking with 5.2.1 as there wasn't any real need to jump to 5.2.2 right now and 'flat="true"' is necessary for our app design to work.

[...]

PR #7077, opened 2019-03-27
fix(ios): flat action bar incorrect layout after navigation

MartoYankov:
Fixes https://github.com/NativeScript/NativeScript/issues/7065


MartoYankov:
test package_version#latest

PR #7082, opened 2019-03-28
fix: revert fromUrl deprecate warning

MartoYankov:


MartoYankov:
test package_version#latest

PR #7092, opened 2019-03-31
feat(android): implement a 'activityNewIntent' event

### LINKED ISSUE #2747: [FEATURE] onNewIntent event listener
_From @roblav96 on September 17, 2016 1:59_

I'm not too sure how hard it would be to implement, but having a `onNewIntent` event would be awesome.

I'm trying to make a custom url plugin but in order for it to work properly I need to set `android:launchMode="singleTask"` on the `com.tns.NativeScriptActivity` in `AndroidManifest.xml` because if I don't, then the intent launches a whole new instance of my app on top of the already existing one.

I tried this with every android event available:

``` typescript
application.android.on(application.AndroidApplication.activityResumedEvent, function(args: application.AndroidActivityEventData) {
    global.tnsconsole.info('application.AndroidApplication.activityResumedEvent >')
    let intent: android.content.Intent = args.activity.getIntent()
    global.tnsconsole.log('intent.getData()', intent.getData())
    global.tnsconsole.log('intent.getDataString()', intent.getDataString())
    global.tnsconsole.log('intent.getScheme()', intent.getScheme())
    global.tnsconsole.log('intent.getExtras()', intent.getExtras())
})
```

and they all come up empty.

According to [EddyVerbruggen/Custom-URL-scheme](https://github.com/EddyVerbruggen/Custom-URL-scheme/blob/master/src/android/nl/xservices/plugins/LaunchMyApp.java#L73), he calls `onNewIntent` which is exactely what I'm looking to do. I also tried making my custom activity and overwriting the event, but it doesn't exist in `ui/frame/AndroidActivityCallbacks`.

This addition would be greatly appreciated!

_Copied from original issue: NativeScript/android-runtime#558_

EddyVerbruggen:
This resolves #2747 by implementing a 'new intent' event that gets triggered if the Android activity receives an `onNewIntent` event. Because Android won't invoke this method when an Intent is used to coldstart the app, we also check for an event in `onCreate`.

Notable usecases for having this event are:

- Using Custom URL schemes
- Using Dynamic Links (fi. [through the Firebase plugin](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/692))
- Depending on the plugin/implementation it may help catching the notification data when the app was not running (killed).
- Any other reason an app or plugin has to override the Android activity in order to 'catch' the intents will no longer have to override the activity. For instance, using [nativescript-nfc](https://github.com/EddyVerbruggen/nativescript-nfc/tree/7e811b4aae9033968fed2d513d7e208c0c707eeb#android-setup) will become a lot easier and less error prone.

### DIFF
--- a/tns-core-modules/ui/frame/activity.android.ts
+++ b/tns-core-modules/ui/frame/activity.android.ts
@@ -27,11 +27,27 @@ class NativeScriptActivity extends android.support.v7.app.AppCompatActivity {
         }
 
         this._callbacks.onCreate(this, savedInstanceState, super.onCreate);
+
+        const intent = this.getIntent();
+        if (intent && intent.getAction()) {
+            appModule.android.notify(<appModule.AndroidActivityNewIntentEventData>{
+                eventName: appModule.AndroidApplication.activityNewIntentEvent,
+                object: appModule.android,
+                activity: this,
+                intent
+            });
+        }
     }
 
     public onNewIntent(intent: android.content.Intent): void {
         super.onNewIntent(intent);
         super.setIntent(intent);
+        appModule.android.notify(<appModule.AndroidActivityNewIntentEventData>{
+            eventName: appModule.AndroidApplication.activityNewIntentEvent,
+            object: appModule.android,
+            activity: this,
+            intent
+        });
     }
 
     public onSaveInstanceState(outState: android.os.Bundle): void {
--- a/tns-core-modules/ui/frame/activity.android.ts
+++ b/tns-core-modules/ui/frame/activity.android.ts
@@ -26,28 +26,11 @@ class NativeScriptActivity extends android.support.v7.app.AppCompatActivity {
             setActivityCallbacks(this);
         }
 
-        this._callbacks.onCreate(this, savedInstanceState, super.onCreate);
-
-        const intent = this.getIntent();
-        if (intent && intent.getAction()) {
-            appModule.android.notify(<appModule.AndroidActivityNewIntentEventData>{
-                eventName: appModule.AndroidApplication.activityNewIntentEvent,
-                object: appModule.android,
-                activity: this,
-                intent
-            });
-        }
+        this._callbacks.onCreate(this, savedInstanceState, this.getIntent(), super.onCreate);
     }
 
     public onNewIntent(intent: android.content.Intent): void {
-        super.onNewIntent(intent);
-        super.setIntent(intent);
-        appModule.android.notify(<appModule.AndroidActivityNewIntentEventData>{
-            eventName: appModule.AndroidApplication.activityNewIntentEvent,
-            object: appModule.android,
-            activity: this,
-            intent
-        });
+        this._callbacks.onNewIntent(this, intent, super.setIntent, super.onNewIntent);
     }
 
     public onSaveInstanceState(outState: android.os.Bundle): void {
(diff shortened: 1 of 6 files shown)


tns-core-modules/ui/frame/activity.android.ts
```diff
@@ -27,11 +27,27 @@ class NativeScriptActivity extends android.support.v7.app.AppCompatActivity {
         }
 
         this._callbacks.onCreate(this, savedInstanceState, super.onCreate);
+
+        const intent = this.getIntent();
+        if (intent && intent.getAction()) {
+            appModule.android.notify(<appModule.AndroidActivityNewIntentEventData>{
```
MartoYankov: We fire other such events from the callbacks. I'd suggest moving this code to the `onCreate` callback for consistency and readability. You can add the `intent` as a parameter. All the callbacks are located in `frame.android.ts` under `ActivityCallbacksImplementation`.

tns-core-modules/ui/frame/activity.android.ts
```diff
@@ -27,11 +27,27 @@ class NativeScriptActivity extends android.support.v7.app.AppCompatActivity {
...
+                activity: this,
+                intent
+            });
+        }
     }
 
     public onNewIntent(intent: android.content.Intent): void {
         super.onNewIntent(intent);
         super.setIntent(intent);
+        appModule.android.notify(<appModule.AndroidActivityNewIntentEventData>{
```
MartoYankov: Same as the above, in this case you can create a new callback since we currently don't have one for `onNewIntent`.

MartoYankov:
Btw, what do you think about the `super.setIntent()` call inside the `onNewIntent()` callback? Right now, we call it always and change the app's Intent, but as far as I know, there can be cases where you don't need or want this behavior.

tns-core-modules/ui/frame/activity.android.ts
EddyVerbruggen: Good catch 👍

EddyVerbruggen:
@MartoYankov thanks for the quick review and feedback! I've tried to implement the changes per your suggestions.

Regarding the `setIntent` call, I don't have a strong opinion about that but either way I don't want to add that change to this PR.

MartoYankov:
@EddyVerbruggen I agree. The reason why I was asking, was because now that the developer will have access to this new intent, they have control and might not want to set it as the main intent. I guess, we'll see if this comes up.

tns-core-modules/ui/frame/frame.android.ts
```diff
@@ -944,7 +944,7 @@ class ActivityCallbacksImplementation implements AndroidActivityCallbacks {
     }
 
     @profile
-    public onCreate(activity: android.support.v7.app.AppCompatActivity, savedInstanceState: android.os.Bundle, superFunc: Function): void {
+    public onCreate(activity: android.support.v7.app.AppCompatActivity, savedInstanceState: android.os.Bundle, intent: android.content.Intent, superFunc: Function): void {
```
MartoYankov: I understand the reason for the `superFunc` to be the last parameter, but this now creates a breaking change for apps that use a custom Activity.

We discussed this internally and we think the best way to do this now is to create an overload for `onCreate` in the d.ts and deprecate the one with the 3 parameters. Then in the implementation you would infer the arguments from the number of arguments. We will then remove the deprecated method overload with 6.0. Does this make sense?

See how this is done with the `showModal` method for reference - https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/core/view-base/view-base.d.ts#L206

EddyVerbruggen: Solid point. I've done my best to implement that with my latest commit.

PR #7098, opened 2019-04-03
chore: add deprecation console warnings

MartoYankov:


tns-core-modules/ui/core/view/view-common.ts
```diff
@@ -259,7 +259,14 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
             if (args.length === 2) {
                 options = <ShowModalOptions>args[1];
             } else {
-                // TODO: Add deprecation warning
+                if (args[0] instanceof ViewCommon) {
+                    console.log("showModal(view: ViewBase, context: any, closeCallback: Function, fullscreen?: boolean, animated?: boolean, stretched?: boolean) " +
+                        "is deprecated. Please call showModal(view: ViewBase, modalOptions: ShowModalOptions) instead.");
```
manoldonev: In other deprecations we use just `Use ...` instead of `Please call ...`

PR #7129, opened 2019-04-10
export es2015 module to allow tree shaking in webpack

farfromrefug:
- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#running-unit-tests.
- [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/WritingUnitTests.md.

## What is the current behavior?
produced js files use ```require```

## What is the new behavior?
produced js file use ```import```, enabling module concatenation in Webpack

Also it keeps the comments in produced ```.js```. this is very important to allow ```terser``` to remove unused classes.
Related Typescript issues https://github.com/Microsoft/TypeScript/issues/15857 https://github.com/rollup/rollup/issues/1763#issuecomment-347272579 https://github.com/Microsoft/TypeScript/issues/10467
On its own keeping the comments wont help in removing unused classes. This is because Tyepscript uses the ```/** @class */``` comment while ```terser``` is looking for ```PURE```
however a simple addition of the a ``loader``` in webpack like this:
```javascript
{
                    test: /\.js$/,
                    exclude: /node_modules/,
                    use: [
                        babelLoader,
                        {
                            loader: 'string-replace-loader',
                            options: {
                                search: '\\/\\*\\* @class \\*\\/',
                                replace: '/*@__PURE__*/',
                                flags: 'g'
                            }
                        }
                    ]
                },
```
makes it work. I have tested it and unused classes are removed from my app bundle.

This is related to the tree shaking issue https://github.com/NativeScript/NativeScript/issues/7127

Issue #7127, opened 2019-04-10
True tree shaking support


farfromrefug:
**Environment**
- CLI:  all
 - Cross-platform modules: none
 - Android Runtime: all
 - iOS Runtime: all
 - Plugin(s): none

**Describe the bug**
{N} is not ```tree shaking``` compatible.
Even though it is mentioned on the ```nativescript-dev-webpack``` page that it supports tree-shaking, you wont get any tree shaking in {N} (almost none!)

This is kind of a big issue with a lot of sparated issues:
- {N} js product files for ```tns-core-modules``` or any other plugin are using commonjs ```require``` and not ```es2015``` ```import```. Which makes them incompatible with webpack [concatenation module](https://webpack.js.org/plugins/module-concatenation-plugin/)
- a LOT of ```tns-core-modules``` files have ```export * from '../../core/view``` or something like this. This create redondancy which implies a lot bigger concatenated modules (when module concatenation is possible). 
- [bundle-entry-points](https://github.com/NativeScript/NativeScript/blob/b436ecde3605b695a0ffa1757e38cc094e2fe311/tns-core-modules/bundle-entry-points.ts) requires all tns widgets even if they are not used. Because of that everything is always included, even the app only use half of the widgets

All 3 aspects can be handled /  fixed separately. But all of them are really necessary to get tree shaking to work as it should. Getting this to work would have multiple benefits:
- lower app bundle size
- lower vendor.js size in bundle => smaller snapshot / less memory / faster runtime
- faster start time as it is much faster to require everything. 

I have already started to work on the first 2 points. Two PRs are coming:
- es2015 module is small and without any real consequence
- fixing all {N} imports / exports to not use ```export * from ...``` is a bigger one with consequences for the user

The 3rd point needs to be addressed by the CLI in the sense that it should find out what widgets are used by the app (parsing XML, VUE files) and create a ```bundle-entries.js``` in consequence only importing what's really used

**To Reproduce**
build with webpack using args ```--display-used-exports --display-modules --display-optimization-bailout```

vakrilov:
Hey @farfromrefug, it is awesome that you have opened this issue. This is definitely the direction we want to go with the `tns-core-modules` package (and for all the plugins in the NS eco-system as well). 

Here are my thoughts on the 3 issues that you have listed:

### Shipping `tns-core-modules` as `es2015` modules
[...]

PR #7130, opened 2019-04-10
Remove all export * from ....

farfromrefug:

Related to the tree shaking issue https://github.com/NativeScript/NativeScript/issues/7127

right now a lot of ```tns-core-modules``` modules use ```export * from...```. I suppose this is done out of practicality. However in webpack this has a big consequences when you are trying to use [tree shaking] (https://webpack.js.org/guides/tree-shaking/)
Tree shaking use module concatenation to bundle an app to its smallest size. However the fact that {N} is using all those ```export * from...``` make it so that all those exports need to be referenced in all modules, ending with a much bigger ```vendor.js```.

As a simple example, apply that PR https://github.com/NativeScript/NativeScript/pull/7129 to {N} then try to bundle an app in production with webpack. Thanks to the PR tree shaking will be working in {N} and so will module concatenation. However because of all those ```export * from``` you will end up with a bigger ```vendor.js```. In my test :
- without es2015 and thus without any tree shaking => ```vendor.js```: 1.2Mo
- with es2015 and thus with any tree shaking => ```vendor.js```: 1.5Mo => 20% increase.

Now using this PR combined with the one for es2015 module i end up with a ```vendor.js``` of 1.2Mo. You can then argue that there is no gain there!!! Well yes and that's because of the ```bundle-entries.ts``` file which ensure that all {N} widgets are always bundled.
Those 2 PRs will show their true effects once ```bundle-entries.js```  will be generated at compilation based on the app used widgets.
However the faster we apply those 2 PRs, the less work it requires.

## WARNING

This PR has bad side effects for the user. Some imports might not be good anymore.
If the user use auto completion, it will often end up with imports like this :
``` import { Property } from 'tns-core-modules/core/view```
However that PR breaks all those imports. Instead the user must change it too
``` import { Property } from 'tns-core-modules/ui/core/properties```

I know this is kind of a big deal, however i really feel this is a necessary step into improving {N} apps performances. Maybe this could be part of Nativescript 6.0

Maybe  a migration script could change all those imports?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,6 @@
         "**/*.js": true,
         "**/*.js.map": true
     },
-    "tslint.configFile": "./build/tslint.json"
+    "tslint.configFile": "./build/tslint.json",
+    "typescript.tsdk": "node_modules/typescript/lib"
 }
--- a/apps/app/modaltest/main-page.ts
+++ b/apps/app/modaltest/main-page.ts
@@ -1,4 +1,5 @@
-import { View, EventData } from "tns-core-modules/ui/core/view";
+import { View } from "tns-core-modules/ui/core/view";
+import { EventData } from "tns-core-modules/data/observable";
 
 export function onNavigatingTo(args: EventData) {
     const page = args.object;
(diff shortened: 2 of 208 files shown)


manoldonev:
test

farfromrefug:
@manoldonev working on the tslint errors right now

manoldonev:
test

vakrilov:
Closing this one for now as it requires a bit more testing if it is really needed (will reopen if this is the case).
Keep the discussion in #7127

farfromrefug:
As long as the discussion remains open, i am ok with this. Actually found issues with pure javascript apps

PR #7136, opened 2019-04-11
feat(android): elevation shadow support

### LINKED ISSUE #550: Implement CSS Box Shadow

edusperoni:
## PR Checklist

No elevation support

## What is the new behavior?
Any view can have a shadow

Fixes/Implements/Closes #550.


edusperoni:
As requested by @manoldonev (https://github.com/NativeScript/NativeScript/pull/7055#issuecomment-482074141)

This is a simple way to support elevation + statelistanimator on android

For some reason, `setElevation` and some other functions throw TS errors (it seems to be using API 17 declaration).


## TODO

1. check API version >=21 (running this on API <21 will result in a crash without it)
2. remove `any` casts
3. implement iOS version (some views set clipsToBounds, which makes the shadow disappear. nativescript-ngx-shadow solves this by wrapping the view in a `StackLayout`)

tns-core-modules/ui/styling/style-properties.ts
```diff
@@ -1004,6 +1004,12 @@ function opacityConverter(value: any): number {
 export const opacityProperty = new CssAnimationProperty<Style, number>({ name: "opacity", cssName: "opacity", defaultValue: 1, valueConverter: opacityConverter });
 opacityProperty.register(Style);
 
+export const elevationProperty = new CssAnimationProperty<Style, number>({ name: "elevation", cssName: "elevation", defaultValue: 0, valueConverter: parseInt });
```
manoldonev: No need to declare the new properties as CssAnimationProperty (or CssProperty for that matter). With NativeScript all simple properties (i.e. declared as Property<...>) can be set through CSS by converting their naming to kebab case (looks-like-this).

Our guideline would be to declare CssProperty only for properties that are actually part of the CSS specification (and yes, there is an `elevation` property in css spec but it does something completely different).

Also, as obviously we haven't communicated this feature very well (the ability to set simple properties via CSS) we are currently discussing an enhancement that should make this more obvious (and also clearly mark simple vs css properties when used in CSS) -- by using something like CSS vendor prefixes (-ns- in our case).

Currently you can do this in css:
``` css
.btn {
    color: red;
    elevation: 8;
    some-other-simple-prop: 0;
}
```
We are considering also enabling this syntax:
``` css
.btn {
    color: red;
    -ns-elevation: 8;
    -ns-some-other-simple-prop: 0;
}
```

tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -709,6 +715,101 @@ export class View extends ViewCommon {
         this.nativeViewProtected.setAlpha(float(value));
     }
 
+    [elevationProperty.getDefault](): number {
+        console.log("elevationProperty.getDefault");
+        return undefined;
```
manoldonev: We usually return -1; in such scenarios.

tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -709,6 +715,101 @@ export class View extends ViewCommon {
...
+    }
+
+    [elevationProperty.setNative](value: number) {
+        if (sdkVersion() < 21) { return; }
+        console.log("elevationProperty.setNative, value: " + value);
+        const nativeView: any = this.nativeViewProtected;
+        if (value !== undefined) {
```
manoldonev: Better switch check to ` if(value != null)`

tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -709,6 +715,101 @@ export class View extends ViewCommon {
...
+        } else {
+            // reset to original value
+            if (this._originalElevation !== undefined) {
+                nativeView.setElevation(this._originalElevation);
+            }
+        }
+        this._refreshStateListAnimator();
+    }
+
+    [androidPressedZProperty.getDefault](): number {
```
manoldonev: [Material design](https://material.io/design/environment/elevation.html#elevation-in-material-design) calls this dynamic elevation offset -- maybe rename ours to dynamicElevation or dynamicElevationOffset?

Also, you have added the android prefix to this property but not to `elevation` -- if dynamic elevation would be eventually supported for iOS as well, we can drop the prefix.

tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -709,6 +715,101 @@ export class View extends ViewCommon {
...
+    }
+
+    [androidPressedZProperty.setNative](value: number) {
+        if (sdkVersion() < 21) { return; }
+        console.log("androidPressedZProperty.setNative, value: " + value);
+        this._refreshStateListAnimator();
+
+    }
+
+    _refreshStateListAnimator() {
```
manoldonev: I was testing the outline button scenario with this markup (btn and btn-outline are classes from nativescript-theme-core):
``` xml
<Button class="btn btn-outline" text="tap" tap="onTap" />
```
But it does not work out-of-the-box despite default elevation is set to 0.

This works:
``` xml
<Button class="btn btn-outline" elevation="0" text="tap" tap="onTap" />
```

edusperoni: the "default" elevation in `undefined` or "not specified", in which case we don't override the view's default. The android button has an elevation by default (and a statelistanimator), so I opted to just override when explicitly setting elevation/androidPressedZ.

This is similar as when we set the button color, just the color changes, but when we set a border-radius, the whole background is replaced by a BorderDrawable.

tns-core-modules/ui/core/view/view.android.ts
edusperoni: I'm still wondering about this whole "elevation" for both platforms. React Native exposes the full properties (https://ethercreative.github.io/react-native-shadow-generator/), but doesn't allow for a simple `elevation` property, or a dynamic elevation offset. This is something that we should probably consider before adding the iOS implementation.

But I agree that setting it to dynamicElevationOffest seems more intutive.

tns-core-modules/ui/styling/style-properties.ts
edusperoni: I'm also using style to check if the property is being overwritten or not. Should I just move the property to the view and switch the checks to `if(elevation < -1)` for example? This also raises the issue that the elevation offset could be negative, so checking if it doesn't exist becomes an issue

tns-core-modules/ui/core/view/view.android.ts
manoldonev: Yeah, I guess default elevation being "unset" makes sense in the broader context of a NativeScript View -- to achieve the desired outline button we will just need to set `elevation: 0;` in the btn-outline class in nativescript-theme-core.

tns-core-modules/ui/styling/style-properties.ts
```diff
@@ -1004,6 +1004,12 @@ function opacityConverter(value: any): number {
...
+elevationProperty.register(Style);
+
+export const androidPressedZProperty = new CssAnimationProperty<Style, number>({ name: "androidPressedZ", cssName: "android-pressed-z", defaultValue: 0, valueConverter: parseInt });
```
manoldonev: As discussed above -- let's switch defaultValue to `null`.

tns-core-modules/ui/styling/style-properties.ts
manoldonev: Btw noticed that android elevation is floating point number -- maybe switch the valueConverter to parseFloat?

tns-core-modules/ui/core/view/view.android.ts
manoldonev: I think if we get it to work in a similar fashion with simple API (elevation & dynamicElevationOffset) on both platforms we should be good to go based on the voting [here](https://github.com/NativeScript/NativeScript/issues/550#issuecomment-448634541).

Or you don't think we can achieve similar appearance with this API on iOS?

tns-core-modules/ui/core/view/view.android.ts
manoldonev: Let's replace the `undefined` usages with `null` (and the `undefined` checks with != null).

tns-core-modules/ui/core/view/view.android.ts
```diff
@@ -709,6 +715,101 @@ export class View extends ViewCommon {
...
+                nativeView.setStateListAnimator(this._originalStateListAnimator);
+            }
+            this._originalStateListAnimator = undefined;
+            return;
+        }
+        const sla = new (android.animation as any).StateListAnimator();
+
+        const ObjectAnimator = android.animation.ObjectAnimator;
+        const AnimatorSet = android.animation.AnimatorSet;
+        const shortAnimTime = android.R.integer.config_shortAnimTime;
```
manoldonev: With respect to https://github.com/NativeScript/android-runtime/issues/1330 -- I don't think we really need to access android.R constants directly here -- we can store the necessary values as constants in TypeScript:
``` ts
const config_shortAnimTime = 17694720;
const state_pressed = 16842919;
const state_enabled = 16842910;
```

edusperoni:
@manoldonev I've switched the properties to the view. Unfortunately, when we unset the properties, they are never unset from the view (they never return to undefined). Is there any way to force them to return to undefined when unset?

tns-core-modules/ui/core/view/view.android.ts
edusperoni: We can achieve it, just wondering because we had the request to allow for extra iOS properties on `nativescript-ngx-shadow` (https://github.com/JoshDSommer/nativescript-ngx-shadow/issues/13), so we could allow it in this implementation.

manoldonev:
About unsetting local values -- seems like this is not implemented currently https://github.com/NativeScript/NativeScript/blob/4e56c89f7d63f69a8c8bd77a11314941851c1107/tns-core-modules/ui/styling/style-scope.ts#L472.

Will have to think about it a bit.

edusperoni:
My naive implementation worked:

```typescript
            if (!(key in newPropertyValues)) {
                if (key in this.view.style) {
                    this.view.style[`css:${key}`] = unsetValue;
                } else {
                    // TRICKY: How do we unset local value?
                    const camelCasedProperty = key.replace(/-([a-z])/g, function (g) { return g[1].toUpperCase(); });
                    this.view[camelCasedProperty] = unsetValue;
                }
            }
```

if the property has been defined as a `Property`, this should work. This won't work when it's not a `Property`, though. If there's a way to check that, we should have a solution for properties with default values, for example.

manoldonev:
Hey @edusperoni 

I've been brainstorming a bit on the proposed elevation support implementation and after discussions with the team here I've made a commit with the following propositions:
- rename the public APIs to designate that currently we provide Android only functionality (i.e. androidElevation and androidDynamicElevationOffset); when iOS equivalent is implemented we will deprecate the android specific properties and introduce common elevation and dynamicElevationOffset properties for both platforms.
- drop the undefined / null convention for unset androidElevation / androidDynamicElevationOffset (as it is really confusing to the end user) and provide the underlying android values instead
- restore your original implementation for androidElevation / androidDynamicElevationOffset to be CssProperties (vs view Properties). We will not be trying to patch the missing functionality for unsetting local values for view properties as part of this PR.
- Drop the caching of the original stateListAnimator as part of the refreshStateListAnimator() method

edusperoni:
@manoldonev  Sounds good! But won't removing the null/undefined checks actually cause an issue with plugins like `nativescript-floatingactionbutton` and `nativescript-cardview`, that actually have custom elevations set by other means, or have non-defaults `dynamicElevationOffest` (since we're no longer caching the original animator, which was similar to what we do with android `background`).

SvetoslavTsenov:
test ios#rc

manoldonev:
@edusperoni do you have a specific scenario in mind? If one is not using the new androidElevation / androidDynamicElevationOffset APIs explicitly we are not tampering with the original animators so those plugins should be working as expected.

SvetoslavTsenov:
test ios#rc

Issue #7050, opened 2019-03-21
Using "UIModalPresentationStyle.Popover" for showModal causes the `closeCallback` to not be called


VladimirAmiorkov:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 5.2.0
 - Cross-platform modules: 5.3.0-next-2019-03-21-142557-02
 - Android Runtime:
 - iOS Runtime: 5.3.0-2019-03-21-073518-01
 - Plugin(s):

**Describe the bug**
When setting the `presentationStyle` of the `ShowModalOptions` to `Popover`:
```
ios: {
    presentationStyle: UIModalPresentationStyle.Popover
}
```
the `closeCallback` of the `ShowModalOptions` is never called if you tap outside of the modal view to close it. 

**To Reproduce**
Change the https://github.com/NativeScript/NativeScript/blob/master/e2e/modal-navigation/app/home/home-page.ts#L38-L41 to:

```
let options: ShowModalOptions = {
    context: "context",
    closeCallback: () => console.log("home-page modal page closed"),
    fullscreen: true,
    ios: {
        presentationStyle: UIModalPresentationStyle.Popover
    },
};
view.showModal(frame, options);
```

**Expected behavior**
The `closeCallback` of the `ShowModalOptions` is called when you tap outside of the shown modal view in order to close it.

PR #7161, opened 2019-04-17
fix(ios-modal): closeCallback not being called with popover presentation style

VladimirAmiorkov:
## PR Checklist

## What is the new behavior?

Fixes https://github.com/NativeScript/NativeScript/issues/7050

## Developer notes
This PR introduces a new event `popoverClosed` to the `View` class. This event is raised when a modal using the `presentationStyle: UIModalPresentationStyle.Popover` is closed by tapping outside of the popup area.

### DIFF
--- a/e2e/modal-navigation/app/home/home-page.ts
+++ b/e2e/modal-navigation/app/home/home-page.ts
@@ -1,7 +1,7 @@
 import * as application from "tns-core-modules/application";
 import { StackLayout } from "tns-core-modules/ui/layouts/stack-layout";
 import { NavigatedData, Page } from "tns-core-modules/ui/page";
-import { View, EventData } from "tns-core-modules/ui/core/view";
+import { View, EventData, ShowModalOptions } from "tns-core-modules/ui/core/view";
 import { Frame } from "tns-core-modules/ui/frame";
 
 export function onNavigatingTo(args: NavigatedData) {
@@ -27,6 +27,20 @@ export function onModalNoPage(args: EventData) {
         "context",
         () => console.log("home-page modal frame closed"),
         false);
+}
+
+export function onPopoverModal(args: EventData) {
+    const view = args.object as View;
+    let options: ShowModalOptions = {
+        context: "context",
+        closeCallback: () => console.log("home-page modal popover frame closed"),
+        animated: false,
+        ios: {
+            presentationStyle: UIModalPresentationStyle.Popover
+        }
+    }
+
+    view.showModal("modal-no-page/modal-no-page", options);
 }
 
 export function onModalFrame(args: EventData) {
--- a/tns-core-modules/ui/core/view/view.ios.ts
+++ b/tns-core-modules/ui/core/view/view.ios.ts
@@ -29,6 +29,7 @@ const majorVersion = iosUtils.MajorVersion;
 export class View extends ViewCommon {
     nativeViewProtected: UIView;
     viewController: UIViewController;
+    private _popoverPresentationDelegate: ios.UIPopoverPresentationControllerDelegateImp;
 
     private _isLaidOut = false;
     private _hasTransfrom = false;
@@ -416,6 +417,8 @@ export class View extends ViewCommon {
 
             if (presentationStyle === UIModalPresentationStyle.Popover) {
                 const popoverPresentationController = controller.popoverPresentationController;
+                this._popoverPresentationDelegate = ios.UIPopoverPresentationControllerDelegateImp.initWithOwner(new WeakRef(this), options.closeCallback);
+                popoverPresentationController.delegate = this._popoverPresentationDelegate;
                 const view = parent.nativeViewProtected;
                 // Note: sourceView and sourceRect are needed to specify the anchor location for the popover.
                 // Note: sourceView should be the button triggering the modal. If it the Page the popover might appear "behind" the page content
@@ -906,7 +909,7 @@ export namespace ios {
 
         public viewDidLoad(): void {
             super.viewDidLoad();
-    
+
             // Unify translucent and opaque bars layout
             // this.edgesForExtendedLayout = UIRectEdgeBottom;
             this.extendedLayoutIncludesOpaqueBars = true;
@@ -979,6 +982,30 @@ export namespace ios {
             const owner = this.owner.get();
             if (owner && !owner.parent) {
                 owner.callUnloaded();
+            }
+        }
+    }
+
+    export class UIPopoverPresentationControllerDelegateImp extends NSObject implements UIPopoverPresentationControllerDelegate {
+        public static ObjCProtocols = [UIPopoverPresentationControllerDelegate];
+
+        private owner: WeakRef<View>;
+        private whenClosedCallback: Function;
+
+        public static initWithOwner(owner: WeakRef<View>, whenClosedCallback: Function): UIPopoverPresentationControllerDelegateImp {
+            const instance = <UIPopoverPresentationControllerDelegateImp>super.new();
+            instance.owner = owner;
+            instance.whenClosedCallback = whenClosedCallback;
+
+            return instance;
+        }
+
+        public popoverPresentationControllerDidDismissPopover(popoverPresentationController: UIPopoverPresentationController) {
+            const owner = this.owner.get();
+            if (owner) {
+                if (typeof this.whenClosedCallback === "function") {
+                    this.whenClosedCallback();
+                }
             }
         }
     }
--- a/e2e/modal-navigation/app/home/home-page.xml
+++ b/e2e/modal-navigation/app/home/home-page.xml
@@ -20,6 +20,7 @@
             <Button text="Show Modal Without Page" tap="onModalNoPage" textAlignment="left" />
             <Button text="Show Modal Page With Frame" tap="onModalFrame" textAlignment="left" />
             <Button text="Show Modal Page" tap="onModalPage" textAlignment="left" />
+            <Button text="Show 'popover' modal" tap="onPopoverModal" textAlignment="left" />
             <Button text="Show Modal Layout" tap="onModalLayout" textAlignment="left" />
             <Button text="Show Modal TabView" tap="onModalTabView" textAlignment="left" />
             <Button text="Android Back Btn Events" tap="onAndroidBackEvents" textAlignment="left" />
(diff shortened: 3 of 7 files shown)


tns-core-modules/ui/core/view/view.ios.ts
```diff
@@ -26,9 +26,33 @@ const PFLAG_LAYOUT_REQUIRED = 1 << 2;
 
 const majorVersion = iosUtils.MajorVersion;
 
+class UIPopoverPresentationControllerDelegateImp extends NSObject implements UIPopoverPresentationControllerDelegate {
```
MartoYankov: There is an 'ios' namespace at the bottom of this file where we put native iOS implementations. See https://github.com/NativeScript/NativeScript/blob/ca873ae69f078e3e08e49d52605a836c3d02f468/tns-core-modules/ui/core/view/view.ios.ts#L938.

tns-core-modules/ui/core/view/view.ios.ts
```diff
@@ -26,9 +26,33 @@ const PFLAG_LAYOUT_REQUIRED = 1 << 2;
...
+    public static ObjCProtocols = [UIPopoverPresentationControllerDelegate];
+
+    private _ownerRef: WeakRef<View>;
```
MartoYankov: Using an owner property is sort of a pattern in the core modules. We name it simply "owner" or "_owner" in other implementations. See https://github.com/NativeScript/NativeScript/blob/ca873ae69f078e3e08e49d52605a836c3d02f468/tns-core-modules/ui/core/view/view.ios.ts#L957.

tns-core-modules/ui/core/view/view.ios.ts
```diff
@@ -26,9 +26,33 @@ const PFLAG_LAYOUT_REQUIRED = 1 << 2;
...
+
+    public static initWithOwner(owner: WeakRef<View>): UIPopoverPresentationControllerDelegateImp {
+        const instance = <UIPopoverPresentationControllerDelegateImp>super.new();
+        instance._ownerRef = owner;
+
+        return instance;
+    }
+
+    public popoverPresentationControllerDidDismissPopover(popoverPresentationController: UIPopoverPresentationController) {
+        let ownerView = this._ownerRef.get();
```
MartoYankov: Should be `const`. Also, we call this simply `owner`.

tns-core-modules/ui/core/view/view.ios.ts
```diff
@@ -26,9 +26,33 @@ const PFLAG_LAYOUT_REQUIRED = 1 << 2;
...
+        if (ownerView) {
+            let modalClosingArgs: EventData = <EventData>{
```
MartoYankov: Should be `const`.

tns-core-modules/ui/core/view/view.d.ts
```diff
@@ -113,6 +113,11 @@ export abstract class View extends ViewBase {
      */
     public static showingModallyEvent: string;
 
+    /**
+     * String value used when hooking to popoverClosed event.
+     */
+    public static popoverClosedEvent: string;
```
MartoYankov: If this event should be used only internally for the time being when we don't have closing events, it should be private.

tns-core-modules/ui/core/view/view.ios.ts
```diff
@@ -449,10 +480,19 @@ export class View extends ViewCommon {
...
+            return;
+        }
+
         const parentController = parent.viewController;
         const animated = (<any>this.viewController).animated;
 
-        parentController.dismissViewControllerAnimatedCompletion(animated, whenClosedCallback);
+        if (this.viewController.popoverPresentationController && this.viewController.popoverPresentationController instanceof UIPopoverPresentationController) {
+            whenClosedCallback();
+            parentController.dismissViewControllerAnimatedCompletion(animated, null);
```
MartoYankov: I think this is a breaking change. The closed callback function will be called before the popover modal is closed?

tns-core-modules/ui/core/view/view.ios.ts
```diff
@@ -449,10 +480,19 @@ export class View extends ViewCommon {
             return;
         }
 
+        if (!this.viewController) {
```
MartoYankov: In which case would there be no view controller?

tns-core-modules/ui/core/view/view.d.ts
VladimirAmiorkov: At the moment it is not restricted to only internal use, for example I am using it in the nativescript-angular code to handled its ModalService closing implementation. I was thinking of implementing additional delegate methods of the `UIPopoverPresentationControllerDelegate` like this `popoverPresentationControllerShouldDismissPopover` and an related event that the user could handled in order to "dismiss" the closing but it felt like a bigger feature than this fix. I think its better suited for a different PR with that new functionality.

tns-core-modules/ui/core/view/view.ios.ts
VladimirAmiorkov: I do not think so. The actual popover modal has already been closed before hitting the new `popoverPresentationControllerDidDismissPopover` which calls the `_closeModalCallback` which finally calls this code.

tns-core-modules/ui/core/view/view.ios.ts
VladimirAmiorkov: In the case of `nativescript-angular`. This is something that may be better implemented not via events I am not sure but due to relying on event and their handlers we need to handled the `popoverClosedEvent` inside `nativescript-angular` but the way events work is that both that angular specific handler and the default one in `tns-core-modules` will both run so the first pass coming from the handler of the `popoverClosedEvent` in `nativescript-angular` will have a `viewControler` and execute correctly but the next pass coming from the `tns-core-modules` handler of `popoverClosedEvent` will be for an already closed modal.

If we want to use "event" to notify the `View` that an 'popover' has been dismissed by the OS we need to do the above code. One other implementation is to not use an `event` but some other technique that will allow is to notify an "observer" that a popover has been dismissed so that it can handled its own modal closing logic (for example the logic of ModalDialogService in `nativescript-angular`

tns-core-modules/ui/core/view/view.ios.ts
VladimirAmiorkov: Yes I will move this.

tns-core-modules/ui/core/view/view.ios.ts
VladimirAmiorkov: Got it, I will rename it accordingly to `owner`.

tns-core-modules/ui/core/view/view.ios.ts
VladimirAmiorkov: Got it, I will rename it accordingly and change it to `const`

tns-core-modules/ui/core/view/view.ios.ts
VladimirAmiorkov: I will change it to const

e2e/modal-navigation/app/home/home-page.ts
```diff
@@ -29,6 +29,20 @@ export function onModalNoPage(args: EventData) {
         false);
 }
 
+export function onPopoverModal(args: EventData) {
```
manoldonev: Actual test cases that use the test app are added here -- https://github.com/NativeScript/NativeScript/tree/master/e2e/modal-navigation/e2e

tns-core-modules/ui/core/view/view.ios.ts
```diff
@@ -982,4 +999,27 @@ export namespace ios {
...
+            instance.owner = owner;
+
+            return instance;
+        }
+
+        public popoverPresentationControllerDidDismissPopover(popoverPresentationController: UIPopoverPresentationController) {
+            const owner = this.owner.get();
+            if (owner) {
+                let modalClosingArgs: EventData = <EventData>{
+                    eventName: ViewCommon.popoverClosedEvent,
```
manoldonev: Do we really need to expose a new and very specific event for this? We do not have a modalClosedEvent at the moment and exposing a popover specific one seems strange (not to mention it is platform specific too).

Can't we just call `owner.closeModal()` here?

VladimirAmiorkov: Calling the `owner.closeModal()` would work for Vanila NativeScript apps, but for apps with `nativescript-angular` which has its own `ModalDialogService` that extends the logic of the modals from tns-core-modules it will not work. In that case the ModalDialogService itself has a logic to close any modal opened via its API and when a "popover" modal is closed we need to pass that "event" to the `ModalDialogService `. I have made a PR in `nativescript-angular` that uses the new event, you can find it [here](https://github.com/NativeScript/nativescript-angular/pull/1782/files).

I do get what you are saying tho and I have already discussed a possible alternative to using an event. We could use an "internal" function that is going to be simply called instead of raising an event, lets call that function "popoverModalCloser()". This will be a public field to the `View `class that is going to have an `undefined `initial value and we are going to assign it to a function in nativescript-agnular that will be able to call the closing logic of the `ModalDialogService`. The end result will be the same as using an event that the `ModalDialogService` will listen for but will can avoid having an public event. The thing that I don't like about such approach is that it is a bit of a "hidden" functionality due to the nature of calling an function that is set to undefined just in case someone outside of the class has set it.

What do you think?

e2e/modal-navigation/app/home/home-page.ts
VladimirAmiorkov: Hi @manoldonev ,

I talked with @SvetoslavTsenov to add one after the PR but I could try to implement one myself.

VladimirAmiorkov:
Closing in favor of new PR to the release branch: https://github.com/NativeScript/NativeScript/pull/7189

tns-core-modules/ui/core/view/view.ios.ts
VladimirAmiorkov: Refactored the implementation to not require the above code, made new PR from release branch [here](https://github.com/NativeScript/NativeScript/pull/7189).

tns-core-modules/ui/core/view/view.ios.ts
VladimirAmiorkov: Refactored the implementation to not require the above code, made new PR from release branch [here](https://github.com/NativeScript/NativeScript/pull/7189).

Issue #7160, opened 2019-04-17
Issues on IOS when getting XML in http GET request.


NickIliev:
@jockehaddock commented on [Mon Apr 15 2019](https://github.com/NativeScript/nativescript-angular/issues/1781)

**Environment**
 - Android Runtime: 5.2.1
 - iOS Runtime: 5.2.0
 - Plugin(s): --
 - NativeScript-Angular: 7.2.1
 - Angular: 7.2.0
 
**Describe the bug**
When doing a HTTP GET request for a XML resource in IOS the response has status 200 OK but request crashes during parsing even though its not supposed to do any parsing:
"status": 200,
"statusText": "Unknown Error",
ok": false,
"name": "HttpErrorResponse",
"message": "Http failure during parsing for https://www.domain.com/file.xml",
"error": {
"line": 173,
"column": 61,
"sourceURL": "file:///app/tns_modules/tns-core-modules/http/http-request/http-request.js"
ResponseType is set to 'text' for the request. In Android everything works fine.

I have debugged and found that http-request.ios.ts is the source of the crash at line 204:
return NSString.alloc().initWithDataEncoding(data, code).toString();

For some reason NSString.alloc().initWithDataEncoding returns null which crashes .toString()

**Expected behavior**
To get back text string with the XML file data.

---

@NickIliev commented on [Mon Apr 15 2019](https://github.com/NativeScript/nativescript-angular/issues/1781#issuecomment-483159844)

@jockehaddock can you share a REST API that returns  XML file on GET (so I could test the issue locally)?
Also, are you using the HttpClient? if possible please share a sample app or[ Playground ](https://play.nativescript.org/)demonstrating the issue.

---

@jockehaddock commented on [Mon Apr 15 2019](https://github.com/NativeScript/nativescript-angular/issues/1781#issuecomment-483179045)

For instance if i try to consume the api described here: `https://www.apfy.me/api/data.bytbil.com/ianreil/`

By doing:
`this.http.get('http://data.bytbil.com/sellers/satramotorcenter/xml/satramotorcenter.xml', { responseType: 'text' }).subscribe(res => console.log(res));`

I set up a simple app in the playground:
https://play.nativescript.org/?template=play-ng&id=TSEuED
(Not sure if its possible to link the playgrounds directly)

I get this error:
```
ERROR: ERROR {
"headers": {
"normalizedNames": {},
"lazyUpdate": null,
"headers": {}
},
"status": 200,
"statusText": "Unknown Error",
[...]

PR #7175, opened 2019-04-25
fix(ios-http-get): http failure during parsing xml

ADjenkov:
Unable to parse xml file that contains bytes encoded with ISO 8859-1. Check [here](https://stackoverflow.com/questions/3485190/nsstring-initwithdata-returns-null) for more info

Fix: https://github.com/NativeScript/NativeScript/issues/7160


SamoB25:
Merge merge merge :) I am anxiously waiting to use this :)

Issue #7226, opened 2019-05-09
Crash(iOS): TypeError: undefined is not an object (evaluating 'measureSpec.setColumn')


m-abs:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 5.3.1
 - Cross-platform modules: 5.3.2
 - Android Runtime: Not applicable
 - iOS Runtime: 5.3.1
 - Plugin(s): nativescript-angular@7.2.3

**Describe the bug**

In the past month we've tracked 20+ crashes with this error:
```
TypeError: undefined is not an object (evaluating 'measureSpec.setColumn')
```

**To Reproduce**

For now we only have the crash logs from Google Analytics.

**Expected behavior**

The app to not crash.

**Additional context**

I think problem is caused in this function:
https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/layouts/grid-layout/grid-layout.ios.ts#L73-L97

And I'm guessing the child layout doesn't have a `measureSpec` in the `GridLayout.map` here:
https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/layouts/grid-layout/grid-layout.ios.ts#L138

I suggest adding a check for `undefined` in `updateMeasureSpecs` and adding the missing `measureSpec` to the map..

m-abs:
Hi @NickIliev 
I see you've requested more info.

Unfortunately I haven't been able to write a test project that reproduce the problem.

I can reproduce it in our private app, but I can't share the code.

I've create a small PR that fixes the problem locally, but I'd really wish I had a proper way to reproduce it.

m-abs:
I've identified the components in our project that causes the problem.

The View tree is like this
```
Frame(2)
  Page(8)
     ProxyViewComponent -> library tabs component
        GridLayout(13) rows="*, auto, auto"
          ProxyViewComponent -> bookshelf component
             ProxyViewComponent -> lyt-failed-retry.component
                Stacklayout(44) < This is the child that is missing the MeasureSpecs in the GridLayout(13)
                   Stacklayout with *ngIf="false"
                       2x Label
```

Here is the template for `lyt-failed-retry.component`:
```html
<StackLayout>
  <StackLayout *ngIf="failed" class="p-5">
    <Label [text]="failedText" textWrap="true" class="t-14 text-center text-danger"></Label>
[...]

PR #7230, opened 2019-05-10
fix(ios): undefined is not an object (evaluating 'measureSpec.setColumn')

m-abs:
We, the rest of the NativeScript community, thank you for your
contribution! 
To help the rest of the community review your change, please follow the instructions in the template.
-->

## PR Checklist

- ~[ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/WritingUnitTests.md~

In the past month we've tracked 20+ crashes with this error:
```
TypeError: undefined is not an object (evaluating 'measureSpec.setColumn')
```

We haven't been able to consistently reproduce the crash locally, but I think it is clear where it does wrong.

## What is the new behavior?

`GridLayout(ios).onMeasure(...)` will create `MeasureSpecs` if missing the for `child` view when it calls `this.eachLayoutChild(...)`

Fixes: #7226

### DIFF
--- a/tns-core-modules/ui/layouts/grid-layout/grid-layout.ios.ts
+++ b/tns-core-modules/ui/layouts/grid-layout/grid-layout.ios.ts
@@ -137,8 +137,7 @@ export class GridLayout extends GridLayoutBase {
         this.eachLayoutChild((child, last) => {
             let measureSpecs = this.map.get(child);
             if (!measureSpecs) {
-                this._registerLayoutChild(child);
-                measureSpecs = this.map.get(child);
+                return;
             }
 
             this.updateMeasureSpecs(child, measureSpecs);
--- a/tns-core-modules/ui/layouts/grid-layout/grid-layout.ios.ts
+++ b/tns-core-modules/ui/layouts/grid-layout/grid-layout.ios.ts
@@ -136,6 +136,11 @@ export class GridLayout extends GridLayoutBase {
 
         this.eachLayoutChild((child, last) => {
             let measureSpecs = this.map.get(child);
+            if (!measureSpecs) {
+                this._registerLayoutChild(child);
+                measureSpecs = this.map.get(child);
+            }
+
             this.updateMeasureSpecs(child, measureSpecs);
             this.helper.addMeasureSpec(measureSpecs);
         });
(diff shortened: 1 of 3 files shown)


tns-core-modules/ui/layouts/grid-layout/grid-layout.ios.ts
```diff
@@ -136,6 +136,11 @@ export class GridLayout extends GridLayoutBase {
 
         this.eachLayoutChild((child, last) => {
             let measureSpecs = this.map.get(child);
+            if (!measureSpecs) {
```
manoldonev: I was reviewing our code and I am not sure trying to fix this exceptional situation by registering layout child here is a good idea. How about just skipping this specific layout child instead (as obviously something is not right with its current state)?
``` ts
if (!measureSpecs) { 
    continue;
}
```

m-abs: I've finally managed to reproduce the problem see: https://github.com/NativeScript/NativeScript/issues/7226#issuecomment-491634072

I'm not sure what the best solution is or if we just use components+layouts incorrectly.

Just skipping as you suggest results in the component not being rendered.

manoldonev: hey @m-abs thank you isolating the problem!
So just skipping the iteration when measureSpec is undefined is not an option, ok. Will investigate the code on the nativescript-angular side a bit but if I don't find a better solution I am going to merge yours as is.

Meanwhile, there is also a workaround that you can use -- wrap the contents of your sub1 component (sub1.component.html) in any kind of layout like this:
``` html
<GridLayout>
    <ns-sub2></ns-sub2>
</GridLayout>
```

manoldonev: @m-abs I think we found the root of the problem in the ProxyViewContainer class -- see my commit here: https://github.com/NativeScript/NativeScript/pull/7230/commits/3acfc70ed49d3eca19666a603cfc62d50489ba66

manoldonev:
test

tns-core-modules/ui/layouts/grid-layout/grid-layout.ios.ts
m-abs: Looks good to me.

PR #7232, opened 2019-05-10
refactor: update access modifier

manoldonev:
We are no longer using StyleScope.changeCssFile() method internally but apparently there was a discrepancy where method implementation was private but d.ts listed it as public -- synced both to public now.


manoldonev:
upload_test_reports

manoldonev:
test upload_test_reports

PR #7296, opened 2019-06-02
feat: cancel contradictory gesture events

### LINKED ISSUE #7281: Gestures: tap and double tap on the same View
**Environment**
 - CLI: latest
 - Cross-platform modules:  5.4.0
 - Android Runtime:  5.4.0
 - iOS Runtime: 5.4.0

**Describe the bug**
When you have a tap and double tap gesture recognizer on a view and you attempt to double tap, the tap handler fires. The tap gesture event should not fire if you're attempting a double tap. Take a look at the "Additional context" for the commit that allowed this.

**To Reproduce**

```
export class TestLayout extends StackLayout {
    onLoaded() {
        super.onLoaded();

        this.on(GestureTypes.tap, () => console.log('Tapped'));
        this.on(GestureTypes.doubleTap, () => console.log('Double tapped'));
    }
}
```

If you attempt to double tap on the `TestLayout`, the console will show:
> `Tapped`
> `Double tapped`

**Expected behavior**
If you try to double tap on a view with a tap gesture recognizer and a double tap gesture recognizer, the double tap event should fire and the tap should not. 

**Additional context**

I have been working through this issue with @MartoYankov .

[Here is the commit that allowed simultaneous gesture events](https://github.com/NativeScript/NativeScript/commit/fe72a43e49cc318d7a4cc2488d66aaf9e07feab1)

[Here are the docs to the delegate method that could be implemented for iOS](https://developer.apple.com/documentation/uikit/uigesturerecognizerdelegate/1624229-gesturerecognizer?language=objc)
- I have already implemented this ^ for iOS (see below). The issue now is getting the same functionality for Android.

**Potential solution (iOS)**

`gestures.ios.ts`
```
class UIGestureRecognizerDelegateImpl extends NSObject implements UIGestureRecognizerDelegate {
    public static ObjCProtocols = [UIGestureRecognizerDelegate];

    public gestureRecognizerShouldRecognizeSimultaneouslyWithGestureRecognizer(gestureRecognizer: UIGestureRecognizer, otherGestureRecognizer: UIGestureRecognizer): boolean {
        // If both gesture recognizers are of type UITapGestureRecognizer, do not allow
        // simultaneous recognition.
        if (gestureRecognizer instanceof UITapGestureRecognizer && otherGestureRecognizer instanceof UITapGestureRecognizer) {
            return false;
        }
        return true;
    }

    public gestureRecognizerShouldRequireFailureOfGestureRecognizer(gestureRecognizer: UIGestureRecognizer, otherGestureRecognizer: UIGestureRecognizer): boolean {
        // If both gesture recognizers are of type UITapGestureRecognizer & one of them is a doubleTap,
        // we must require a failure.
        if (gestureRecognizer instanceof UITapGestureRecognizer
            && otherGestureRecognizer instanceof UITapGestureRecognizer
            && otherGestureRecognizer.numberOfTapsRequired === 2) {
            return true;
        }
        return false;
    }
}
```

MCurran16:
## PR Checklist

When you have a tap and double tap gesture recognizer on a view and you attempt to double tap, the tap handler fires. The tap gesture event should not fire if you're attempting a double tap.

## What is the new behavior?

For both platforms, the `tap` event will not be fired when you're attempting a `doubleTap` gesture.

Implements #7281 

## Additional Context

@MartoYankov 

For android, I decided to not use `onSingleTapConfirmed` due to the delay that is described in this SO answer: https://stackoverflow.com/a/48844866 . I did try using `onSingleTapConfirmed` and there was consistently a noticeable delay and sometimes it wasn't even fired.

For iOS, the solution in the attached issue was implemented.

I also tried to get those files I changed to follow the NS code convention a bit more... hoping that's okay!

BREAKING CHANGES:

Old behavior:
 - iOS/Android:
    - double tap: child tap -> parent tap -> child double tap -> parent double tap
    - tap: child tap -> parent tap

New behavior:
 - iOS
    - double tap: child double tap
    - tap: child tap
 - Android
    - double tap: child double tap -> parent double tap
    - tap: child tap -> parent tap

Migration steps:
Move event handlers accordingly.


tns-core-modules/ui/gestures/gestures.android.ts
```diff
@@ -36,29 +43,40 @@ function initializeTapAndDoubleTapGestureListener() {
...
-                let args = _getArgs(GestureTypes.tap, this._target, motionEvent);
-                _executeCallback(this._observer, args);
+        public onDown(motionEvent: android.view.MotionEvent): boolean {
+            const tapTime = Date.now();
+            if ((tapTime - this._lastTapTime) <= TapAndDoubleTapGestureListenerImpl.DoubleTapTimeout) {
+                this._handleDoubleTap(motionEvent);
+            } else {
+                this._handleSingleTap(motionEvent);
             }
+            this._lastTapTime = tapTime;
```
MartoYankov: According to docs - https://developer.android.com/reference/android/view/ViewConfiguration.html#getDoubleTapTimeout() - this timeout should be between the first event's up and the down of the second event. In this case you are detecting the `_lastTapTime` on down, which can result in strange behavior.

MartoYankov:
Also, I think this should be considered as behavior breaking change.

tns-core-modules/ui/gestures/gestures.android.ts
MCurran16: Fixed. PR has been updated. Thanks!

tns-core-modules/ui/gestures/gestures.ios.ts
```diff
@@ -8,16 +8,33 @@ import { GesturesObserverBase, toString, TouchAction, GestureStateTypes, Gesture
...
 }
 
 class UIGestureRecognizerDelegateImpl extends NSObject implements UIGestureRecognizerDelegate {
     public static ObjCProtocols = [UIGestureRecognizerDelegate];
+
     public gestureRecognizerShouldRecognizeSimultaneouslyWithGestureRecognizer(gestureRecognizer: UIGestureRecognizer, otherGestureRecognizer: UIGestureRecognizer): boolean {
+        // If both gesture recognizers are of type UITapGestureRecognizer, do not allow
+        // simultaneous recognition.
+        if (gestureRecognizer instanceof UITapGestureRecognizer && otherGestureRecognizer instanceof UITapGestureRecognizer) {
+            return false;
```
edusperoni: I imagine this is going to change the behavior where a `tap` was recognized by both parent and child. This change makes it so only one `tap` gesture is recognized on iOS, but seems to be unchanged on Android, which may lead to inconsistent behavior that was previously patched out (for more context see #7310)

Old behavior:

iOS and Android:
double tap: child tap -> parent tap -> child double tap -> parent double tap
tap: child tap -> parent tap

New behavior (iOS):
double tap: child double tap
tap: child tap

New behavior (Android):
double tap: child double tap -> parent double tap
tap: child tap -> parent tap

MCurran16: @edusperoni It is already inconsistent between the platforms and this PR was not intending to fix child/parent intercepting/cancelling/bubbling.

My understanding of the iOS implementation was that the `UIGestureRecognizerDelegateImpl` was for a single view as the NS view is passed in which can be seen [here](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/gestures/gestures.ios.ts#L10). It sounds like that is an incorrect understanding. See here:

edusperoni: Actually, the current implementation of events (and bubbling) seem to be consistent in both platforms (all events bubble and the same events trigger).

I'm aware this PR does not intend to fix child/parent bubbling, I'm pointing out that this is a behavior breaking change. The same delegate is used by all views, as seen [here](https://github.com/NativeScript/NativeScript/blob/c5db112b8d13285057b8a154b905f9a25e1368aa/tns-core-modules/ui/gestures/gestures.ios.ts#L22), so that line will affect child/parent bubbling and break consistency.

MCurran16: Okay that is fair. I'm understanding now. 

In your mind, where do we go from here? What would you like me to do?

Maybe @MartoYankov can weigh in on next steps?

edusperoni: I don't believe this is a negative change, but I can think of 3 options:

1- if we're detecting a tap gesture, bubble this information up to views on android and prevent tap recognition up the chain
https://github.com/NativeScript/NativeScript/blob/cf533a7b6d6607aefd39c06978578bb2b453eb3d/tns-core-modules/ui/core/view/view.android.ts#L397-L399
2- Compare targets and required taps in `if (gestureRecognizer instanceof UITapGestureRecognizer && otherGestureRecognizer instanceof UITapGestureRecognizer) {`, so events will keep bubbling
3- Document it as a breaking change

I believe 1 is out of scope of this PR and 2 may add unneeded complexity. Maybe the best option is really 3 until we reach a consensus on how gesture disambiguation should work in NS?

The impact of the change should be really low. The only plugin I know to handle multiple taps (`nativescript-windowed-modal`) only does it to prevent a parent tap when the child was tapped (using Hit Testing). My own plugin (`nativescript-ng-ripple`, soon to become one with `nativescript-ripple`) uses `touch`, so it probably won't be affected.

MartoYankov: We discussed this internally and we agree that 3 is probably the best option for now.

MartoYankov:
test

tns-core-modules/ui/gestures/gestures.ios.ts
manoldonev: @MCurran16 @edusperoni @MartoYankov Seems we have a problem with this specific change in UIGestureRecognizerDelegateImpl.gestureRecognizerShouldRecognizeSimultaneouslyWithGestureRecognizer(...) that breaks ActionBar usage scenario in all master/detail templates so I am reverting it for the time being -- see https://github.com/NativeScript/NativeScript/pull/7400.

PR #7400, opened 2019-06-24
fix(ios-actionbar): parent/child gesture recognition

manoldonev:
Changes related to parent/child gesture recognition in iOS introduced with https://github.com/NativeScript/NativeScript/pull/7296 break an ActionBar usage scenario in all master/detail templates [here](https://github.com/NativeScript/nativescript-app-templates/blob/a7df226418c85c54ae1788a9c01e11c06f297ae0/packages/template-master-detail-ts/app/cars/car-detail-edit-page/car-detail-edit-page.xml#L14):
``` xml
    <ActionBar class="action-bar">
       ...

        <ActionItem ios.position="right" android.position="right">
            <Label text="Done" class="action-item" verticalAlignment="center" tap="onDoneButtonTap" 
                isEnabled="{{ car.isModelValid }}"
                isUserInteractionEnabled="{{ car.isModelValid }}" />
        </ActionItem>
    </ActionBar>
``` 

Where tap event is not raised for the ActionItem Label anymore.


MCurran16:
:(

Issue #5835, opened 2018-05-18
textfield textchange event not working


giorgi-abashidze:
textfield textchange event not working....

tsonevn:
HI @giorgi-abashidze,
Please follow the template and provide the requested info.
->
```
<!--
We appreciate your feedback! While we are doing all we can to take care of every issue, sometimes we get overwhelmed. That's why
 - issues that are not constructive or describe problems that cannot be reproduced will be closed
 - feature requests or bug reports with unanswered questions regarding the behavior/reproduction for more than 20 days will be closed 
-->

### Did you verify this is a real problem by searching the [NativeScript Forum](http://forum.nativescript.org) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?

### Tell us about the problem
Please, ensure your title is less than 63 characters long and starts with a capital
letter.

### Which platform(s) does your issue occur on?
iOS/Android/Both

### Please provide the following version numbers that your issue occurs with:
- CLI: (run `tns --version` to fetch it)
- Cross-platform modules: (check the 'version' attribute in the
`node_modules/tns-core-modules/package.json` file in your project)
- Runtime(s): (look for the `"tns-android"` and `"tns-ios"` properties in the
`package.json` file of your project)
- Plugin(s): (look for the version number in the `package.json` file of your
project)

### Please tell us how to recreate the issue in as much detail as possible.
The best approach would be to get your code running in the [NativeScript Playground](https://play.nativescript.org) and share the link with us, along with any additional details or steps to reproduce needed for examining the issue there. 

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.

```

NickIliev:
@giorgi-abashidze the `textChange` event is available in an Angular base event. In NativeScript Core use two-way data binding to detect changes in the text field. More about text filed events in NativeScript Core bcan be found in these examples

Issue #3971, opened 2017-04-10
[TNS 3.0 RC] propertyChangeEvent not firing on views nor observables


NordlingArt:
**Platform:** iOS
**TNS:** 3.0 RC

TextView won't fire a propertyChange event:

**UPDATE:** At first I thought this had to do with TextView specifically, but it doesn't even work for an observable.

```xml
<Page xmlns="http://schemas.nativescript.org/tns.xsd" navigatingTo="onNavigatingTo">
  <TextField id="inputField" />
</Page>
```

```js
const Observable = require("data/observable").fromObject;

exports.onNavigatingTo = function(args) {
  const page = args.object;
  const inputField = page.getViewById("inputField");
  const inputBindings = Observable({ text: "" });
  
  inputField.bind({ targetProperty: "text", sourceProperty: "text" }, inputBindings);
  inputBindings.on("propertyChange", e => {
    console.log(e.propertyName); // Never fired
  });
};
```

Is there something that's changed with event listeners in the 3.0 RC release?

tsonevn:
Hi @NordlingArt,
While using @rc modules, you should specify the property name.
For example, if you would like to handle the text change event, you should listen for textChange event.
Example:
```
inputField.on("textChange", (e:PropertyChangeData) => {
    console.log(e.propertyName); 
  });
```

NordlingArt:
@tsonevn - Got it. That worked. So how would I observe a property change of any property? Say if I change `inputField.height` to something else. `propertyChange` is no longer in use from what I can understand.

toddanglin:
Here's another problem I've hit upgrading my app from 2.5 to 3.0. :)

Is this an intentional breaking change? Or should the `propertyChange` syntax that worked in 2.5.x still work in 3.0?

This workaround does work for me, BUT only when configuring the binding via JavaScript. It does NOT seem to work when trying to configure the binding via XML. Feels like another regression we should prioritize.

tsonevn:
Hi @toddanglin.
Could you provide sample project or some code snippet, where the problem could be reproduced?

toddanglin:
Will work on a small project tonight. I should add that my usage is occurring within a modal page on iOS, and I know modals [currently have some other issues](https://github.com/NativeScript/NativeScript/issues/4151) within iOS. Might be related to the XML binding not working here.

Nikoms:
I have the same problem on Android.
Same problem in the xml like this. "onTextChanged" is never called:
```
[...]

PR #7403, opened 2019-06-25
fix: restore TextField.textChange and Switch.checkedChange event syntax in xml

NathanaelA:
## PR Checklist

-  [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.

`<TextField textChange="someFunc"/> `
`SomeFunc` does not fire.   When NativeScript was upgraded to v3.0.0 somehow when the renaming the events from `properyChange` to `textChange` on the text editable fields; this event was not properly declared.  This became a nasty breaking change for NS-Core apps, as there is then now no way to wire up the "change" events via xml anymore at all.    This really becomes problematic with repeaters and listviews (i.e. anything that has templates that can repeat)

## What is the **FIXED** behavior?
`<TextField textChange="someFunc"/> `
`someFunc` will now be wired up properly from the XML builder so that this event will properly fire.  This fixes the **regression** in NS 3.0 that has causes a lot of issues for those who are using NS Core for the last couple versions of NS.

#5835, #3971 https://github.com/NativeScript/nativescript-angular/issues/79

I would really like to see this show up in v6.00 -- It was broke in 3.00 and we (ns-core) have limped along with some significantly **CRAPPY** work arounds -- its time to get it fixed...  ;-)  

Btw, I can also add `checkedChange` (i.e. `Switch` regressed) event to this PR or another if you prefer.  It is another item that's even was regressed in v3.0 of NS.


NathanaelA:
Can we make sure we get some visibility on this so it doesn't get dropped.   
@manoldonev @vchimev @vakrilov @EmilStoychev 

I just ended up upgrading a second NS-Core 2.x app to NS 5 tonight and ran into this again.   However, this time the client needs to be able to build the app; which complicates this drastically as on the first app I can easily modify my own tns-core-modules -- and build it from that.    But for this second one I'd rather not be distributing a custom tns-core-modules just to fix a bug that you so simply can commit and have fixed for all future versions...   Please can we fix this!

manoldonev:
test

manoldonev:
test

manoldonev:
@NathanaelA we'd really like to find a more general solution (probably in the builder module) but for the time being we decided to accept this change as I am not sure we'll be able to investigate this further within the NativeScript 6.0 release timeframe.

PR #7426, opened 2019-06-29
fix(observable-array): splice(...) to return correct amount of added items

### LINKED ISSUE #7425: ObservableArray does not notify correctly when splice replaces elements
**Environment**
 - CLI: next
 - Cross-platform modules: next
 - Android Runtime: next
 - iOS Runtime: next

**Describe the bug**
When you call ObservableArray.splice(), the event it emits fails to account for replaced items. This means it accounts only for removals/inserts, but any replace returns an added count of 0

**To Reproduce**
https://play.nativescript.org/?template=play-ng&id=jVKVft

**Expected behavior**
`addedCount` should take into account the replaced items

**Additional context**
Maybe we should change `addedCount` to `this._array.length + removed.length - oldLength`

I ran into this issue trying to splice `FormattedString.spans`. The only way I could make it work was by either:

1. removing and adding again
2. applying the `addedCount` fix I proposed

Since ObservableArray is used almost everywhere, this change could be breaking.

edusperoni:
## PR Checklist

ObservableArray `splice` does not notify the number of updated items and it can't be inferred from the passed arguments.

## What is the new behavior?
ObservableArray `splice` adds all updated items to `addedCount`, using a logic of "x items removed, y items added in their place"

The amount of change in the length of the array can be always inferred by `addedCount - removed.length`.

This fixes a bug that would cause `FormattedString.spans.splice` to keep removing spans and never adding any back.

Fixes #7425.

BREAKING CHANGES:

The `addedCount` variable from ObservableArray's `splice` change event is always the amount of added items.

Migration steps:
Ideally, developers should handle the case where items are replaced by `splice`. The old `addedCount` can be obtained by

`const addedCount = event.addedCount - event.removed.length`

### DIFF
--- a/tests/app/data/observable-array-tests.ts
+++ b/tests/app/data/observable-array-tests.ts
@@ -410,7 +410,7 @@ export const test_ObservableArray_spliceShouldRemoveAndInertSpecifiedNumberOfEle
         // args.action is "splice".
         // args.index is the start index.
         // args.removed.length is equal to the number of deleted items.
-        // args.addedCount is equal to the delta between number of inserted items and number of deleted items but not less than 0.
+        // args.addedCount is equal to the amount of added and replaced items.
 
         // >> (hide)
         result = args;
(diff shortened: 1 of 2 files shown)


edusperoni:
Oops, my mistake, Some weird commits went with it. Let me fix it

manoldonev:
test

PR #7428, opened 2019-06-29
fix: full Unicode support in xml

edusperoni:
## PR Checklist

Full unicode can have up to 21 bits, but the current decoding method ( `String.fromCharCode`) can only handle 16 bits. HTML entities like `&#x1f923;` (17 bits) will fail to render properly as `&#xf923;` (highest-order bits are stripped).

## What is the new behavior?
 `&#x1f923;` is correcly rendered as the byte sequence `f0 9f a4 a3` instead of `ef a4 a3` by using `String.fromCodePoint`, which generates a string/char from any number of bytes


manoldonev:
test

Issue #5661, opened 2018-04-11
[Feature] Add openFile method for android


PeterStaev:
### Did you verify this is a real problem by searching the [NativeScript Forum](http://forum.nativescript.org) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?
Yes

### Tell us about the problem
Currently under iOS we can easily open a file by using `utils.ios.openFile()` method. Under android there is no similar method. There should be a similar method for android enabling the user to open local files. Code similar to this can be used in order to achieve it:
https://github.com/PeterStaev/nativescript-telerik-reporting/blob/b57db9e79f4c4c7ef71fdb070cebd3063e106995/demo/app/main-page.ts#L85-L91
The only catch is that the file must be present on the external storage in order other apps to have access to it. 

### Which platform(s) does your issue occur on?
Android

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.4
- Cross-platform modules: 3.4
- Runtime(s): 3.4

NickIliev:
@PeterStaev perhaps we can create an explicit check if there is an external storage access and implement the method with this as requrment (to be used only for files from external strorage). 

For example somthing like:
```TypeScript
export function openFile(filePath: string): boolean {
    const context = ad.getApplicationContext();
    try {
        if (isExternalStorageAvailable() && !isExternalStorageReadOnly()) {
            const fsa = new FileSystemAccess();
            const mimeTypeMap = android.webkit.MimeTypeMap.getSingleton();
            const mimeType = mimeTypeMap.getMimeTypeFromExtension(fsa.getFileExtension(filePath).replace(".", "").toLowerCase());
            const intent = new android.content.Intent(android.content.Intent.ACTION_VIEW);
            intent.setDataAndType(android.net.Uri.fromFile(new java.io.File(filePath)), mimeType);

            context.startActivity(android.content.Intent.createChooser(intent, "Open File..."));

            return true;
        } else {
            // TODO: no external storage access logic
        }
    } catch(e) {
        traceWrite("Error in openFile", traceCategories.Error, traceMessageType.error);
    }
    return false;
}

function isExternalStorageReadOnly(): boolean {
    const extStorageState = android.os.Environment.getExternalStorageState();
    if (android.os.Environment.MEDIA_MOUNTED_READ_ONLY === extStorageState) {
        return true;
    }
    return false;
}


function isExternalStorageAvailable(): boolean {
[...]

PR #7431, opened 2019-07-01
fix: added missing openFile method in ios utils

vakrilov:
Export `openFile` method for in IOS utils.
Related to #5661


tns-core-modules/utils/utils.ios.ts
```diff
@@ -146,6 +146,8 @@ export module ios {
 
 }
 
+export const openFile = ios.openFile;
```
manoldonev: Maybe we should deprecate ios.openFile then?

vakrilov: Moved implementation to the root and deprecated the `utils.ios.openFile()` method

SvetoslavTsenov:
test

vakrilov:
@cla-bot check

PR #7441, opened 2019-07-02
fix(tabs): current TabContentItem index not set

ADjenkov:
[Android] Fix crash when switching between tab with Frame to tab without Frame and vice versa.
[Android] Fix crash when app suspend/resume while on a tab with Frame


dtopuzov:
test package_version#rc

manoldonev:
test package_version#rc

PR #7453, opened 2019-07-03
fix(bottom-navigation): current TabContentItem index not set

ADjenkov:
[Android] Fix crash when switching between tabs with or withouth Frame
[Android] Fix crash when app suspend/resume then selecting/reselecting tab


ADjenkov:
test package_version#rc

manoldonev:
test package_version#rc

PR #7462, opened 2019-07-04
refactor: HMR and webpack improvements

vakrilov:
In this PR:
- Changes in `global.loadModule`. Modules loaded from builder are loaded with `global.loadModule(moduleName, true)`. This will make the `global._isModuleLoadedForUI()` function return `true` for this module. The [webpack hot-loader](https://github.com/NativeScript/nativescript-dev-webpack/blob/file-qualifiers/hmr/hot-loader.ts#L26) inserts a check for each module and enables self-accept (`module.hot.accept()`) if this module was loaded form the builder. This means that all modules can be processed build-time by the `hmr-loader` but only those that were loaded (runtime) by the builder will be self-accepted and will trigger HMR logic.
- HMR for custom-components: Whenever custom component code or markup is updated it will cause refresh of the page containing the custom-component.


tns-core-modules/ui/core/view/view-common.ts
```diff
@@ -192,6 +192,20 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
...
+        // Handle script/markup changes in custom components by falling back to page refresh
+        if (viewMatchesModuleContext(this, context, ["markup", "script"]) &&
+            this.page &&
+            this.page.frame) {
+
+            if (traceEnabled()) {
+                traceWrite(`Change Handled: Changing ${context.type} for ${this} inside ${this.page}`, traceCategories.Livesync);
+            }
+
+            if (this.page && this.page.frame) {
```
vchimev: This is not needed - it has been already verified above.

tns-core-modules/utils/utils-common.ts
```diff
@@ -35,7 +36,7 @@ export function convertString(value: any): any {
 export function getModuleName(path: string): string {
```
vchimev: It looks like `sanitizeModuleName()` eliminates the need for `getModuleName()` -  not used anymore/could be deleted.

vakrilov: Yup - it was released "officially" in 5.4 though - do you think it's OK to remove it?

tns-core-modules/globals/register-module-helpers.ts
```diff
@@ -0,0 +1,128 @@
...
+
+global.moduleExists = function moduleExists(name: string): boolean {
+    return modules.has(name);
+};
+
+global.loadModule = function loadModule(name: string, isUIModule: boolean = false): any {
+    const moduleInfo = modules.get(name);
+    if (moduleInfo) {
+        if (isUIModule) {
+            // console.log(`--> load module for UI: ${moduleInfo.moduleId}`);
```
DimitarTachev: forgotten `console.log`

Issue #7435, opened 2019-07-01
[Tabs] tabstripitem tap does not change tabs


vchimev:
**Environment**
 - tns-core-modules modules: 6.0.0-rc-2019-06-28-175837-02

**Describe the bug**

Tap on a `TabStripItem` does not change to its `TabContentItem` on iOS.

**To Reproduce** - a [sample tabs page](https://github.com/NativeScript/NativeScript/commit/808fbd414d6a01ebad34fcdbbc816af82223b858).

Tap on *TabStripItem 3*.

**Expected**

![expected](https://user-images.githubusercontent.com/12251337/60441863-8d98d200-9c20-11e9-9d22-b9ea01f15864.png)

**Actual**

![actual](https://user-images.githubusercontent.com/12251337/60441875-94bfe000-9c20-11e9-9711-033296cc9683.png)

PR #7468, opened 2019-07-05
fix(ios-tabs): tabstripitem tap does not change tabs

ADjenkov:
Note 1: `tabBar.delegate` was cleared after some time. Storing it in a local var and clearing on dispose.
Note 2: Enable tabStrip items selection only if no current animation in progress.

Fix https://github.com/NativeScript/NativeScript/issues/7435
Fix -> Changing fast tabStrip items causes glitches and content items problem

### DIFF
--- a/tns-core-modules/ui/tabs/tabs.ios.ts
+++ b/tns-core-modules/ui/tabs/tabs.ios.ts
@@ -34,7 +34,15 @@ class MDCTabBarDelegateImpl extends NSObject implements MDCTabBarDelegate {
     }
 
     public tabBarShouldSelectItem(tabBar: MDCTabBar, item: UITabBarItem): boolean {
-        return true;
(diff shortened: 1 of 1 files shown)


tns-core-modules/ui/tabs/tabs.ios.ts
```diff
@@ -34,7 +34,15 @@ class MDCTabBarDelegateImpl extends NSObject implements MDCTabBarDelegate {
     }
 
     public tabBarShouldSelectItem(tabBar: MDCTabBar, item: UITabBarItem): boolean {
-        return true;
+        const owner = this._owner.get();
+        const shouldSelectItem = owner._canSelectItem;
```
manoldonev: Check whether `owner` is not null prior to using it (weak reference)

PR #7514, opened 2019-07-13
feat(file-system): run asynchronous methods on seperate thread

### LINKED ISSUE #2239: Make the file-system module run in a background thread [$300]
Calling file-sytem -> writeText is freezing the UI of my apps until the write process is complete. Even the ActivityIndicator freezes during that time.
This is a big issue when saving big chunks of data which can take between 10 to 60 seconds to write.

I have tested it on Android with NativeScript 2.0.1

To solve that it would be great if the write functionality could run in a background thread, just like it is done for the http module.

Here is the scenario where I need to save big chunks of data.
The app I am building uses Lokijs for Offline Storage mechanism, where I store 50k records from different tables. When I call save database, the LokiNativeScriptAdapter uses file-system module to save the data to a file, which takes up to 60 seconds.

<bountysource-plugin>

---
There is a **[$300 open bounty](https://www.bountysource.com/issues/34832190-make-the-file-system-module-run-in-a-background-thread?utm_campaign=plugin&utm_content=tracker%2F12908224&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F12908224&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mudlabs:
The asynchronous methods of `file-system` _(i.e. writeText)_,  are executed on the main _UI_ thread. This can lead to poor performance _(i.e. UI thread freezing)_ when handling large chunks of data. (See #2239)

## What is the new behavior?
All asynchronous methods are dispatched to a Worker, freeing up the main thread. The Worker is only created once a call to one of these methods is made. 

By default the Worker will close after a period of inactivity, I've arbitrarily set this period to 2 minutes. But developers are able to turn this inactivity observer off and `terminate` at their own time by using the new `worker` sub-module of `file-system`. 

For now the `worker` module only provides a `terminate` method to kill the Worker, and a `keepAlive` method to prevent the default close on inactivity behaviour. But we could expand on this later (allowing devs to set their own period etc...).
* __Note:__ Even if `keepAlive` is called the Worker will still be closed if the application raises a `lowMemmoryEvent`, or the application itself is closed.

---

Closes #2239


PeterStaev:
I don't think it is appropriate to run such a thing in a worker, considering there have been problems with workers and webpack. Much better to add native async methods and then call those from the core modules.

edusperoni:
Adding to @PeterStaev's comment, you can't spawn workers from workers, so this implementation prevents the filesystem module from being used inside a worker, and you can only access one file at a time.

manoldonev:
@mudlabs Thank you for this PR! For this specific feature we preferred a different solution that relies on native async methods https://github.com/NativeScript/NativeScript/pull/7671 as discussed above by @PeterStaev and @edusperoni. Nevertheless, we appreciate your time and effort and we hope you will continue to contribute to NativeScript in the future.

You can sign up in the form here https://goo.gl/forms/jkOpclSNMHeaRg4B2 so we can send you some stickers as a small token of our gratitude.

Issue #4864, opened 2017-09-20
CSS Variables support


bundyo:
Nearly all major browsers support [CSS variables](http://caniuse.com/#search=css%20variables) natively now (except Edge, which still has some bugs in 15). Support of CSS variables can greatly simplify the styling of an app and provide better support for theming when SASS is not used.

Maybe it is time to implement them in NativeScript?

NateVonSmith:
pleeeeeease  🙏 🙏 🙏

rafaband:
If NativeScript add SASS support and solve the bugs for development for sure i migrate all my ionic app to NS

bundyo:
NativeScript works with SASS from long time ago. Most CLI start templates are with SASS set up and used.

crowmagnumb:
What about the original variable issue mentioned here? I am currently converting a big chunk of my angular web components to shared web/nativescript components and the lack of variable support is plaguing me. Everywhere I use a CSS variable I have to make a special case for nativescript and use a scss variable. This is not only sad and annoying but now means that my nativescript versions of these components are hard-colored to my default theme. Would **love** to have the css variable support! Cheers.

bundyo:
While we wait for official support, you can use PostCSS for CSS variable support. Tried it and it works.

m-abs:
We could also use this in our projects, so I've started implementing it.

I intend to add support for both `calc` and `css-variables`.

bundyo:
CSS variables support was already merged, so it will be released with 6.1.

mfcarneiro:
Thanks @bundyo. I will wait for `6.1` version to start my app :sweat_smile:

AnthonyLenglet:
Support has landed, however I'm still wondering one thing about this

when working on websites, people usualy use `::host` in order to store all of their variables

what would be the equivalent here ?

m-abs:
@AnthonyLenglet 

You should also define them in `ns-modal`, if you're using modal-views in your application.

I'm not sure about other renders for `nativescript`, but `nativescript-angular` does support `:host` in the component stylesheet.

Like this `my.component.scss`:
```scss
:host {
  --scarlet: #FF2400;
}

.bg_red {
  background-color: var(--scarlet);
}
```

AnthonyLenglet:
> I'm not sure about other renders for `nativescript`, but `nativescript-angular` does support `:host` in the component stylesheet.
> 
> Like this `my.component.scss`:
> 
> ```css-scss
> :host {
[...]

PR #7553, opened 2019-07-19
feat: implement css-variables and css-calc

m-abs:
## PR Checklist

No support for [css-variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) or for [css-calc](https://developer.mozilla.org/en-US/docs/Web/CSS/calc)

## What is the new behavior?

Adds ~basic~ support for [css-variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) and [css-calc](https://developer.mozilla.org/en-US/docs/Web/CSS/calc)

```css
.button {
   --custom-color: black; /** text-color: black **/
   color: var(--custom-color);
}

.button.got-the-blues {
   --custom-color: blue; /** text-color: blue **/
}

.button.red {
   --custom-color: red;  /** text-color: red **/
}
```

Adds support for `calc(...)` in CSS, via [reduce-css-calc](https://github.com/MoOx/reduce-css-calc).

```css
StackLayout {
    --my-factor: 1;  /** width: 100% **/
   width: calc(100% * var(--my-factor));
}

StackLayout.slim {
    --my-factor: 0.1;  /** width: 10% **/
}

StackLayout.wide {
    --my-factor: 1.25; /** width: 125% **/
}
```

### What is still missing:
- ~Updating the variables via `view.style` is broken, if already defined in CSS.~
- ~For CSS-variables to be really useful, support for `calc(var(--my-variable) * 2)` is needed. I intend to add this to the PR.~
- ~clean up~

Implements #4864


exejutable:
this is gonna be merged?

m-abs:
@exejutable
I hope so, but I need to do some clean up and it needs to be properly reviewed and tested.

themounthead:
Can this merged please !!! I'm migrating my web app to a shared codebase and this is last missing piece to handle universal Theming.

bundyo:
I've tested this PR a bit and it seems to be working rather nice.

manoldonev:
test

tests/app/ui/styling/style-tests.ts
```diff
@@ -1426,6 +1426,143 @@ export function test_CascadingClassNamesAppliesAfterPageLoad() {
     });
 }
 
+export function test_css_calc() {
```
manoldonev: Would it be useful to add a test for nested `calc`?

m-abs: Yes, I've added two more tests for nested `css-calc` and `css-calc`+`css-variables`.

m-abs:
@manoldonev 
Is it okay to depend on `reduce-css-calc` or should I try to write my own parser?

`reduce-css-calc` doesn't take up much space and work as advertised, but it is a new extra dependency plus its dependencies.

m-abs:
@manoldonev 
I tried out my own implementation on https://github.com/m-abs/NativeScript/tree/feat/css-variables-own-calc.

Seems to work fine, but it needs a little clean up.

manoldonev:
test

manoldonev:
@m-abs I think we'll be fine with using reduce-css-calc (license is ok, tests are passing, has only a couple dependencies).

tns-core-modules/ui/core/properties/properties.ts
```diff
@@ -56,6 +58,94 @@ export function _getStyleProperties(): CssProperty<any, any>[] {
...
+export function isCssCalcValue(value: string) {
+    return cssCalcRegexp.test(value);
+}
+
+export function isCssValueUsingCssVariable(value: string) {
+    return cssVarValueRegexp.test(value);
+}
+
+export function _cssVariableConverter<T>(view: ViewBase, cssName: string, value: string | T): string | T {
+    let res = [] as string[];
```
manoldonev: We generally use the Array<String> syntax in the codebase (also as a general rule we try to avoid abbreviations like `res`) so this would look like:
``` ts
let result = new Array<String>();
```

manoldonev: Also, better initialize it after the first two if-checks (as they are not using it).

tns-core-modules/ui/core/properties/properties.ts
```diff
@@ -56,6 +58,94 @@ export function _getStyleProperties(): CssProperty<any, any>[] {
...
+
+export function isCssCalcValue(value: string) {
+    return cssCalcRegexp.test(value);
+}
+
+export function isCssValueUsingCssVariable(value: string) {
+    return cssVarValueRegexp.test(value);
+}
+
+export function _cssVariableConverter<T>(view: ViewBase, cssName: string, value: string | T): string | T {
```
manoldonev: Naming this converter is a bit confusing to me in the context of what NativeScript refers to as converters. How about renaming this to `_evaluateCssVariable` or even `_tryEvaluateCssVariable` / `tryEvalCssVariable`?

tns-core-modules/ui/core/properties/properties.ts
```diff
@@ -56,6 +58,94 @@ export function _getStyleProperties(): CssProperty<any, any>[] {
...
+        }
+
+        if (!isCssValueUsingCssVariable(part)) {
+            // part is not using css-variable, use as is.
+            res.push(part);
+            continue;
+        }
+
+        // Each sub-section can use more than one css-variable, each must be processed
+        for (let m = cssVarValueRegexp.exec(part), lastValue: string; lastValue !== part && m; m = cssVarValueRegexp.exec(part)) {
```
manoldonev: In my opinion using `for` as a control structure in this scenario is hard to follow -- what do you think about replacing it with `while`?

tns-core-modules/ui/core/properties/properties.ts
m-abs: Naming is hard :) I've renamed the function to `_evaluateCssVariable`

tns-core-modules/ui/core/properties/properties.ts
m-abs: I think, I found a better way to process the variables in `_evaluateCssVariable`.

The nested loops wasn't needed. I just keep looping until the output-string doesn't change anymore.

I'm not sure what is the best way to handle missing values?

At the moment missing values will be added as a `null` and if the output is the string `"null"` `_evaluateCssVariable` return a null, to avoid two printing two errors like this:
```bash
[1] JS: Style: Failed to get value for css-variable "--my-background-color-1565280860851" used in "css:background-color"=[var(--my-background-color-1565280860851)] to Label(181)
[1] JS: Error: Failed to apply property [background-color] with value [var(--my-background-color-1565280860851)] to Label(181). Error: Invalid color: null
```

m-abs: We could have a case like this:

```css
.broken {
   --my-font-family: 'san-sarif';
   font-family: var(--my-none-existing-font), var(--my-font-family);
}
```

This will become:
```css
.broken {
   --my-font-family: 'san-sarif';
   font-family: null, 'san-sarif';
}
```

tns-core-modules/ui/core/properties/properties.ts
```diff
@@ -56,6 +58,83 @@ export function _getStyleProperties(): CssProperty<any, any>[] {
...
+export function isCssCalcValue(value: string) {
+    return cssCalcRegexp.test(value);
+}
+
+export function isCssValueUsingCssVariable(value: string) {
+    return cssVarValueRegexp.test(value);
+}
+
+export function _evaluateCssVariable<T>(view: ViewBase, cssName: string, value: string | T): string | T {
+    if (typeof value !== "string" || !value) {
```
manoldonev: `if (typeof value !== "string")` is enough -- no need to check for !value

tns-core-modules/ui/core/properties/properties.ts
```diff
@@ -56,6 +58,83 @@ export function _getStyleProperties(): CssProperty<any, any>[] {
...
+        return value;
+    }
+
+    if (!isCssValueUsingCssVariable(value)) {
+        // Value is not using css-variable(s)
+        return value;
+    }
+
+    let output = `${value}`.trim();
```
manoldonev: No need to use template string in this case -- better use the simpler `let output = value.trim();`

tns-core-modules/ui/core/properties/properties.ts
```diff
@@ -56,6 +58,83 @@ export function _getStyleProperties(): CssProperty<any, any>[] {
...
+
+    // Evaluate every css-variable in the value.
+    let lastValue: string;
+    while (lastValue !== output) {
+        lastValue = output;
+
+        let m = cssVarValueRegexp.exec(output);
```
manoldonev: Guess `m` is for `match`? We prefer avoiding such abbreviations -- I'd suggest renaming it to `match`, also it can be initialized outside the while { ... } structure just like `lastValue`.

tns-core-modules/ui/core/properties/properties.ts
```diff
@@ -56,6 +58,83 @@ export function _getStyleProperties(): CssProperty<any, any>[] {
...
+
+        const newValue = `${output.substr(0, matchIndex)}${cssVariableValue}${output.substr(matchIndex + matchLength)}`;
+        if (newValue === output) {
+            break;
+        }
+
+        output = newValue.trim();
+    }
+
+    if (output === "null") {
```
manoldonev: How can we end up in this scenario?

m-abs: If can happen if.
1. value = "var(--my-custom-value)"
2. "--my-custom-value" doesn't exist.

I'm not sure what the best way to handle missing values are.

If I don't return `null` here two errors will be logged:
```bash
[1] JS: Style: Failed to get value for css-variable "--my-custom-value" used in "css:background-color"=[var(--my-custom-value)] to Label(181)
[1] JS: Error: Failed to apply property [background-color] with value [var(--my-custom-value)] to Label(181). Error: Invalid color: null
```

But it might be a css-property field like `font-family`
```css
.broken {
   --my-font-family: 'san-sarif';
   font-family: var(--my-none-existing-font), var(--my-font-family); /** null, 'san-sarif' **/
}
```

Alternatively I could just leave the `var(...)`:
```css
.broken {
   --my-font-family: 'san-sarif';
   font-family: var(--my-none-existing-font), var(--my-font-family); /** var(--my-none-existing-font), 'san-sarif' **/
}
```

tns-core-modules/ui/core/properties/properties.ts
manoldonev: Maybe missing value should be substituted with `unset` -- https://developer.mozilla.org/en-US/docs/Web/CSS/unset?

tns-core-modules/ui/core/properties/properties.ts
m-abs: That was meant to check for empty string.

tns-core-modules/ui/core/properties/properties.ts
m-abs: I went with the latter `var(--my-none-existing-font)` is now left as is.

tns-core-modules/ui/core/properties/properties.ts
manoldonev: Ok, got it.

tns-core-modules/ui/core/properties/properties.ts
manoldonev: I wrote it above in another comment but what do you think about using css `unset` value instead of null? Does it make more sense?

tns-core-modules/ui/core/properties/properties.ts
```diff
@@ -56,6 +58,68 @@ export function _getStyleProperties(): CssProperty<any, any>[] {
...
+            }
+
+            return cssVariableValue;
+        });
+    }
+
+    return output;
+}
+
+export function _cssCalcConverter<T>(value: string | T) {
```
manoldonev: I think for consistency it's better to rename this as well -- how about `_evaluateCalcExpression(...)`?

tns-core-modules/ui/core/properties/properties.ts
```diff
@@ -56,6 +58,68 @@ export function _getStyleProperties(): CssProperty<any, any>[] {
...
+const cssCalcRegexp = /\bcalc\(/;
+const cssVariableNameRegexp = /^--[^,\s]+?$/;
+const cssVarValueRegexp = /\bvar\(\s*(--[^,\s]+?)(?:\s*,\s*(.+))?\s*\)/;
+const cssVarAllValuesRegexp = /\bvar\(\s*(--[^,\s]+?)(?:\s*,\s*(.+))?\s*\)/g;
+
+export function isCssVariableName(property: string) {
+    return cssVariableNameRegexp.test(property);
+}
+
+export function isCssCalcValue(value: string) {
```
manoldonev: If we rename `_cssCalcConverter`(see below) probably a better name for this would be `isCssCalcExpression`?

tns-core-modules/ui/core/properties/properties.ts
m-abs: I missed that.

Yes of cause, unset is a much better choice.

manoldonev: Actually using `unset` value right away is probably not the best solution -- we need to first exhaust the options for custom property fallback values (see https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#Custom_property_fallback_values, also I think your css variable regexes already take into account the possibility of fallback values), and only if after that the var() substitution is invalid --> use css `unset` value.

So here are several examples that I think we need to support:
``` css
.one {
    --my-var: blue;
    color: var(--my-var); // resolved as color: blue;
}

.two {
    --my-var: blue;
    color: var(--my-var, red); // resolved as color: blue;
}

.three {
    color: var(--my-var); // resolved as color: unset;
}

.four {
    color: var(--my-var, red); // resolved as color: red;
}

.five {
    --my-fallback-var: lime;
    color: var(--my-var, var(--my-fallback-var)); // resolved as color: lime;
}

.six {
    color: var(--my-var, var(--my-fallback-var, yellow)); // resolved as color: yellow;
}

.seven {
    color: var(--my-var, var(--my-fallback-var)); // resolved as color: unset;
}
```

tns-core-modules/ui/styling/style-properties.ts
```diff
@@ -116,6 +117,10 @@ export namespace PercentLength {
         }
         if (typeof fromValue === "string") {
             let stringValue = fromValue.trim();
+            if (isCssCalcExpression(stringValue)) {
```
manoldonev: Evaluating the calc expressions as part of the Length.parse(...) / PercentLength.parse(...) methods seems a bit strange (out of place?). My first concern is that these parse methods have specific purpose and second -- why are we separating the evaluation of css variables from the evaluation of calc expressions?

What do you think about executing both in the properties.ts setters (like you do in style-scope.ts):
``` ts
let boxedValue = _evaluateCssCalcExpression(_evaluateCssVariable(view, cssName, inputBoxedValue));
```

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -522,32 +524,52 @@ export class CssState {
...
+
             const value = newPropertyValues[property];
+            if (oldProperties && property in oldProperties && oldProperties[property] === newPropertyValues[property] && !isCssValueUsingCssVariable(value)) {
+                // Skip unchanged values unless they use css-variables
+                continue;
+            }
+
             try {
                 if (property in view.style) {
                     view.style[`css:${property}`] = value;
```
manoldonev: Actually is there a reason why we are not evaluating the css variables & calc expressions here like in the "else" case?
If we could do that I think it would be better for consistency and would isolate the changes too.

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -819,16 +841,32 @@ function resolveFilePathFromImport(importSource: string, fileName: string): stri
 export const applyInlineStyle = profile(function applyInlineStyle(view: ViewBase, styleStr: string) {
     let localStyle = `local { ${styleStr} }`;
     let inlineRuleSet = CSSSource.fromSource(localStyle, new Map()).selectors;
-    const style = view.style;
+
+    // Reset unscoped css-variables
+    view.style.clearCssVariable(false);
```
manoldonev: As far as I can see this is the only place clearCssVariable(...) is called -- my suggestion is to remove the optional `scoped` argument at this point (moreover, I think the three-way check you are using true/false/undefined is a bit confusing). Also, probably `clearCssVariables()` is a better name for this method.

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -819,16 +841,32 @@ function resolveFilePathFromImport(importSource: string, fileName: string): stri
...
+    });
 
     inlineRuleSet[0].declarations.forEach(d => {
         // Use the actual property name so that a local value is set.
-        let name = d.property;
+        let property = d.property;
         try {
-            if (name in style) {
-                style[name] = d.value;
+            if (isCssVariableName(property)) {
```
manoldonev: Generally this "if" statement can be placed outside the try/catch block.

tns-core-modules/ui/core/properties/properties.ts
m-abs: Good catch.
I've added support for fallback and a test based on your example:
https://github.com/NativeScript/NativeScript/blob/53ae2d7969bfaacd3ce04b0f7d653c363940257b/tests/app/ui/styling/style-tests.ts#L1653-L1730

tns-core-modules/ui/styling/style-properties.ts
m-abs: I've moved this into `tns-core-modules/ui/styling/style-scope.ts` and reverted my changes to `tns-core-modules/ui/styling/style-properties.ts`.

I think it is much cleaner evaluate the CSS expressions in `style-scope.ts`.

tns-core-modules/ui/styling/style-scope.ts
m-abs: I started out evaluating the CSS expressions here, but ran into some issues.
I've solved those and moved it back.

tns-core-modules/ui/styling/style-scope.ts
m-abs: You're right. It is only used to clear unscoped css-vars,

I'm not sure about the name `clearCssVariables()`, since it is only for unscopes css-vars, but I don't have a better suggestion at the moment.

tns-core-modules/ui/styling/style/style.ts
```diff
@@ -29,6 +39,50 @@ export class Style extends Observable implements StyleDefinition {
         }
     }
 
+    public setCssVariable(varName: string, value: string, scoped: boolean): void {
+        this.cssVariables.set(makeCssVariableName(varName, scoped), value);
+    }
+
+    public unsetCssVariable(varName: string, scoped: boolean): void {
```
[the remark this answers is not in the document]
m-abs: I renamed this function to `unsetScopedCssVariable(...)` to make it more clear what it does.

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -59,6 +61,32 @@ try {
     //
 }
 
+/**
+ * Evaluate css-variable and css-calc expressions
+ */
+function evaluateCssExpressions(view: ViewBase, property: string, value: string) {
+    if (isCssVariableExpression(value)) {
```
manoldonev: I think we can drop this check -- it is already asserted in the beginning of the `_evaluateCssVariableExpression` method.

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -59,6 +61,32 @@ try {
...
+        const newValue = _evaluateCssVariableExpression(view, property, value);
+        if (newValue === "unset") {
+            return unsetValue;
+        }
+
+        value = newValue;
+    }
+
+    if (isCssCalcExpression(value)) {
```
manoldonev: Same here -- asserted in `_evaluateCssVariableExpression`

tns-core-modules/ui/styling/style/style.d.ts
```diff
@@ -168,6 +168,32 @@ export class Style extends Observable {
...
+    /**
+     * Get value of the css-variable.
+     * If the value is not set on this style-object, try the parent view. 
+     */
+    public getCssVariable(varName: string): string | null;
+
+    /**
+     * Remove all unscoped css-variables
+     */
+    public resetUnscopedCssVariables(): void;
```
manoldonev: I think `clearUnscopedCssVariables` or `removeUnscopedCssVariables` might be better options here.

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -522,22 +550,68 @@ export class CssState {
...
+        for (const property in newPropertyValues) {
+            if (isCssVariable(property)) {
+                const value = newPropertyValues[property];
+
+                view.style.setScopedCssVariable(property, value);
+            }
+        }
+
+        // Unset removed css-variables for the scope
+        for (const property in oldProperties) {
```
manoldonev: Do we need to unset "old" scoped css variables one by one? Can't we just call (hypothetic) `view.style.clearScopedCssVariables()` before we iterate on `newPropertyValues` above and set the current scoped css variables?

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -522,22 +550,68 @@ export class CssState {
...
+            if (!(property in newPropertyValues)) {
+                if (isCssVariable(property)) {
+                    view.style.unsetScopedCssVariable(property);
+                }
+            }
+        }
+
+        // Now the css-variables are up-to-date, evalute css-expressions to get the latest values.
+        for (const property in newPropertyValues) {
```
manoldonev: When iterating on `newPropertyValues` above to set the current scoped variables we can initialize a flag `isCssVariablesInUse` or something and only iterate over the `newPropertyValues` once again if this flag is in place?

m-abs: We need to check the value for all `newPropertyValues` in the first, to see if they are css-calc/variable expression.
I added the flag `isCssExpressionInUse` for this.

tns-core-modules/ui/styling/style-scope.ts
m-abs: Agreed.
For consistency I named it `view.style.resetScopedCssVariables()`

tns-core-modules/ui/styling/style-scope.ts
manoldonev: Agreed, I was thinking we could raise the flag inside the `if(isCssVariable(property){...}` statement but this is not correct -- the non-presence of a css variable does not guarantee we should not be evaluating css expression so your implementation is correct.

As for savings -- it will be smaller in this case but due to the short-circuit OR evaluation `isCssVariableExpression(...)` & `isCssCalcExpression(...)` will still be evaluated only if the flag is false.

manoldonev:
test

SvetoslavTsenov:
test

m-abs:
test

m-abs:
Hi @manoldonev,
Thank you for approving the PR.

Unfortunately I have discovered a limitation with the `css-calc` support.

Expressions like `calc(100% - 30px + 20px)` are reduced to `calc(100% - 10px)` because we don't know what the relative size of `100%` is at the time.

Can we leave it as a known limitation and print a warning if/when it happens?

SvetoslavTsenov:
Hey @m-abs, this feat will be available in the next version of `tns-core-modules@next` and 
in the next official version of `ins-core-modules@6.1.0`

manoldonev:
@m-abs can you log this limitation https://github.com/NativeScript/NativeScript/pull/7553#issuecomment-522764014 as a separate issue and we will mark it with the "known issues" label for the time being?

manoldonev:
@m-abs Thank you for the awesome contribution and for taking time to address all comments!

PR #7501, opened 2019-07-10
chore: fix unit tests

manoldonev:

PR #7563, opened 2019-07-22
refactor: Enable missing css check and test

vakrilov:
Enable missing CSS check and that were disabled in #7499 and #7501


tns-core-modules/module-name-resolver/non-bundle-workflow-compat.ts
```diff
@@ -66,43 +87,37 @@ function processFolder(path: string) {
...
-    // moduleName is file - load all files in it's folder
+    // moduleName is file - load all files in its parent folder
     const parentName = moduleName.substr(0, moduleName.lastIndexOf(fs.path.separator));
     const parentFolderPath = fs.path.join(fs.knownFolders.currentApp().path, parentName);
     if (fs.Folder.exists(parentFolderPath)) {
         processFolder(parentFolderPath);
+        folderFound = true;
+    }
 
+    if (folderFound) {
```
MartoYankov: Why do we want to traverse the parent folder if we know the module is a folder? The same applies to the code below.

tns-core-modules/application/application-common.ts
```diff
@@ -1,5 +1,5 @@
 // Require globals first so that snapshot takes __extends function.
-require("globals");
+import "../globals";
```
MartoYankov: What is difference here? We have this in a lot of doc files in the API reference i believe.

tns-core-modules/trace/trace.ts
```diff
@@ -131,21 +131,16 @@ export module categories {
...
-        + Animation + separator
-        + Transition + separator
-        + Livesync;
+    export const All = [
+        VisualTreeEvents, Layout, Style,
+        ViewHierarchy, NativeLifecycle,
+        Debug, Navigation, Test, Binding,
+        Error, Animation, Transition, Livesync,
+        ModuleNameResolver]
+        .join(separator);
```
MartoYankov: 💯 😄

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -40,6 +40,8 @@ function ensureKeyframeAnimationModule() {
 }
 
 import * as capm from "./css-animation-parser";
+import { sanitizeModuleName } from "../builder/module-name-sanitizer";
+import { resolveModuleName } from "../../module-name-resolver";
```
MartoYankov: It's a bit weird that these are in different modules

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -336,8 +344,10 @@ applicationCommon.on("cssChanged", onCssChanged);
 applicationCommon.on("livesync", onLiveSync);
 
 export const loadAppCSS = profile("\"style-scope\".loadAppCSS", (args: applicationCommon.LoadAppCSSEventData) => {
-    loadCss(args.cssFile);
-    applicationCommon.off("loadAppCss", loadAppCSS);
+    if (applicationCommon.hasLaunched()) {
```
MartoYankov: This has already been checked below at line 353. Why is this necessary?

tns-core-modules/application/application-common.ts
vakrilov: Webpack is strange and creates code for all "import" statements at the **start** of the module. So all the imports that are **after** `require("globals")` actually get executed before it. 
This happens with when modules are linked and TS is compiled by webpack/

tns-core-modules/module-name-resolver/non-bundle-workflow-compat.ts
vakrilov: It might be an empty folder. This was actually what happens sometimes in the playground - because syncing new app removes the files, but does not remove folders. You might have `app-root` folder which is empty and `app-root.xml` file. 

Just to be safe we traverse both the folder and the parent dir. We do caching by folder so we make sure we won't traverse folders more than once ;)

tns-core-modules/ui/styling/style-scope.ts
vakrilov: `sanitizeModuleName` is a utility function and is used in many places. We can probably move it to utils

tns-core-modules/ui/styling/style-scope.ts
vakrilov: Good question!
Because of the webpack snapshot plugin which inserts a code calling [`loadAppCss()`](webpack/blob/master/load-application-css.js#L7) directly. This code is not snap-shot safe and should not be executed if the application is not running (during snapshot time) - so we need the check.

tns-core-modules/application/application-common.ts
MartoYankov: Maybe that's the reason we got so many circular dependencies with the webpack plugin for circular deps. @manoldonev

PR #7665, opened 2019-08-09
chore: update modal-navigation app to use local modules via symlink

VladimirAmiorkov:
Update the e2e/modal-navigation` application to work with local `tns-core-modules`

### DIFF
--- a/e2e/modal-navigation/tsconfig.json
+++ b/e2e/modal-navigation/tsconfig.json
@@ -21,7 +21,12 @@
             ]
         }
     },
+    "include": [
+        "../tns-core-modules",
+        "**/*"
+    ],
     "exclude": [
+        "../tns-core-modules/node_modules",
         "node_modules",
         "platforms",
         "e2e"
--- a/e2e/modal-navigation/tsconfig.json
+++ b/e2e/modal-navigation/tsconfig.json
@@ -26,7 +26,6 @@
         "**/*"
     ],
     "exclude": [
-        "../tns-core-modules/node_modules",
         "node_modules",
         "platforms",
         "e2e"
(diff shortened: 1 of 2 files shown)


e2e/modal-navigation/tsconfig.json
```diff
@@ -21,7 +21,12 @@
             ]
         }
     },
+    "include": [
+        "../tns-core-modules",
+        "**/*"
+    ],
     "exclude": [
+        "../tns-core-modules/node_modules",
```
MartoYankov: I think the last we agreed is to remove/not use these includes. We are not sure how this works and what it does, but the path is wrong and adding configurations that don't do anything isn't a good idea.

VladimirAmiorkov: I could remove it and.or update it but there is the same here: https://github.com/NativeScript/NativeScript/blob/master/apps/tsconfig.json#L25 . Should we also remove that?

MartoYankov: I believe we discussed that all these should be cleaned up together, but it's not critical at the moment.

MartoYankov: Btw, the app that you are linking to doesn't exist anymore. The apps there have been moved to various places.

VladimirAmiorkov: What do you mean it does not exist anymore? Are you talking about this app: https://github.com/NativeScript/NativeScript/tree/master/e2e/modal-navigation

VladimirAmiorkov: I have removed this above line from the `exclude`

PR #7671, opened 2019-08-11
feat(file-system): async read/write

### LINKED ISSUE #2239: Make the file-system module run in a background thread [$300]

PeterStaev:
## PR Checklist

Current `file-system` methods for read/write files are synchronous which makes the UI thread block if a big file is read/written.

## What is the new behavior?
True async methods have been added that use native code inside `tns-core-modules-widgets` execute the read/write operations in a separate thread. 

Closes #2239 .


NathanaelA:
Unfortunately; they probably won't release this until version 7.   Breaking changes are typically sat on until the next major release. 

So maybe instead of replacing `readText`.   Make `readText ` call `readTextSync` function, and rename this new function to be `readTextAsync` that way this could be released and isn't a breaking change.  But could be marked as depreciated.   Then in NS 7 swap `readText` to point to `readTextAsync`.   This way we have access to the async functions now, but no breaking change occurs.   :D

PeterStaev:
Yes, this is a possible solution, if breaking changes in minor versions are considered that bad :)

etabakov:
Hey, @PeterStaev. Regarding the breaking changes policy, our team is committed to following the [SemVer Specification](https://semver.org/), which means that all changes in a minor version should be introduced in a backward-compatible manner. With that said, Nathanael's suggestion is a good one. Would you mind adjusting your PR according to his proposal, so that we can proceed with the technical review?

Thanks for jumping on this issue!

PeterStaev:
@etabakov , updated and removed the breaking changes. Removed from the above description as well :)

isabelh340:
Alot

tns-core-modules/file-system/file-system.ts
```diff
@@ -242,23 +292,38 @@ export class File extends FileSystemEntity {
...
+            } catch (ex) {
+                reject(ex);
 
-            const content = this.readTextSync(localError, encoding);
-            if (!hasError) {
-                resolve(content);
+                return;
+            }
+
+            try {
```
manoldonev: Is the try/catch block necessary here?

tns-core-modules-widgets/android/widgets/src/main/java/org/nativescript/widgets/Async.java
```diff
@@ -9,17 +9,25 @@
 import android.util.Log;
 
 import java.io.BufferedInputStream;
+import java.io.BufferedReader;
```
manoldonev: Unused import.

tns-core-modules/file-system/file-system.ts
```diff
@@ -277,23 +342,38 @@ export class File extends FileSystemEntity {
...
+            } catch (ex) {
+                reject(ex);
 
-            this.writeTextSync(content, localError, encoding);
-            if (!hasError) {
-                resolve();
+                return;
+            }
+
+            try {
```
manoldonev: Same as above -- I think we do not need the try/catch block here in the asynchronous variant?

PeterStaev: 👍 Indeed they are not needed, looking at the code now. Initially I thought that something might throw leaving the lock flag rased.

manoldonev:
test

tns-core-modules/file-system/file-system.d.ts
```diff
@@ -96,6 +96,11 @@ export class File extends FileSystemEntity {
      */
     readTextSync(onError?: (error: any) => any, encoding?: string): string;
 
+    /**
+     * Reads the binary content of the file asynchronously.
+     */
+    read(): Promise<any>;
```
manoldonev: Can you add an async variant of the following unit test https://github.com/PeterStaev/NativeScript/blob/625eba21e690425f187f23d1e1172502965d1722/tests/app/file-system/file-system-tests.ts#L178-L202 to test the new async binary read/write methods?

PeterStaev: Not a problem, but can you point me to a sample async test available in the repo. Not sure how to use `TKUnit` in case of an async operation. Or is it just the same?

manoldonev: Basically the same -- check the previous test in the same file https://github.com/PeterStaev/NativeScript/blob/625eba21e690425f187f23d1e1172502965d1722/tests/app/file-system/file-system-tests.ts#L141 that asserts async text read/write.

manoldonev:
test

manoldonev:
test

SvetoslavTsenov:
test

manoldonev:
@PeterStaev Looking good now and I have merged the PR. Thanks!
Will check what exactly is the process with https://www.bountysource.com/issues/34832190-make-the-file-system-module-run-in-a-background-thread and someone with the NativeScript team will follow up with you.

PR #7681, opened 2019-08-15
fix(ios): handle tabs extended layout for ios 10

MartoYankov:
The new Tabs component crashed on iOS 10 and 9, because of usage of APIs that were removed on these versions (the safe area insets APIs).

Fixes iOS 10 and 9 tests on CI. Should test on these versions.


SvetoslavTsenov:
test --ignore uitests@${sha1} ios10

SvetoslavTsenov:
test --ignore uitests#${sha1} ios10

PR #7689, opened 2019-08-17
fix(css): parse css selectors with escape sequences

### LINKED ISSUE #7688: CSS selectors don't support escape sequence
**Environment**

 - CLI: 6.0.3
 - Cross-platform modules: 6.0.6
 - Android Runtime: 6.0.1
 - iOS Runtime: 6.0.2
 - Plugin(s): [nativescript-tailwind](https://github.com/rigor789/nativescript-tailwind), ... (N/A)

**Describe the bug**

Given we have a css selector like this:
```css
.w-1\/2 {
  width: 50%;
}
```
This is fully supported in browsers, however in NativeScript when the css parser parses it, the selector is parsed as `.w-1` completely ignoring the `\/2` part. 

In practice you would use these selectors as `<Element class="w-1/2" />`.

The issue with them parsing as just `.w-1` is that if you already have a `.w-1` class, it will override it's properties and lead to undesired results.

**To Reproduce**

Add the following css to any NativeScript app
```css
.w-1 {
  width: 25;
}
.w-1\/2 {
  width: 50%;
}
```

And use this simple layout to see the issue
```HTML
<StackLayout>
  <Label class="w-1" backgroundColor="red" />
  <Label class="w-1/2" backgroundColor="blue" />
</StackLayout>
```

**Expected behavior**

I would expect the `red` label to be `width: 25` and the `blue` label to be `width: 50%`.

What is actually happening:
 * `red` is `width: 50%`
 * `blue` is `width: auto` (or whatever is default)

**Sample project**

https://play.nativescript.org/?template=play-vue&id=kMFahL

**Additional context**

I have tracked the issue down to the css parser, more specifically this part:

https://github.com/NativeScript/NativeScript/blob/c0b8db40982825f8e14b70f5a5d2fece17d06337/tns-core-modules/css/parser.ts#L737-L750

I'm working on a PR to fix this issue.

rigor789:
## PR Checklist

https://github.com/NativeScript/NativeScript/issues/7688

## What is the new behavior?
With the changes, the escaped css selectors are parsed correctly and can be applied to elements.

Fixes/Implements/Closes #7688.


tns-core-modules/css/parser.ts
```diff
@@ -743,7 +743,7 @@ export function parseSimpleIdentifierSelector(text: string, start: number = 0):
     }
     const end = simpleIdentifierSelectorRegEx.lastIndex;
     const type = <"#" | "." | ":" | "">result[1];
-    const identifier: string = result[2];
+    const identifier: string = result[2].replace(/\\/g, '');
```
NathanaelA: @rigor789 - Wouldn't this strip out the '\' you kept on?    Is that what you really want to do?

rigor789: @NathanaelA Unfortunately not!

For example, if the input is
```css
.w-6\/12 {
  width: 50%
}
```
The parser receives both the `\` backslash and the `/` forward slash. If the regex doesn't match the `\` anything after it gets ignored, including the `/12` part, so we actually need to allow it, but later remove it, since it is just an escape character.

I don't think it's possible to capture a group with skipping characters in the middle.

NathanaelA: Ah, I see what you are saying...   Went back and had to look at the spec.  :-D     Problem really is that this is unfortunately just a small hack to accept this one corner case; in all reality the `\` should be parsed to see if it is a hex number following it or any literal character, if hex number it should be converted i.e. "\000020" and "\20" should turn in to a space.   Where `\&` turns into a `&`.      This CSS might need to go through the same process as the XML where the Escapes are parsed first...

manoldonev:
test

PR #7691, opened 2019-08-18
feat: introduce fixed font icons to tab navigation

MartoYankov:
Add new property `fixedIcons` that defaults to `true` that fixes icon sizes to the spec size - 24x24 with max width 31 and max height 28 allowed. Set to `false` to use the actual size of the images.


tns-core-modules/ui/bottom-navigation/bottom-navigation.android.ts
```diff
@@ -579,6 +500,131 @@ export class BottomNavigation extends TabNavigationBase {
...
+        } else {
+            // TODO
+            // traceMissingIcon(iconSource);
+        }
+    
+        return imageDrawable;
+    }
+    
+    private fixIconSize(image: android.graphics.Bitmap): android.graphics.Bitmap {
+        const minSide = 24;
```
manoldonev: minSide -> minSize typo?

manoldonev: Also minIconSize, maxIconWidth, maxIconHeight can be initialized outside of this method.

tns-core-modules/ui/bottom-navigation/bottom-navigation.android.ts
```diff
@@ -579,6 +500,131 @@ export class BottomNavigation extends TabNavigationBase {
...
+            imageDrawable = new android.graphics.drawable.BitmapDrawable(application.android.context.getResources(), image);
+        } else {
+            // TODO
+            // traceMissingIcon(iconSource);
+        }
+    
+        return imageDrawable;
+    }
+    
+    private fixIconSize(image: android.graphics.Bitmap): android.graphics.Bitmap {
```
manoldonev: Naming is a bit misleading -- how about renaming it to getFixedSizeIcon(...) or getScaledIcon(...) ?

tns-core-modules/ui/bottom-navigation/bottom-navigation.android.ts
manoldonev: Even better -- extract the common cross-platform logic that calculates the scaled image size and only leave here the platform specific image creation?

tns-core-modules/ui/tab-navigation-base/tab-navigation-base/tab-navigation-base.d.ts
```diff
@@ -44,6 +44,11 @@ export class TabNavigationBase extends View {
      */
     selectedIndex: number;
 
+    /**
+     * Gets or sets whether the icons in the tab navigation tab strip are fixed based on specs or are actual size. Defaults to true.
+     */
+    fixedIcons: boolean;
```
manoldonev: `fixedIcons` naming might be misleading -- how about renaming this property to `fixedIconSize` or `isIconSizeFixed`?

tns-core-modules/ui/tab-navigation-base/tab-navigation-base/tab-navigation-base.d.ts
```diff
@@ -44,6 +44,11 @@ export class TabNavigationBase extends View {
      */
     selectedIndex: number;
 
+    /**
+     * Gets or sets whether the icons in the tab navigation tab strip are fixed based on specs or are actual size. Defaults to true.
+     */
+    isIconSizeFixed: boolean;
```
manoldonev: Wondering now whether this property should be exposed here or on the `TabStrip` class as the latter exposes another icon related property `iosIconRenderingMode` (that we seem not to respect at the moment but that is another matter).

manoldonev:
This scenario still does not look OK out-of-the-box but we'll investigate it after feature freeze.
Screenshot_1566317899

manoldonev:
@MartoYankov Bottom Navigation font icons scenario -- third font icon gets clipped upon activation (will log it):

Simulator Screen Shot - iPhone Xʀ - 2019-08-20 at 19 43 56
Simulator Screen Shot - iPhone Xʀ - 2019-08-20 at 19 44 03

manoldonev:
@MartoYankov iOS Tabs font icons scenario looks broken -- font icon color is wrong (will log it):

Simulator Screen Shot - iPhone Xʀ - 2019-08-20 at 19 47 42

manoldonev:
@MartoYankov iOS tabs icon change scenario still does not change the icon (will log it):

Simulator Screen Shot - iPhone Xʀ - 2019-08-20 at 19 48 30

MartoYankov:
@manoldonev 

1. First one should be investigated.

2. Regarding the clipped icon - I believe this is expected. If you look closely the icon is clipped before the tab item is activated too. This is the case when the font is wrong and doesn't support icons. It's not an issue with the scaling, but with the font itself. If you do this in a `Label` the result will be the same.

3. Regarding the iOS Tabs icon colors, this is a decision/limitation I imposed. Since the color property can't apply on the titles (because of MDCTabBar limitation) I disabled it on the icons as well. Now for both the color property value of the `TabStrip` is the valid one. It's possible to do it for the icons since they are simply images, but looks weird.

manoldonev:
test api19

manoldonev:
@MartoYankov is the last (fourth) issue also a limitation of the MDCTabBar usage (about iOS tabs icon change example not working) or should I log it?

Issue #570, opened 2015-08-11
Binary data support for fetch and XMLHttpRequest


enchev:
https://fetch.spec.whatwg.org/#dom-body-arraybuffer
https://xhr.spec.whatwg.org/

satish-pokala:
writesync is not working with binary data any update on this ?

kumaran-dena:
Hey @enchev & @NickIliev  Any update on this feature request?

May i know when we all can expect this feature avail?

Issue i have faced and my personal need of this feature!
I need to download the byte array from the server through http post api call. The byte array will be hold different type of attachments files/documents which will sizes around 60 to 100 MB's. Once i download & decompress the byte array to local file system then only i can proceed further in my application.

Still i cannot get any proper solution/idea on this. I got stucked in this step. I got confused with this.

Can you people suggest me anything in this moment !!

caiusCitiriga:
Any news on this?

surdu:
This was fixed in #7707

Issue #1314, opened 2016-01-01
http gzip support missing


NathanaelA:
The http module in nativescript apparently doesn't support basic things like gzip (at least on Android).  I have a server that by default is sending a gzipped response (to save major bandwidth as the data is rather large) and of course then the http module will die a horribile death and throw a Java EOF error.

This is actually something that is very surprising that it is totally broken in NS as anything mobile we want to save the amount of data being transferred.   So by default we should be supporting http data compression.

If anyone else runs into this issue; you can send the header: **accept-encoding: identity** which will force it to be a text based uncompressed stream protocol.

atanasovg:
We may research whether the okhttp library for Android already handles this. Relates to https://github.com/NativeScript/android-runtime/issues/209. Probably a custom http plugin that enables okhttp would be a viable solution.

valentinstoychev:
Yes - we should check this as and provide a recommended way for consumption of "zipped" http content in NativeScript as this really will increase the perf of the app. cc: @enchev

enchev:
Technically you/we can add support even know in [our JavaScript wrapper](https://github.com/NativeScript/NativeScript/blob/master/http/http-request.android.ts#L44) or simply in the application [using this approach](http://stackoverflow.com/questions/1573391/android-http-communication-should-use-accept-encoding-gzip) however in this case unzipping will be in the UI thread. 

We plan to move [our Android HTTP implementation](https://github.com/NativeScript/android-runtime/blob/master/src/src/com/tns/Async.java) from the runtime to separate plugin similar to [widgets](https://github.com/NativeScript/android-widgets) where we can add this functionality and our modules will depend on the new Async module [similar again to widgets](https://github.com/NativeScript/NativeScript/blob/master/package.json#L39).

What do you think about this plan? :)

jacargentina:
@enchev  On iOS i have no problem with the gzip response. Seems to be related only to the Android. Any news on this?

mshanak:
you can use nativescript http with angular , following example:

first import http serves in the top of the service component
`var http = require("http");
`

in the sevice.ts file ,you need to convert the http promise to observable .
```
//GET Example
	getList(): Observable<News[]> {
		return Observable.fromPromise(
					**http.getJSON(serverConfig.apiUrl+"news")**
				).map(function(res:any) {
					console.log('finished sendign request'); 
					return res as News[] 
				} );

	}


//POST EXAMPLE

submitFormData(data): Observable<Forms> {
[...]

PR #7707, opened 2019-08-21
feat(http): better binary support & XHR support

### LINKED ISSUE #7632: Protobuf support
**Describe the solution you'd like**

Protobuf based frameworks like [gRPC](https://grpc.io/) and [Twirp](https://github.com/twitchtv/twirp) (made by Twitch) are very popular because of the efficiencies gained in front->backend handshake clarity, straightforward for teams of devs to understand and manage, and size of payload over the wire (extra appealing to mobile).

Ultimately the solution I'd like is to send protobuf (binary) from the [NS HTTP Framework module](https://docs.nativescript.org/ns-framework-modules/http) and be able to receive a protobuf (binary) response.

My hope is the market trend towards protobuf based technologies will give further motivation to supporting sending and receiving these (small) binary payloads.

**Describe alternatives you've considered**

Writing protobuf to then filesystem then sending via [nativescript-background-http](https://github.com/NativeScript/nativescript-background-http/) in the background.  Not only is this super messy, it is very error prone (and hard to recover from).  Not good for the mission critical path of your app interfacing with backend. 

**Questions for starting discussion**

I know there are a few outstanding issues (see below) around supporting binary, my hope is this issue can bring them together with a solid use case (Protobuf).

1.  Support to accept gzip (binary) incoming payloads was added [here](https://github.com/NativeScript/NativeScript/issues/1314).  Can this methodology be used to accept `application/protobuf`?
1.  [protobuf.js](https://github.com/protobufjs/protobuf.js) is rock solid and works in node and browser.  What are the specific technical road blocks preventing it from being used in NS?  In searching and looking at code I can't tell what they are (I'm green to native dev).

**Additional context**

-  Related NS Issues: 
    - https://github.com/NativeScript/nativescript-angular/issues/1202
    - https://github.com/NativeScript/NativeScript/issues/570
    - partial blob support in [fetch](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/fetch/fetch.js)
-  My Q [posted](https://stackoverflow.com/questions/56622076/how-can-i-use-protobuf-in-nativescript) on SO.

charsleysa:
## PR Checklist

HTTP module does not support uploading binary information, only text based information.
XHR doesn't follow spec.

## What is the new behavior?

HTTP module now supports uploading binary information through the use of ArrayBuffer.
XHR now better follows spec, now has support for Blob and ArrayBuffer.

Implements #570.
Closes #7632.
Closes NativeScript/nativescript-angular#1202.

BREAKING CHANGES:

Possibly a breaking change if anyone is using the `org.nativescript.widgets2.Async.Http` Java class directly as `RequestOptions.content` has changed from a type `String` to type `java.nio.ByteBuffer`, though I'm not sure if this is considered to be an internal API.

Also, XHR is now more compliant with the spec which means some code that may have previously worked (or failed silently) will now cause an error.

Migration steps:
The content string must be converted to a `java.nio.ByteBuffer` instead of setting it directly to the request options. Here's an example of converting a content string:
```ts
        const contentString = '...blah...';
        const nativeString = new java.lang.String(contentString);
        const nativeBytes = nativeString.getBytes('UTF-8');
        const nativeBuffer = java.nio.ByteBuffer.wrap(nativeBytes);
        myhttpRequestOptions.content = nativeBuffer;
```


rynop:
Thanks @charsleysa , really appreciate this.  Few things:

- How are you handling `application/protobuf` (binary) responses? AFAIK NS does not currently support (I think confirmed by the https://github.com/NativeScript/nativescript-angular/issues/1202 you mention)
-  Do you have any examples/testcases? From looking at code should just need to set `content` to an `ArrayBuffer` however having something concrete to start from would save me time.  Sounds like you use Angular - Angular ex would be perfect.

FWIW I'm willing to test and write the docs

charsleysa:
Hi @rynop 

Regarding your first point, you are correct that binary download support is not currently available through angular due to XHR not having binary response support. I will add in arraybuffer responseType support to XHR and include it in this PR.

As for your second point, any angular tutorial for arraybuffer should work. Here's the http snippet from our app when we upload a user profile image:
```ts
                            this.http
                                .request(
                                    new HttpRequest('PUT', uploadUrl, data, {
                                        headers: new HttpHeaders({
                                            'Content-Type': fileType,
                                            'Content-Disposition': `attachment; filename="${fileName}"`
                                        })
                                    })
                                )
```


If you're able to write tests and docs that'd be a great help!

charsleysa:
@rynop I've implemented full support (within limitations) for Blob and ArrayBuffer in both upload and download for XHR, updated the Fetch API polyfill. I've also added some tests but this PR will probably need a bit of testing against a few apps to make sure nothing is horribly broken.

tests/app/fetch/fetch-tests.ts
```diff
@@ -54,6 +54,30 @@ export var test_fetch_formData = function (done: (err: Error, res?: string) => v
     // << fetch-formdata
 };
 
+export var test_fetch_blob = function (done: (err: Error, res?: string) => void) {
+    // >> fetch-blob
+    fetch("https://httpbin.org/get").then(response => response.blob()).then(function (r) {
+        // Argument (r) is Blob object!
```
rynop: https://httpbin.org/get doesn't return binary data correct? Wondering if this is a valid test. I know the JSON response is Gzipped, but I'm concerned its deflated via https://github.com/NativeScript/NativeScript/issues/1314  

Or are you just testing that you can convert a string response from fetch to a blob?

tests/app/fetch/fetch-tests.ts
```diff
@@ -54,6 +54,30 @@ export var test_fetch_formData = function (done: (err: Error, res?: string) => v
...
+        done(null);
+        // << (hide)
+    }).catch(failOnError(done));
+    // << fetch-blob
+};
+
+export var test_fetch_arraybuffer = function (done: (err: Error, res?: string) => void) {
+    // >> fetch-arraybuffer
+    fetch("https://httpbin.org/get").then(response => response.arrayBuffer()).then(function (r) {
+        // Argument (r) is ArrayBuffer object!
```
rynop: Same here, https://httpbin.org/get returns Gzipped string (JSON) not binary data...

tns-core-modules/http/http.ts
```diff
@@ -58,3 +58,17 @@ export function getFile(arg: any, destinationFilePath?: string): Promise<any> {
             }, e => reject(e));
     });
 }
+
```
rynop: Need to support more than the `GET` method.  For example, `POST`ing protobuf and getting a RPC protobuf (binary) response is very typical.

I think ideally all the existing NS http module [request](https://docs.nativescript.org/ns-framework-modules/http) methods would inspect the response determine if its is binary automatically.

Said another way, when using XHR/fetch from browser I don't have to know ahead of time if the response will be string or not - I just get back a string or an ArrayBuffer.  If this isn't possible, we'd need `[post|put|..]Binary` methods as well.

It looks like my outlined logic IS what you are doing in your new changes to tns-core-modules/xhr/xhr.ts.  If so I'm confused why you need this `getBinary()` method.  So I'm probably missing something here....

tests/app/fetch/fetch-tests.ts
charsleysa: This is just testing the Fetch API to see if it will return the proper type.

tests/app/fetch/fetch-tests.ts
charsleysa: Same as above.

tns-core-modules/http/http.ts
charsleysa: Http lib is only a basic API. If you want full support you must use Fetch of XHR. Also, as per spec for XHR you must specify ahead of time what type of response you want, XHR may not change the response type. You can get close to the functionality you are after by using the Fetch API but it's still not quite the same. Both XHR and Fetch aim to meet spec, the http lib is an underlying platform abstraction which I don't recommend using directly.

rynop: Makes sense, thanks.  I like you, am a big fan of Angular. You recommend staying away from any abstraction? Your [snippet above](https://github.com/NativeScript/NativeScript/pull/7707#issuecomment-523660311) is using the [@angular/common/http](https://angular.io/api/common/http) abstraction of `XMLHttpRequest` correct? This abstraction safe to use in NS in your experience (I'll obv test)?

charsleysa: Angular HttpClient uses XHR under the hood so Angular's abstraction should work completely fine. One note (which if you go through the code you'll be able to see) is that the lower level platform abstractions don't provide progress reports so XHR only reports progress once on completion of the request.

charsleysa:
@rynop did you solve your problem and get this running?
I have not tested linking the repo to another NS app project.

rynop:
@charsleysa nope, I think there is a `tns-core-modules` bug in master (and therefore your fork/branch).  Reproduction of bug and tracking of it is in "Update DevelopmentWorkflow.md for v6" link above.

bundyo:
The problem with linking is that Angular/TS project tend to grab the linked TS files in core modules and try to compile them without the references needed. I've only managed to run it by compiling the core modules with outDir and linking that (.d.ts, package.json and other files, except TS ones should be copied over too). Maybe the team can consider moving to an outDir compilation. :)

rynop:
@charsleysa It was a long (30+) day journey but I _finally_ was able to verify the changes in the PR work great when using binary protobuf payloads.  Can't tell you how much I appreciate this PR, I really think it is a big step forward for NS and gives a competitive advantage over React Native (which last time I checked has problems passing binary data over their JS bridge).

Now the details.

**Workaround so I could test:**
[As suspected](https://github.com/NativeScript/NativeScript/issues/7745), the [development workflow](https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md) for contributing to `tns-core-modules` and `tns-platform-declarations` do not work when verifying using an Angular flavor NS app.  After lots of investigation and debugging I have documented a workaround in https://github.com/NativeScript/NativeScript/pull/7861 (see `files changed`).  It is nasty and the NS team agrees.  They are working on a better approach.

**What I tested**
Binary support you've added fetch/XHR is awesome.  No longer limited to base64 encode a [background thread](https://github.com/NativeScript/nativescript-background-http) which both have obv drawbacks.  My main drive for binary support was to [get protobuf working in NS](https://github.com/NativeScript/NativeScript/issues/7632).  

In my test, I ran against a [Twirp RPC](https://github.com/twitchtv/twirp) server.  For those curious, I used https://github.com/rynop/abp-sam-twirp almost verbatim.

I then created an Angular NS project, and an Angular service with the method below, which invokes the code in this PR.

```
  public createTwirpRPCAdapter(servicePathPrefix: string): $protobuf.RPCImpl {
    return (serviceMethodName: $protobuf.Method, requestData: Uint8Array, callback: Function) => {
      const req = new HttpRequest(
        'POST',
        servicePathPrefix + TwirpService.getServiceMethodName(serviceMethodName),
        // required to get an arraybuffer of the actual size, not the 8192 buffer pool that protobuf.js uses
        // see: https://github.com/dcodeIO/protobuf.js/issues/852
        requestData.slice().buffer,
        {
          headers: new HttpHeaders({
            'Accept': 'application/protobuf',
            'Content-Type': 'application/protobuf'
          }),
          responseType: 'arraybuffer',
        }
      );

      this.http.request<ArrayBuffer>(req)
        .subscribe(
          event => {
            console.log('event', event);
            if (event instanceof HttpResponse) {
              callback(null, new Uint8Array(event.body));
            }
          },
          (error: HttpErrorResponse) => {
            return callback(TwirpError.isTwirpError(error) ? new TwirpError(error) : error, null);
          }
        );
    };
  }
```

I successfully tested both green and error (500/404/400 etc) paths.

There is quite a bit more to the protobuf aspect, but I don't want to bloat this PR.  I'll blog and/or post more details for others wanting this use case in https://github.com/NativeScript/NativeScript/issues/7632

@bundyo the only issue I see is the potentially breaking change (`org.nativescript.widgets2.Async.Http`) mentioned above.  Is this considered an internal API? If not do you know someone at Telerik / maintainer that can make the call? As stated above I think this PR has some significant value, so would hate to see it get held up too long.

bundyo:
@manoldonev Any chance this hitting 6.2?

JoeyBG:
We really need this as well! Simply trying to upload a raw image to S3..

manoldonev:
@bundyo most probably will not ship with 6.2 but this PR is definitely on our radar and we will do our best to process it as soon as possible (and release it officially with 6.3)

rynop:
I should have noted that my test above uses `HttpRequest` from [@angular/common/http](https://angular.io/api/common/http) which under the covers uses `XMLHttpRequest`.

For completeness sake, I just tested `POST`ing protobuf via [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) like so:

```javascript
  public createTwirpRPCAdapter(servicePathPrefix: string, headerOverrides?: HttpHeaders): $protobuf.RPCImpl {
    return (serviceMethodName: $protobuf.Method, requestData: Uint8Array, callback: Function) => {
      fetch(servicePathPrefix + TwirpService.getServiceMethodName(serviceMethodName),
        {
          method: 'POST',
          headers: {
            'Accept': 'application/protobuf',
            'Content-Type': 'application/protobuf'
          },
          body: requestData.slice().buffer,
        })
        .then(response => response.arrayBuffer())
        .then(body => callback(null, new Uint8Array(body)))
        .catch(error => {
          callback(TwirpError.isTwirpError(error) ? new TwirpError(error) : error, null);
        });
    };
  }
```

Sending binary in the request body and getting binary in the response body works in my tests via `XMLHttpRequest` and `fetch`.

acivier-serial:
Hello!
Could you confirm that this feature will be present in the next release? (6.3)

rynop:
@manoldonev since this did not make it into v6.3 as hoped, can you confirm it is slated to go into v6.4?

vahidvdn:
Is the PR stuck for some reason?

rynop:
From my perspective it is ready to go.  I tested multiple scenarios on multiple devices/emulators/simulators.   I'm guessing the team wants to get more testing (wise).

vtrifonov:
test

vtrifonov:
test

vtrifonov:
test

vtrifonov:
test

rynop:
Wahooo!!! Thanks for everyone's hard work on this! Especially you @charsleysa

vibos:
I'm trying to upload image in binary mode using Angular's `HttpClient`. I tried different options, also tried using fetch and xhr, but nothing works. 
I have `ImageSource` as input, then saving it to filesystem and need to send POST request with `multipart/form-data` type.
Could anyone please provide code snippet how to do that?

PR #7711, opened 2019-08-22
feat(TabStrip): add itemTap event

vchimev:
## PR Checklist

## What is the new behavior?

Add *itemTap* event of `TabStrip`.
Add *TabStripItemEventData* containing `TabStripItem`'s index, for which the event is raised.


tns-core-modules/ui/tab-navigation-base/tab-strip/tab-strip.d.ts
```diff
@@ -36,6 +36,34 @@ export class TabStrip extends View {
...
+    /**
+     * Raised when an item inside the TabStrip is tapped.
+     */
+    on(event: "itemTap", callback: (args: ItemEventData) => void, thisArg?: any);
+}
+
+/**
+ * Event data containing information for the index associated with a tab strip item.
+ */
+export interface ItemEventData extends EventData {
```
manoldonev: As far as I can see ListView already uses this interface name for its own purposes -- I think it is not a good idea to use the same name for two different things (also with scoped packages we would be exporting all entities on root level so I don't think this will work at all).

Issue #7709, opened 2019-08-21
Changing rootView className replaces all platform classes


bundyo:
**Environment**
 - Cross-platform modules: 6.1.0-next-2019-08-21-142040-01
 
**Describe the bug**
Since all changes to platform classes in #7606 are done through cssClasses property, they are not reflected in className property, thus changing className will overwrite them.

**To Reproduce**
Change root view className after running the app.

**Expected behavior**
CSS classes should be added both in cssClasses and className properties.

PR #7725, opened 2019-08-23
fix-next(css): className to preserve root views classes

vchimev:
## PR Checklist

When you set `className` to a (modal) root view, it *overwrites* view's default CSS classes.

## What is the new behavior?

When you set `className` to a (modal) root view, it *preserves* view's default CSS classes.

Fixes https://github.com/NativeScript/NativeScript/issues/7709.

### DIFF
--- a/tns-core-modules/css/system-classes.d.ts
+++ b/tns-core-modules/css/system-classes.d.ts
@@ -1,0 +1,24 @@
+/**
+ * @module "system-classes"
+ */ /** */
+
+/**
+ * String value "ns-" used for CSS system class prefix.
+ */
+export const CSS_CLASS_PREFIX: string;
+
+/**
+ * Gets CSS system class for modal root view.
+ */
+export function getModalRootViewCssClass(): string;
+
+/**
+ * Gets CSS system classes for root view.
+ */
+export function getRootViewCssClasses(): string[];
+
+/**
+ * * Appends new CSS class to the system classes, and returns the new length of the array.
+ * @param value New CSS system class.
+ */
+export function pushToRootViewCssClasses(value: string): number;
--- a/tns-core-modules/css/system-classes.d.ts
+++ b/tns-core-modules/css/system-classes.d.ts
@@ -18,7 +18,13 @@ export function getModalRootViewCssClass(): string;
 export function getRootViewCssClasses(): string[];
 
 /**
- * * Appends new CSS class to the system classes, and returns the new length of the array.
+ * * Appends new CSS class to the system classes and returns the new length of the array.
  * @param value New CSS system class.
  */
 export function pushToRootViewCssClasses(value: string): number;
+
+/**
+ * Removes CSS class from the system classes and returns it.
+ * @param value
+ */
+export function removeFromRootViewCssClasses(value: string): string;
--- a/tns-core-modules/css/system-classes.d.ts
+++ b/tns-core-modules/css/system-classes.d.ts
@@ -5,7 +5,7 @@
 /**
  * String value "ns-" used for CSS system class prefix.
  */
-export const CSS_CLASS_PREFIX: string;
+export const CLASS_PREFIX: string;
 
 /**
  * Gets CSS system class for modal root view.
--- a/tns-core-modules/css/system-classes.ts
+++ b/tns-core-modules/css/system-classes.ts
@@ -1,0 +1,21 @@
+const MODAL = "modal";
+const ROOT = "root";
+
+export const CLASS_PREFIX = "ns-";
+
+const modalRootViewCssClass = `${CLASS_PREFIX}${MODAL}`;
+const rootViewCssClasses = [`${CLASS_PREFIX}${ROOT}`];
+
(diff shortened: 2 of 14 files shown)


tns-core-modules/ui/core/view-base/view-base.ts
```diff
@@ -1036,11 +1038,23 @@ bindingContextProperty.register(ViewBase);
 export const classNameProperty = new Property<ViewBase, string>({
     name: "className",
     valueChanged(view: ViewBase, oldValue: string, newValue: string) {
-        let classes = view.cssClasses;
-        classes.clear();
+        const cssClasses = view.cssClasses;
+
+        const shouldAddModalRootViewCssClass = cssClasses.has(MODAL_ROOT_VIEW_CSS_CLASS);
+        const shouldAddRootViewCssClasses = cssClasses.has(ROOT_VIEW_CSS_CLASSES[0]);
```
bundyo: What if only some of them are missing and the first is there? :)

tns-core-modules/ui/core/view-base/view-base.ts
```diff
@@ -1036,11 +1038,23 @@ bindingContextProperty.register(ViewBase);
...
+
+        cssClasses.clear();
```
bundyo: Why do we clear this at all? Maybe modifying the add method to not add a cssClass if it already exists would skip all of this adding and removing classes? 

Also modifying cssClassses should update className.

bundyo: Actually just saw that cssClasses is a Set, so it should have only unique items, thus no need for even checking on add(). :slightly_smiling_face:

bundyo: Yep, deleting all the new code and cssClasses.clear() works okay.

bundyo: Now we are left with only one problem - all the new root classes are not reflected in className, thus the user can't see them in the string or in DevTools.

bundyo: Okay, since cssClasses is not synced with className, it has to be cleared every time before adding classes, to remove any classes that are not in className...

tns-core-modules/ui/core/view-base/view-base.ts
```diff
@@ -1036,11 +1038,23 @@ bindingContextProperty.register(ViewBase);
...
+
+        if (shouldAddModalRootViewCssClass) {
```
manoldonev: Why only check for root and modal css classes? Can't see the logic that ensures platform, device type, and orientation classes will be preserved if I only change `className` property?

tns-core-modules/ui/core/view-base/view-base.ts
```diff
@@ -1036,11 +1038,23 @@ bindingContextProperty.register(ViewBase);
 export const classNameProperty = new Property<ViewBase, string>({
     name: "className",
```
manoldonev: So as discussed offline with @vchimev and @bundyo currently there is a discrepancy between `className` and `cssClasses` -- if you try to get the `className` value it will not include any values that were added through `cssClasses` directly. One potential solution could be to update the `className` getter to return a concatenated string of `cssClasses` values.

manoldonev: We will log a separate issue for this as apparently we will need to expose public API for customizing property getter/setter.

tns-core-modules/css/system-classes.d.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * @module "system-classes"
+ */ /** */
+
+/**
+ * String value "ns-" used for CSS system class prefix.
+ */
+export const CSS_CLASS_PREFIX: string;
```
manoldonev: Should be CLASS_PREFIX I think

tns-core-modules/ui/core/view-base/view-base.ts
```diff
@@ -22,10 +25,9 @@ export { isIOS, isAndroid, layout, Color };
 export * from "../bindable";
 export * from "../properties";
 
+import * as dnm from "../../../debugger/dom-node";
```
manoldonev: Is there a reason for moving this import?

vchimev: Nope, just grouping this section of imports.

tns-core-modules/css/system-classes.ts
```diff
@@ -0,0 +1,21 @@
+const MODAL = "modal";
+const ROOT = "root";
+
+export const CLASS_PREFIX = "ns-";
+
+const modalRootViewCssClass = `${CLASS_PREFIX}${MODAL}`;
+const rootViewCssClasses = [`${CLASS_PREFIX}${ROOT}`];
```
manoldonev: I think this field will get out-of-sync in orientation change scenarios. Currently the `orientationChanged` handler will properly update the orientation css class directly on the root element but this field will hold the "old" orientation value.

Issue #7713, opened 2019-08-22
[BottomNav] Clipped tabstripitem title content when iconSource specified


manoldonev:
**Environment**
 - CLI: 6.0.3
 - Cross-platform modules: 6.0.6-next
 - Android Runtime: 6.0.1

**Describe the bug**
TabStripItem title gets clipped when both iconSource and title are specified.

**To Reproduce**
Erroneous behavior can be observed in e2e/ui-tests-app -> bottom-navigation -> icon-change example

**Expected behavior**
TabStripItem title should not be clipped in any scenario.

**Additional context**
![Screenshot_1566317899](https://user-images.githubusercontent.com/2650247/63510141-825a7800-c4e6-11e9-8309-e2e1d4e37dce.png)

PR #7737, opened 2019-08-27
feat(android): fit tab resource icon size based on spec

MartoYankov:
In order to have the fit to spec tab icons, we need to remove this optimization.

Fixes https://github.com/NativeScript/NativeScript/issues/7713


MartoYankov:
test nativescript-dev-webpack#next

Issue #7692, opened 2019-08-19
[Tabs][BottomNavigation] Crash with TabStrip configured in code-behind


tsonevn:
**Environment**
 - CLI: 6.0.3
 - Cross-platform modules: 6.1.0-next-2019-08-18-231654-03
 - iOS Runtime: 6.1.0-2019-08-15-135151-01

The problem is related to incorrect handling of a scenario where TabStrip and TabStripItems are created dynamically in code behind. The issue is reproducible only on iOS. Error:
```
CONSOLE INFO file:///node_modules/nativescript-dev-webpack/hot.js:3:0 HMR: Hot Module Replacement Enabled. Waiting for signal.
CONSOLE LOG file:///app/ns-ui-widgets-category/tabs/properties/properties-page.js:96:0 >>> Loaded [Content 1]
CONSOLE LOG file:///app/ns-ui-widgets-category/tabs/properties/properties-page.js:96:0 >>> Loaded [Content 2]
CONSOLE LOG file:///app/ns-ui-widgets-category/tabs/properties/properties-page.js:96:0 >>> Loaded [Content 3]
CONSOLE LOG file:///app/ns-ui-widgets-category/tabs/properties/properties-page.js:41:0 tabs.nativeView:  <_UIPageViewControllerContentView: 0x7fb6e7647bc0; frame = (0 0; 375 812); clipsToBounds = YES; opaque = NO; autoresize = W+H; layer = <CALayer: 0x6000019bbd40>>
***** Fatal JavaScript exception - application has been terminated. *****
Native stack trace:
1   0x1015e362a NativeScript::reportFatalErrorBeforeShutdown(JSC::ExecState*, JSC::Exception*, bool)
2   0x10161d108 NativeScript::FFICallback<NativeScript::ObjCMethodCallback>::ffiClosureCallback(ffi_cif*, void*, void**, void*)
3   0x101fc3a86 ffi_closure_unix64_inner
4   0x101fc44aa ffi_closure_unix64
5   0x10558bd22 -[CALayer layoutSublayers]
6   0x1055909fc CA::Layer::layout_if_needed(CA::Transaction*)
7   0x10559cd58 CA::Layer::layout_and_display_if_needed(CA::Transaction*)
8   0x10550c24a CA::Context::commit_transaction(CA::Transaction*)
9   0x105543606 CA::Transaction::commit()
10  0x1062ea99c _afterCACommitHandler
[...]

PR #7743, opened 2019-08-28
fix(ios-tabs): crash when add tabstrip in loaded event

MartoYankov:
Fixes https://github.com/NativeScript/NativeScript/issues/7692


MartoYankov:
test package_version#next

PR #7789, opened 2019-09-05
fix(css-android): restore elements' native features if offending css is removed while the app is running

### LINKED ISSUE #7788: Buttons not getting ripple back if border-radius is set back to 0 with additional CSS
**Environment**
 - Android Runtime: 6.1.0

**Describe the bug**
If you set border-width, border-radius, background-image or clip-path, the Button ripple is gone, and can be got back by removing them later. However, if the CSS is added on the fly, the ripple is not restored. Something like this:

```css
Button {
    border-radius: 1;
	background-color: #ccc;
}
```
Then add this CSS later:
```css
.something Button {
    border-radius: 0;
}
```

**To Reproduce**
https://play.nativescript.org/?template=play-js&id=vjUfza&v=2

**Expected behavior**
The ripple to be restored

bundyo:
(e.g. border-width > 0, border-radius > 0, background-image, etc.)

## PR Checklist

When the user sets border-width, border-radius or others, the native features of the control are removed, but can't be restored when adding additional CSS later.

## What is the new behavior?
The native features are restored by just applying the filter on top of a new drawable.

Fixes/Implements/Closes #7788.


tns-core-modules/ui/styling/background.android.ts
```diff
@@ -56,6 +56,16 @@ export module ad {
             && !background.clipPath
             && !background.image
             && background.color) {
+
+            if (drawable instanceof org.nativescript.widgets.BorderDrawable) {
```
manoldonev: It seems this condition can never be satisfied as on line 53 there is the opposite check?

bundyo: Oops, forgot to copy that change over.

Issue #7785, opened 2019-09-05
ListPicker not working on API 29


vmutafov:
Hi,
It seems that the `ListPicker` is not working on API 29. The issue seems to be because of using a Java field through reflection which is 'hidden' in the API 29 SDK and causes a `NoSuchFieldException`: https://github.com/NativeScript/NativeScript/blob/4f39fb728b6bcacdafed38dd0a80137fcc9fb1e9/tns-core-modules/ui/list-picker/list-picker.android.ts#L67

vmutafov:
Is this `selectorWheelPaintField` field actually used somewhere?

manoldonev:
It is used by the property system https://github.com/NativeScript/NativeScript/blob/4f39fb728b6bcacdafed38dd0a80137fcc9fb1e9/tns-core-modules/ui/list-picker/list-picker.android.ts#L160 and https://github.com/NativeScript/NativeScript/blob/4f39fb728b6bcacdafed38dd0a80137fcc9fb1e9/tns-core-modules/ui/list-picker/list-picker.android.ts#L174

manoldonev:
@vmutafov  I was able to reproduce the exception on api29 but only with tns-android@6.1.0. With tns-android@6.0.2 the same app works as expected on api29 — do you have any idea what causes the different behavior from runtime’s perspective?

vmutafov:
Yeah, it's used, my bad I didn't see it. As for the question about the runtime version, 6.1.0 contains fixes in the runtime for using target and compile sdk 29. In the older runtimes, there was an issue when targeting sdk 29. I suppose you've tried it on an API 29 emulator but with an app built with target/compile sdk 29. Correct me if me presumption is wrong.

pclokcer:
reduce app.gradle like :

`android {
  compileSdkVersion 29
  buildToolsVersion "29.0.2"
  defaultConfig {
    minSdkVersion 17
    generatedDensities = []
  }
  aaptOptions {
    additionalParameters "--no-version-vectors"
  }
}
`

and reduce to manifest like : 

android:versionCode="7"
android:versionName="505357"

and solved

PR #7790, opened 2019-09-05
fix(android-list-picker): NoSuchFieldException on api29

manoldonev:
Note that Android ListPicker.color property had some weird get/set arguments and return values that I have now removed because as far as I can see:
- These were never documented as public API, could not find any sample usage, or references anywhere
- Setting different values for wheelColor and textColor worked only on initial load of the compoent, once you started scrolling textColor was lost and the input text was painted with wheelColor
- Now that api29 introduces proper public methods for manipulating this functionality, it exposed a single NumberPicker.setTextColor(...) native method that sets both wheelColor and textColor (in the older terms).

Fixes: https://github.com/NativeScript/NativeScript/issues/7785


dtopuzov:
test package_version#latest

PR #7804, opened 2019-09-10
fix(button-android): use setMinWidth to set TextView widgets min-width

### LINKED ISSUE #7803: Android Buttons have min-width that can't be removed
**Environment**
 - Cross-platform modules: 6.1.0

**Describe the bug**
If an Android Button is not stretched and has short text, there seems to be some min-width set that can't be seen or removed from CSS or element properties.

**To Reproduce**
Create a Button, align it left and set its text to "Go".

**Expected behavior**
The Button to be as big as the text.

**Sample project**
https://play.nativescript.org/?template=play-vue&id=jrAllC&v=2

bundyo:
## PR Checklist

Aligned Buttons with short text have a min-width that can't be removed.

## What is the new behavior?
Removes the min-width. According to Android docs:

> The value given with TextView.setMinWidth(int) is different than View.setMinimumWidth(int). Between setMinWidth and the value set in View.setMinimumWidth(int), the greater one is used to decide the final width.
	

Thus for TextViews as the Button, both should be set.

https://developer.android.com/reference/android/widget/TextView.html#setMinWidth(int)

Fixes/Implements/Closes #7803.


vakrilov:
There seems to be a behavior change with this PR. The default min-width of the android buttons is no longer present. 
Here is a screenshot of one fo the test app to get the idea:

Before:
dialogs_11_alert_expected

After:
dialogs_11_alert_actual

This has a potential of breaking client apps. @bundyo do you think there is a way to make this back-compatible?

bundyo:
We might introduce a default min-width for the property?

vakrilov:
test

vakrilov:
test

vtrifonov:
test

vtrifonov:
test

PR #7805, opened 2019-09-10
feat(ActionBar): add Android inset ActionBar property to control the titleView content insets

### LINKED ISSUE #6133: custom ActionBar with GridLayout has a small gap on the left
Yes, there's issue #4843 

### Tell us about the problem
There's a gap in a custom ActionBar on the left (GridLayout inside ActionBar), which can't be closed, even when margin/padding is 0 

Android
(Might be also iOS, haven't tested yet)

- CLI: 4.1.0
- Cross-platform modules: 4.1.1
- Runtime(s): 4.0.0
- Plugin(s): version number of `tns-android` is 4.1.3

The best approach would be to get your code running in the [NativeScript Playground](https://play.nativescript.org) and share the link with us, along with any additional details or steps to reproduce needed for examining the issue there. 

1. Create a new angular project using a blank template: `tns create myActionBarIssue --ng`
2. On the path `app/item/items.component.html`:

Replace the:

```
<ActionBar title="My App" class="action-bar">
</ActionBar>
```

With:

```
<ActionBar title="My App" class="action-bar">
    <GridLayout rows="*" columns="*" backgroundColor="green">
        <Label text="Test"></Label>        
    </GridLayout>
</ActionBar>
```

3. See that there's a small gap between the label and the edge on the left, even though there shouldn't be
4. Now add `marginLeft="-15"` to the `ActionBar` element like so:

```
<ActionBar title="My App" class="action-bar" marginLeft="-15">
    <GridLayout rows="*" columns="*" backgroundColor="green" >
        <Label text="Test"></Label>        
    </GridLayout>
</ActionBar>
```

5. See that the label is placed correctly more or less

No code is involved

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61442651-custom-actionbar-with-gridlayout-has-a-small-gap-on-the-left?utm_campaign=plugin&utm_content=tracker%2F12908224&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F12908224&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bundyo:
## PR Checklist

When a custom titleView is assigned to the ActionBar in Android, its contents are offset with some space on both sides, despite that the padding is set to 0.

## What is the new behavior?
The user can set inset property to "leftInset rightInset" in order to control this ghost padding in Android.

Fixes/Implements/Closes #6133.


tns-core-modules/ui/action-bar/action-bar.ios.ts
```diff
@@ -425,6 +426,8 @@ export class ActionBar extends ActionBarBase {
         }
     }
 
+    [insetProperty.setNative](value: string) {}
```
manoldonev: As the property is Android only, no changes should be required to `action-bar.ios.ts`

tns-core-modules/ui/action-bar/action-bar-common.ts
```diff
@@ -365,3 +365,6 @@ visibilityProperty.register(ActionItemBase);
 
 export const flatProperty = new Property<ActionBarBase, boolean>({ name: "flat", defaultValue: false, valueConverter: booleanConverter });
 flatProperty.register(ActionBarBase);
+
+export const insetProperty = new Property<ActionBarBase, boolean>({ name: "inset", defaultValue: false });
```
manoldonev: Several comments on this one:
- property is declared as boolean (`Property<ActionBarBase, boolean>`) but is actually a string
- property is Android only; while historically we did not always named those consistently my suggestion is to rename it with `android` prefix
- property should be exposed in action-bar.d.ts

tns-core-modules/ui/action-bar/action-bar.android.ts
```diff
@@ -421,6 +421,12 @@ export class ActionBar extends ActionBarBase {
             }
         }
     }
+
+    [insetProperty.setNative](value: string) {
+        const { 0: left, 1: right } = value.split(" ");
```
manoldonev: This implementation does not feel correct and seems a bit restrictive to me; also does not follow the same principle as css shorthand properties e.g. if I specify inset as "7, 11", it will break the logic.

My suggestion is to refactor this to either of the following two approaches:
- Remove `insetProperty` and introduce `androidContentInsetLeftProperty` & `androidContentInsetRightProperty` (properties are declared as `Property` of type `Length`); this is simple but requires the developer to set two properties
- Rework `insetProperty` to `androidContentInsetProperty` declared as `ShorthandProperty<Style, string | Length>`; also introduce `androidContentInsetLeftProperty` & `androidContentInsetRightProperty` (declared as `CssProperty` of type `Length` (check margin & padding implementation in `ui/styling/style-properties.ts`); this would allow you to set the insets either through the short-hand property as your original implementation but follow the same principle as existing css shorthand properties -- "drawback" would be that those properties should be registered on `Style` class so in a way they will be polluting it but in reality we have already crossed that line many times so I think it's ok.

I would go for option 2 -- let me know what do you think.

manoldonev:
Also, currently CI reports a tslint error to fix:
```
npm run tslint
> NativeScript@ tslint /home/travis/build/NativeScript/NativeScript
> tslint --project tsconfig.tslint.json --config build/tslint.json
ERROR: /home/travis/build/NativeScript/NativeScript/tns-core-modules/ui/action-bar/action-bar.ios.ts:429:46 - block is empty
```

--> `npm run tslint` :)

tns-core-modules/ui/action-bar/action-bar.android.ts
bundyo: There is a problem with option 2 - in order to do that, ActionBarBase should inherit Style in order to match its type.

Issue #7745, opened 2019-08-28
Update DevelopmentWorkflow.md for v6


rynop:
I'm trying to test a [contribution](https://github.com/NativeScript/NativeScript/pull/7707) to the v6 `tns-core-modules` and `tns-core-modules-widgets`, and running into problems following the [DevelopmentWorkflow.md](https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md).  I think it is outdated.

In short, I want to modify a `.ts` file in `tns-core-modules` and be able to test the modification in a v6 NativeScript app in another directory.

A few issues with the current [DevelopmentWorkflow.md](https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md):
1.  The initial setup (`npm install`) does not npm link
1.  [Running Another App](https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#running-another-app) indicates changes require TS rebuild (tsc), however it does not specify which package.json `script` target to run.  My hunch is it is `dev-tsc-all` but this script has a bug in it (pr below) AND after running, the `.js` files end up in `bin/dist/tns-core-modules/` without a `package.json` (so `npm link` will not work).  `compile-all` script does the same.
1.  `dev-link-tns-core-modules-widgets` script references a `dist/package` dir that DNE.

I'm happy to do the work to PR an update to the docs, however I need some clarity on the right way to do it 1st.

Here is my workflow - what am I doing wrong?

1.  Fork and clone
1.  `git checkout -b my-feature/fix-branch`
1.  `npm install`
1.  `npm run compile-all` 
1.  From root of my NS proj: `npm link ../<where I cloned>/tns-core-modules-widgets/ && npm link ../<where I cloned>/tns-core-modules`
1.  `tns debug ios --debug-brk --env.sourceMap`

NickIliev:
@rynop the developer workflow document is now being updated via [this PR](https://github.com/NativeScript/NativeScript/pull/7750/files) - can you try the simplified steps and let us know if that is works for you.

rynop:
Unfortunately, that did not work either. Reproduction steps:

```
$ node --version
v10.16.0
$tns --version
6.0.3
$ git clone https://github.com/NativeScript/NativeScript
$ tns create v6test --ng --appid com.rynop.v6test
$ cd v6test
$ npm install --save ../NativeScript/tns-core-modules
+ tns-core-modules@6.2.0
[...]

PR #7861, opened 2019-09-24
chore: update dev workflow for Angular app

rynop:
## PR Checklist

When trying to contribute to `tns-core*`, and using an external NS Angular to develop/test these contributions, the compile step fails when following the instructions currently outlined in [DevelopmentWorkflow.md](https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#running-another-app) (Running Another App).

@bundyo [discovered](https://github.com/NativeScript/nativescript-dev-webpack/issues/1044#issuecomment-533397897) this workaround, I'll paste his findings here:

> The problem with linking is that Angular/TS project tend to grab the linked TS files in core modules and try to compile them without the references needed. I've only managed to run it by compiling the core modules with outDir and linking that (.d.ts, package.json and other files, except TS ones should be copied over too). Maybe the team can consider moving to an outDir compilation. Another way would be dev-webpack to forbid Angular/ts-loader for traversing node_modules.

## What is the new behavior?

This PR contains a workaround for Angular that moves `tns-core*` to a new `dist` directory.

This workflow is obviously not ideal.  Would love the NativeScript team's thoughts on a cleaner solution.

Fixes https://github.com/NativeScript/nativescript-dev-webpack/issues/1044 and https://github.com/NativeScript/NativeScript/issues/7745

### DIFF
--- a/DevelopmentWorkflow.md
+++ b/DevelopmentWorkflow.md
@@ -81,6 +81,47 @@ tns run ios
 tns run android
 ```
 
+### Running Another Angular App
+
+When linking within a Angular/TS project, `tns run` will grab the linked TS files in core modules and try to compile them without the references needed.  To work around this, from the root of your `NativeScript` clone do the following:
+
+1. Create a copy of `tns-core-modules` folder via:
+```bash
+mkdir dist && cp -r tns-core-modules dist/
+# Should look like this:
+NativeScript
+     |
+     +--> dist
+            |
+            +--> tns-core-modules
+                        |
+                        +--> ...
+                        +--> ...
+                        +--> ...
+```
+2. Delete all `.ts` files, keeping `d.ts` ones via:
+```bash
+find dist/tns-core-modules -name "*.ts" -not -name "*.d.ts" -delete
+```
+3. Add `--outDir dist` to the `tsc-w` script in `package.json`
+```
+"scripts": {
+...
+"tsc-w": "npm run tsc -- --skipLibCheck -w --outDir dist",
+...
+```
+4. Run `npm run tsc-w`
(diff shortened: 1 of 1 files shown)


MartoYankov:
While this works it requires lots of manual work, which is not ideal. Imho, it should work the same for Angular and TS project. Here are my thoughts:

1. The Angular compiler seems to ignore or fail to resolve the references.d.ts in the tns-core-modules folder. If you add `tns-platform-declarations` and a `references.d.ts` to your app, the required references will be there. We can add this to the steps.
2. It seems that even when you do 1. the Angular compiler still doesn't like the typescript in the core modules. We should perhaps investigate these errors. They seem to be connected to the interface exports in the d.ts files.

bundyo:
I'll make some tests where the problem is, since it is happening only in Angular (though I might have seen it in Vue+TS) and then we can address it in dev-webpack or, if not possible - create an automated script to deploy to another folder.

MartoYankov:
@bundyo I followed the steps from "running another app" plus one additional step - adding the `tns-platform-declarations` packages to the app. The errors that the ng compiler producers are with types and exports. I suspect these still happen in the regular `ts-loader`, but as warnings. The ng compiler is producing them as errors.

The reason why we don't want to promote alternative dev setups is because we can't support them. We adhere to the dev setup in the nativescript plugin seed to make sure the `tns-core-modules` have a similar setup to all plugins. What I mean is, we should fix the current setup or if not possible - change the dev setup in the plugin seed first.

manoldonev:
@rynop as elaborated by @MartoYankov I have logged a separate issue https://github.com/NativeScript/NativeScript/issues/7905 that we will need to address to fix the dev workflow in this scenario. I will close this PR for the time being.

PR #7880, opened 2019-09-30
feat: adding support for the new selectedSegmentTintColor property

tsonevn:
Adding support for the new SegmentedControl's selectedSegmentTintColor property. Introduced with iOS 13.

### DIFF
--- a/tns-core-modules/ui/segmented-bar/segmented-bar.ios.ts
+++ b/tns-core-modules/ui/segmented-bar/segmented-bar.ios.ts
@@ -3,6 +3,7 @@ import {
     SegmentedBarItemBase, SegmentedBarBase, selectedIndexProperty, itemsProperty, selectedBackgroundColorProperty,
     colorProperty, fontInternalProperty, Color
 } from "./segmented-bar-common";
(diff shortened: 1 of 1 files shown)


tns-core-modules/ui/segmented-bar/segmented-bar.ios.ts
```diff
@@ -68,11 +67,19 @@ export class SegmentedBar extends SegmentedBarBase {
     }
 
     [selectedBackgroundColorProperty.getDefault](): UIColor {
-        return this.ios.tintColor;
+        const currentOsVersion = parseFloat(UIDevice.currentDevice.systemVersion);
```
MartoYankov: We have an internal way of getting the major version. It's better to use it. See - https://github.com/NativeScript/NativeScript/blob/4f39fb728b6bcacdafed38dd0a80137fcc9fb1e9/tns-core-modules/ui/search-bar/search-bar.ios.ts#L10

MartoYankov:
@tsonevn I think this can be released as a patch.

SvetoslavTsenov:
@cla-bot check

PR #7896, opened 2019-10-02
feat(segmented-bar-ios): add support for new selectedSegmentTintColor

manoldonev:
Cherry-picked from https://github.com/NativeScript/NativeScript/pull/7880


manoldonev:
Need iOS13 typings, will release with 6.2.0

PR #7907, opened 2019-10-05
feat(css): add attribute scoped css without global refresh

joshcomley:
As per the problem I reported here:
https://github.com/NativeScript/theme/issues/179

When using component scoped CSS there is no need to trigger a full global refresh of the styles.

This change does *not* affect performance or function of using `encapsulation: ViewEncapsulation.None` on a component.

This PR goes hand in hand with a [modification to `nativescript-angular`](https://github.com/NativeScript/nativescript-angular/pull/1999) to employ the below changes (PR coming up right after this).

## PR Checklist

## What is the new behavior?

### DIFF
--- a/tns-core-modules/application/application.android.ts
+++ b/tns-core-modules/application/application.android.ts
@@ -177,11 +177,13 @@ export function run(entry?: NavigationEntry | string) {
     _start(entry);
 }
 
-export function addCss(cssText: string): void {
+export function addCss(cssText: string, scoped?: boolean): void {
     notify(<CssChangedEventData>{ eventName: "cssChanged", object: androidApp, cssText: cssText });
-    const rootView = getRootView();
-    if (rootView) {
-        rootView._onCssStateChange();
+    if(!scoped) {
+        const rootView = getRootView();
+        if (rootView) {
+            rootView._onCssStateChange();
+        }
     }
 }
 
--- a/tns-core-modules/application/application.android.ts
+++ b/tns-core-modules/application/application.android.ts
@@ -177,9 +177,9 @@ export function run(entry?: NavigationEntry | string) {
     _start(entry);
 }
 
-export function addCss(cssText: string, scoped?: boolean): void {
+export function addCss(cssText: string, attributeScoped?: boolean): void {
     notify(<CssChangedEventData>{ eventName: "cssChanged", object: androidApp, cssText: cssText });
-    if(!scoped) {
+    if (!attributeScoped) {
         const rootView = getRootView();
         if (rootView) {
             rootView._onCssStateChange();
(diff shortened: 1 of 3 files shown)


joshcomley:
I am happy to resubmit this PR following the above guidelines (I am not entirely sure how to compose the PR title, for example), but would somebody be able to review the code to ensure it's a valid PR otherwise? Many thanks

tns-core-modules/application/application.android.ts
```diff
@@ -177,11 +177,13 @@ export function run(entry?: NavigationEntry | string) {
     _start(entry);
 }
 
-export function addCss(cssText: string): void {
+export function addCss(cssText: string, scoped?: boolean): void {
     notify(<CssChangedEventData>{ eventName: "cssChanged", object: androidApp, cssText: cssText });
-    const rootView = getRootView();
-    if (rootView) {
-        rootView._onCssStateChange();
+    if(!scoped) {
```
charsleysa: Missing space between `if` and `(`

charsleysa:
For PR title, follow this guideline: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages

You could would end up with something like `feat(css): add scoped css without global refresh`

tns-core-modules/application/application.android.ts
```diff
@@ -177,11 +177,13 @@ export function run(entry?: NavigationEntry | string) {
     _start(entry);
 }
 
-export function addCss(cssText: string): void {
+export function addCss(cssText: string, scoped?: boolean): void {
```
MartoYankov:
```suggestion
export function addCss(cssText: string, attributeScoped?: boolean): void {
```

This implementation is needed only when the css is global, but scoped by attributes (like in Angular and Vue). To differentiate this from the {N} Core implementation with style scope objects I think it's a better naming.

tns-core-modules/application/application.ios.ts
```diff
@@ -371,11 +371,13 @@ export function run(entry?: string | NavigationEntry) {
     _start(entry);
 }
 
-export function addCss(cssText: string): void {
+export function addCss(cssText: string, scoped?: boolean): void {
```
MartoYankov:
```suggestion
export function addCss(cssText: string, attributeScoped?: boolean): void {
```

tns-core-modules/application/application.android.ts
MartoYankov: You should add this new parameter to the `d.t.s` too explaining it can be used when you add global css that is scoped by attribute for example.

joshcomley:
Hi there,

Thanks for the suggestions, I've implemented these and updated the PR.

MartoYankov:
test --ignore

MartoYankov:
test --ignore

Issue #7800, opened 2019-09-10
[Tracking] Dark Mode Support


vchimev:
## Overview

Android 10 and iOS 13 introduce system-wide dark appearance for all user interface elements, known as Dark theme for Android and Dark Mode for iOS. They allow users to toggle their interface and focus on the application’s content based on ambient lighting conditions.

Dark theme for Android:
https://developer.android.com/guide/topics/ui/look-and-feel/darktheme

Dark Mode for iOS:
https://developer.apple.com/documentation/appkit/supporting_dark_mode_in_your_interface

## Design

A summary for the Dark Mode support in NativeScript can be found here.

TODO: Add a link. 

## Contribution

If you want to work on a task from the list, tell us and we will do our best to help you!
If you have another great idea on how to make the Dark Mode support even cooler, share it in the comments!

## Discussion

Please, don't report problems here. Instead, open a new issue and link it to this one.
Let's use this discussion for suggestions and improvement ideas. We would love to hear from you!

roi-kol:
I saw that the if the android 10, when it an dark mode all the color are negative... how i force light mode?

Skillnter:
How to change icons/resources for theme change...dark icons will not show on the dark theme or vice versa

PR #7946, opened 2019-10-11
[WIP] fix(css-classes): increase application CSS selector version at runtime

vchimev:
## PR Checklist

On a runtime change of orientation or system appearance,
the respective root view CSS class updates to reflect the changes.

At the same time, increase the application CSS selector version,
so that to update styles on back navigation.

The check for back navigation is in `onNavigatingTo` method:
https://github.com/NativeScript/NativeScript/blob/6.1.0/tns-core-modules/ui/page/page-common.ts#L108-L113

Relates to https://github.com/NativeScript/NativeScript/issues/7800.


vchimev:
test --ignore

tns-core-modules/ui/styling/style-scope.ts
```diff
@@ -760,6 +760,10 @@ export class StyleScope {
         return this._getSelectorsVersion();
     }
 
+    public increaseApplicationCssSelectorVersion(): number {
```
manoldonev: Now that we are introducing this method I think we need to update the `applicationCssSelectorVersion++;` usage in `mergeCssSelectors` as well. This will ensure that if we modify some aspect of the "increase css selector version" step in the future, it will be applied to all relevant places.

tns-core-modules/application/application-common.ts
```diff
@@ -147,6 +147,7 @@ export function orientationChanged(rootView: View, newOrientation: "portrait" |
     if (!rootView.cssClasses.has(newOrientationCssClass)) {
         ORIENTATION_CSS_CLASSES.forEach(cssClass => removeCssClass(rootView, cssClass));
         applyCssClass(rootView, newOrientationCssClass);
+        rootView._styleScope.increaseApplicationCssSelectorVersion();
```
manoldonev: Thinking out loud -- calling out this api here directly does not feel right. At first glance (yup, I know it's probably not that simple) it seems like it should be part of the `rootView._onCssStateChange()` implementation. Also, on the other hand, there are other `rootView._onCssStateChange()` calls in platform-specific application files -- should we increase the application css selector version there as well?

tns-core-modules/ui/styling/style-scope.ts
manoldonev: Also, I think this method should be public private e.g. with underscore `_increaseApplicationCssSelectorVersion` -- in this way we can remove it in the future if we are able to refactor the codebase (the other comment).

vchimev:
`test --ignore` so not to overload the CI as my current changes do not affect existing functionality.

tns-core-modules/application/application-common.ts
vchimev: We shouldn't increase the application CSS selector version explicitly anywhere else because:
- `_onCssStateChange` in the platform-specific application files executes after loading new styles where the version will increase,
- `_onCssStateChange` in the common application file executes on `livesync` where application styles load on each sync.

vchimev: Furthermore, should root view CSS orientation/system appearance class change notify about `cssChanged`? In this scenario, the application does not load new styles, it changes selectors.

tns-core-modules/ui/styling/style-scope.ts
vchimev: I agree - for this purpose `_increaseApplicationCssSelectorVersion` needs to be static.

vchimev:
I'm closing this pull request in favor of https://github.com/NativeScript/NativeScript/pull/7992.

PR #7947, opened 2019-10-11
feat(css): add attribute scoped css without global refresh (#7907)

VladimirAmiorkov:
This is a patch PR for https://github.com/NativeScript/NativeScript/pull/7907

### DIFF
--- a/tns-core-modules/application/application.android.ts
+++ b/tns-core-modules/application/application.android.ts
@@ -177,11 +177,13 @@ export function run(entry?: NavigationEntry | string) {
     _start(entry);
 }
 
-export function addCss(cssText: string): void {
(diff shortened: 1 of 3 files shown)


MartoYankov:
This PR is basically adding a new feature. I'm not sure this is good for a patch. Also, I've communicated with the guy who initially reported the issue and helped me isolate it that it will come out with {N} 6.2 at the end of the month.

manoldonev:
@MartoYankov Generally I agree with you but as this change is triggered by a nativescript-angular fix, we need to release it with a patch in tns-core-modules as with 6.2.0 we'll be releasing a patch version of nativescript-angular (nativescript-angular versioning is tied to angular major and minor releases) that will rely on this fix and otherwise it will break all users.

PR #7955, opened 2019-10-13
fix(android): BottomNavigation TabStripItem compat with prettier

NathanWalker:

Using `.prettierrc` for auto codebase formatting will often lowercase `Image` and `Label` component tag names and hence break `BottomNavigation`'s `TabStripItem`.

## What is the new behavior?

It now works with both cases (with or without prettier).

## Notes

Caught me by surprise. Was working in codebase that used prettier to format entire codebase including .html templates and BottomNavigation was just not showing anything in the TabStrip. Was related to this.

### DIFF
--- a/tns-core-modules/ui/tab-navigation-base/tab-strip-item/tab-strip-item.ts
+++ b/tns-core-modules/ui/tab-navigation-base/tab-strip-item/tab-strip-item.ts
@@ -172,14 +172,15 @@ export class TabStripItem extends View implements TabStripItemDefinition, AddChi
     }
 
     public _addChildFromBuilder(name: string, value: any): void {
-        if (name === "Image") {
+        // ensure compat with .prettier formatters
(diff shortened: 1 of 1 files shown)


MartoYankov:
Hey @NathanWalker, just out of curiosity - why would prettier format `image` and `label` to lower case, but not `TabStrip` and `TabStripItem` for example? I see the same problem there and even with the old `TabViewItem`, which is probably the reason why there is a problem with the new ones.

I don't know why we check for the name instead of checking for the type like we do in all other cases. I will look into it.

NathanWalker:
Lowercase component names usually aren't an issue but I think the key issue here is that the builder was setup to look at literal exact case match whereas using a type check like you all do other places may be best overall.

MartoYankov:
@NathanWalker Closing this PR in favor of https://github.com/NativeScript/NativeScript/pull/8016

Issue #4201, opened 2017-05-16
backgroundPosition with percent/number is not working


sis0k0:
Let's say you have a background image, and you try to apply backgroundPosition through css. The following is working:
```
background-position: center;
```

However, neither of these is:
```
background-position: 50%;
background-position: 50;
background-position: 0.5;
```

I'm attaching a stripped version of [groceries](https://github.com/NativeScript/NativeScript/files/1003727/groceries.tar.gz). The `app/groceries/groceries.component.html` contains a single GridLayout with a backgroundImage and class - list-page. The list-page class is defined in `app.css`.

Related to https://github.com/NativeScript/nativescript-dev-webpack/issues/135

airandfingers:
This works for me now, but only if I specify it as `background-position: 14 8;`.

Values that didn't work: `left 14` and `left 14 top 8`.

tsonevn:
The issue is still reproducible with NativeScript 4.1.
[Archive.zip](https://github.com/NativeScript/NativeScript/files/2236095/Archive.zip)

abozanona:
@sis0k0 Did you find any solution for this problem?

mahmoudajawad:
I came across this issue when I tried using `background-position-y`, but then realised I can stick to using `background-position` with two values. However, I also tried the OP brought-up bug and can confirm the merged PR resolves this issue. I guess it can be closed.

janoshrubos:
Closing this issue since the implementation is merged to master. 
If this error is still relevant, feel free to open a new issue!

PR #7958, opened 2019-10-14
feat(css-bkg-pos): Added possibility to declare background pos by single numeric value 

surdu:
## PR Checklist

You can't declare a background-position attribute in CSS using only one numeric value:

```css
background-position: 50%;
background-position: 50;
background-position: 0.5;
```

## What is the new behavior?
Now you can declare `background-position` as above mentioned, and additionally in the following format:

```css
background-position: <number|percent> bottom|center;
```
 

Implements  #4201


surdu:
I'm afraid I'm going to need a little bit of help with the tests for this PR: what I need to do is see if the drawing of the background Canvas/UIImage was done correctly, and I can't figure out how to achieve this. Other options would be to spy on some of the methods that draw the background, but that would not be possible on Android, where all the action happens on Android side in `tns-core-module-widgets`.

Could anybody please point me in the right direction on how to write tests for this? I tested this manually and all works great.

tns-core-modules-widgets/android/widgets/src/main/java/org/nativescript/widgets/BorderDrawable.java
```diff
@@ -713,6 +713,18 @@ private BackgroundDrawParams getDrawParams(float width, float height) {
...
 
+                    if ("center".equals(vy.getString().toLowerCase(Locale.ENGLISH))) {
+                        res.posY = spaceY / 2;
+                    } else if ("bottom".equals(vy.getString().toLowerCase(Locale.ENGLISH))) {
+                        res.posY = spaceY;
+                    }
+                } else if ("number".equals(vx.getType()) && "ident".equals(vy.getType())) {
+                    if ("%".equals(vx.getUnit())) {
+                        res.posX = spaceX * vx.getValue() / 100;
+                    } else if ("number".equals(vx.getType()) || "px".equals(vx.getUnit()) || vx.getUnit() == null || vx.getUnit().isEmpty()) {
```
vmutafov: Do we need the check for `"number".equals(vx.getType())`, it is checked a couple of lines before?

tns-core-modules-widgets/android/widgets/src/main/java/org/nativescript/widgets/BorderDrawable.java
```diff
@@ -713,6 +713,18 @@ private BackgroundDrawParams getDrawParams(float width, float height) {
                         res.posX = spaceX;
                     }
 
+                    if ("center".equals(vy.getString().toLowerCase(Locale.ENGLISH))) {
```
vmutafov: Couldn't this logic be reused with the one below? This `if` introduces code duplication. Also, perhaps it's a good idea to extract pieces like `vy.getString().toLowerCase(Locale.ENGLISH)` into separate variables. This would increase readability, easier debugging and would not perform multiple method calls.

nativescript-core/ui/styling/background.ios.ts
```diff
@@ -322,6 +327,18 @@ function getDrawParams(this: void, image: UIImage, background: BackgroundDefinit
...
                     res.posY = spaceY;
                 }
+            } else if (v.x.type === "number" && v.y.type === "ident") {
+              if (v.x.unit === "%") {
+                res.posX = spaceX * v.x.value / 100;
+              } else if (v.x.type === "number" || v.x.unit === "px" || v.x.unit === "") {
+                res.posX = v.x.value;
+              }
+
+              if (v.y.string.toLowerCase() === "center") {
```
vmutafov: Same suggestion as in the Java code comment about reusing code and not duplicating.

vtrifonov:
test

tns-core-modules-widgets/android/widgets/src/main/java/org/nativescript/widgets/BorderDrawable.java
surdu: I'm afraid that without strong unit tests around this area, an important part of the parser could brake. I could write those unit tests, but as stated above, I would need some help with that 😞

vtrifonov:
test

vtrifonov:
test

vtrifonov:
test

PR #7987, opened 2019-10-21
fix(7986): nested css-calc and css-variables with fallback give incorrect result

### LINKED ISSUE #7986: nested css-calc and css-variables with fallback give incorrect result
**Environment**
 - CLI: 6.1.2
 - Cross-platform modules: 6.1.x (whatever playground is running)
 - Android Runtime: 6.1.2
 - iOS Runtime:
 - Plugin(s):

**Describe the bug**

Nested `css-calc` and `css-variables` with fallback value give incorrect result.

**To Reproduce**

Try this expression:
```css
.css-calc-nested {
  width: calc(calc(var(--undefined-var, 16) / 2) * 2)
}
```

This give the incorrect `width = 8`, but it should be `16`.

**Expected behavior**

I'd expect the result to be the equivalent of:
```
.css-calc-nested {
   width: calc(calc(16 / 2) * 2); // e.g. 16
}
```

**Sample project**

https://play.nativescript.org/?template=play-ng&id=JvqmiA&v=2

**Additional context**

This is caused by the regexp for the css-variables.
I'm working on a solution to the problem.

m-abs:
## PR Checklist

Nesting css-calc expressions with css-variable with fallback give the wrong result.

Assuming input value: `calc(calc(var(--undefined-var, 16) / 2) * 2)`.

1. `_evaluateCssVariableExpression` returns `calc(calc(16) / 2) * 2)` (note: the missing end-`)`)
2. `_evaluateCssCalcExpression` returns `8 * 2)`.
3. Then the value is cast to number with `parseFloat` which results in the number `8`.

The problem is caused by `cssVariableExpressionRegexp` finding the wrong `)`.

## What is the new behavior?

Fixes #7986.

1. No longer uses regexp to process `css-variable` expressions.
2. Switch to simple string operations e.g. `includes`, `indexOf` and `lastIndexOf`.
3. Process `css-variable` expressions is now done from the end of the string, before it was done from the start.
3b. This means the inner `css-variable` is a nested expression is processed first, but this was the simplest way to find the correct end-`)` of those expressions.


manoldonev:
test

manoldonev:
test

manoldonev:
test

PR #7992, opened 2019-10-21
fix-next(css-classes): increase application CSS selector version at runtime

vchimev:
On a runtime change of orientation or system appearance,
the respective root view CSS class updates to reflect the changes.

At the same time, increase the application CSS selector version,
so that to update styles on back navigation.

The check for back navigation is in `onNavigatingTo` method:
https://github.com/NativeScript/NativeScript/blob/6.1.0/tns-core-modules/ui/page/page-common.ts#L108-L113

Relates to #7800.

### DIFF
--- a/nativescript-core/application/application-common.ts
+++ b/nativescript-core/application/application-common.ts
@@ -142,6 +142,13 @@ function removeCssClass(rootView: View, cssClass: string) {
     rootView.cssClasses.delete(cssClass);
 }
 
+function increaseStyleScopeApplicationCssSelectorVersion(rootView: View) {
+    const styleScope = rootView._styleScope || (<any>rootView).currentPage._styleScope;
+    if (styleScope) {
+        styleScope._increaseApplicationCssSelectorVersion();
+    }
+}
+
 export function orientationChanged(rootView: View, newOrientation: "portrait" | "landscape" | "unknown"): void {
     if (!rootView) {
         return;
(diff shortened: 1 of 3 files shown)


nativescript-core/application/application-common.ts
```diff
@@ -142,6 +142,13 @@ function removeCssClass(rootView: View, cssClass: string) {
     rootView.cssClasses.delete(cssClass);
 }
 
+function increaseStyleScopeApplicationCssSelectorVersion(rootView: View) {
+    const styleScope = rootView._styleScope || (<any>rootView).currentPage._styleScope;
```
MartoYankov:
```suggestion
    const styleScope = rootView._styleScope || ((<any>rootView).currentPage && (<any>rootView).currentPage._styleScope);
```

MartoYankov: Also you can add comment about the second option - that it's needed for `application.start()`.

nativescript-core/ui/styling/style-scope.ts
```diff
@@ -760,6 +760,10 @@ export class StyleScope {
         return this._getSelectorsVersion();
     }
 
+    public _increaseApplicationCssSelectorVersion(): number {
```
MartoYankov: There is no need for this to have a return type or to return the value below.

Issue #7988, opened 2019-10-21
[IOS] unecessary layout steps on subview size change


farfromrefug:
{N} 6.1.2

Playground: 
https://play.nativescript.org/?template=play-vue&id=yQTS5W&v=3

In {N} if we change the width of a View a ```requestLayout``` is called. In the view-base implementation a ```requestLayout``` is made on the parent. So whenever you change a view, whatever the level, the whole page(if not frame) is always re-layout. Which means re-measure / re-draw.
This is very expensive, especially on complex layouts(even though it is not so complex) like this one.
![image](https://user-images.githubusercontent.com/655344/67204395-af0b0d00-f40d-11e9-8e9e-b12eb0ca9a6d.png)
In my case this makes the animation stutter without real need for it

You wont see that issue using Nativescript animation module because the animation is made natively and not through the property, so a ```requestLayout``` is not made on every step of the animation.
However you will see in the playground that i linked before. Simply enable ```Layout``` trace to see the layout called on the whole page on button size change.

So 2 things:
- why do need to animate the size and not simply a transformation? Because i want the buttons to always remain center during the whole animation
- how can we fix this?

Well the fix is actually straight forward. You only need to call the parent ```requestLayout``` if the parent has it's size depending on the child size. If not there is no need to to "propagate" the event.
I have not yet started to write the test for "fixed" size but it should be straight forward

farfromrefug:
Actually it is not as easy as. i thought it was. If i try to stop propagating the ```requestLayout``` then there is no update of layout at all!
And i think it is because ```setNeedsLayout``` is used which can only be "heard" from a viewController https://github.com/NativeScript/NativeScript/blob/cc97a1680009f1bf6dbf97c421f6e8dc535295b5/nativescript-core/ui/core/view/view.ios.ts#L748
I need to find a way to make any view request a layout and redo the measure/layout

farfromrefug:
So i got something partially working. The idea is this:
- create a custom ```UICustomLayoutView``` to be able to override ```layoutSubViews```
- in ```layoutSubViews``` call a ```forceReLayout``` on the attached {N} ```View```. The ```forceReLayout``` will measure and layout children
- use ```UICustomLayoutView``` as the nativeView of ```CustomLayoutView```
[...]

PR #7996, opened 2019-10-22
testing improvements on layout

farfromrefug:
Related to this issue https://github.com/NativeScript/NativeScript/issues/7988

For now this PR only checks for auto in width or height. This is clearly not enough.
But the idea is to check if it is necessary to requestLayout on the parent. Most of the times it is not necessary.

In the case of a complex app the improvements is huge!. It makes apps using animations (not using {N} animation system which is too limited) far more snappier.

This is only a POC!


vtrifonov:
test

vtrifonov:
You needed to run `npm run api-extractor` in the root folder which will regenerate the NaitiveScript.api.md file as there is a change in the public API

farfromrefug:
@vtrifonov will try and do that as soon as possible

farfromrefug:
@vtrifonov just pushed the update

nativescript-core/ui/core/view-base/view-base.d.ts
```diff
@@ -310,7 +310,7 @@ export abstract class ViewBase extends Observable {
     /**
      * Invalidates the layout of the view and triggers a new layout pass.
      */
-    public requestLayout(): void;
+    public requestLayout(calledFromParent?:boolean): void;
```
vtrifonov: The same here, maybe the parameter name should be **calledFromChild**
```suggestion
    public requestLayout(calledFromChild?:boolean): void;
```

nativescript-core/ui/core/view/view-common.ts
```diff
@@ -819,9 +819,9 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
         }
     }
 
-    public requestLayout(): void {
+    public requestLayout(calledFromChild): void {
```
vtrifonov: it should have a default value here as in the view-base
```suggestion
    public requestLayout(calledFromChild = false): void {
```

api-reports/NativeScript.api.md
```diff
@@ -2915,7 +2915,7 @@ export abstract class ViewBase extends Observable {
     public _removeViewCore(view: ViewBase): void;
     // (undocumented)
     _removeViewFromNativeVisualTree(view: ViewBase): void;
-    public requestLayout(): void;
+    public requestLayout(calledFromParent?:boolean): void;
```
vtrifonov: I guess this should be **calledFromChild** (as it is in the view-base.ts and view-common.ts files) but not **calledFromParent**
```suggestion
    public requestLayout(calledFromChild?:boolean): void;
```

farfromrefug: @vtrifonov actually already changed that here. Need to pus hit!

farfromrefug:
@vtrifonov as discussed with @vakrilov i will close that PR and open a new one as Draft PR.
This feature is a POC and needs a bit more work. We need to discuss this

PR #7998, opened 2019-10-22
chore(hook): Check for min required webpack version

vakrilov:
## PR Checklist

## What is the new behavior?
Add a hook that will throw an error if building Angular project without the required version fo `nativescript-dev-webpack` (1.3.0)

### DIFF
--- a/nativescript-core/cli-hooks/before-checkForChanges.js
+++ b/nativescript-core/cli-hooks/before-checkForChanges.js
@@ -1,13 +1,62 @@
 const semver = require("semver");
 
+const webpackPackageName = "nativescript-dev-webpack";
+
 module.exports = function ($staticConfig, hookArgs) {
     const cliVersion = semver.parse($staticConfig.version);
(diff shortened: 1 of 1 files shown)


nativescript-core/cli-hooks/before-checkForChanges.js
```diff
@@ -1,13 +1,62 @@
 const semver = require("semver");
```
DimitarTachev: We should not rely on a semver dependency in the app

vakrilov: The hook that was generated by the `nativescript-hooks` plugin is actually requiring a file *inside* the `@nativescript/core` package. Here is the code of the hook:
```
module.exports = require("@nativescript/core/cli-hooks/before-checkForChanges.js");
```

The `nativescript-core/cli-hooks/before-checkForChanges.js` is actually inside the `@nativescript/core` package which has a dependency to `semver`. So the `require` will work as expected.

PR #8003, opened 2019-10-22
refactor: android.R to const for perf reasons

manoldonev:
Kudos to @vakrilov for tracking this!


SvetoslavTsenov:
test --ignore uitests#${sha1} api19 api24

PR #8008, opened 2019-10-23
feat(css-classes): add system appearance CSS class to modal root view

vchimev:
Depends on https://github.com/NativeScript/NativeScript/pull/8005.


MartoYankov:
The fact that we are changing the API of the system classes module now makes me think we might want or need to change it in the future too. I suggest all exported methods start with underscore and document the module as internal use only.

vchimev:
I am reworking this and going to open a new pull request.

Issue #8094, opened 2019-11-13
[Android] Nested frame wrong background after parent navigation


ADjenkov:
Steps: 
1. Start nested-frame-navigation application
2. Go to "Page W/ Frame"
3. Go to "Navigate to some page" (any transition)
4. Back
5. Go to nested "Player One (flip transition)"

Expected: white background instead of:
![image](https://user-images.githubusercontent.com/5860190/68766198-1c821600-0627-11ea-8736-00bfb147fe5a.png)

PR #8095, opened 2019-11-13
fix(frame): nested frame wrong background after parent navigation

ADjenkov:
Fix https://github.com/NativeScript/NativeScript/issues/8094


manoldonev:
test --ignore api24

Issue #7374, opened 2019-06-19
Support requestAnimationFrame


edusperoni:
**Is your feature request related to a problem? Please describe.**
Currently, there isn't a way to call a function when a frame "ticks". This also makes using RxJS's `animationFrameScheduler` throw an error.

**Describe the solution you'd like**
Implement `requestAnimationFrame` and `cancelAnimationFrame`.

**Describe alternatives you've considered**
`setInterval(() => ..., 16);`

**Additional context**
I've already implemented it 😅 

https://play.nativescript.org/?template=play-ng&id=62Ge6S&v=3

I've tested on android, might need `zonedCallback`. Worked out of the box + using `animationFrameScheduler`.

I'd open a PR, but I'm not sure if this falls into `timer` or `fps-meter`, as I use `fps-meter/fps-native` and `timer` does not have a `common` implementation.

Edit: updated the demo for a case where a callback would be canceled by another callback and we'd end up with undefined in the for loop

tsonevn:
Hi @edusperoni,
I reviewed the attached project, however, I am not sure that  I am understanding, where the issue is. On that matter, can you provide some more info about the problem and what you are trying to achieve as a final result?

Thank you in advance for your cooperation.

edusperoni:
Hi @tsonevn, thanks for taking the time.

Say we're processing an event (like touch/drag) or manually animating some property. If we do processing every tick (setInterval with 0 or 1), we'd be wasting a lot of processing on properties that only change ever frame.

We could use setInterval 16 (60fps = 16-17ms per frame), but that could cause a weird issue where you animate twice in one frame or skip a frame.

That's why we use `requestAnimationFrame`, so the app will call the function on the next frame. RxJS has `animationFrameScheduler` so we can discard events that emit more than once a frame, for example.

My sample project didn't have any practical example for it, just the implementation and the output (it shows how long it took for the function to be called, 16 to 17ms each time, or more if there are frame skips)

Without the two functions I implemented, `animationFrameScheduler` will throw an error.

edusperoni:
I've updated the example with a better comparison between a timeout and a animationframe:

at 60 fps, the animation frame will emit 59-60 events. setTimeout will emit around 120 events with a 0 delay and around 55-56 with a 16ms delay.

https://play.nativescript.org/?template=play-ng&id=62Ge6S&v=4

alexziskind1:
[...]

PR #8112, opened 2019-11-18
feat: support requestAnimationFrame

edusperoni:
## PR Checklist

requestAnimationFrame is not polyfilled

## What is the new behavior?
requestAnimationFrame is polyfilled

Please note: https://medium.com/@paul_irish/requestanimationframe-scheduling-for-nerds-9c57f7438ef4

Implements #7374.


nativescript-core/fps-meter/animation-native.android.ts
```diff
@@ -0,0 +1,4 @@
+
+export function getTimeInFrameBase(): number {
+    return java.lang.System.nanoTime() / 1000000;
```
vakrilov: Is there a reason not to use the `import { time } from "../profiling";` for android also.
You can also use the `global.__time()` (which is the same thing). It should be more performant and accurate than  `java.lang.System.nanoTime() / 1000000`

vakrilov:
Thanks for the PR! Awesome work as always. I have a couple of small suggestions - but the implementation seems solid.

nativescript-core/fps-meter/animation-native.android.ts
edusperoni: I started with `time`, but Java uses a different base for `nanoTime`.

>The value returned represents nanoseconds since some fixed but arbitrary time (perhaps in the future, so values may be negative).

So if I use time, I'd get the system time in ms, but the time android sends to `doFrame(currentTimeMillis: number)` is based off `nanoTime`, which means the arguments won't be consistent.

This is similar to the web (`performance.now()`). We could just evaluate `time` in `doFrame` and pass that, but I believe we'd lose some accuracy.

Regarding the file location, we were unsure where to put it in the issue discussion. I'll move it to `nativescript-core/animation-frame`, as it does indeed make more sense.

edusperoni:
Thanks! I refactored the code to move to its own module. Unfortunately I'm currently unable to test the iOS implementation regarding timestamps, the rest of the implementation was tested previously. I believe iOS uses the same timestamp for `time` and the `doFrame` callback, but I'm not 100% sure.

vakrilov:
test

vakrilov:
test

Issue #7365, opened 2019-06-13
Unable to debug with `--debug-brk` on iOS when linked `tns-core-modulеs`


vchimev:
**Describe the bug**

Unable to debug with `--debug-brk` on iOS when linked `tns-core-moduels`.

**To Reproduce**

```
git clone -b bundle-workflow https://github.com/NativeScript/NativeScript.git
cd NativeScript && npm install
cd ../tns-core-modules && npm install
cd ../tests && npm install
tns debug ios --debug-brk
```

**Additional context**

Error:
```
TypeError: _application_application_common__WEBPACK_IMPORTED_MODULE_7__["on"] is not a function. (In '_application_application_common__WEBPACK_IMPORTED_MODULE_7__["on"]("cssChanged", onCssChanged)', '_application_application_common__WEBPACK_IMPORTED_MODULE_7__["on"]' is undefined)
(anonymous) @ vendor.js:47943
../../tns-core-modules/ui/styling/style-scope.ts @ vendor.js:48381
__webpack_require__ @ bootstrap:74
exports @ vendor.js:56281
exports @ vendor.js:56266
(anonymous) @ app.ts:1
./app.ts @ bundle.js:125
__webpack_require__ @ bootstrap:74
checkDeferredModules @ bootstrap:43
webpackJsonpCallback @ bootstrap:30
anonymous @ bundle.js:2
evaluate @ [native code]:0
moduleEvaluation @ [native code]:0
promiseReactionJob @ [native code]:0
Show 4 more frames
```

vakrilov:
Looks like the problem is related to [`inspector_modules.ios.ts`](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/inspector_modules.ios.ts) being loaded first when `--debug-brk` is on. It might be related to a circular dependency. More investigation needed.

vchimev:
Caused by:
```
ERROR in Circular dependency detected:
../NativeScript/tns-core-modules/ui/styling/style-scope.ts ->
../NativeScript/tns-core-modules/ui/core/view-base/view-base.ts ->
../NativeScript/tns-core-modules/ui/core/bindable/bindable.ts ->
../NativeScript/tns-core-modules/application/application-common.ts ->
../NativeScript/tns-core-modules/globals/globals.ts ->
../NativeScript/tns-core-modules/globals/polyfills/dialogs/dialogs.ts ->
../NativeScript/tns-core-modules/ui/dialogs/dialogs.ts ->
[...]

PR #8114, opened 2019-11-19
chore: remove critical circular dependencies

MartoYankov:
## PR Checklist

There are circular dependencies that prevent us from debugging iOS apps with linked modules and debug-brk/stopOnEntry.

## What is the new behavior?
Here is a list of the important refactoring points:

- Removed dependency from `application` to `frame` modules by removing the option to create a default root frame. This has been deprecated and deleted as an API, but was used internally for the `ui-tests-app`.
- Separated the `utils/utils` module's internal modules, so that they can be imported separately. These are `layout`, `ios` and `android`.
- Extract the `ios` namespace from `view.ios.ts` in a separate module, so that it can be imported individually.
- Extract some modules interfaces and types to a separate file containing only such TS constructs that are for type checking only, i.e. the file will disappear when transpiled. This is to ensure no accidental circular dependencies happen when importing types/interfaces.

Fixes https://github.com/NativeScript/NativeScript/issues/7365


nativescript-core/application/application.android.ts
```diff
@@ -187,11 +189,10 @@ export function _start(entry?: NavigationEntry | string) {
 }
 
 export function _shouldCreateRootFrame(): boolean {
-    return createRootFrame.value;
+    return false;
```
vakrilov: Breaking Change?
As far as I can tell - this will break applications that still us the `application._start()` API.

vakrilov: That might be OK, as we left the `_start()` method only because of the test project. 
I think it's better to remove the whole `_start()` API altogether

nativescript-core/utils/layout-helper/layout-helper.android.ts
```diff
@@ -0,0 +1,52 @@
+import { ad } from "../native-helper";
+
+export * from "./layout-helper-common";
+
+let density: number;
+
+// cache the MeasureSpec constants here, to prevent extensive marshaling calls to and from Java
+// TODO: While this boosts the performance it is error-prone in case Google changes these constants
+const MODE_SHIFT = 30;
```
vakrilov: `MODE_SHIFT` and `MODE_MASK` constants are the same as in `layout-helper-common` - reuse them from there.

nativescript-core/utils/layout-helper/layout-helper.ios.ts
```diff
@@ -0,0 +1,39 @@
+import * as layoutHelperCommon from "./layout-helper-common";
+
+export * from "./layout-helper-common";
+
+const MODE_SHIFT = 30;
```
vakrilov: MODE_SHIFT and MODE_MASK constants are the same as in layout-helper-common - reuse them from there.

nativescript-core/application/application-common.ts
```diff
@@ -1,13 +1,29 @@
...
+
+// Requires
+import { Observable } from "../data/observable";
 import {
     trace as profilingTrace,
     time,
     uptime,
     level as profilingLevel,
 } from "../profiling";
+import * as bindableResources from "../ui/core/bindable/bindable-resources";
```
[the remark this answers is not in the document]
MartoYankov: This feature as a whole "bindable resources" should be re-evaluated in my opinion. The bindings to parents context should work.

vakrilov:
test

vakrilov:
test

Issue #8151, opened 2019-11-28
[IOS] lineHeight breaks textAlignment on Button label when multiline


SylannBin:
I expect the following code to render a button with text centered inside.
But on IOS, the text aligns left as soon as there is more than one line.

```html
<Button textWrap="true" textAlign="center">
  <FormattedString>
    <Span text="Start of text" />
    <Span text=" " />
    <Span text="next part of text" />
  </FormattedString>
</Button>
```

see screenshot below:

![image](https://user-images.githubusercontent.com/17208592/69828151-f52c6b00-121a-11ea-9727-e09e44e03f35.png)

1st case: content wraps inside the button and is aligned left

2nd case: content does not wrap and is aligned center as expected

EDIT:
The problem also happen in a simple button. Is it apple that does not support centered multiline label on buttons?

```html
<Button text="a long text that wraps on a second line" textWrap="true" textAlign="center" />
```

![image](https://user-images.githubusercontent.com/17208592/69828478-51dc5580-121c-11ea-86df-9fba46cab53e.png)

NickIliev:
@SylannBin the property is `textAlignment` - on my side it is working as expected (see[ this Playground demo](https://play.nativescript.org/?template=play-tsc&id=mnTd9h))

SylannBin:
I really used a class in my case, so my bad for the wrong attribute name on this thread, but it didn't work with the css property "text-align: center" @NickIliev 
I am trying your playground right now

SylannBin:
I have reproduced most of my use case in the playground and don't have the problem here.
On the other hand, when I copy/paste your example in my app, the bug appears again.

So there must be something in my app, but I don't know what.

SylannBin:
Example playground with the bug:
https://play.nativescript.org/?template=play-tsc&id=mnTd9h&v=2

SylannBin:
Ok I've taken some time to check this myself. I've lurked a bit in the Nativescript code and managed to fix the problem which is caused by the following line:
https://github.com/NativeScript/NativeScript/blob/3199a392b45635606e06119845099965246395c3/nativescript-core/ui/text-base/text-base.ios.ts#L142

Funnily enough, a similar bug was addressed previously by a previous fix. (https://github.com/NativeScript/NativeScript/pull/4909)

Sadly this fix didn't took care of the `UIButton` which accesses `textAlignment` through its `titleLabel` object and not directly on itself.

[...]

PR #8181, opened 2019-12-08
fix(ios-text-base): button textAlignment on IOS (UIButton)

SylannBin:
## Introduction
When setting the format of a "text view", we need to recover existing values for properties that will get default values otherwise.

The "textAlignment" of a UIButton is available through a "titleLabel" object instead of directly on itself (see IOS documentation fo UIButton).
Currently, when setting a new format on a UIButton, the recovered value for textAlignment would always be `undefined`, and therefore the default value for textAlignment would always be applied when instantiating a new paragraphStyling.

## What is the new behavior?
After this commit, the previously set textAlignment (if any) is properly recovered and reused.

fixes pr #8151

### DIFF
--- a/nativescript-core/ui/text-base/text-base.ios.ts
+++ b/nativescript-core/ui/text-base/text-base.ios.ts
@@ -139,7 +139,12 @@ export class TextBase extends TextBaseCommon {
             const paragraphStyle = NSMutableParagraphStyle.alloc().init();
             paragraphStyle.lineSpacing = this.lineHeight;
             // make sure a possible previously set text alignment setting is not lost when line height is specified
-            paragraphStyle.alignment = (<UITextField | UITextView | UILabel>this.nativeTextViewProtected).textAlignment;
+            if (this.nativeTextViewProtected instanceof UIButton) {
+                paragraphStyle.alignment = (<UIButton>this.nativeTextViewProtected).titleLabel.textAlignment;
+            } else {
+                paragraphStyle.alignment = (<UITextField | UITextView | UILabel>this.nativeTextViewProtected).textAlignment;
+            }
+
             if (this.nativeTextViewProtected instanceof UILabel) {
                 // make sure a possible previously set line break mode is not lost when line height is specified
                 paragraphStyle.lineBreakMode = this.nativeTextViewProtected.lineBreakMode;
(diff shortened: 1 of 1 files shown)


romain20100:
@cla-bot check

vmutafov:
test

nativescript-core/ui/text-base/text-base.ios.ts
```diff
@@ -192,7 +197,12 @@ export class TextBase extends TextBaseCommon {
             const paragraphStyle = NSMutableParagraphStyle.alloc().init();
             paragraphStyle.lineSpacing = style.lineHeight;
             // make sure a possible previously set text alignment setting is not lost when line height is specified
-            paragraphStyle.alignment = (<UITextField | UITextView | UILabel>this.nativeTextViewProtected).textAlignment;
+            if (this.nativeTextViewProtected instanceof UIButton) {
```
vmutafov: Couldn't we extract the logic of this `if` statement to a separate function or introduce some polymorphic behavior for `UIButton` as these lines are duplicated thus increasing the maintenance cost?

vtrifonov:
test

nativescript-core/ui/text-base/text-base.ios.ts
SylannBin: I'm not sure it's a good idea in that case. It's only repeated twice and there is a lot of context.

Or maybe there is a bigger picture, but I don't see it.

vmutafov: Why do you think it's not a good idea and what do you mean by context?
The bigger picture is that, in my opinion, the codebase is starting to have a lot of code duplications and it's getting common to fix a bug at multiple places. To mitigate this problem I think we should be conservative about adding new code duplications and refactor existing ones whenever possible.

SylannBin: If you are talking about rethinking this whole file to avoid even having this if statement, I agree. But moving just this if statement in a function to avoid repeating 2 lines of code, is a perfect example of over abstraction in my opinion.

However, I do want to say that I have little experience with nativescript, since I almost just discovered this project. So I don't pretend to have a genuine opinion in here. I'm just sharing my "general experience" on what looks like a lot of work for a not so important abstraction.

vmutafov: Alright then, I'll approve your changes. We'd perhaps refactor this whole file so we'll abstract stuff at a later point. Thank you for the contribution :)

vmutafov:
test

Issue #7469, opened 2019-07-05
Reusable NativeScript views


edusperoni:
**Is your feature request related to a problem? Please describe.**
Removing a NativeScript view from the visualtree results in all native views being disposed on Android. On iOS, some native views are not disposed, but their delegates are. This behavior seems to come from https://github.com/NativeScript/NativeScript/pull/3912 but I haven't found an explanation for why it is that way.

Since the view is destroyed, there is no way to do, for example:

```typescript
const parent1 = stack1;
const parent2 = stack2;
const st  = new StackLayout();
const bt = new Button()
st.addChild(bt);
parent1.addChild(st);
// st.android != null
parent1.removeChild(st);
// st.android == null
parent2.addChild(st);
// st.android != null
```

Without the view being destroyed and created again.

This also makes hard to use some plugins like `nativescript-popup`. We've tried creating an angular template and passing the NS views to the plugin, but once we detached the view from it's parent, the native views became null and the popup was empty. Our workaround was to manually use the native views that were created and set `view.parent = null` otherwise the app would crash when closed.

**Describe the solution you'd like**
NativeScript views could be reusable:

```typescript
const parent1 = stack1;
const parent2 = stack2;
const st  = new StackLayout();
st.reusable = true;
const bt = new Button()
st.addChild(bt);
parent1.addChild(st);
const native = st.android;
parent1.removeChild(st);
native === st.android // true
parent2.addChild(st);
native === st.android // true
```

This is different from https://github.com/NativeScript/NativeScript/issues/4189, since we're not just reusing the native views, but the whole NS view, so no properties have to be reset. Maybe an additional method (like `view.inflate()`) could be provided to inflate views without the need to attach to the "dom".

**Additional context**
I've only used core and angular flavors, so I'll contextualize it using them:

Angular/Vue/React/etc flavors of nativescript would create all their views as `reusable`. The angular renderer has [destroyNode](https://angular.io/api/core/Renderer2#destroyNode) method that is unimplemented by `nativescript-angular` because the core already destroys the view when it's detached. [This method is called when the view is not being used anymore](https://github.com/angular/angular/blob/7ca611cd1266ede26a26b17cb31748988ff8d98c/packages/core/src/render3/node_manipulation.ts#L90-L97), so the view can be safely detached and reused. I'm sure the other frameworks have similar approaches.

What could be accomplished with this:
```
<Button *ngFor="items;"></Button>
```
[...]

PR #8184, opened 2019-12-09
DO NOT MERGE: feat: reusable views

edusperoni:
## PR Checklist

Views and view trees are destroyed when removed from the view tree.

## What is the new behavior?
Views can be marked as `reusable` and will not be destroyed when removed from view tree. Removing these nodes require the developer to do: 
```
view.destroyNode();
// which calls
view.reusable = false;
view._tearDownUI();
```

For implementation details on how frameworks should deal with it, see my work in `nativescript-angular`: https://github.com/edusperoni/nativescript-angular/tree/reusable-views

Implements #7469.

-------
## To be defined:

What happens when a reusable view with children gets removed? The current implementation does not call _tearDownUI on children, but we should still suspendNativeUpdates down the chain (`this._suspendNativeUpdates(SuspendType.UISetup);`).

Maybe we should we create a new SuspendType (Detached)? Maybe this isn't even needed as the view should be unloaded by `_removeViewCore` which eventually calls `this._suspendNativeUpdates(SuspendType.Loaded);` on all children.

## Possible refactoring needed (not directly related to this PR)

The current logic of `_resumeNativeUpdates` will call `initNativeView`. In most places I checked, `initNativeView` creates delegates that are only removed in `disposeNativeView`. This behavior is probably already an issue on loading/unloading views (like navigation) because of `this._resumeNativeUpdates(SuspendType.Loaded);`, meaning `initNativeView` is called multiple times during the application lifecycle, but the delegates are only cleaned up once.

During my tests with WebView, calling `initNativeView` did not result in a performance impact


edusperoni:
After further working on this, `SuspendType.Loaded` seems to be sufficient. One of my commits had actually messed with the logic resulting in a crash (trying to re-add children to the native visual tree, which they already were added to). All child views are unloaded on parent removal.

I'll also weigh in that I believe child views should not be destroyed even when detached simply because it's already the behavior on [tabviews](https://github.com/NativeScript/NativeScript/blob/a61de80a354cf9633a4dc8fc6ac2e97f26ffe5f8/nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts#L349) (which soft detaches elements)

vtrifonov:
test

vtrifonov:
test

NathanWalker:
@edusperoni I'm curious what are your current thoughts on this PR? Does it still require changes?

edusperoni:
@NathanWalker I believe this just requires some testing and maybe treatment of edge cases. For example: if the view is detached and reusable is set to false, should it run _tearDownUI? Other than that, I think it'd be good as is.

NathanWalker:
Flagging for 7.0 to test more with latest changes planned.

NathanWalker:
@edusperoni Let's update this to latest master either from a new branch or rebase.

PR #8380, opened 2020-02-26
fix(css): parse issue with new Webpacked AST css/scss files

NathanaelA:
## PR Checklist

- [X] All existing tests are passing: 

Webpack 1.50 (1.40?) now pre-parses CSS/SCSS using the 'css-loader' and includes the AST for the CSS instead of the raw CSS.   This now causes breaking issues when you use `addTaggedAdditionalCSS` as the file that is webpacked is now no longer CSS, but an AST tree.
All apps that used require(css/scss) with `addTaggedAdditionalCSS` are now broken.   

## What is the new behavior?
This changes the `addTaggedAdditionalCSS` to actual run the same detection code that is used in the `fromURI`, and abstracts that code into a new function `fromDetect`  Now both `fromURI` and `addTaggedAdditionalCSS` call `fromDetect` so they share the same logic to detect which way the css was handled.

This PR FIXES a BREAKING change caused by a change in the Webpack configuration...


vmutafov:
test

dtopuzov:
test

PR #8402, opened 2020-03-02
fix: dont create an actionbar if not necessary

farfromrefug:
Also i changed the android check to behave like iOS


vakrilov:
Hey @farfromrefug - this is kind of a breaking change as in Android you get an ActionBar with the name of the app by default. With your PR it is hidden by default. You can see the change in the `e2e/ui-tests-app`:

Screenshot from `master`:
Screenshot_1583484089

Screenshot from PR branch:
Screenshot_1583484050

farfromrefug:
@vakrilov ok i thought it would be. I ll fix it !

farfromrefug:
@vakrilov  i improved it.
Now it should be created only if needed in the sense of `actionBarHidden !== true`
If `actionBarHidden` is true then the actionBar will not be created on `onLoaded` but will be if `actionBarHidden` is changed to false.

vtrifonov:
@farfromrefug would you update your branch, please!

vtrifonov:
test

Issue #8407, opened 2020-03-03
Tap event data should include touch location


mudlabs:
**Is your feature request related to a problem? Please describe.**
 
As it stands getting the _x/y_ location of a tap event requires diving into native code. But touch events are already supplying this utility. This should be extended to tap events.

**Describe the solution you'd like**

The `touch` gesture event data methods of `getX`, and `getY` should be included in tap event data.

Also a TapGestureEventData interface should be available for type assignment.

```ts
export function tapped(args: TapGestureEventData): void {
    const tapped_x = args.getX(); // The X axis tap location in DIP
    const tapped_y = args.getY(); // The Y axis tap location in DIP
}
```

**Describe alternatives you've considered**

Currently you have to dive into platform specific native code to get the tap location. But they are also not in DIP, so that conversion must be done.
```ts
import { screen } from "tns-core-modules/platform";

export function tapped(args: GestureEventData): void {
    if (args.ios) {
        const location = args.ios.locationInView(args.object.ios); //CGPoint in DP.
        const x = location.x / screen.mainScreen.scale; // The X axis tap location in DIP
        const y = location.y / screen.mainScreen.scale; // The Y axis tap location in DIP
    } else {
        const x = args.android.getX() / screen.mainScreen.scale; // The X axis tap location in DIP
        const y = args.android.getY() / screen.mainScreen.scale; // The Y axis tap location in DIP
    }
}
```

**Additional context**

Could also extend the `getPointerCount`, and `getAllPointers` methods available to `touch` events. This would give a simple way to handle single or multiple pointer events when you don't require the longer lived event cycle of `touch` events.

NathanaelA:
This is a good idea --  However it probably should be a helper function/getter; like something like:
`const tapLocation = args.getLocation()` which returns it.   We probably don't want to run the conversion code on every tap event, but only when the developer actually needs the tap location..

mudlabs:
Yeah, that's how `touch` does it.

Looks like there's some inconsistency with how `getX` and `getY` work too.
- The `getX` and `getY` methods of the `touch` event data are in **DP**, on both platforms.
- But if you get a `touch` event Pointer _(i.e. args.getActivePointers()[0])_, its `getX` and `getY` methods return **DP** on _iOS_, and **DIP** on _Android_.

NickIliev:
Marking this issue as an enhancement (for the tap event data) and also as a good-first-issue as most of the code is already implemented in the touch event data and can be used as a reference.

PR #8415, opened 2020-03-05
fix: update tap event data

mudlabs:
# PR Checklist

Tap and doubleTap events do not provide first-class methods for obtaining the tap location or pointer count, unlike touch events.

## What is the new behavior?
There is now a `TapGestureEventData` interface, available to `tap` and `doubleTap` events.

The `getX()`, `getY()`, and `getPointerCount()` methods of `touch` events are now available to `tap`, and `doubleTap` events.

```ts
export function myTapEvent(args: TapGestureEventData): void {
  const [ x, y ] = [args.getX(), args.getY()];
  const pointers = args.getPointerCount();

  console.log(`Tap location: (${x}, ${y}). Number of Pointers: ${pointers}`);
}
```

Implements #8407 

#### BREAKING CHANGES:
None


NathanaelA:
Do to the fact this is a breaking change; I would recommend you create new functions for this like "getDIPX()"; that way it can be merged.   Then in NS7 we can possibly "fix" the getX() to use the proper one.   The NativeScript team won't take breaking changes until major release; and getting something like this in would be worth using a new name.  :)

mudlabs:
@NathanaelA ok.

What about the conflict with _iOS_ vs _Android_ `touch` Pointers?

As it stands:
```ts
function myTouchEvent(args: TouchGestureEventData): void {
  const pointer = args.getActivePointers()[0];
  const x = pointer.getX();
 
  // x is in DP on iOS, but DIP on Android
  console.log(x);
}
```

Should I leave _iOS_ as **DP** for now. Or correct it to match _Android_. They should both be the same.

P.S. I think all properties and methods provided by the NativeScript API should default to **DIP**, unless otherwise noted.

farfromrefug:
Agree everything should return dip! I wrote a plugin for better gestures based on the great react native plugin. This was one of the reason. I think that PR should be a amazing addition

tests/package.json
```diff
@@ -5,7 +5,7 @@
       "version": "next"
     },
     "tns-ios": {
-      "version": "next"
+      "version": "6.5.0-2020-02-19-163529-01"
```
vmutafov: Why is this hardcoded here?

mudlabs:
@vmutafov , Sorry about the formatting. My VSCode just does that every time I save a file. I must have installed a formatter at some point. Not sure how to deactivate it, we'll look into it.

I will rework the branch to remove _breaking changes_.

#### As for the method implementation do we prefer:
```ts
// A)  Same method names, but an optional boolean specifies the format.
getX(DIP?: boolean): number;

// B) Method names specify format, then property.
getDIPX(): number;

// C) Method names specify property, then format.
getXDIP(): number;
```

Personally I prefer either **A** or **C**. If there's already a convention where object _properties_ would follow **C**, then it could make sense to do likewise with methods. Although, if the methods return the same type then **A** could be the go, as it follows the same convention, only in a method centric way, and with one less function declaration.

<br/>

@NathanaelA @farfromrefug

tests/package.json
mudlabs: Not sure. This must be related to me auto formatting on save issue.

vmutafov:
Personally, I'm not a fan of methods starting with `get` having a boolean parameter. It seems like a code smell to me, so I would prefer B) or C)

farfromrefug:
@vmutafov i understand your point
 But to me if we want not to break anything and still be consistent with the way it works everywhere else A seems the best to me

NathanaelA:
I'm with B or C also. Also in this specific case -- since their is a discrepancy between platforms and being able to get both values is very useful -- maybe the thing is that you add a `gitXPixel` and `getXDIP` and then we just mark getX as depreciated.   That way all platforms and events could easily become consistent and we don't worry about trying to make `getX` consistent, just eliminate it in NS 7...

mudlabs:
@NathanaelA  Does it clutter the API do have both `getXDIP()` and `getXPixels()` etc..?

I feel like I've seen comments from the NativeScript team indicating they only want their API to offer properties/methods that have I high need from most users.

I think most will probably just need **DIP**. Unless you are diving down into native code **DIP** is what you'll need to update the UI; if that's why you need the coordinates.

Devs needing pixels (**DP**) are likely implementing native code, and probably know about `layout.getDisplayDensity()`, `screen.mainScreen.scale`, or how to get the values from the underlying native object.

This is why I think **DIP** should be the standardised default throughout the API. And why my initial PR was just an override of `getX()`/`getY()`, because if it's all standardised to **DIP** we don't need to specify format in property/method names.

farfromrefug:
@mudlabs I totally agree. I don't see the need for both.

NathanaelA:
I can understand your thinking; I guess I've ran into many cases over the years where I needed access to both values.   A simple low code helper for a fairly common item, seems worth it for me...   In fact a Chart I was working on the other day; I needed both values in the handlers; I would have been a lot happier had I been able to just simple say getDipX, getPixelX and not worry having to grab the `platform`, and figure out how to get the scale value again from it and then testing if it was a multiplication or division to scale the values properly...  :grinning: 

My vote is to put both in place; but I won't complain if you don't -- my main concern is I do want the consistency in all the handlers, but we can't break existing code...

mudlabs:
### Found more inconsistencies
On _iOS_ the return values from `getFocusX()`/`getFocusY()` of Pinch events, and the values for `deltaX`/`deltaY` of Pan events are all in **DP** format.

On _Android_ these are all in **DIP** format.

---

I think I'm going to just keep this PR to implementing the `getX()`/`getY()` methods for tap events. These will return **DIP**.

I'll undo any other changes so we don't have any breakages. And leave in all the inconsistencies for now.

Then I'll open a new issue outlining our desire to see a standardised use of **DIP** throughout the API. And then create a new PR with breaking changes for _v7_.

mudlabs:
@vmutafov any chance of getting that review closed?

vtrifonov:
test

vtrifonov:
test

PR #8510, opened 2020-04-07
fix(android-tabs): tab backgroundColor cache can be wrong - DO NOT MERGE

NathanaelA:
## PR Checklist

1. Currently on android if you set the Tab & TabView's background color; it cache's the color when you leave the page and resets it when you return to the page.  (Good behavior so far).   However, if you navigate away; and then change the global CSS from say light theme to a dark theme, then return back to the page with a tab/tab-view the tab attempts to use the old cached color, despite the CSS system passing a new backgroundColor into the control on refresh. 

2. On tab-views; the ViewPager can cache the background color; so we need to force it to change the color also.

3. Crashing issue with `updateAndroidItemAt` during startup of control.   

## What is the new behavior?
1. This clears the `_originalBackground` cached value if a new backgroundColor is set.

2. sets the viewPagers color also

3. Adds a safety check to make sure updateAndroidItemAt can't update something that doesn't exist.

Fixes/Implements/Closes #
Fixes two issues repoted in NativeScript-Themes repo with the "tabs" and "tab-view" controls.

### DIFF
--- a/nativescript-core/ui/tab-view/tab-view.android.ts
+++ b/nativescript-core/ui/tab-view/tab-view.android.ts
@@ -698,7 +698,9 @@ export class TabView extends TabViewBase {
     }
 
     public updateAndroidItemAt(index: number, spec: org.nativescript.widgets.TabItemSpec) {
-        this._tabLayout.updateItemAt(index, spec);
+        if (index < this._tabLayout.getItemCount()) {
+            this._tabLayout.updateItemAt(index, spec);
+        }
     }
 
     [androidOffscreenTabLimitProperty.getDefault](): number {
--- a/nativescript-core/ui/tab-view/tab-view.android.ts
+++ b/nativescript-core/ui/tab-view/tab-view.android.ts
@@ -734,6 +734,7 @@ export class TabView extends TabViewBase {
 
         if (value instanceof Color) {
             this._tabLayout.setBackgroundColor(value.android);
+            this.nativeViewProtected.viewPager.setBackgroundColor(value.android);
         } else {
             this._tabLayout.setBackground(tryCloneDrawable(value, this.nativeViewProtected.getResources()));
         }
--- a/nativescript-core/ui/tab-view/tab-view.android.ts
+++ b/nativescript-core/ui/tab-view/tab-view.android.ts
@@ -736,7 +736,7 @@ export class TabView extends TabViewBase {
 
         if (value instanceof Color) {
             this._tabLayout.setBackgroundColor(value.android);
-            this.nativeViewProtected.viewPager.setBackgroundColor(value.android);
+            this._viewPager.setBackgroundColor(value.android);
         } else {
             this._tabLayout.setBackground(tryCloneDrawable(value, this.nativeViewProtected.getResources()));
         }
--- a/nativescript-core/ui/tab-view/tab-view.android.ts
+++ b/nativescript-core/ui/tab-view/tab-view.android.ts
@@ -730,6 +730,8 @@ export class TabView extends TabViewBase {
         return this._tabLayout.getBackground();
     }
     [tabBackgroundColorProperty.setNative](value: android.graphics.drawable.Drawable | Color) {
+        this._originalBackground = null;
+
         if (value instanceof Color) {
             this._tabLayout.setBackgroundColor(value.android);
         } else {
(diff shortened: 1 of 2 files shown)


vtrifonov:
test

vtrifonov:
test

nativescript-core/ui/tab-view/tab-view.android.ts
```diff
@@ -730,8 +732,11 @@ export class TabView extends TabViewBase {
         return this._tabLayout.getBackground();
     }
     [tabBackgroundColorProperty.setNative](value: android.graphics.drawable.Drawable | Color) {
+        this._originalBackground = null;
+
         if (value instanceof Color) {
             this._tabLayout.setBackgroundColor(value.android);
+            this._viewPager.setBackgroundColor(value.android);
```
vtrifonov, line 739: We shouldn't set the background color of the viewPager here as this should change only the background of the tab bar, but not the tab page as well. You can check the following page in the UI test project - https://github.com/NativeScript/NativeScript/blob/master/e2e/ui-tests-app/app/tab-view/all-page.xml

nativescript-core/ui/tabs/tabs.android.ts
```diff
@@ -783,14 +783,18 @@ export class Tabs extends TabsBase {
     // }
 
     public updateAndroidItemAt(index: number, spec: org.nativescript.widgets.TabItemSpec) {
-        this._tabsBar.updateItemAt(index, spec);
+        if (index < this._tabsBar.getItemCount()) {
```
vtrifonov, line 786: I guess the same changes should be applied here - https://github.com/NativeScript/NativeScript/blob/master/nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts#L703

nativescript-core/ui/tabs/tabs.android.ts
```diff
@@ -783,14 +783,18 @@ export class Tabs extends TabsBase {
...
+            this._tabsBar.updateItemAt(index, spec);
+        }
     }
 
     public getTabBarBackgroundColor(): android.graphics.drawable.Drawable {
         return this._tabsBar.getBackground();
     }
 
     public setTabBarBackgroundColor(value: android.graphics.drawable.Drawable | Color): void {
+        this._originalBackground = null;
```
vtrifonov, line 796: and this should be applied here - https://github.com/NativeScript/NativeScript/blob/master/nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts#L711

vtrifonov:
test

vtrifonov:
@NathanaelA did you have a chance to view the comments?

NathanaelA:
No, for some reason Github didn't send them to me... Looking at them now...  Thanks for pinging me...

nativescript-core/ui/tab-view/tab-view.android.ts
NathanaelA: Ah, I think I must have misunderstood this property; this is only the "Tab" not the tab pages.  Correct?

I'm almost sure this effected the tab page on first load; which is why I got confused...  

 I'll play with my demo a bit more to isolate down the issue and make some PR changes...     :grinning:

nativescript-core/ui/tabs/tabs.android.ts
NathanaelA: Oh, good catch -- I didn't even think to check the Bottom Navigations...   I'll add them to my example app and verify everything works...

vtrifonov:
@NathanaelA you can use [e2e/ui-tests-app](https://github.com/NativeScript/NativeScript/tree/master/e2e/ui-tests-app) as an example app.

NathanaelA:
@vtrifonov - This PR will be fixed later as some code is definitely not relevant anymore...  :grinning:   

But it needs to wait until we can figure out the best solution for this issue: https://github.com/NativeScript/NativeScript/issues/8547

cjohn001:
@NathanaelA 
I have seen that you have already partly implemented a fix for the issue. As the issue seems to be a difficult one, and as it is somewhat older already, I was wondering, if you might recommend a workaround for it as long as the bug is open.The bug (background lost and white bg instead) appears for me if I use a goBack navigation which goes back to a tab view. Is there a way I could force to refresh the background on the tab control before it is rendered? Thank you very much for your help!

Best regards,
 Christoph

NathanaelA:
@cjohn001 - If you store and reset the background manually (not using css) on leave and enter; then the bug won't occur; the issue occurs because it tries to do this internally and then breaks the css from being able to re-apply it.   This PR actually is no longer valid as I need to fix the underlying issue first.

NathanWalker:
Further investigation into CSS property handling will be done separately.

PR #8514, opened 2020-04-09
chore: refactor for easy subclassing and new features

farfromrefug:
This is a simple refactoring in the sense that now the intelligence of the `UITextFieldDelegate` is in the `TextField` class.

This is a really nice addition for plugin developers like me.
For example in `nativescript-material-textfield` i wanted to add a feature to limit digits that a textfield can accept.
This has to be handled in the delegate. Right now i have to rewrite and duplicate the whole delegate code from {N}  to keep the feature of the {N} Textfield. It also means maintaining the delegate through {N} updates.

This makes it possible to simply "override" the delegate methods from your subclass of `TextField`

The only unknown there is, should those delegate functions be added in the typings


vtrifonov:
test

Issue #8523, opened 2020-04-12
TextField : setting secure to false has not effect if editable set to false


Rahul1d:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 6.5
 - Cross-platform modules:6.5
 - Android Runtime:6.5
 - iOS Runtime:6.5
 - Plugin(s):

**Describe the bug**
Hi,
In TextField, editable is set to false. I am trying to toggle between secure and not secure. But Unfortunately, text is hidden and not visible even if secure is set to false.

**To Reproduce**
https://play.nativescript.org/?template=play-ng&id=qGgSQQ&v=4

NickIliev:
Confirming the issue as a bug - once **editable** is set to **false** there is no toggling the **secure** option (in fact, the value is changed but the UI is not updated and the TextField stays secure)

PR #8525, opened 2020-04-13
fix(android/TextField): set focusable, clickable to false if not editable

Rahul1d:
Do not clear key listener when editable is false This prevents the input type to be changed.
Instead, we can set focusable, clickable to false.
This allows to change input Type i.e (secure) at run time when editable is false.

Reference issue: https://github.com/NativeScript/NativeScript/issues/8523

## PR Checklist

In android platform, In TextField if editable is set to false. We cannot change or toggle secure property. This occurs because we are setting KeyListener to false. Instead we can set Focusable, clickable to false.

## What is the new behavior?
Now, one can toggle between secure property, even if editable is set to false.


vtrifonov:
test

Issue #6659, opened 2018-12-04
Nested frames (fragments) disappear when their parent frame (fragment) is animated


manoldonev:
**Environment**
 - CLI: 5.0.2
 - Cross-platform modules: 5.0.5
 - Android Runtime: 5.0.0

**Describe the bug**
Nested frames (i.e. fragments) disappear when their parent is animated. Related to https://github.com/NativeScript/NativeScript/issues/6641 but the problem is not specific to TabView and is more general. Caused by https://issuetracker.google.com/issues/36970570 -- arcane issue that Google somehow marked as obsolete & won't fix.

**To Reproduce**
Tap "navigate to some page" -- notice how the "Players" part of the page disappears immediately instead of transitioning away with the rest of the view.

**Sample project**
Playground project for reference: https://play.nativescript.org/?template=play-tsc&id=BZfk8A&v=5

manoldonev:
Reopening this issue as it could occur in a similar scenario but when navigating backwards from a nested-frame page.

**To reproduce:**
1. Define a frame root for your app (default page contains no frames i.e. we have single frame scenario)
2. Navigate to a different page that contains additional frame (i.e. we now are in a nested frame scenario)
3. Navigate back with android hardware button -- nested frame/fragment disappears immediately.

Note this problem can only be observed after https://github.com/NativeScript/NativeScript/pull/7011 is merged

farfromrefug:
The fix you implemented where you dispose the current fragment on `onUnloaded` has bad consequences for apps with map for example.
Doing so create unecessary OpenGL redrawing (becuase the mapview is recreated on every pause/resume).

PLus running the sample app i actually cant see any issue as explaned here. 
I will create a PR to remove that behavior fix. Hoping for you guys to accept it.

PR #8561, opened 2020-05-05
fix(android): remove the hack for nested fragments

farfromrefug:
As explained here https://github.com/NativeScript/NativeScript/issues/6659
That hack has bad consequences.
Also i actually cant see any wrong behavior using the sample provided in the issue and upgrading to latest N.

### DIFF
--- a/nativescript-core/ui/frame/frame.android.ts
+++ b/nativescript-core/ui/frame/frame.android.ts
@@ -251,11 +251,6 @@ export class Frame extends FrameBase {
 
     onUnloaded() {
         super.onUnloaded();
-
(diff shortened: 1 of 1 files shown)


vtrifonov:
test

vtrifonov:
test

NathanWalker:
@NathanaelA can you provide feedback here? Breaking change or do you feel we should add to 7.0 plan or safe for patch core module release?

NathanaelA:
@NathanWalker - We need to test the two sample apps before and after and see if this PR will break, fix or does nothing in the scheme of the original issue that @farfromrefug is eliminating....

PR #8609, opened 2020-06-03
fix(ios): dialog handling with top view controller

NathanWalker:


diegovincent:
@NathanWalker Do you think this was the reason why I cannot use @Akylas bottom sheet plugin with a `page-router-outlet` inside? (See [this](https://github.com/Akylas/nativescript-material-components/issues/143)). Even after upgrading, I cannot manage to use `page-router-outlet` instead of `router-outlet` without the app crashing in both platforms.

PR #8617, opened 2020-06-05
fix(text-base): update text color for ios

dosomder:
## PR Checklist

- [ ] You have signed the [CLA](http://www.nativescript.org/cla). -> Link does not work

Updating the color property on a button in iOS does not change its text color.
Looks like the NSForegroundColorAttributeName takes precedence over any color later applied by setTitleColorForState.

## What is the new behavior?
Updating the color property on a button in iOS does change its text color.

Fixes/Implements/Closes https://github.com/NativeScript/nativescript-angular/issues/1453

### DIFF
--- a/nativescript-core/ui/text-base/text-base.ios.ts
+++ b/nativescript-core/ui/text-base/text-base.ios.ts
@@ -141,13 +141,7 @@ export class TextBase extends TextBaseCommon {
     }
     [colorProperty.setNative](value: Color | UIColor) {
         const color = value instanceof Color ? value.ios : value;
-        const nativeView = this.nativeTextViewProtected;
-        if (nativeView instanceof UIButton) {
-            nativeView.setTitleColorForState(color, UIControlState.Normal);
-            nativeView.titleLabel.textColor = color;
-        } else {
-            nativeView.textColor = color;
-        }
+        this._setColor(color);
     }
 
     [fontInternalProperty.getDefault](): UIFont {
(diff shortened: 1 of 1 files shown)


dosomder:
The CLA link does not work.

rigor789:
@dosomder thanks for the heads up - looking into the missing CLA form!

nativescript-core/ui/text-base/text-base.ios.ts
```diff
@@ -306,14 +310,6 @@ export class TextBase extends TextBaseCommon {
...
         }
 
-        if (dict.size > 0 || isTextView) {
-            if (style.color) {
-                dict.set(NSForegroundColorAttributeName, style.color.ios);
-            } else if (majorVersion >= 13 && UIColor.labelColor) {
-                dict.set(NSForegroundColorAttributeName, UIColor.labelColor);
-            }
-        }
-
```
NathanWalker, line 316: @dosomder thank you for this contribution - curious on the removal of this block here. Looks like the `dict` where these properties are set is used right below there to set the attributes:
```
if (dict.size > 0 || isTextView) {
            if (isTextView) {
                // UITextView's font seems to change inside.
                dict.set(NSFontAttributeName, this.nativeTextViewProtected.font);
            }

            const result = NSMutableAttributedString.alloc().initWithString(source);
            result.setAttributesRange(<any>dict, { location: 0, length: source.length });
```
If `NSForegroundColorAttributeName` is not added to the `dict` it would be missing from the attributes when they are set on the `result` there in that case.

Or am I missing something?

NathanWalker:
Ended up resolving the comment in question via another pr - thanks again for another great contribution @dosomder 🤗

dosomder:
I see you added the
```
        if (style.color) {
            dict.set(NSForegroundColorAttributeName, style.color.ios);
        } else if (majorVersion >= 13 && UIColor.labelColor) {
            dict.set(NSForegroundColorAttributeName, UIColor.labelColor);
        }
```
again. I'm going to test this but I believe we don't want to set `NSForegroundColorAttributeName` because it takes precende over the way we are changing color in `_setColor`

dosomder:
@NathanWalker This bug is still not fixed in 6.5.5 since you added `NSForegroundColorAttributeName`. Do you need me to do something here again?

NathanaelA:
@dosomder,

> The CLA link does not work.

Thanks for letting us know, The CLA link should be working again.   Can you please sign it for us...   https://nativescript.org/cla so that you are in the system.

Issue #4892, opened 2017-09-27
Issue with css letter-spacing in ios


priyangamani:
Dear friend, we, the rest of the NativeScript community really
appreciate your feedback! While we are doing all we can to take care of every
issue, sometimes we get overwhelmed. Because of that, we will consider issues
that are not constructive or problems that cannot be reproduced "dead".
Additionally, we will treat feature requests or bug reports with unanswered
questions regarding the behavior/reproduction for more than 20 days "dead". All
"dead" issues will get closed.

### Tell us about the problem
Please, ensure your title is less than 63 characters long and starts with a capital
letter.

### Which platform(s) does your issue occur on?
iOS/Android/Both

### Please provide the following version numbers that your issue occurs with:
- CLI: (run `tns --version` to fetch it)
- Cross-platform modules: (check the 'version' attribute in the
`node_modules/tns-core-modules/package.json` file in your project)
- Runtime(s): (look for the `"tns-android"` and `"tns-ios"` properties in the
`package.json` file of your project)
- Plugin(s): (look for the version number in the `package.json` file of your
project)

tsonevn:
Hi @priyangamani,
First of all, thank you for your interest in NativeScript.
Regarding your request, the `letter-spacing` property is already supported in NativeScript as it is mentioned here in this issue. For your convenience, I am attaching sample code and screenshot for the simulator.
XML
```
<StackLayout class="p-20">
        <Label text="Tap the button" textWrap="true" class="h1 text-center"/>
    </StackLayout>
```

CSS
```
Label{
    letter-spacing: 1;
}
```
![screen shot 2017-09-27 at 9 11 04 am](https://user-images.githubusercontent.com/17448734/30898388-d66ce2da-a363-11e7-9ef6-66f9418fbc89.png)

priyangamani:
Thanks For ur reply @tsonevn 
I tried to set letter-spacing property in Textfield but it won't work in ios (In android working fine..) . I am attaching sample code and screenshot for the simulator also please check with textfield.

![simulator screen shot 27 sep 2017 3 02 50 pm](https://user-images.githubusercontent.com/6038069/30900237-d22432f0-a395-11e7-9861-6c76f6024eba.png)
"/>

XML
<StackLayout class="p-20">
        <Textfield text="1234"  class="field
 </StackLayout>

CSS
.field{
    letter-spacing: 1;
}

priyangamani:
Thanks For ur reply @tsonevn 
[...]

PR #8257, opened 2020-01-18
feat(text-base): Add Span vertical-align support

bundyo:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [ ] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#running-unit-tests.
- [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/WritingUnitTests.md.

## What is the current behavior?
Currently there's no way to align vertically the Spans in a FormattedString

## What is the new behavior?
Spans in FormattedString support the vertical-align CSS property.

Implements #4795.

PR #8618, opened 2020-06-06
fix(text-base): apply letter spacing in ios text field

dosomder:
## PR Checklist

- [ ] You have signed the [CLA](http://www.nativescript.org/cla)

If the value of the text field is changed at runtime, the configured letter spacing is not respected.

## What is the new behavior?
Letter spacing is respected if text in text field is changed at runtime.

Fixes/Implements/Closes https://github.com/NativeScript/NativeScript/issues/4892

### DIFF
--- a/nativescript-core/ui/text-base/text-base.ios.ts
+++ b/nativescript-core/ui/text-base/text-base.ios.ts
@@ -281,7 +281,11 @@ export class TextBase extends TextBaseCommon {
         }
 
         if (style.letterSpacing !== 0) {
-            dict.set(NSKernAttributeName, style.letterSpacing * this.nativeTextViewProtected.font.pointSize);
+            const kern = style.letterSpacing * this.nativeTextViewProtected.font.pointSize
+            dict.set(NSKernAttributeName, kern);
+            if (this.nativeTextViewProtected instanceof UITextField) {
+                this.nativeTextViewProtected.defaultTextAttributes.setObjectForKey(kern, NSKernAttributeName);
+            }
         }
 
         const isTextView = this.nativeTextViewProtected instanceof UITextView;


NathanWalker:
Thank you for the contribution @dosomder - this will go out with `6.5.5` likely Monday.

dosomder:
Thanks @NathanWalker It appears though as the merging of https://github.com/NativeScript/NativeScript/pull/8257 removed the fix again. There is also a missing semicolon on the line, that's why the current build is failing

NathanWalker:
Thanks for heads up, we’re doing some cleanup there this evening. I’ll try without that block entirely and see if can find any regression.

NathanWalker:
I noticed that your change here was a mutation on a NSDictionary which usually won’t work since setObjectForKey can only be called on NSMutableDictionary. Which is partly why some post cleanup was done.

NathanWalker:
Your change made it to master though, it’s right here:
https://github.com/NativeScript/NativeScript/blob/7823cd4d220267c0bafb297d360fbb526fab61a0/nativescript-core/ui/text-base/text-base.ios.ts#L289

dosomder:
The fix is actually the part
```
            if (this.nativeTextViewProtected instanceof UITextField) {
                this.nativeTextViewProtected.defaultTextAttributes.setObjectForKey(kern, NSKernAttributeName);
            }
```
The `defaultTextAttributes` are applied for new text and somehow the letter spacing is not already set there (while for example other properties like font and color is set).

If it's immutable (weird that it didn't give an error) we can do a `mutableCopy()` before.

NathanWalker:
@dosomder Would it be possible to add another label or textfield to the ui-tests-app here:
https://github.com/NativeScript/NativeScript/blob/master/e2e/ui-tests-app/app/fonts-tests/label-page.xml
which contains the styling for the letter spacing cases that you expect? That'll help us lock in the behavior.

NathanWalker:
https://stackoverflow.com/a/20991693/2192332

NathanWalker:
Since the `dict` would have properties it would end up setting `attributedText` here as mentioned in the SO post above:
https://github.com/NativeScript/NativeScript/blob/master/nativescript-core/ui/text-base/text-base.ios.ts#L322-L334
So not sure which case you may be seeing issue but perhaps if you add to the ui-tests-app for the case you're expecting we can work together on it.

dosomder:
Yes `attributtedText` is set with the letter spacing property. The problem is in a `TextField` with no text and then the user inputs something in the text field. In that case the properties which were set on `attributedText` do not apply but the `defaultTextAttributes` do. I'll see if I can figure out how to write a test for this

dosomder:
@NathanWalker I hope we can solve this asap

NathanWalker:
@dosomder correct on the text-field-page - exactly - just add a textfield with the styling and perhaps post a PR with that and I can collab with you there to find a suitable fix.

NathanWalker:
That way there's a page in the repo which locks in that behavior and thus a place to test against now but also in the future.

Issue #8580, opened 2020-05-13
getActiveNetworkInfo() and NetworkInfo were deprecated in API level 29


edurenye:
**Describe the bug**
`connectivityManager.getActiveNetworkInfo()` (https://developer.android.com/reference/android/net/ConnectivityManager#getActiveNetworkInfo()) and `NetworkInfo` (https://developer.android.com/reference/android/net/NetworkInfo) were deprecated in API level 29, and we use it in: https://github.com/NativeScript/NativeScript/blob/master/nativescript-core/connectivity/connectivity.android.ts

**Expected behavior**
I'm not sure how to get the network type but to get the SSID we should use https://developer.android.com/reference/android/net/wifi/WifiManager `wifiManager.getConnectionInfo().getSSID()`

NickIliev:
@edurenye confirming this one as a bug that prevents proper detecting of active connections on newer Android API levels.

With the current implementation, devices with newer API levels won't detect that there is active internet connection when the connection type is WIFI.

For example:
- Turn WIFI on the device,
- Turn of Mobile Data

as a result:
`getConnectionType` and `startMonitoring` will both return **connectionType = 0** (no Internet connection) while it should be of type 1 (WIFI connection active).

jarmani:
@triniwiz @NathanWalker I think the issue can be closed

PR #8652, opened 2020-06-17
fix(connectivity): getActiveNetworkInfo and NetworkInfo modern compliance #8580

triniwiz:


NathanWalker:
iOS changes lgtm - @NathanaelA can approve android changes.

triniwiz:
Existing apps using the connectivity module will need to add the following to the `AndroidManifest.xml` which should be located in `app/App_Resources/Android/src/main/AndroidManifest.xml`
```xml
<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
```

farfromrefug:
If i may jump in. I think it would be best not to use the extend syntax (android NetworkCallback). But to use constructor implementation instead. Same behavior behind but in a sense it is more readable (typings) as not seen as class extension. Also it allows to pass object for which instance methods can be used as implementation. Means it prevents instantiating functions everytime. 
Finally you should use method shorthands for the methods definitions for NetworkCallback. Like  'onUnvailable:notifyCallback'

farfromrefug:
@triniwiz as we discussed yesterday, `extend`is the best way to go as it could be the same syntax on iOS.
Rest the point about not creating funcs every time
And using `extend` it means you should store the newly created class for later use.

NathanWalker:
Released with https://github.com/NativeScript/NativeScript/releases/tag/6.5.9 now

PR #8704, opened 2020-07-10
feat(tabs): add animationEnabled property

### LINKED ISSUE #8553: iOS Tabs component issue when changing selectedIndex by tapping the desired Tab + TabContent is RadListView
**Environment**
 - CLI: 6.5.0
 - Cross-platform modules: 6.51
 - Android Runtime: 6.5.0
 - iOS Runtime: 6.5.0
 - Plugin(s): nativescript-ui-listview 8.1.2 (also tested 7.1.0)

**Describe the bug**
When changing the Tab selectedIndex bei tapping on the desired Tab within the TabStrip, the component is not always snapping to the right content (sometimes only 95% of the desired tab-content visible).
ios_tabs_issue_selectedindex

In this case, you have to swipe the content a little bit and tap again, then it works.
The following GIF shows our real app, where you can see this effect easier than in the sample app.
ios_tabs_snapping_fanfiktion

Note: it is only reproducable to us, when using RadlistViews within the Tabs content. We can not reproduce it ourselves when using ListVews instead (but we need RadListView in our real app).
ios_tabs_correct_listview

When swiping through the Tabs (without tapping the TabStripItem), everything seems to work as expected on iOS (even if the Tabs contain RadListViews).

**To Reproduce**
1.) Open the sample app
2.) navigate to the RadListView page 
3.) Tap on a TabStripItem -> the content is changing correctly
4.) Tap on another TabStripItem -> nothing happens

**Expected behavior**
Tabs component should behave correctly when tapping the TabStripItems.

**Sample project**
[ns-tabs-issues.zip](https://github.com/NativeScript/NativeScript/files/4657451/ns-tabs-issues.zip)

rigor789:
## PR Checklist

## What is the new behavior?

implements / fixes #8553

### DIFF
--- a/nativescript-core/ui/tabs/tabs-common.ts
+++ b/nativescript-core/ui/tabs/tabs-common.ts
@@ -45,5 +45,5 @@ export type IOSTabBarItemsAlignment = "leading" | "justified" | "center" | "cent
 export const iOSTabBarItemsAlignmentProperty = new Property<TabsBase, IOSTabBarItemsAlignment>({ name: "iOSTabBarItemsAlignment", defaultValue: "justified" });
 iOSTabBarItemsAlignmentProperty.register(TabsBase);
 
-export const iOSAnimationEnabledProperty = new Property<TabsBase, boolean>({ name: "iOSAnimationEnabled", defaultValue: true, valueConverter: booleanConverter });
-iOSAnimationEnabledProperty.register(TabsBase);
+export const animationEnabledProperty = new Property<TabsBase, boolean>({ name: "animationEnabled", defaultValue: true, valueConverter: booleanConverter });
+animationEnabledProperty.register(TabsBase);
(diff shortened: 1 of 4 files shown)


farfromrefug:
Shouldnt it be better to be `animationEnabled`? If you start using with ios it means that when you ll add android we ll have two separated props. And it won't be east to go back as it will be a breaking change for users

rigor789:
> 
> 
> Shouldnt it be better to be `animationEnabled`? If you start using with ios it means that when you ll add android we ll have two separated props. And it won't be east to go back as it will be a breaking change for users

Right, we were debating the name too - more specifically the `ios` vs `iOS` prefix... we have both across the different properties, and it's a little inconsistent.

nativescript-core/ui/tabs/tabs-common.ts
```diff
@@ -43,4 +43,7 @@ tabsPositionProperty.register(TabsBase);
 
 export type IOSTabBarItemsAlignment = "leading" | "justified" | "center" | "centerSelected";
 export const iOSTabBarItemsAlignmentProperty = new Property<TabsBase, IOSTabBarItemsAlignment>({ name: "iOSTabBarItemsAlignment", defaultValue: "justified" });
-iOSTabBarItemsAlignmentProperty.register(TabsBase);
\ No newline at end of file
+iOSTabBarItemsAlignmentProperty.register(TabsBase);
+
+export const iOSAnimationEnabledProperty = new Property<TabsBase, boolean>({ name: "iOSAnimationEnabled", defaultValue: true, valueConverter: booleanConverter });
```
NathanWalker, line 48: Let's rename to `animationEnabled`

rigor789:
@farfromrefug would you mind checking my changes too? :D

farfromrefug:
@rigor789 you actually implemented it on android :) awesome! Seems perfect to me

Issue #8698, opened 2020-07-08
feat(iOS-dialogs): Setting the size of iOS popup dialog thru CSS or element like Android


slolam:
**Is your feature request related to a problem? Please describe.**
Currently, you need to pass the size of the dialog using the dialog options for iOS and for Android you have set it using CSS or height & width attributes. This has been quite frustrating that I have to make changes in 2 different places for different mobile platforms. From a consistency standpoint, it would be great if we could just use the same mechanism for sizing of dialog for both iOS and Android platform. 

**Describe the solution you'd like**
To make both Android and iOS consistent when it comes to dialog size and piggyback off on CSS or width & height attributes of the dialog root element.

**Describe alternatives you've considered**
The current alternative is to pass extra size information specific to iOS in the dialog options while opening the dialog. 

**Additional context**
Since, I am building my app to work on Android as well so I see the pain to adjust across multiple screen sizes unlike iOS screen aspect ratio.

PR #8710, opened 2020-07-14
feat: (iOS dialog size using CSS) Set dialog size using CSS or attribute like android

slolam:
## PR Checklist

To size a dialog in iOS, one has to pass extra iOS specific attributes to ModalDialogService unlike Android. Android uses CSS or dialog root element attributes to size the dialog.

## What is the new behavior?
Like Android, this change will use CSS or attributes on the dialog root element to determine the size of the dialog. The existing code that uses iOS specific dialog size information passed to ModalDialogService will continue to work. However, any CSS or attribute size will override the existing size information passed using ModalDialogService parameters.

#8698


farfromrefug:
Option width and height should be preferred. Consequently it should be an if else. Also you should handle percentage too as you can compute it from the parent controller

slolam:
> Option width and height should be preferred. Consequently it should be an if else. Also you should handle percentage too as you can compute it from the parent controller

@farfromrefug, I could move the code into else block, however, I wanted make an headway for the people to use CSS & attributes over Options to be consistent with Android. Since, I was working only on NativeScript iOS, I came up with the Options for width & height back then in my previous pull request for dialog size. Now that I work on both iOS and Android I don't appreciate the inconsistency between them both the platforms when it comes to dialog size.

As I said, the purpose is to make it consistent with Android. Android does not support percentage width and height. Trying to do it for iOS does not make a lot of sense. 

Please let me know if you want me to move the code to else block.

Thanks.

farfromrefug:
OK i get it. I would go for android changes to. 
One thing is that when using options width and height I like not to have that handler running for nothing. It can hurt perfs in a sense, not a heavy compute but the sum of little things can turn up to be a big thing!
About percentage same thing would be best to have it on both.

That being said if this PR only brings the same behavior I guess it is good enough. Maybe a todo comment ?

slolam:
> OK i get it. I would go for android changes to.
> One thing is that when using options width and height I like not to have that handler running for nothing. It can hurt perfs in a sense, not a heavy compute but the sum of little things can turn up to be a big thing!
> About percentage same thing would be best to have it on both.
> 
> That being said if this PR only brings the same behavior I guess it is good enough. Maybe a todo comment ?

Added TODO comment, I have already moved the handler code into else block in my previous commit.

slolam:
Hey @farfromrefug, anything you are waiting from my side or this PR is good to go?

farfromrefug:
@slolam sorry I don't the rights to merge your PR right now. Nathan will look at this. Simply a lot of things happening right now ;)

NathanWalker:
Nice work @slolam sorry for delay, appreciate the contribution!

PR #8729, opened 2020-07-23
Feat/monorepo esm

farfromrefug:
This PR moves the core to a full es2017 js build.

* it adds a new `@NativeClass` decorator
* it uses a ts transformer to detect that decorator and transform those classes to es5
* the decorator is removed from the generated es5 (to not require changes to the runtimes)

To apply the transformer using `tsc` we need a modified version of `typescript`. To do that i use [ts-patch](https://github.com/nonara/ts-patch). I added it to the `setup` npm task. That's enough!

Here you go with that the core is build in es2017 ...

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "nx": "nx",
     "format": "nx format:write",
-    "setup": "npx rimraf -- hooks node_modules package-lock.json && npm i && nx run core:setup"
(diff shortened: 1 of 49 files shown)


farfromrefug:
@cla-bot check

NathanWalker:
This is 🔥 @farfromrefug thank you!

PR #8732, opened 2020-07-26
DO NOT MERGE - Feat/monorepo native

farfromrefug:
NativescriptActivity and Fragment are now native


NathanaelA:
**DO NOT MERGE** this until some clarity is provided.   :grinning:   

IMHO, If I understand this change (Which I admit I might not, so I'm asking!!!  :grinning:)  This, IMHO, has a significant chance to break many applications, so before it is merged I want to make sure we don't break any applications, or make this any more difficult.

1st, what is the purpose for making the Activity Native?     This seems to be a negative change unless I'm missing something.

By default the SBG builder will take your JS code and create a native  activity; this makes it EASY to extend the Activity from JS/TS, and not have to use any Java.   Example:

I created:
1. Activity.android.js
image


2. In it is code like so (basically copy the exising JS activity file, and make the changes I needed to it like added additional listeners):
image

By default it generated this NATIVE Activity with my new listeners:
image

WE DO NOT want to lose the easy ability to extend the activity from JavaScript with very trivial code...

farfromrefug:
@NathanaelA you dont loose anything. I dont see why you cant do that with that PR? 
YOu can still extend the NativescriptActivity from JS and same thing with Fragment. 

Now i am not saying we really need this. This was a try with @NathanWalker for es2017. He had some issues with application/activity init with es2017.
This code was made for another try as es2017 without using `@NativeClass`

NathanaelA:
I recall in NS 3 (maybe 2/1?) when the Activity was Java and so making any the changes to the Activity required rebuilding the widgets.   I do not want to go back to the stage where we have to hack the activity in Java to make it work.    I've used the ability to add events multiple times over the years, and just gave a client code about two weeks ago to fix something they were running into using that technique...     So when I see us moving it back to Java it worries me greatly  :grinning: because I remember the significant pain...  :grinning:

farfromrefug:
OK i see. On that tns modules widgets note I would like to get rid of it! I would like us to publish the java / iOS sources directly in the platforms folder of the core (in ui module in a future release where we would split ui) . that no depency on a another module requiring a new build. You see what I mean?

PR #8750, opened 2020-08-07
[android] dont use replace transaction anymore.

farfromrefug:
So this is a long coming fix for me in all my apps.
Kind of in the same vibe as https://github.com/NativeScript/NativeScript/pull/8561

Until now when transitioning between fragment we were using `replace` which means that moving to a new fragment the current fragment was destroyed and re created on back.
This was bad for maps and any opengl rendered views.
Here is an example of what it looked like before
file1

This PR changes that behavior by using `add` and `remove`.
I hacked the transition life handlers a bit the get the entry we are moving to and the entry we are moving from.
Another plus of this is that back navigations should be faster as we dont recreate the views.
We will also keep the positions in any scroll view!!!

Here is what it looks like now:

file

### DIFF
--- a/packages/core/ui/frame/fragment.transitions.android.ts
+++ b/packages/core/ui/frame/fragment.transitions.android.ts
@@ -31,10 +31,11 @@ interface ExpandedTransitionListener extends androidx.transition.Transition.Tran
 
 interface ExpandedAnimator extends android.animation.Animator {
 	entry: ExpandedEntry;
+	backEntry?: BackstackEntry;
 	transitionType?: string;
 }
 
-interface ExpandedEntry extends BackstackEntry {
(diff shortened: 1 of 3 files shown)


farfromrefug:
Thanks would be good to test it a bit. The ui app did not that many tests for navigation. And the navigation app was not building here

NathanaelA:
@farfromrefug - This is being manually reverted out; it breaks navigation in multiple places.     We really want the smoother look (which looks awesome!).   But it can't break the unloaded/loaded events for elements inside the fragments as this is behavior many app rely on.  

The Toolbox app that runs through a million tests; it will fail in random places when this is implemented; after adding a lot of logging, finally found out that unloaded and loaded weren't being called properly on a lot of the navigation.

PR #8787, opened 2020-08-23
Feat/monorepo fix fragment transitions 2

farfromrefug:
Another try at fixing this.
Tested with clearHistory and multiple in between fragments while going back.

One note though on this: `unloaded` event wont be called anymore  on fragment 1 when going "forward" from fragment 1 to fragment 2.  The reason is simply that fragment 1 is not unloaded anymore (so that we dont see the black thing while going back).

@NathanaelA can you guide on how to test this the way you did?

### DIFF
--- a/apps/automated/app/ui/page/page-tests-common.ts
+++ b/apps/automated/app/ui/page/page-tests-common.ts
@@ -181,7 +181,7 @@ function _test_PageNavigation_EventSequence(withTransition: boolean) {
 	helper.navigateWithEntry(navigationEntry);
 	helper.goBack();
 
-	const expectedEventSequence = ['navigatingTo', 'loaded', 'navigatedTo', 'navigatingFrom', 'unloaded', 'navigatedFrom'];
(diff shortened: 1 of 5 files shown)


farfromrefug:
I tested this on iOS and Android with modals and normal pages and everything seems fine now.
Why did i test on iOS? because of that code https://github.com/NativeScript/NativeScript/pull/8787/commits/a6336b082c2e261b6813fd580af459030471d3e7 which was also running on iOS. that code does not seem necessary anymore. In all my tests the unloaded event was called correctly. On iOS was already done in `viewDidDisappear`. On Android it is correctly called from `Fragment` `onDestroy`

Now a side note on the fact that now the behavior is different on iOS and Android. On iOS pages are unloaded when navigating forward. I dont see any side effect of that but we might have to look at this in the future for custom transitions
On android that PR makes it so that pages are NOT unloaded when navigating forward to get correctly transition animations and no "black effect"

NathanWalker:
@farfromrefug thanks for this - can you target `master` now with this PR 👍

PR #8791, opened 2020-08-27
[android] fragment transactions

farfromrefug:
This PR changes the behavior of android fragment transactions to use `add` instead of `replace` on forward navigation

It is a breaking change in the sense that it changes the internal behavior of Android navigation:
* while navigating forward, the page navigated from is not unloaded anymore
* events order is changed in the sense that now `unloaded` happens after `navigatedFrom` instead of before

There are multiple plus side o this:
* no more black views on navigation when using opengl (maps, ...)
* navigation is faster, especially the navigation back! We dont need to recreate the page anymore. But navigation forward also gets faster as we dont unload the previous page anymore.
* `navigatedFrom` event happens faster

This the default behavior used by most of the android native apps.


NathanaelA:
What is the difference between this an #8793, if they are the same, can you close one of them and we can discuss any issues on the correct issue.  :)

farfromrefug:
They are not related and do not fix the same issue.
The other one is about fragment released too soon and not following android lifecycle

NathanaelA:
Ok, thanks for the clarification.  :grinning:   I really like the idea of this patch...

For this one, I would like to see if we can get this into NS 7.1 or 7.2   

I honestly have serious misgivings about the unloaded event not being fired.   And I understand the reasons, however I am willing to run several solid client Angular, Vue and PAN apps to see if any failures occur; if all our client apps work perfectly; then I'll be happy.   In addition I want to make a change to my `event` testing app to test a couple corner cases that concern me and see what happens, because I think their might be a couple corner cases that still need to be handled.     If everything works; we will merge.  If things fail, we will post back what the issue is so that you can resolve it first.     :grinning:

   However, since we are targeting release of 7.0 for next Monday I honestly do NOT have the time to do a deep dive into testing each of these apps with this change and get the other things on my plate finished.   :frowning_face:

farfromrefug:
Merge whenever you want. Already using it in multiple prod apps from my fork ;)

farfromrefug:
@NathanWalker i guess you are looking at those right? again jut want to confirm i have multiple productions app running this.

farfromrefug:
@rigor789 @NathanaelA @NathanWalker Thanks to that PR and a new commit https://github.com/Akylas/NativeScript/commit/a4d7674abaac9f95f91112d0611d4f94eeb543dd
I managed to make Maps work correctly with fragment transitions.
This is what we get today without that PR for `GLSurfaceView`. It is known that `GLSurfaceView` opacity cant be animated.
https://user-images.githubusercontent.com/655344/108607902-a345a700-73c3-11eb-82a8-aa143fb74a7b.mp4

This is what we get today without that PR for `GLTextureView`. There is "blink" on closing fragments with `GLTextureView`.  It seems to happen only when using android `Transition`
https://user-images.githubusercontent.com/655344/108607939-ce2ffb00-73c3-11eb-9e7a-7b8eb8bb4f32.mp4

Now this is what i get with `GLTextureView` and this PR and the latest commit
https://user-images.githubusercontent.com/655344/108607967-f91a4f00-73c3-11eb-8969-b00b7835b664.mp4

Both the PR and the commit are needed to achieve this. I hope you are going to look at this.

NathanWalker:
@farfromrefug we are going to get this into 8.1 - there's only 1 conflicted file related to automated test, could you just resolve that one conflict and then we can merge into a 8.1 prep branch we are working through.

farfromrefug:
@NathanWalker great to hear. Will fix it on monday. Soon enough?

NathanWalker:
Absolutely thank you!

vallemar:
I think it would be important to review this pr since it was reverted, the views with maps emit a flash with the navigation back

Siergiej29:
After all these years, there's still nothing new on this topic. Maybe it would be worth implementing this ready-made solution?

PR #8792, opened 2020-08-27
onListenerAdded and onListenerRemoved

farfromrefug:
really useful to for example turn on gps when a `location` event is added

### DIFF
--- a/packages/core/data/observable/index.d.ts
+++ b/packages/core/data/observable/index.d.ts
@@ -163,6 +163,19 @@ export class Observable {
 	 */
 	public _isViewBase: boolean;
 	//@endprivate
+
+	/**
+	 * This method is intended to be overriden by inheritors to be notified when a new listener is added
+	 * count is the number of listeners for that event after addition
+	 */
+	onListenerAdded(eventName: string, count: number);
+
+	/**
+	 * This method is intended to be overriden by inheritors to be notified when a new listener is removed
+	 * count is the number of listeners for that event after removal
+	 */
+	onListenerRemoved(eventName: string, count: number);
+	
 }
 
 /**
--- a/packages/core/data/observable/index.ts
+++ b/packages/core/data/observable/index.ts
@@ -98,7 +98,9 @@ export class Observable implements ObservableDefinition {
 	public off(eventNames: string, callback?: any, thisArg?: any): void {
 		this.removeEventListener(eventNames, callback, thisArg);
 	}
-
+	onListenerAdded(eventName: string, count: number) {}
+	onListenerRemoved(eventName: string, count: number) {}
+	
 	public addEventListener(eventNames: string, callback: (data: EventData) => void, thisArg?: any): void {
 		if (typeof eventNames !== 'string') {
 			throw new TypeError('Events name(s) must be string.');
@@ -117,6 +119,7 @@ export class Observable implements ObservableDefinition {
 				callback: callback,
 				thisArg: thisArg,
 			});
+            this.onListenerAdded(event, list.length );
 		}
 	}
 
@@ -143,9 +146,11 @@ export class Observable implements ObservableDefinition {
 						delete this._observers[event];
 					}
 				}
+				this.onListenerRemoved(event, list ? list.length : 0);
 			} else {
 				this._observers[event] = undefined;
 				delete this._observers[event];
+				this.onListenerRemoved(event, 0);
 			}
 		}
 	}


NathanaelA:
The more I think about this the more I don't see the reason to create a slight performance hit for **all** events for a single use case.   For this single use case you just do this in that specific location component:
```
addEventListener(event, ...) { 
   super.addEventListener(event, ...)
   // Do whatever logic you decide to acivate gps...
}

removeEventListener(event, ...) {
  super.removeEventListener(event, ...);
  // Do whatever logic you decide to de-activate gps...
}

```

So all you are doing is chaining the `addEventListener` instead of a new `onListenerAdded`, code works identically; but know slow downs added to the event system....

farfromrefug:
Wow closed pretty quickly. Like always the point is to shared same logic between components instead of rewriting it all the time. 
Already use this as a component for 4 components. 
Seemed best to be in the core...

NathanaelA:
Yep, trying to clean up PR's and/or merge them.  ;-)    Hence the other PR I asked you about which one was which one; so we could close the unneeded one was because I want to try and make sure to merge PR's asap if possible in the future -- I hate PR's getting stale or ignored, which was a big issue in the past...  :grinning:

---

However in this specific case you gave me a single example.   For a single example we for sure aren't adding a performance hit to the **entire event system**.     Remember with your change every single add/remove of an event will have to call what 99.99% of the time will be a empty function with parameters, this is overhead that is unneeded 99.99% of the time, buttons, navigation, just think of ALL the events that are added/removed during just setup, let alone during app running and switching pages.    So this was very easy to close with you only providing a single example and their being a very simple way to do the exact same thing without effecting performance for the rest of the system...

Now with that said, can you tell me or point me to the code for the other three components; because 4 is a bit stronger of a reason, however to add any event performance hit we really want to have to have some **awesome** justification. 

 However in all four depending on what they are you sill could have:  (assuming a view, but you could replace View with the lower component in the tree if this isn't a view component.)
``` 
class ListenerNotifierView extends View { 
    addEventListener() { super.addEventListener(); onAddedEventListener(); }
    onAddedEventListener() { }
}
```
then
Your component extends from ListenerNotifierView like so:
```
class someCoolViewItem extends from ListenerNotifierView {
   onAddedEventListener() { /* do whatever magic you need */ }
}
```

Then the performance hit is only seen by the components that need it which should be minimal verses the entire event system and you are sharing your code.  ;-)

farfromrefug:
OK about cleaning up PRs. Would be good to wait a bit more for just created PRs. Others to clean first I think.

Now what performance hit are you talking about? Calling an empty function with const parameter has no consequence at all I think. Plus it is not like addEventListener is call a thousand times.
And your `onAddedEventListener` won't solve anything. The whole point is to have the count to for example enable / disable GPS or Bluetooth state broadcast depending on count.
I use that already in my gps comp and nativescript-bluetooth.
I know I have other but don't have the remember which ones right now.

But don't merge it if you think it won't benefit others. I have a plugin doing it. However it has a perf hit and an api issue as I have to use private api and count again even though as you can see in this PR it is a one liner in the core.

NathanaelA:
> Now what performance hit are you talking about? Calling an empty function with const parameter has no consequence at all I think.

ALL calls to functions in JavaScript (or any language that I'm aware of) have setup and tear down costs, even empty functions are still considered a function and have actually significantly more costly than any other standalone statement.   

FYI for actual costs: A EcmaScript engine running a function is here: https://tc39.es/ecma262/#sec-ecmascript-function-objects-call-thisargument-argumentslist  (9.2.1) which included almost all steps in 9.2.1.1 thru  9.2.1.3 and any sub-links to other description actions - as you can see this is NOT a zero cost item, in fact it is very costly compared to almost any thing else, so to add this to EVERY single Add/Remove is something we must balance with what it actually offers.     And yes I realize their are not 1000's of calls.   But every single one that is called, adds time, which does add up.   So why add this extra time for two(?) plugins that can TRIVIALLY do it themselves and not add any additional time to everybody else. 

This is why you see in many places in the NS codebase a `if` statement wrapping a diagnostic call (for example almost all Trace calls does a `if` check BEFORE calling he trace function, the cost of a `if` even though it adds some code bloat, is so significantly faster than moving the if inside the trace functions.

---

>OK about cleaning up PRs. Would be good to wait a bit more for just created PRs. Others to clean first I think.

Well, unless the PR has a DO NOT MERGE or WIP; this should mean the PR being pushed here is ready for merging otherwise it should stay on your local tree until you are happy with it.  :grinning:   Both Nathan and I were cleaning PR's and asking people to rebase some of them to the new master...  

---
> And your onAddedEventListener won't solve anything. The whole point is to have the count to for example enable / disable GPS or Bluetooth state broadcast depending on count.

Why do you need a count on add, if someone added a event, then you know you need to trigger the enable on GPS, right????   And being as good of a developer as you are, what is stopping you from calling `this._getEventList()` from the  `removeEventListener` override class function and then using your exact same code for the count.      The cost to your plugin is minuscule, as how frequently is removeEventListener really called on your location or bluetooth plugin?     So nothing that I can see is stopping you from creating a simple wrapping class like I was summarizing, and as such for a single (or two) use case -- my suggestion seems like still the easiest way to create shared code that does exactly the same thing with no performance penalty to the rest of the event system.    And if it really is two plugins, then just drop the same "add/remove" code in both plugins; you don't have to share code and sometimes it is overkill for only a couple instances and just creates more complex code to maintain and more moving pieces...


---

>But don't merge it if you think it won't benefit others. I have a plugin doing it. However it has a perf hit and an api issue as I have to use private api and count again even though as you can see in this PR it is a one liner in the core.

I do understand your use case.    I don't see how two trivial cases can justify a performance hit for the entire system.   As for your perf hit; do you really need a count on `add` what really is the difference between 1 and 10 for add?   You can mitigate pretty much any perf hit in your code on Add by eliminating the count.   On Remove; unless you think people are going to be calling add/remove frequently; then your remove's count performance hit should be insignificant as it would only still be called so rarely it won't matter..  

 That is the difference in the proposed solutions, my suggestion has a insignificant hit in a two rarely used calls; your PR adds a small hit to every single add/remove in the entire application...

A PR merged should offer to fix something or additional functionality; however their are pieces of the system you have to be careful and balance features with extra code and/or performance penalties.    In this case the penalty actually is very small; but for two use cases that you raised -- (which I do think the use cases are valid!); but the plugin can easily do it itself with close to zero performance impact to anything.    So, balancing a your use cases vs adding a minor performance hit seems to be a still like the balance falls on don't merge pretty strongly imho.

PR #8793, opened 2020-08-27
fix(android): dont dispose fragment on onloaded

farfromrefug:
As explained here #6659
That hack has bad consequences.
Also i actually cant see any wrong behavior using the sample provided in the issue and upgrading to latest N.

### DIFF
--- a/packages/core/ui/frame/index.android.ts
+++ b/packages/core/ui/frame/index.android.ts
@@ -241,11 +241,6 @@ export class Frame extends FrameBase {
 
 	onUnloaded() {
 		super.onUnloaded();
-
-		// calling dispose fragment after super.onUnloaded() means we are not relying on the built-in Android logic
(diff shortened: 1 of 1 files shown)


NathanaelA:
Thanks for the PR; I need to run some tests before we can actually merge this -- as the prior issue was a crashing issue which you now say it causes some slow downs with the patch.   I would rather have some slower code (i.e. having to redraw) than crashes.   So merging something that undoes that and might bring back crashes; needs some good testing just to verify that it will not create a new crashing issue.  :grinning:

farfromrefug:
The initial crashing sample is not crashing anymore with this PR. At least I could not get it to crash

DimitrisRK:
The removal of this line of code fixes issue #8828 where Frames are used as content items in BottomNavigation.

NathanWalker (approved):

PR #8797, opened 2020-08-28
fix(ios): stop using artifical state handler via animated setter on uiviewcontroller

NathanWalker:


farfromrefug:
Can you explain the reason of that change ? Seems weird and less error prone than storing it in the corresponding controller

NathanWalker:
@farfromrefug what was being done here is considered bad practice with {N}. `animated` setter on a uiviewcontroller instance has no meaning and was merely being temporarily stored there for usage when modal closes as a matter of state. Further using a generic property on a native instance object is dangerous as any arbitrary property could end up being *readonly* as was the case here and indeed caused on app crash under v8 iOS.

farfromrefug:
I knew it was related to a crash in the v8 runtime :). I can get the fact we should not use existing native properties though i would prefer the crash being handled on the V8 side ;) 
Thanks for the explanation. At least now it is commented here why this change was made (which can get important in the future)

PR #8808, opened 2020-09-02
default typings for android to 29

farfromrefug:
most plugins / apps depends on newer APIs than 17. Plus it wont break anything, will just add more typings (i think)

### DIFF
--- a/packages/types-android/src/lib/android.d.ts
+++ b/packages/types-android/src/lib/android.d.ts
@@ -1,1 +1,1 @@
-/// <reference path="./android-17.d.ts" />
+/// <reference path="./android-29.d.ts" />


NathanWalker:
I've wanted this for awhile 👍  -- @triniwiz @NathanaelA thoughts?

triniwiz:
I've thought about this for a while but the one issue or a couple that may arise is if devs do not look up what api is support at what version ... other than that LGTM 👍

NathanaelA:
I have to say this is probably **not** a good idea for the general users.   The issue is that it can easily cause someone to use a recent API as their are a lot of newer API's after 17 and if you use **any** of them on a app that remains targeted to Android 17 (the default for all NS app's), it will either fail to build with weird build errors or much more likely just crash the during the attempted use of the API.   So, imho, his would probably not be something worth doing at this point in time on the main references file.

 HOWEVER this would be awesome in the new plugin seed and make sure the plugin seed has in the main readme that it is using the latest api and that if they use anything from a later API; they will need to ship a gradle file that targets the later API so that the plugin can force the app to a later version..

---

Another,  idea that just occurred to me for a future nativescript feature is since we already scan the source code for sbg building.   
- We create some "diff" files of the apis that are available in API 18+ so that when the SBG is scanning the source code if it detects a api that it knows is newer than the current target it can throw a build error and tell them how to fix it easily; then this could be easily always be set to always to the latest API.     

But at this point, weird unknown cause errors are something I think we want to avoid...

farfromrefug:
I can get the idea. What it says to me is that we should (but it is a lot of work) would be to auto generate typings decorators to warn the users. In android studio that is what happen.
Btw about the target thing. From now on we are forced to target 29 :(

Just to explain where that PR comes from. I use "pnpm" which broke with the new typings because of the way native types are referenced within @nativescript/types. That s why I stumbled upon this. @nativescript/types is unusable for most plugin devs as it points to very old android APIs.
No easy solution but I have to solutions still. The best solution is to not use @nativescript/types but @nativescript/types-ios and android directly. So I ll do that In all my plugins / apps (even though I like the idea of @nativescript/types).

NathanaelA:
@farfromrefug - Martin, many of your statements really confuse me...

> I can get the idea. What it says to me is that we should (but it is a lot of work) would be to auto generate typings decorators to warn the users. In android studio that is what happen.
Btw about the target thing. From now on we are forced to target 29 :(

The number of people who want their apps to support even earlier API's than NS actually supports is still amazing to me, so defaulting the API to 29 would just create a massive amount of additional support.     So bumping the minimum in the build is not something that would be (imho) worth it, we would get so many people asking why the app no longer runs on Android 5, 6, 7, 8, and 9...   So if you are really proposing we bump the target; that would be suicide by issues.  :grinning:    

IMHO, to keep support issues to the minimum; since we allow the app to run on Android 4.4, the best typings to continue to DEFAULT use to eliminate any confusion for new users is the typings that match the target.  :grinning:  


> Just to explain where that PR comes from. I use "pnpm" which broke with the new typings because of the way native types are referenced within @nativescript/types. That s why I stumbled upon this. @nativescript/types is unusable for most plugin devs as it points to very old android APIs.

Not sure why pnpm would break it; because pnpm should retain the structure of @nativescript/types which even your PR still references; the only change in your PR was to change from 17 to 29, no pathing differences, so I'm totally confused as to how this PR would change anything in your environment related to pnpm...  

>No easy solution but I have to solutions still. The best solution is to not use @nativescript/types but @nativescript/types-ios and android directly. So I ll do that In all my plugins / apps (even though I like the idea of @nativescript/types).

This confuses me greatly; I don't see how this PR actually changes anything.  If `@nativescript/types` is not installing correctly but `@naivescript/types-android` is -- then how exactly is pnpm actually working, that seems like really buggy behavior?     And why is this pr, for `@nativescript/types` just to bump from 17 to 29?    I can't figure out how this PR would help any buggy pnpm behavior???

Finally why don't your just update your own references.d.ts inside your own app/plugin to directly point to api-29, then you already know the risk if you use a API 29 call; your app will only run on Android 10.

farfromrefug:
@NathanaelA i have to disagree on your point explaining why we should stick to 17. Most users of typings are actually plugin devs. Those kind of know (and test ) what works and not. And most of the app/plugins will want to use features from 21 at least. And if (and most of us do) want to support .older sdk, they "if" for it. 
Honestly i dont see any difference in setting 17 or 29 except for us plugin dev. You wont get much more support from this. 
And at one point you will have to do it. When then?

Now you talked about "targeting 17" i might have misunderstood but today all android apps must "target" 29. It is a requirement.

About the rest of my explanation it is not directly related to that PR. It explained why i ended up here and why @nativescript/types are not that useful as they target 17.  None of my plugin can really use it because of that (and because my plugins use some feature  > 17, like most).

About pnpm issue it is related to using "path" in typescript references but not only. `@nativescript/types-ios` and android are now not "direct" dependencies, thus pnpm does not put them at root, thus typescript reference cant find them either using "path" or "types".

I think the whole "should we merge that PR" is not related to how many people would get confused by api not working because of the target, but to how many projects(apps, plugins) can actually use `@nativescript/types` (and thus NOT use ANY api > 17). My POV is really not a lot

NathanaelA:
I have worked with multiple clients who have me review their code and the vast majority of them do NOT create plugins, but just use the api's natively inside their app (and many of them lots of api calls).   Half of them didn't even realize how simple it was to refactor and make all this "native" code into a plugin.    These are the people I worry about, not the power users like you or me -- we can spend two minutes and make our environment match what we want.       They are the people who will create issues when the app starts crashing because they are unaware that gradle by default is targeting API 17 as the minimum sdk it can run on.   This means that SDK items from 18+ can't be used in the app unless they if/case them and verify they are running on a device using API of that level.    
  
We DO NOT get the compile time safety with NativeScript so if they are using something from API 20 in the JS code, then they will now see weird crashes about X being missing on Android 19.   Rather than the compiler saying hey you have API 20 here, and you are targeting API 17; they see a crash and then that will become a support issue that someone has to deal with...

This is also the reason I suggested the PLUGIN SEED be bumped to API 29+ as these are the users who probably have some idea about how things work and min-sdk vs running skd and what can cause certain issues, so bumping the new plugin seed with API 29+ is a good compromise for everyone imho.


> Now you talked about "targeting 17" i might have misunderstood but today all android apps must "target" 29. It is a requirement.

This is what I'm talking about, inside the gradle file:
image
The min-sdk is extremely important; this means you can freely use API 1 - 17; anything past 17 has to be special cased to detect if the app is on a device that actually supports that API.    If we bump this value to 29; then the app will not run on any devices below Android 10.  So imho, keeping the default API === min-sdk is the most logical to eliminate any support issues....


> I think the whole "should we merge that PR" is not related to how many people would get confused by api not working because of the target, but to how many projects(apps, plugins) can actually use @nativescript/types (and thus NOT use ANY api > 17). 

Unfortunately, we do have to consider the community at large and not the power users with some decisions.   And since it is two minutes of your time to fix your environment to target api-29 and NO **additional support** for anyone else vs merging something that WILL generate support issues; I tend to lean very much towards we don't want to deal with the additional support issues this will create. 

---

For that matter, are you aware you can create your own templates.   Clone the blank template of your favorite flavor; bump it to API 29, then use it for your `tns create projectName --template=@farfromrefug/blank` and then everyone is happy, your apps default to API 29, and their is no extra support issues.  :grinning:

NathanWalker:
I do agree with @NathanaelA arguments here - I recently started using android 29 in plugins and it makes a lot of sense there however given {N} framework usage across many different uses cases probably better to support lowest possible we can without creating huge maintenance burden to the TSC but we can probably do a better job making it more clear how to increase the android typings to use newer api target if a team wanted to. It's been stated in some blog posts but perhaps should be stated up front with all project creation that setting that to their desired typings for their development scenario would be nice and clear.

farfromrefug:
OK i get it.

PR #8809, opened 2020-09-02
fix(core): type collisions with namespace

NathanWalker:

Currently while using `import { Enums } from '@nativescript/core';` in es2017 there was no clear way to delineate between `type` definition and `namespace`/`module` as they were overloaded throughout.

```
import { Enums } from '@nativescript/core';

const left = Enums.TextAlignment.left; // works
const leftAssignment: Enums.TextAlignment = 'left'; // doesn't work
```

There was a workaround that you could do:

```
const left: typeof Enums.TextAlignment;
```

...but that's pretty inconvenient.

Also the barrel was missing a few ui property getter/setters for plugin usage.

## What is the new behavior?

`Type` has been added to suffix any exported `type` named symbols for better clarity and usage while allowing the `namespace`/`module` to work as normal as well.

```
import { Enums } from '@nativescript/core';

const left = Enums.TextAlignment.left; // works
const leftAssignment: Enums.TextAlignmentType = 'left'; // works
```

This also adds the additional getter/setters for plugin usage:

```
import { 
  letterSpacingProperty, 
  textAlignmentProperty, 
  textDecorationProperty, 
  textTransformProperty, 
  whiteSpaceProperty, 
  lineHeightProperty 
} from '@nativescript/core';
```

/cc @PeterStaev


NathanWalker:
There's a few other adjustments to make here so will leave this open for few days to vet through the remaining areas.

NathanWalker:
This has been published as an `rc` to begin some testing against plugins and other projects: `7.1.0-rc.0`

PR #8811, opened 2020-09-03
fix(webpack): apply-css-loader not applying css

rigor789:
Now that everything is using ES modules, the `exports` object no longer exists.

`css-loader` adds the `___CSS_LOADER_EXPORT___` variable to the module, and then exports it. This allows us to hook into it, and apply the css to a NativeScript app. 

This fixes Vue apps in NativeScript 7. See https://github.com/nativescript-vue/nativescript-vue/issues/715

I believe `require("@nativescript/core/ui/styling/style-scope")` is no longer needed - because globals are handled differently (testing the change in an app works fine) - but @NathanWalker please take a look!

The change has been released under `3.0.1-rc.0`.

### DIFF
--- a/packages/webpack/helpers/apply-css-loader.js
+++ b/packages/webpack/helpers/apply-css-loader.js
@@ -27,7 +27,7 @@ module.exports = function (content, map) {
 
     content += `
     const { Application } = require("@nativescript/core");
-    //require("@nativescript/core/ui/styling/style-scope");
+    require("@nativescript/core/ui/styling/style-scope");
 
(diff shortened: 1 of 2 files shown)


NathanWalker (approved):

romandragan:
Seems like broken again in 3.0.2 release.

rigor789:
@romandragan has not been released yet - only in `3.0.1-rc.0`

rigor789:
Released in 3.0.3!

PR #8823, opened 2020-09-07
Npm scripts

farfromrefug:
This PR brings back npm scripts for people not wanting to use `nx`. 
It also brings a nice change in the sense that it allows you to use npm/pnpm/yarn as your package manager.

for nativescript(root) 
```sh
npm config set nativescript:npm_alias pnpm
```

for webpack 
```sh
npm config set @nativescript/webpack:npm_alias yarn
```

I needed to change the transformer path for the core for it to build.
I am not sure how it could work with `nx` before. `nx` simply does not work here for me so i cant really test that the core still builds.

I did not port the nx test commands as they dont work anymore ( jest config files removed).

Also to point out(so that it can be found later) that `nx` does not work with pnpm or yarn. It seems not to work with their node_modules structure.

PS: Maybe we should now change the nx commands to simply run `npm run ...` now? That way people using `nx` would also benefit from choosing npm/pnpm/yarn. And it would "unify" commands among tools.


packages/core/tsconfig.lib.json
```diff
@@ -17,7 +17,7 @@
     "rootDir": "./",
     "types": ["node"],
     "plugins": [
-      { "transform": "../../../packages/webpack/transformers/ns-transform-native-classes.ts", "type": "raw" }
```
NathanaelA, line 20: Are you sure his path is supposed to change?

NathanaelA (commented):
1. What happens when you do a "npm start" do the commands still work from the npm menus?
2. What happens when you do a "npm run core:build" does it create the output .tgz file properly?

packages/core/tsconfig.lib.json
farfromrefug: That what I asked. Now that I know why nx did not work I can try

farfromrefug:
core:build does not generate the tgz, core:build.npm does and it does with npm scripts
Npm start did not change

farfromrefug:
BTW some scripts actually works better now with npm scripts (could be ported to nx). For example types build commands clear dist package before building it to avoid issues

farfromrefug:
any news on this? the more ways to build you add the easier you make it for people to contribute

NathanWalker:
Closing in favor of this simplified approach which is now on master:
https://github.com/NativeScript/NativeScript/commit/d7b2c84da87f9f9381d04886a3ddb0c3f031abfa

PR #8827, opened 2020-09-08
feat: boolean to disable systemAppearanceChanged (theme)

farfromrefug:
With es2017 `@nativescript/theme` wont work because of an error like that 
```
Cannot set property systemAppearanceChanged of #<Object> which has only a getter
```

Now i think it is best to not "override" that method but to have a boolean to disable


NathanWalker (approved):
👍

PR #8830, opened 2020-09-09
feat: global support for Global Events

NathanaelA:
## PR Checklist

Using "Tap" as the example event, but this actually applies to **ALL** events.

The new Global Events, does not allow you to capture __ALL__ "Tap" events easily, it is currently setup to allow you to only get all "Tap" events on a specific class you use its `on` function.  So if you do `Button.on('tap', ...)` you will catch all Tap events on all Buttons, but no other elements "Tap" events will be captured without you manually adding them also using there `on` handler.  

## What is the new behavior?
This updates the code so you can now ALSO do `Observable.on('tap', ...)` and you will see ALL `Tap` events on ALL elements since you are registering the event on the base observable class.

### DIFF
--- a/packages/core/data/observable/index.ts
+++ b/packages/core/data/observable/index.ts
@@ -36,9 +36,7 @@ export class WrappedValue implements WrappedValueDefinition {
 
 const _wrappedValues = [new WrappedValue(null), new WrappedValue(null), new WrappedValue(null), new WrappedValue(null), new WrappedValue(null)];
 
-
 const _globalEventHandlers = {};
-
 
 export class Observable implements ObservableDefinition {
 	public static propertyChangeEvent = 'propertyChange';
@@ -163,14 +161,14 @@ export class Observable implements ObservableDefinition {
 			throw new TypeError('callback must be function.');
 		}
 
-		const eventClass = this.name;
+		const eventClass = this.name === 'Observable' ? '*' : this.name;
 		if (!_globalEventHandlers[eventClass]) {
 			_globalEventHandlers[eventClass] = {};
 		}
 		if (!Array.isArray(_globalEventHandlers[eventClass][eventName])) {
 			_globalEventHandlers[eventClass][eventName] = [];
 		}
-		_globalEventHandlers[eventClass][eventName].push({callback, thisArg, once: true});
+		_globalEventHandlers[eventClass][eventName].push({ callback, thisArg, once: true });
 	}
 
 	public static off(eventName: string, callback?: any, thisArg?: any): void {
@@ -186,10 +184,12 @@ export class Observable implements ObservableDefinition {
 			throw new TypeError('callback must be function.');
 		}
 
-		const eventClass = this.name;
+		const eventClass = this.name === 'Observable' ? '*' : this.name;
 
 		// Short Circuit if no handlers exist..
-		if (!_globalEventHandlers[eventClass] || !Array.isArray(_globalEventHandlers[eventClass][eventName])) { return; }
+		if (!_globalEventHandlers[eventClass] || !Array.isArray(_globalEventHandlers[eventClass][eventName])) {
+			return;
+		}
 
 		const events = _globalEventHandlers[eventClass][eventName];
 		if (thisArg) {
@@ -221,7 +221,6 @@ export class Observable implements ObservableDefinition {
 		if (keys.length === 0) {
 			delete _globalEventHandlers[eventClass];
 		}
-
 	}
 
 	public static addEventListener(eventName: string, callback: any, thisArg?: any): void {
@@ -233,41 +232,52 @@ export class Observable implements ObservableDefinition {
 			throw new TypeError('callback must be function.');
 		}
 
-		const eventClass = this.name;
-		 if (!_globalEventHandlers[eventClass]) {
-		 		_globalEventHandlers[eventClass] = {};
-		 }
-		 if (!Array.isArray(_globalEventHandlers[eventClass][eventName])) {
-			 	_globalEventHandlers[eventClass][eventName] = [];
-		 }
-		 _globalEventHandlers[eventClass][eventName].push({callback, thisArg});
+		const eventClass = this.name === 'Observable' ? '*' : this.name;
+		if (!_globalEventHandlers[eventClass]) {
+			_globalEventHandlers[eventClass] = {};
+		}
+		if (!Array.isArray(_globalEventHandlers[eventClass][eventName])) {
+			_globalEventHandlers[eventClass][eventName] = [];
+		}
+		_globalEventHandlers[eventClass][eventName].push({ callback, thisArg });
 	}
 
-	public notify<T extends EventData>(data: T): void {
-		const eventClass = this.constructor.name;
+	private _globalNotify<T extends EventData>(eventClass: string, eventType: string, data: T): void {
+		// Check for the Global handlers for JUST this class
 		if (_globalEventHandlers[eventClass]) {
-			const event = data.eventName + "First";
+			const event = data.eventName + eventType;
 			const events = _globalEventHandlers[eventClass][event];
 			if (events) {
 				Observable._handleEvent(events, data);
 			}
 		}
 
-		const observers = <Array<ListenerEntry>>this._observers[data.eventName];
-		if (observers) {
-			Observable._handleEvent(observers, data);
-		}
-
-		if (_globalEventHandlers[eventClass]) {
-			const events = _globalEventHandlers[eventClass][data.eventName];
+		// Check for he Global handlers for ALL classes
+		if (_globalEventHandlers['*']) {
+			const event = data.eventName + eventType;
+			const events = _globalEventHandlers['*'][event];
 			if (events) {
 				Observable._handleEvent(events, data);
 			}
 		}
 	}
 
+	public notify<T extends EventData>(data: T): void {
+		const eventClass = this.constructor.name;
+		this._globalNotify(eventClass, 'First', data);
+
+		const observers = <Array<ListenerEntry>>this._observers[data.eventName];
+		if (observers) {
+			Observable._handleEvent(observers, data);
+		}
+
+		this._globalNotify(eventClass, '', data);
+	}
+
 	private static _handleEvent<T extends EventData>(observers: Array<ListenerEntry>, data: T): void {
-		if (!observers) { return; }
+		if (!observers) {
+			return;
+		}
 		for (let i = observers.length - 1; i >= 0; i--) {
 			const entry = observers[i];
 			if (entry.once) {


packages/core/data/observable/index.ts
```diff
@@ -233,41 +232,52 @@ export class Observable implements ObservableDefinition {
...
+			const events = _globalEventHandlers['*'][event];
+			if (events) {
+				Observable._handleEvent(events, data);
+			}
+		}
+	}
+
+	public notify<T extends EventData>(data: T): void {
+		const eventClass = this.constructor.name;
+		this._globalNotify(eventClass, 'First', data);
```
NathanWalker, line 267: Is that ‘First’ intentional?

farfromrefug:
I have to say it. It seems to me that that PR has much more performance consequence s than my PR on Observable that was refused out of performance worries.
And I think it would be even less used.
What is the use case for this?

That being said I don't think it is the best approach on how to write this. A lot of duplicated intelligence. 
I would simply hold a static "Observable" instance created only when the static on method is called.
And then simply call notify on that static instance from the Observable notify method.
Also isn't the static off method missing ?
Less code and much simpler logic

packages/core/data/observable/index.ts
NathanaelA: Yes, the `First` is intended.     Global events has always supported you getting notified BEFORE the default handler, so events happen in this order:

`TapFirst` <- Global Handler 
`Tap` <- The Normal Class Handlers that everyone uses on each Object
`Tap` <- Global Handler  

If you have a better name for "First" I have no issues switching it.   This is an event that is needed for one or two of my heavily used plugins as they needed to catch the event before the normal handler ran.  In addition, a couple people had also requested this feature over the period of time for stuff they were doing so I don't know how they were using it (I had never documented it initially, so people contacted me on how to do it...)     

FYI, this part of the code already exists in NS7 (i.e. `First`).  These changes was to clean up the code to be much easier to maintain and mainly to allow the developer to catch ALL event of a type on all elements so that they can do analytics much easier as they can also now tie onto a event globally for all elements rather than just globally per element.

rigor789: I think this would be a great opportunity to bring event handling closer to DOM-spec, and instead of calling the event `TapFirst` we could introduce a pseudo-capture phase by listening as `.on('Tap', () => {}, { capture: true })`. 

We can keep the `First` suffix for compatibility reasons, and print a warning to use the `capture: true` way instead, and at some point we can drop the suffixed event.

NathanaelA:
>I have to say it. It seems to me that that PR has much more performance consequence s than my PR on Observable that was refused out of performance worries.

You provided a case where it was wanted and I showed you how to TRIVIALLY do it in your own plugin code without impacting any other part of the system.   Big difference in the reasons for this PR vs that PR.   


>And I think it would be even less used.
>What is the use case for this?
Well the absolute Global events tied onto Observable; you MIGHT be correct --  I'll be honest -- I really struggled with adding this specific PR (it's been done for over a week) .     And just like your changes required a REVIEW, mine are getting a REVIEW also.  :grinning:    

However, the reasoning that finally made me decide to push this for review and adding this was because a several people have asked me about how to track analytics application wide; and their is no easy solution.  And no easy way for them to do it.    This PR makes it trivial to do so.   However, I will NOT be upset if the rest of the team decides this PR isn't worth it.     I've personally never used something like this; but having the ability to track every tap, navigation and keyboard click can allow a user to actually create event recordings for remote bug reporting, macros, and other advanced features.

>That being said I don't think it is the best approach on how to write this. A lot of duplicated intelligence.
I would simply hold a static "Observable" instance created only when the static on method is called.
And then simply call notify on that static instance from the Observable notify method.
Also isn't the static off method missing ?
Less code and much simpler logic

Your looking at the changes this does to the system to clean up some existing code to make it easier to maintain and remove duplicate code.   The majority of the code (including the static on/off/once) already exists in NS7.   NS7 already has Global Events built in, this PR just expands the Global events ability to allow you to catch a single event on all objects that fire it.   :grinning:

packages/core/data/observable/index.ts
NathanaelA: Igor mentioned in a chat that we should maybe refactor this to be {capture: true}.  Make it more like DOM.    I'm in agreement; however that does NOT impact this PR as it stands now; `First` already exists.   So I think the refactoring should be a different PR which then add a depreciation notice for `First`.

farfromrefug:
@NathanaelA My review is this. This PR should be changed to remove that global thing all together. Make a plugin for it (using mixins). Analytics features should not  have performance consequences for users not even using analytics !

PS: There was nothing trivia about solving my PR case, you simply did not get the use case. It requires private API access which was the whole point of the PR. I just stopped trying to explain.

NathanaelA:
@farfromrefug -

>There was nothing trivia about solving my PR case, you simply did not get the use case. It requires private API access which was the whole point of the PR. I just stopped trying to explain.

 I don't want to get into a fight over how trivial your use case was, but considering it was what something like 5 simple lines of code needed and NO monkey patching to be added to your plugin to support it; I think most people would consider that extremely trivial and then it didn't effect anything else for anybody else and it was a very maintainable solution by you even if it used a private api call (which you already use private apis in several of your plugins, so not sure why that is listed as a complaint).      I would appreciate it if we can drop the argument over your PR; as it has no bearing on if this PR is worth while, this PR needs to stand or fall on its own merits not on if I nixed a PR that I felt was not helpful in moving things forward and imho had a extremely trivial and easily maintainable solution.    

---

> My review is this. This PR should be changed to remove that global thing all together. Make a plugin for it (using mixins). Analytics features should not have performance consequences for users not even using analytics 

Are you saying Mixin as in monkey patching the Observable.notify event?    That is a **valid** possibility!    This is something my Global-Events plugin does, however I'll say there are several corner cases with Angular support where monkey patching fails for Global Events and it can't be used and so those events can't be used in plugins on Angular properly.    So, there are some areas I don't like Monkey patching if I can avoid it.

The real question to answer for whoever accepts or declines this PR Is having this built in for use by _all_ plugins and all applications to be able to use, is that more valuable than not?     The only reason I submitted this PR is I personally believe having a much much better event system is critical for the future applications and growth of the framework.   So having this built in to the core is worth much more than making people continue to monkey patch it in.   In addition, are you aware that my Global Events plugin appears to be the 10th most used plugin in the ENTIRE eco-system.   So obviously a better event system is something the majority of the ecosystem also wants.  :grinning:  This PR is the start to allow a more extensive event system in the future that will allow capture and cancelling of events.   So imho, this PR is worthy of being added, otherwise I wouldn't have submitted it.   :grinning:

farfromrefug:
I will just say here that putting into the core is not a question of how many people use this. Everyone uses UI and we still almost all agree it needs to be separated.
To it is is  really a question of do we want users not using it to be impacted by it? Right now they are as there is a performance toll. 
I agree not a big one on tap event (but still there is one!)
But for people using real.time events like sensors for example, it becomes much more of a toll. Personally I already removed that in my fork!

And I have to mention this appeared in 7.0 without everyone knowing about it, and no single commit pointi g to it. That s no good! In that sense I am glad you did that PR :)

NathanaelA:
>I will just say here that putting into the core is not a question of how many people use this.

Actually this is where I would 100% disagree with you.   If nobody uses it -- it probably shouldn't be in the core and should probably be removed from core to reduce support.   If the majority of the eco-system uses it; it should be put in to core (if it doesn't exist), we WANT core to be what is core to the majority of all programs, especially if it is foundational like events, css, parsing, and other key pieces that everything depends on.   The core should really contain pieces that the vast majority of applications need to work (or use).    IMHO, the majority of the plugins should be either addons (i.e. firebase, sqlite, secure-storage, https, etc) or UI (i.e. RadSideView, RadListView, Calendar, Video, etc).    


>To it is is really a question of do we want users not using it to be impacted by it? Right now they are as there is a performance toll. I agree not a big one on tap event (but still there is one!)   But for people using real.time events like sensors for example, it becomes much more of a toll.

LOL, The impact of this PR on real time sensors will be close to 0 if not actually 0 (i.e. totally un-measurable) in your example.  You seem to forget the overhead for a **single** sensor event going over the bridge from Java/ObjC to JS will so vastly out weigh **ANY** potential added overhead here, your talking milli-seconds (for bridge) vs micro-seconds (the added global event code)...    

> Personally I already removed that in my fork!

Awesome, we love people actually digging into the core modules.     


>And I have to mention this appeared in 7.0 without everyone knowing about it, and no single commit pointi g to it. That s no good! In that sense I am glad you did that PR :)

1st, you are correct **everyone** doesn't know about it; hell their are still people who don't even know NS 7 has been released or supports ES7...   As for **others** knowing about it, a blog post will be forth coming about this feature, just like a blog post about ES7 support, and how to do plugin upgrades.    This isn't something that impacts people, and we just have been busy with everything else like NS7 plugin updates and fixing any issues that are reported from the NS7 upgrade...    Like I said earlier it is the 10th most popular plugin in the entire ECO-SYSTEM.

NathanaelA:
@farfromrefug -

THANK you for making me actually benchmark it.   I grabbed one of my SLOWEST Android devices, a 32bit Arm Android 5.1.1 device that I rarely use because it is so slooooooooow, but it makes a awesome benchmark device...    And do you know what I discovered after benchmarking all three variations of the event code?   :grinning: :grinning: :grinning:

(Taking the best times from multiple random separate runs, the same # of runs were performed) 
1. Original NS 6 Event handler; 534ms total time taken for 1000 calls, so the avg is ~0.53ms per event.
2. NS 7.0 w/ current global events code, 500ms total time taken for 1000 calls, so the avg is ~0.50ms per event.
3. NS 7.0 w/ this PR, 488ms total time taken for 1000 calls, so the avg is ~0.48ms per event.

Do you see the interesting thing that occurred?  :grinning:    The NEW code is faster than the OLD code?   I have to admit I didn't actually expect that...  :grinning:           However, I had a few tests where the NEW code was slower than the OLD code.   Most the time they both were within MS of each other.     Why do you suppose that is?    Because the difference in the code changes is absolutely un-measurable.    The OS doing other things in the background causes enough jitter to cause one to be slightly faster than the others randomly.    Again we are talking the difference of a fraction of a MS between the three different event notify code bases.

Now the next interesting item, just to be complete is the bridge time alone to just get the nano-seconds from the Android device took almost 50% of the above listed time...    So in all reality each of those are inflated by 50% because of the benchmarking code, which that alone might actually explain the slight jitter in timings.      

To be complete in case you want to duplicate his -- the code is:
```
now() {
        return java.lang.System.nanoTime() / 1000000;
    }
 notify(data) {
        let timer = this.now();
        const eventClass = this.constructor.name;
        this._globalNotify(eventClass, 'First', data);

        const observers = this._observers[data.eventName];
        if (observers) {
            Observable._handleEvent(observers, data);
        }

        this._globalNotify(eventClass, '', data);
        _timeTaken += this.now() - timer;
    }
```
_timeTaken is a global value set to 0 at the start of the test.   The tests start 3 seconds after the app has fully booted (so machine/app is fully not doing anything and has settled down) and runs the notify function 1000 times each 50ms.   Finally after a 1000 tests, it prints out the _timeTaken value.    So the entire time the app is doing NOTHING, no printing, no console logs, absolutely nothing besides every 50ms running another `notify` call a 1000 time.       The code inside the `notify` function was 100% replaced for each test with which ever version i wanted to randomly test that time, to make sure v8 had the best chance to optimize the function.   In addition I used --no-hmr to make sure the app did a fresh reboot each time.    ( Btw, comments in code can cause v8 to not optimize a function, so each function was the original version w/ no changes besides adding the two `now()` calls to benchmark how fast we finished running all the code in the notify function. )

farfromrefug:
Then  why did you answer what you answered in my PR.... Let s stop it here. Add it if you want. It s out of my fork. Too bad honestly that most people will end up with that code they don't need (which adds up to all the other ones you might add in just like that). I wish the community was more in touch will all that s happening right now.

NathanaelA:
>Then why did you answer what you answered in my PR...

Because what you wrote in that PR, I reviewed and nixed was NOT performant; not only were you calling a heavy function on every add or remove of a event, you also were calling a second blank function on every add/remove for absolutely no reason.   This was all extra unneeded overhead that provided absolutely zero additional value since your PR IS totally trivial to implemented in ANY plugin that needed it with no monkey patching, no side effects, no corner cases and no extra maintenance work just adding **Literally 18 simple** lines of code and no private api needed.   
```
addEventListener(event, ...) { 
   super.addEventListener(event, ...); 
   const events = event.split(','); 
   events.foreach( (evt) => {
         typeof this._counter[evt] === 'undefined' ? this._counter[evt] = 1 : this._counter[evt]++; 
         onNewListener(evt, ..., this._counter[evt]); 
   });
} 
removeEventListener(event, ...) { 
   super.removeEventListener(event, ...); 
   const events = event.split(','); 
   events.foreach( (evt) => {
         this._counter[evt]--; 
         onRemoveListener(evt, ..., this._counter[evt]); 
   });
}
```
 and then add `this_counter={}` to your constructor and your done.    So can we finally drop your PR; as bringing up your PR just muddies the water and doesn't add anything to this PR.


>Too bad honestly that most people will end up with that code they don't need (which adds up to all the other ones you might add in just like that).

Well, let me see, this PR code adds such a small amount of time it is unmeasurable to the event routine, it comes from the 10th most used plugin in the **entire** eco-system, and it expands on that plugins abilities massively on what you can do with events.    That seems like a win/win/win in my book.   No speed loss, no extra incompatibilities, no breaking changes -- just an enhanced event system that users can easily use to do even more stuff.

>Let s stop it here.

I agree, unless you have something actually constructive to say regarding this PR, please refrain from further comments.  : I shouldn't have to defend my action about a different PR on a unrelated PR, because you are mad I didn't accept it.    If you can point to a **REAL** technical reason this shouldn't be added, I'm all ears...

farfromrefug:
> This was all extra unneeded overhead that provided absolutely zero additional value since your PR IS totally trivial to implemented in ANY plugin that needed it with no monkey patching, no side effects, no corner cases and no extra maintenance work just adding Literally 18 simple lines of code and no private api needed.
You do know the point of reusable code and not duplicating code? And cant you do exactly the same thing with that PR? Yes!

> I shouldn't have to defend my action about a different PR on a unrelated PR, because you are mad I didn't accept it. If you can point to a REAL technical reason this shouldn't be added, I'm all ears...
Let just be clear: I dont give a damn about my PR, said it multiple times. What i care about is what gets in N. I would hope it is still an OSS framework and that you dont add stuff just because some of you clients need it!!!!

Why dont you start asking the community if that kind of stuff really needs to get in the core? I would love to see what they think.

NathanaelA:
>Why dont you start asking the community if that kind of stuff really needs to get in the core? I would love to see what they think.

Global Events is the **10th** most popular plugin in the **entire community** out of more than 1,400 public plugins.    Repeat after me, 10th!  10th!  10th!    :grinning:

Now I'm not sure how many times I need to say that, because if you had thought about that statement logically the last few times I posted it; it would be obvious how your statement is again not constructive to this conversation.   But since you apparently are having issues understanding the ramifications of my statement, let me help make it crystal clear...

1. 10th means it was downloaded more times than the vast vast vast majority (i.e. more than ~99.9928%) of the other 1,400 plugins.
2. And it didn't get position 10 by some fluke; it was downloaded  over 14,000 times over just the last 30 days.
3. And it didn't get position 10 over 11 by just a minor amount, but the next closest plugin is like a 1,000 downloads behind it.
4. This is a consistent position and has the 10 position also in total downloads over life of all the plugins too; with a whooping 162,000 downloads to date.  
5. These numbers do NOT include any of the ProPlugins counts, and GE has been on ProPlugins for over a year now;  So if we include those numbers it even increases the d/l numbers much higher and might even put it in position 6 or 7, but I've too much to do to go grab those numbers right this second, so I'm content to argue from position 10.   But sufficed to say ProPlugins sees a lot of d/l's also...

Now with those **FACTS**, let discuss the logical ramifications:
If the community uses this plugin more than 99.99% of **ALL** the other plugins, and in most cases by a wide margin more, can you see **LOGICALLY** the community has already **voted** just by its _usage_ that this is a plugin that they want to use since it is used so much more than all other plugins (except the 9 that have more downloads).     So the answer to your statement is the community has spoken each and every month, for the last 4 years that GE is a plugin they WANT to use.   As such it make logical sense and makes it easier for everyone that it should be included into core.     If you still can' reach this logical conclusion, then you might seriously need to re-evaluate if you aren't arguing from emotion, perhaps because you are still pissed at the PR I closed of yours.

---
 I have answered and wasted a huge amount of time answering every issue you raised.   Most of them were not constructive and imho, completely off topic -- but I want you to see I care enough about the community to put in the effort to answer even poorly though out points, however I do have a lot of work to get done -- so let me repeat myself once more.

**Unless you have something actually CONSTRUCTIVE to say regarding this PR, please refrain from any further comments.   If you can point to a REAL TECHNICAL reason this shouldn't be added, I'm all ears...**

NathanWalker (approved):

PR #8851, opened 2020-09-16
fix(android-transitions): remove hard-coded flip transition duration/curve

lochstar:

When navigating with a `flip*` transition, the `duration` and `curve` are not honoured. Duration is hard-coded to 3000 which results in a low transition. This may have been accidentally left in during testing.

## What is the new behavior?

Hard-coded line is removed, the transition now honours the passed options.


lochstar:
@cla-bot check

farfromrefug (approved):

PR #8853, opened 2020-09-16
Ios remove tabs

farfromrefug:
As discussed this is needed to remove the need for the Podfile inside core.
It will also also 3rd party plugins to use newer versions of https://github.com/material-components/material-components-ios

To do that i implemented the tabs plugin in nativescript material components.
For N 7 https://www.npmjs.com/package/@nativescript-community/ui-material-tabs
for N 6.x https://www.npmjs.com/package/nativescript-material-tabs

The change is a bit tricky though. The current version in the podfile is really old. So when this will be merged users will get a new version of the ios lib which requires an updated version of my plugins. However i cant release it now as it wont work with the current version of N.

So i think i ll create a first PR to upgrade the material lib in the Core. And thus update the tabs component.


NathanaelA:
Hey @farfromrefug - AWESOME!   I really really really want to remove that annoying cocoapod, so thanks for stepping up and working on this!  :fire: :fire: :fire: 

Can the PR be changed to moving all this into a @nativescript/tabs plugin rather than a full removal.  NS7 is already published and I for sure do not want to wait until NS8 for this. :grinning:   So the next best idea it to eliminate this as a breaking change as much as possible.  So we can push this into NS7.1/7.2 we can have the migrate/upgrade to add `@nativescript/tabs` as a dependency to the package.json so that it no longer becomes a breaking change.

So can you instead move the code into a NEW plugin called `@nativescript/tabs` that also lives in the packages folder, then you wouldn't have to remove any tests, just have the UI testing app import the @nativescript/tabs plugin.

farfromrefug:
If you want to do it in another plugin you can but I won't maintain it. The idea was as it was a material plugin I was OK maintaining in my material suit. 
I won't maintain it if I don't have too. That component is pretty complicated. And right now even migrating it to latest iOS lib is a pain! Most features are simply broken once you upgrade the pod.
Either way my material suit is simply stuck because of this. 
Let me know if you want to do it and I ll gladly leave to you to do it.

NathanaelA:
> If you want to do it in another plugin you can but I won't maintain it. The idea was as it was a material plugin I was OK maintaining in my material suit.

We don't expect you to maintain it, at all!!!!   My objective here is to eliminate as many bumps in the road to get this merged -- as it stands now we couldn't merge this until NS8 and for moving NS forward that is a horrible outcome as this cocoapod is a pita for almost every user.

So, thinking outside the box what it the best way to get this merged during NS7.  If we move it into a external @nativescript/tabs plugin for the time being, so that their is no major breaking change.  It still is a minor breaking change; but I think the community would accept it if we made sure the migration was virtually seamless and things just worked in their app as-is out of the box as long as they include the @nativescript/tabs.

So rather than delete the code; move it so that it can be installed as a plugin. 
This would
- Allow NS to no longer conflict with your plugin by default
- Be easily replaced by your plugin (for those who want a much better tabs and other material design components)
- Eliminate the CocoaPod from default builds that don't need it :fireworks: 
- Create a fairly seamless upgrade path without it being a major breaking change.

farfromrefug:
@NathanaelA ok i agree you can make it so. It is under @nativescript still, so i ll let you guys do it.

NathanaelA:
@farfromrefug - Well, yep it is true, we can do it; but if you have the time; a PR would be nice otherwise I know it will be a while before I can tackle this as the android issues I'm researching and the existing PR's I need to test have a much higher priority.

farfromrefug:
@NathanaelA sorry dont have time to create another plugin. Tabs plugin is already part of my material comps (updated).
Right now to allow me to move forward i created a hook in my plugin to remove the Podfile and "break" the tabs component of N. Not ideal at all, but i have to move forward (even with 6.x). Have to warn the user also which is good either.

PR #8881, opened 2020-09-22
fix: delegate should be set on nativeTextViewProtected

farfromrefug:
just like textfields. Fixes components extending this one

## PR Checklist

## What is the new behavior?

### DIFF
--- a/packages/core/ui/text-view/index.ios.ts
+++ b/packages/core/ui/text-view/index.ios.ts
@@ -85,6 +85,7 @@ class NoScrollAnimationUITextView extends UITextView {
 @CSSType('TextView')
 export class TextView extends TextViewBaseCommon {
 	nativeViewProtected: UITextView;
+	nativeTextViewProtected: UITextView;
 	private _delegate: UITextViewDelegateImpl;
 	_isShowingHint: boolean;
 	public _isEditing: boolean;
@@ -114,11 +115,11 @@ export class TextView extends TextViewBaseCommon {
 	@profile
 	public onLoaded() {
 		super.onLoaded();
-		this.ios.delegate = this._delegate;
+		this.nativeTextViewProtected.delegate = this._delegate;
 	}
 
 	public onUnloaded() {
-		this.ios.delegate = null;
+		this.nativeTextViewProtected.delegate = null;
 		super.onUnloaded();
 	}
 


NathanaelA (changes requested):
Interesting -- I have to be honest, having `nativeTextViewProtected` confuses me, and I was thinking maybe either something legacy, or potentially a extended class might have it own separate `nativeTextViewProtected`.  

So I just scanned the entire NS source code for `nativeTextViewProtected` to see if their is somewhere that this is actually using a variable and not just the textbase.getter.    It is used 178 times in the code; however **ALL of them** extend text-base-common so they all are using the getter that returns `nativeViewProtected` as the value that is being changed.   

It is used in:
Label, editable-text-base, text-view, text-field, and finally text-base.   However all of them extend text-base; which is where they get access to the getter `nativeTextViewProtected` which is just a getter that returns `nativeViewProtected` 

Can anyone think of a good reason to not rename all instances of `nativeTextViewProtected` back to `nativeViewProtected` to simplify the codebase?  (However leaving the `nativeTextViewProtected` getter so that we don't break any existing code...)

Please note `nativeViewProtected` is used by `View` which is one of the base classes ALL views extend from, so having the name change in a few of the components but be still mapped to the same base value seems confusing to me...

Edit: this comment makes more sense AFTER the review as it was done after the review; so realize that this is after I made comments on the code changes..

farfromrefug:
Please dont do that. That was  done in a effort with the old N team to make it work with material native plugins where a text field/textview is made of 2 views : the layout and the actual text field. So some properties must be applied to one and some to the other. This brings a lot of flexibility to plugins developers. 

By changing that you will break all my material plugins !
That PR simply finishes the work we started with the old team (textview was not in scope at the time)

NathanaelA:
> Please dont do that. That was done in a effort with the old N team to make it work with material native plugins where a text field/textview is made of 2 views : the layout and the actual text field. So some properties must be applied to one and some to the other. This brings a lot of flexibility to plugins developers.

Can you point me to some code in your repo where this adds flexibility; I fail to see how this helps you.

> By changing that you will break all my material plugins !

Shouldn't break any plugins that aren't broken already.   But I am trying to keep an open mind -- can you point me to some code on how it helps.   

We do NOT want to break any plugins, so I really would like to see how this helps; as to me having three different ways to access the same variable seems complicated...

farfromrefug:
I told you it helps in separating what should be applied to the textview and what should be applied to the layout.

And of course you will break plugins if you remove it ! 
btw this PR / fix is needed to fix issue users are having because the textview is not behaving as the text field.

NathanaelA:
>I told you it helps in separating what should be applied to the textview and what should be applied to the layout.

In the current NS source code as it is used today it does not separate ANYTHING.    

`label.ios === label.nativeViewProtected === label.nativeTextViewProtected`    

All it is is additional ways to access `nativeViewProtected` and adds complication to the code base because then the developer has to do the mental gymnastics of realizing all three are the same.  

>And of course you will break plugins if you remove it !
>btw this PR / fix is needed to fix issue users are having because the textview is not behaving as the text field.

This PR doesn't fix a **single** thing in the actual NS code base, but makes it slightly more confusing.   

---

I do not want to argue with you; I want to understand the why and need of it; So I will ask again the exact same question as I did in my prior response twice.

Can **YOU** show me how it fixes something in one of your plugins I am interested to see how it does.     It is a simple request; POINT me to A **SINGLE** piece of code in an existing plugin that this would break if I we did a refactor to remove all references to `nativeTextViewProtected` and just leave the getter as depreciated?      I don't need **hypothetical**; I don't need **words**.    **I need a actual link to source code where it is being used** so I can read it for a couple minutes and understand why you believe this is needed.    Either one of us can be wrong, and I have no problem being wrong.  The code is the source of truth for this matter; so I am asking you for a link to where this is used so I can understand why you believe it is necessary.

farfromrefug:
Really bold, uppercase? Seriously....

It fixes delegate being set on the wrong view on iOS (the layout and not the view).if
I can't show code it fixes in my comp.as it is not code in my comp but in N. For.my  comp nativeViewProtected and nativeTextViewProtected are different. Before this PR the delegate for UITextView was set on the "layout" view (represented by nativeViewProtected)
You should read this
https://github.com/NativeScript/NativeScript/pull/6102
https://github.com/NativeScript/NativeScript/pull/8514

NathanaelA:
Yes, it needed to be bold uppercase to make sure you read and answered properly because I asked **before and you ignored it (and you still did, so I'm having to repeat myself yet again)** and just continued to argue it is needed, which solves nothing as we need to know WHY you think you need it.  :grinning:     

---

>It fixes delegate being set on the wrong view on iOS (the layout and not the view)

That is incorrect; you must have missed my outline earlier in the review notes so let **repeat myself** again... 

https://github.com/NativeScript/NativeScript/blob/master/packages/core/ui/text-view/index.ios.ts#L117
points to (Same file, 9 lines down):
https://github.com/NativeScript/NativeScript/blob/master/packages/core/ui/text-view/index.ios.ts#L126

And `this.nativeTextViewProtected` points to
https://github.com/NativeScript/NativeScript/blob/master/packages/core/ui/text-base/text-base-common.ts#L29-L31

So this means `this.ios` RETURNS `this.nativeViewProtected` and your proposed change
`this.nativeTextViewProtected` **ALSO RETURNS** `this.nativeViewProtected`

In other words for the NativeScript code base this is a NO-OP, and doesn't change anything other than add a different redirection to use a different getter function which is buried even deeper in the source code.   IMHO, the better change as I stated in the review notes would be just to remove all the re-directions and change both places `this.ios` is used to `this.nativeViewProtected` and be done with it.

---

> I can't show code it fixes in my comp.as it is not code in my comp but in N. For.my comp nativeViewProtected and nativeTextViewProtected are different. Before this PR the delegate for UITextView was set on the "layout" view (represented by nativeViewProtected)

I'm starting to wonder if you read my entire responses or just skim it.  I do realize I'm a bit long-winded :grinning:, but I try and give all the information needed to help make a choice...   As I stated in the PRIOR response their is **ABSOLUTELY NOTHING** in NativeScript that needs a separate `nativeTextViewProtected`, I looked at all **178** references to `nativeTextViewProtected` yesterday and every single instance could be changed to `nativeViewProtected` with **ABSOLUTELY NO CHANGE** of functionality in the NativeScript code base.   

And repeating myself **yet again**. WE need to see a VALID reason (i.e. code that actually uses this feature), not arguments about why it is needed, as code that NEEDS this feature is what is important to if it stays.   So **please** link to some actual **CODE** that has a actual **VALID** reason for this code to exist.    We need why is this valid, and exactly what would actually be harmed if we changed it.

---

I actually did look #6102 (& others), early this morning, before my prior response as I was trying to understand the reasoning behind this, and I even saw this comment which apparently you forgot to do in your PR.
>_Also, please add a comment to the nativeTextViewProtected getter on why we are doing this and why you would want to override it._ 

Which would have helped us greatly in this discussion as I still don't see WHY this separation is needed since it points to the EXACT same variable.    And he apparently didn't either, as he asked for a comment to the PR; but apparently both of you forgot about that comment and as such; now we are discussing the why's 2 years later because it was NEVER commented in the source code...   If you can provide a solid reason, we will get the comment into the code for that getter so that this doesn't come up again.  :grinning:    

I just looked at #8514 as it didn't come up in my search early.    It has no effect on this; so either you linked the wrong issue; or you don't seem to understand that we can replace every place that has `this.nativeTextViewProtected` with `this.ios` or `this.nativeViewProtected` and the code will still work the same, so that PR is not affected by a change like this...   I did a search for all issues/PR's that had `nativeTextViewProtected` this morning, and not a single one that I saw explained why this change is needed or even desired/wanted.

---

So I will repeat myself one final time and then I'm done, I don't have time to argue, and I need you to follow this simple request (done multiple times now) for the information that would help everyone understand why.    

- WE NEED, and we mean NEED, to see why we shouldn't cleanup this **mess** up and make it a LOT simpler for ALL future developers to grok the code base, *OR* why we NEED to keep this extra indirection and make good notes about why it exists, so that others can understand the why's when they are working on the code base.   This is critical for maintaining the code base; either simplifying and/or explaining the whys so that people can work on it and understand...

- There is NOTHING, and I mean **ABSOLUTELY NOTHING** in the NativeScript core that needs this extra indirection.  So pointing me to other issues in the NS Core repo is probably going to be unhelpful at best, and more annoying at worst.

- So exactly what plugin(s) uses this functionality and **LINKS** to exactly WHERE in these plugins does it NEED it.  

We don't want to mess you (or any other author) up; but help us help you by providing the information requested;  i.e. some place in ACTUAL code it is actually used that it couldn't be just replaced by `nativeViewProtected` and continue to work as is.     Please point me to the actual component that needs it as I already did a really **quick** scan of the Material Design plugin of yours (I didn't read every bit of code, just randomly picked several components that had `nativeTextViewProtected` in them and read through their code); and I didn't see a single case where you needed the separate functionality either.   All of them would have worked had you used `this.ios` or `this.nativeViewProtected`.   

So **PLEASE** just point me to any component that **NEEDS** this and is actually **USING** this extra indirection; but make sure the component actually needs it.  Example: your material UILabel uses nativeTextViewProtected, but the component wouldn't change a bit if it was replaced by `nativeViewProtected` so verify it is a **REAL** use-case...

I don't know how to make it any plainer what we need from you, we need a actual VALID usage of this functionality to see WHY...

I'll be very honest I really HATE repeating myself and multiple times at that now, it is a huge waste of time; so please read **EVERYTHING, RE-READ EVERYTHING**, then respond with a valid actual link to code that is actually needing this indirection.

NathanaelA:
@farfromrefug - Igor scanned your UI-Material components code and did a more in-depth search than I did and found a use case where you are using this ability:

He linked me too:

https://github.com/nativescript-community/ui-material-components/blob/master/src/textfield/textfield.android.ts#L33-L52

This was **all** we were asking for, a valid use case that shows WHY this feature is needed.

---

Based on reading the code; you have the main view https://developer.android.com/reference/com/google/android/material/textfield/TextInputLayout
Which has the actual TextInput held internally.

Now, the three choices to make this work is
- extend the TextInputLayout control to forward all "TextInput" values internally so that the code could word as is; however this means that in the future if new things are added it adds a maintenance burden as you have to add new setter/getters to forward values.
- Use your own TextInput class; this again means that if we add stuff this increases your support burden.
- Leave the code as is, so that the wrapper can be the `nativeProtectedView` and the internal text view can be assigned to the `nativeTextProtectedView`

So this is a interesting use-case which that code brings up; and we will have to figure out how to document this properly on the getter.  I believe that because of the dual view nature of that component that this PR is valid and we would NOT want to remove the `nativeTextViewProtected` value, we just need documentation of why it exists.

NathanaelA (approved):
Based on the code Igor linked me to; I believe this is a valid use case to eliminate extra work on the plugin author without being a major issue for the framework.   We do need to document this getter and we will be doing a PR for that shortly.

Issue #8778, opened 2020-08-19
iOS crash when running with release/production flag: Can't find variable: __metadata


nabilzhafri:
Simply copy steps from here: https://nativescript.org/blog/upgrading-tips-for-angular-10/

```
npm i -g @angular/cli // You could be needing to prefix sudo on GNU/Linux and other Unix-like OSes.
npm i -g nativescript // You could be needing to prefix sudo on GNU/Linux and other Unix-like OSes.
npm i -g @nativescript/schematics // You could be needing to prefix sudo on GNU/Linux and other Unix-like OSes.

ng new workspace
cd workspace
ng add @nativescript/schematics --skipAutoGeneratedComponent

// start apps:
npm run ios --> this runs fine
npm run ios -- --release   --> this will cause error below
npm run ios -- --env.production   --> this will cause error below
```

Any idea why? Tried to search for solution but can't find any.
Happens to my existing project as well after upgrading to angular 10. Right now I can't seem to deploy the iOS app due to this issue. It's causing the app to crash on launching.

```
JavaScript stack trace:
file:///app/vendor.js:90935:15
at __webpack_require__(file:///app/runtime.js:75:34)
at file:///app/vendor.js:84982:90
at file:///app/vendor.js:85830:34
at __webpack_require__(file:///app/runtime.js:75:34)
at file:///app/vendor.js:42890:97
at __webpack_require__(file:///app/runtime.js:75:34)
at file:///app/vendor.js:62743:87
at __webpack_require__(file:///app/runtime.js:75:34)
at file:///app/vendor.js:35266:89
at __webpack_require__(file:///app/runtime.js:75:34)
at file:///app/vendor.js:64700:87
at __webpack_require__(file:///app/runtime.js:75:34)
at file:///app/vendor.js:66519:95
at file:///app/vendor.js:71822:34
at __webpack_require__(file:///app/runtime.js:75:34)
at file:///app/bundle.js:51:47
at __webpack_require__(file:///app/runtime.js:75:34)
at checkDeferredModules(file:///app/runtime.js:44:42)
at webpackJsonpCallback(file:///app/runtime.js:31:39)
at anonymous(file:///app/bundle.js:2:61)
at evaluate([native code])
at moduleEvaluation([native code])
at [native code]
at asyncFunctionResume([nativ<…>
JavaScript error:
file:///app/vendor.js:90935:15: JS ERROR ReferenceError: Can't find variable: __metadata
[...]

PR #8884, opened 2020-09-23
fix: bundle entry points

farfromrefug:
As requested here is the PR again which fixes

This PR fixes a big issue which appeared with N 7 / esm / webpack

With webpack import / require order is not respected : https://github.com/webpack/webpack/issues/1343

Now in [bundle-config-loader](https://github.com/Nativescript/NativeScript/blob/65a3302a048fb32da78968256c16fc63a5706e3d/packages/webpack/bundle-config-loader.ts#L55) we rely on `require` to load `bundle-entry-points` thus calling `initGlobals` and making tslib functions global.
Because it uses `require` it actually end up at the end of your `main.js` (simply look at a generated bundle.js)

To keep that call up we need to use `import` 
But that is not enough!
in `bundle-entry-points` we use 'require' on 'globals'. And using `import` is not enough either because we call a method on it which will end up after all imports!

The solution to all this is call `initGlobal` inside `global/index.ts` then use `import './global'` for example.
Doing that i needed cleanup a bit and we end up with only on use of `initGlobal`

EDIT: that [app](https://github.com/nativescript-community/gesturehandler/tree/master/demo_vue) will crash on ios because of that. The reason is that i import View at the top of my main.ts which include files using `__metadata`


NathanWalker:
I think we did some tests with import vs. require and were seeing exactly what is commented on that thread here:
https://github.com/webpack/webpack/issues/1343#issuecomment-523076943

Where using require as opposed to import seemed to be guarantee.

farfromrefug:
Imports will all get at the top you are right. Which is why I changed the way initGlobals is called to ensure the right order. 
In bundle config loader using require makes it end up last. Look at the generated code for a current app to see it.

farfromrefug:
After that test with that code and you ll see the order being correct.

rigor789:
LGTM - I'll just run a test and include a diff here with the before/after behavior.

rigor789 (approved):

NathanWalker:
Awesome thanks @farfromrefug and @rigor789 👍

NathanWalker:
Released with 7.0.5 https://github.com/NativeScript/NativeScript/releases/tag/7.0.5-core

asharghi:
Possible to get this fix for NS 6 also?

farfromrefug:
@NathanWalker i think it needs a webpack release too. https://github.com/NativeScript/NativeScript/issues/8778

PR #8888, opened 2020-09-24
support runtime in devDep

farfromrefug:
Right now it seems to be used for snapshot
This is one part of more global fix for snapshots which are broken with latest webpack

### DIFF
--- a/packages/webpack/helpers/androidProjectHelpers.js
+++ b/packages/webpack/helpers/androidProjectHelpers.js
@@ -11,8 +11,12 @@ const ANDROID_CONFIGURATIONS_PATH = join(ANDROID_PROJECT_DIR, "build/configurati
 const getAndroidRuntimeVersion = (projectDir) => {
     try {
         const projectPackageJSON = getPackageJson(projectDir);
-
-        const version = projectPackageJSON["nativescript"]["tns-android"]["version"];
+				let version;
+			  if (projectPackageJSON["nativescript"]) {
+					  version = projectPackageJSON["nativescript"]["tns-android"]["version"];
+				} else {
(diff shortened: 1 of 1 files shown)


packages/webpack/helpers/androidProjectHelpers.js
```diff
@@ -11,8 +11,12 @@ const ANDROID_CONFIGURATIONS_PATH = join(ANDROID_PROJECT_DIR, "build/configurati
 const getAndroidRuntimeVersion = (projectDir) => {
     try {
         const projectPackageJSON = getPackageJson(projectDir);
-
-        const version = projectPackageJSON["nativescript"]["tns-android"]["version"];
+				let version;
+			  if (projectPackageJSON["nativescript"]) {
+					  version = projectPackageJSON["nativescript"]["tns-android"]["version"];
+				} else {
+					  version = (projectPackageJSON["devDependencies"]["tns-android"] || projectPackageJSON["devDependencies"]["@nativescript/android"]).replace(/[~^]/, '');
```
rigor789, line 18: The tricky part with this is that it could be a path to a `tgz` file too. The CLI tries to read from the installed runtimes `package.json` in `node_modules` if that's the case, but it's not "fool-proof". 

I guess it's fine for now - just something we need to keep in mind.

farfromrefug: You are right fdont like it either. Does not support latest or * too....

NathanWalker:
Closing as confirmed operational and not pursuing snapshots.

farfromrefug:
Again total nonsense... Either remove all the snapshot code now or accept fixes

NathanaelA:
Snapshot code will be removed as we have time.  However, we would gladly take any PR's that remove the snapshot code if you want to continue to be helpful.

See: https://github.com/NativeScript/NativeScript/issues/8926 -- This issue has been open for almost a month, and not a **single** community member has commented negatively about this decision...

PR #8889, opened 2020-09-24
fix broken download of mk snapshot tools

farfromrefug:
Part of a work to fix snaphshots for N 7

### DIFF
--- a/packages/webpack/snapshot/android/utils.js
+++ b/packages/webpack/snapshot/android/utils.js
@@ -4,7 +4,7 @@ const { join, relative, isAbsolute } = require("path");
 const os = require("os");
 
 const { mkdir } = require("shelljs");
-// const { get } = require("request");
+const { get } = require("request");
 const { getProxySettings } = require("proxy-lib");
 const semver = require("semver");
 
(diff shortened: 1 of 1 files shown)


NathanWalker:
Closing as not pursuing snapshots further.

farfromrefug:
Makes no sense not to fix this still  when I give it to you... Total nonsense

NathanaelA:
Because as I've personally mentioned to you before (many many times now, & since before NS7 was released), Snapshots were depreciated and we were not planning on supporting them at all in NS 7.  

So we need to be removing all the extra code from the code base, to simplify things -- not continue patching a feature that makes no sense based on the real benchmarks.    We do appreciate the work on the PR, but it doesn't align  with the direction and keeps the code base more messy and complex for no gain.

See: https://github.com/NativeScript/NativeScript/issues/8926 -- This issue has been open for almost a month, and not a **single** community member has commented negatively about this decision and as such, at the TSC meeting today we went through all the PR's and started cleaning things up and trying to get a handle on any outstanding PR's.

farfromrefug:
@NathanaelA I know you want to remove. And it is kind of funny your point about no one reacting to your issue about officially removing. No one knew about it.... Even me who is subscribed to all issues and PRs missed it. I don't think you can say it reflect the opinion of a community if you don't let the community knows...
It s about time your change your way of doing things and really include the community in your decision
I asked you already (many many times as you like to say) to create TSC discussion channels and other ways for the community to react and contribute to your TSC decisions. So please don't use that argument.

PS: as for the true matter of snapshot removal, I believe you are wrong to do it so fast without even having true number about the consequences (run time) , but I also am starting to believe it does indeed does not make a huge difference. It just makes me really sad to not have definite proof (which you should provide) before seeing this gone. We might create a bit hit on N (simply compare a react native cold runtime to N, makes you wonder if you should really use N)

NathanaelA:
I don't know how you missed it:
1. It is the only **Pinned** Issue for the last month  (xcode 12 issue was also pinned for a while)
image

2. I even linked it in the last convo I recall we had about snapshots where I told you snapshots are offtopic for the PR: 8909 as we are discussing the merits of the PR, not snapshots. (https://github.com/NativeScript/NativeScript/pull/8909#issuecomment-701057391)

As for matter of removal; see the issue https://github.com/NativeScript/NativeScript/issues/8926 -- I ran benchmarks on multiple devices from lowest spec to highest spec I had access to.   I reported the numbers on the two sides (slowest & fastest) as the rest of the devices all lined up the EXACT same way (so no need to make the data more complex).    

In addition the prior NativeScript team came to the same conclusion and so did others in the community who actually figured this out a long time ago and  actually sparked the team to run tests.

PR #8890, opened 2020-09-24
fix for android snapshot

farfromrefug:
using metadata will make parameters type appear in js files like for this one
https://github.com/NativeScript/NativeScript/blob/master/packages/core/ui/frame/index.android.ts#L831

```
    profile,
    __metadata("design:type", Function),
    __metadata("design:paramtypes", [androidx.fragment.app.Fragment, Number, Boolean, Number, Function]),
    __metadata("design:returntype", android.animation.Animator)
], FragmentCallbacksImplementation.prototype, "onCreateAnimator", null);
```
Then the snapshot tool will complain about not knowing android / androidx

### DIFF
--- a/packages/core/tsconfig.lib.json
+++ b/packages/core/tsconfig.lib.json
@@ -7,7 +7,7 @@
     "noImplicitAny": false,
     "noImplicitUseStrict": true,
     "removeComments": false,
-    "emitDecoratorMetadata": true,
+    "emitDecoratorMetadata": false,
     "experimentalDecorators": true,
     "diagnostics": true,
     "sourceMap": true,


packages/core/tsconfig.lib.json
```diff
@@ -7,7 +7,7 @@
     "noImplicitAny": false,
     "noImplicitUseStrict": true,
     "removeComments": false,
-    "emitDecoratorMetadata": true,
+    "emitDecoratorMetadata": false,
```
NathanaelA, line 10: This is going to have to be tested really well to make sure it doesn't break other ES2017 support for non snapshot builds.

As even TS has these warnings about this flag:
> Please note that auto-generated type metadata may have issues with circular or forward references for types.
> Decorator metadata is an experimental feature and may introduce breaking changes in future releases.

farfromrefug: Yes it is experimental and was false before N 7 I think. If it was used before we would have had the same issue

farfromrefug:
@NathanaelA You changed this to draft. Any news? I have apps running perfectly with that change. As i said it was like that in 6.x

NathanaelA:
Well their are two things; 
1. The ES2017 might have different issues with Snapshots and this flag; but that is awesome to hear this is working for you.  That is a good data point; thanks for letting us know as I wasn't sure if the PR was completed yet!

2. We need to _verify_ this works for all three platforms, before we can merge, so I have to find some time to carve out to run this on all three OS's and verify that snapshots work from each of them.    Which platform are you running from?

farfromrefug:
Yes you need to test but remember that is actually the actual value which needs testing. It was a change with N 7!
So rolling back is actually the safest route.
And snapshot is working perfectly with 3 production apps here with N 7 and full es2017 (vue and svelte)

NathanWalker:
Closing as not pursuing snapshots further.

farfromrefug:
@NathanWalker really ? You know this change makes the framework much slower? There is no need to decorator metadata and by leaving it like this you marshal a ton of android class even If a user is not even using them...
And remember you added that unnecessary change with the monorepo

farfromrefug:
Every file in N are.imported because of the ui registration. It means every decorator metadata is "loaded in memory" and thus android classes which appeared as parameter marshalled. 
Without those decorator those classes would only be marshalled if "used"
For example android textviee class is only marshalled if you create a label. Look at a profile timeline to see it.
Honestly keeping the decorator metadatas when you actually don't need it (N ever had those before, remember that!) Is a truelly big mistake. 

Now about the snapshot drop. TSC decision without any discussion with the community :) didn't you say the tsc channel on discord would get publicly readable  ? Weird for it not be. Not trace of your TSC discussions anywhere for the community to react...
That being said it still makes no sense to refuse fixes to code still in the repo(s).

NathanaelA:
1. I'll double check all the timings, however with you tying it to being a "Snapshot" issue -- that is why it got closed, if it is a general performance issue that has no negative impact on our new ES6+ support, then we aren't opposed.  But if it is purely for Snapshot, then it is a pointless commit as we don't support snapshots.

2. See  https://github.com/NativeScript/NativeScript/issues/8926 -- This issue has been open for almost a month.   Not sure how much more public it can be.   All the data laid out why we want to do this, and not a single negative comment.  :grinning: 

3. The discord channel we have been discussing opening it up to be read only, I know I forgot about it as I know I'm rarely on Discord, but I recall voting in favor of opening it up.  :grinning:    I'll have to go check to see if we got a consensus on it...

4. Yes it does, adding support for something we are dropping sends mixed signals.   The goal of PR's is to improve the framework, fixing snapshots support is going backwards not forwards.

farfromrefug:
Just about your 1 I discovered it because of the snapshot but I explained quickly it was more than that. You should have triaged and renamed the issue then.

Issue #8893, opened 2020-09-24
[NS7] - Android error extending custom activity 


mapo80:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 7.0.8
 - Cross-platform modules: 7.0.0
 - Android Runtime: 7.0.0
 - iOS Runtime: 7.0.0
 - Plugin(s):

**Describe the bug**
I've a custom application in AndroidManifest.xml, I've followed this guide but it doesn't work:
https://docs.nativescript.org/core-concepts/android-runtime/advanced-topics/extend-application-activity

If I run application I've this error:

`09-24 16:58:29.251 11882 11882 E AndroidRuntime: java.lang.RuntimeException: Unable to instantiate application com.custom.NativeScriptApplication: java.lang.ClassNotFoundException: Didn't find class "com.custom.NativeScriptApplication" on path: DexPathList[[zip file "/data/app/org.nativescript.testnsng-n9D3x_-2ywiadEyA-gCknQ==/base.apk"],nativeLibraryDirectories=[/data/app/org.nativescript.testnsng-n9D3x_-2ywiadEyA-gCknQ==/lib/x86_64, /data/app/org.nativescript.testnsng-n9D3x_-2ywiadEyA-gCknQ==/base.apk!/lib/x86_64, /system/lib64]]
`

**To Reproduce**
I've added a custom application: com.custom.NativeScriptApplication
It' defined in application.android.ts, if I leave com.tns.NativeScriptApplication everything works fine.

**Sample project**
https://github.com/mapo80/nativescript-ns7-custom-application

shansb-dev:
Hi, we are experiencing a similiar issue. We also followed this guide to add a custom android activity.
However it does not seem to work like this with NS7

activity.android.ts (Changed the decorator to @NativeClass()) 
```
@NativeClass()
class MainActivity extends androidx.appcompat.app.AppCompatActivity {
...
```

AndroidManifest.xml
```
android:name="org.test.nativescript.MainActivity"
```

webpack.config.js
```
// Add your custom Activities, Services and other Android app components here.
    const appComponents = [
        "@nativescript/core/ui/frame", "@nativescript/core/ui/frame/activity", resolve(__dirname, "app/activity.android.ts")
    ];
...
```

```
[...]

PR #8898, opened 2020-09-25
Fix extending android application and activity

mapo80:
This change fixes this issue: https://github.com/NativeScript/NativeScript/issues/8893

## PR Checklist

## What is the new behavior?


mapo80:
@cla-bot check

packages/webpack/templates/webpack.angular.js
```diff
@@ -84,12 +83,13 @@ module.exports = env => {
   }
   const entryModule = `${nsWebpack.getEntryModule(appFullPath, platform)}.ts`;
   const entryPath = `.${sep}${entryModule}`;
-  const entries = { bundle: entryPath };
+  const appEntries = Object.assign({ bundle: entryPath }, entries);
```
rigor789, line 86: I think we should just keep the same naming as before. Just saw that the original `env` spread is `const` but, can add

```js
let {
    appComponents = [],
    entries = {}
} = env;

entries = Object.assign({ bundle: entryPath }, entries);
appComponents = appComponents.concat([
    "@nativescript/core/ui/frame", "@nativescript/core/ui/frame/activity"
]);
```

rigor789 (changes requested):
Thanks for the PR!

Left a note with a suggestion, so we keep the config very similar to previous configs - just extending it, rather than renaming the variables.

Other than that, It should be done for all configs for consistency.

packages/webpack/templates/webpack.angular.js
mapo80: Original env is a const, do you want to change it?

rigor789: I think we can just add it below, as a mutable/let spread. 
```js
const {
 // ...
} = env;

let {
    appComponents = [],
    entries = {}
} = env;
```

It could possibly stay a const too - since array operations/object mutations don't replace the variable itself.

So perhaps
```js
const {
 // ...
  appComponents = [];
  entries = {}
} = env;

// assign { bundle: entryPath } to entries which is {} by default, but still assign entries at the end
// to allow overriding the bundle entry too
Object.assign(entries, { bundle: entryPath }, entries);
appComponents.push("@nativescript/core/ui/frame", "@nativescript/core/ui/frame/activity");
```

mapo80 (commented):
For me is ok!

mapo80:
I've pushed changes. If it's ok can you merge?

mapo80:
news?

rigor789:
@mapo80 this is looking good, we just want to add it to all the other configs as well before we merge it. If you can do it, that'd be awesome - otherwise when we have the chance we'll add and then merge.

NathanWalker (approved):

PR #8877, opened 2020-09-21
fix(ios): time-picker and date-picker for iOS 14

asharghi:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [ ] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#running-unit-tests.
- [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/WritingUnitTests.md.

## What is the current behavior?
TimePicker/DatePicker crashes on iOS 14
 
## What is the new behavior?
Does not crash

Fixes/Implements/Closes #8862

PR #8899, opened 2020-09-25
fix(ios): preferredDatePickerStyle property

asharghi:
This is my suggestion for the discussion at https://github.com/NativeScript/NativeScript/pull/8877#issuecomment-698830203

I will add the same property to the datetime element after I have gotten feedback

### DIFF
--- a/packages/core/ui/time-picker/time-picker-common.ts
+++ b/packages/core/ui/time-picker/time-picker-common.ts
@@ -208,6 +208,7 @@ timeProperty.register(TimePickerBase);
 
 export const iosPreferredDatePickerStyleProperty = new Property<TimePickerBase, number>({
 	name: 'iosPreferredDatePickerStyle',
-	defaultValue: 0
+	defaultValue: 0,
+	valueConverter: (v) => parseInt(v),
 });
 iosPreferredDatePickerStyleProperty.register(TimePickerBase);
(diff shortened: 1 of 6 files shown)


NathanWalker (dismissed):

NathanWalker (approved):

packages/core/ui/time-picker/time-picker-common.ts
```diff
@@ -98,6 +98,7 @@ export abstract class TimePickerBase extends View implements TimePickerDefinitio
 	public maxHour: number;
 	public minMinute: number;
 	public maxMinute: number;
+	public iosPreferredDatePickerStyle: number;
```
DepickereSven, line 101: Shouldn't this be `iosPreferredTimePickerStyle` instead of `iosPreferredDatePickerStyle`. Because this is the time-picker component and not the date-picker component

asharghi: It's named after what Apple has called it https://developer.apple.com/documentation/uikit/uidatepicker/3526124-preferreddatepickerstyle

DepickereSven: Okay, then I understand it. First I thought huh is this a mistake? Because it's a time picker component and not a date picker component

PR #8904, opened 2020-09-25
feat(core): queueMacroTask

edusperoni:
## PR Checklist

The only way to schedule a macrotask is by using `setTimeout`, and even that isn't a guaranteed to be executed immediately (ready: before frame) on iOS.

Here's an example (for some reason playground is not saving the new version, so comment the last `setTimeout` of home.component.ts` or you'll freeze for a few seconds at the home page): https://play.nativescript.org/?template=play-ng&id=B4UY5b&v=2

## What is the new behavior?
Defines a proper scheduler for macroTasks by queuing a task on the main thread) that will run after all microTasks have ran. requestAnimationFrame is now in a macroTask to mimic web behavior (this is necessary because angular will use `requestAnimationFrame` to schedule it's change detection).

Looking for input in error handling and if we should schedule multiple tasks instead of scheduling one and iterating through the callbacks.


packages/core/utils/macrotask-scheduler.ts
```diff
@@ -0,0 +1,20 @@
...
+
+let scheduled = false;
+let macroTaskQueue: Array<() => void> = [];
+
+function drainMacrotaskQueue() {
+	const currentQueue = macroTaskQueue;
+	macroTaskQueue = [];
+	scheduled = false;
+	// TODO: error handling?
+	currentQueue.forEach((v) => v());
```
NathanaelA, line 11: Probably add a try / catch around currentQueue.forEach, so it catches failures and then call `uncaughtErrorEvent` so that if they don't catch the error; it calls the uncaught error handler.

edusperoni: I haven't found any example of this. Do you know where I can find one? `global.__onUncaughtError` requires a `NativescriptError` with `nativeError` attached, just want to make sure I'm passing the correct error to it.

NathanaelA: Yeah, maybe the `__onUncaughtError` and `__onDiscardedError` needs to be changed to basic `Error`...   @NathanWalker @rigor789 - Thoughts?

edusperoni:
Updated to latest master. As soon as we settle on error handling this is good to go.

Might be worth mentioning that `setTimeout(fn,0)` will run **after** `queueMacrotask` on iOS as it uses timers. On android ``setTimeout(fn,0)` has the same priority as `queueMacrotask` (whichever gets scheduled first runs first).

NathanWalker:
All tests passing thanks, we can do some testing against frontend flavors - definitely want to check against Vue, React and Svelte.

packages/core/utils/macrotask-scheduler.ts
edusperoni: I decided to follow the path some other core parts use (like CSS) that will just output the error to the error stream. We might have to discuss an API for dealing with these (and other errors from core/plugins) in the future

NathanWalker (approved):

DimitrisRK:
@edusperoni @NathanWalker I know this PR was approved, but it causes 2 errors.
Inside animation-frame and macro-task modules, a bad import of trace module is used.
Should change from this:
`import { Trace } from 'trace';`
to this:
`import { Trace } from '../trace';`

farfromrefug:
@DimitrisRK @NathanWalker created a PR https://github.com/NativeScript/NativeScript/pull/9027.
I think it should be a good thing for you to always run the automated app when approving a PR

Issue #6125, opened 2018-07-27
Feature(iOS): Fonts are not scaled for a11y in nativescript.


m-abs:
### Tell us about the problem
Platform font-size setting not respected by nativescript apps.

This is a huge problem for our target audience, which include the some who are only partially sighted.

### Which platform(s) does your issue occur on?
iOS only. It works as expected on Android.

### Please provide the following version numbers that your issue occurs with:
- CLI: 4.1x
- Cross-platform modules: 4.1.x, 4.0.x, 3.x.
- Runtime(s): 4.1.0
- Plugin(s): none

### Please tell us how to recreate the issue in as much detail as possible.
Create a nativescript app and run it on iOS.

1. Go to `Settings -> General -> Accessibility -> Larger text` and change the font-size.
2. Go back to the `nativescript` app, the text is rendered the same size as before.

On `iOS 10+` UILabel have the property `adjustsFontForContentSizeCategory`
https://developer.apple.com/documentation/uikit/uicontentsizecategoryadjusting/1771731-adjustsfontforcontentsizecategor?language=objc

But my experience from solving this with our `@nota/nativescript-accessibility-ext` is that it only works for the preferred fonts on `iOS` and those are not easily set from `NativeScript`.

The way we solve it in `@nota/nativescript-accessibility-ext` is by [listening](https://github.com/Notalib/nativescript-accessibility-ext/blob/master/src/utils/FontScaleObservable.ios.ts) for the platform setting and scaling the font-size, unfortunately this is slow and gives makes the UI jumpy.

NickIliev:
@m-abs indeed there is no out-of-the-box support for dynamic types. I am marking this one as a feature as even in native iOS not all fonts can be used as dynamic and for many custom fonts [a different techniques](https://medium.com/livefront/practical-dynamic-type-d2b5774e8a8a) are used to scale them properly.

Resources:
- [Dynamic Types from iOS Human INterface Guidelines](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/)

- [Practical dynamic types for iOS 11 with `UIFontMetrics`](https://medium.com/livefront/practical-dynamic-type-d2b5774e8a8a)

- [Using `adjustFontsForContentSizeCategory`](https://useyourloaf.com/blog/auto-adjusting-fonts-for-dynamic-type/)

PatrickLohan:
[...]

Issue #6993, opened 2019-03-06
Feature: Support for accessibility


m-abs:
**Is your feature request related to a problem? Please describe.**
Hi,

I work at [Nota](https://nota.dk), the Danish Library and Expertise Center for people with print disabilities.

Our users have various disabilities that we need to support, some are dyslexic, some have visual impairment and other have cognitive impairment.

It is essential for our use case that we can make the app accessible, unfortunately this in one of the things `NativeScript` doesn’t do very well.

This is the reason we maintain this plugin: [@nota/nativescript-accessibility-ext ](https://www.npmjs.com/package/@nota/nativescript-accessibility-ext).

I had a chat with @sebawita, we agreed I should make this issue, so we at least can have a talk to see, if get built-in support for accessibility in `NativeScript`.

**Describe the solution you'd like**
The solution should make it easier to make NativeScript apps accessible for people with disabilities.
It should smooth out the differences between the platforms, while still provide full access to those features only found on one platform.

The biggest challenge is the differences in the accessibility support of two platforms.

To give an example of a feature which is on both platforms and one that are only on one:
Setting a label on a view, on iOS we can set the string value to `accessibilityLabel` and on Android we set the View's `ContentDescription`.
This label is what VoiceOver(iOS)/TalkBack(android) will read.

iOS also have the property `accessibilityIdentifier`, there is no equivalent on Android. This value is meant to be a single unique value that can be used to find the `UIView` during automated testing.

**Side note:** `NativeScript` binds the property `automationText` to both `accessibilityLabel`  and `accessibilityIdentifier` on iOS, which is incorrect usage see #3150

This feature meant as start of discussion, so I won't into all the accessibility features here.

Is this something you're interested in supporting upstream?
And if so, what are your thoughts on how best to proceed?

**Describe alternatives you've considered**

As mentioned we have a plugin `@nota/nativescript-accessibility-ext` that implements some of the features.
The dream scenario is to deprecate our plugin and get a cleaner API than our more-or-less direct mapping to native-properties. 

**Additional context**

tsonevn:
ping @manoldonev

manoldonev:
@m-abs Thanks for reaching out!

Ideally as both iOS and Android provide APIs for making apps accessible to people with disabilities, NativeScript should be able to leverage these capabilities as well (I guess that is true for any UI framework / library).

[...]

Issue #6123, opened 2018-07-26
Bug(iOS): Escape-gesture doesn't work as expected for VoiceOver users.


m-abs:
### Tell us about the problem
Accessibility users can't dismiss dialogs and go back with the z-gesture*.

* Z-Gesture also known as the escape-gesture

### Which platform(s) does your issue occur on?
iOS only.

### Please provide the following version numbers that your issue occurs with:
- CLI: 4.1.x, 4.0.x, 3,x etc.
- Cross-platform modules: 4.1.0
- Runtime(s): tns-ios@4.1.0
- Plugin(s): none

### Please tell us how to recreate the issue in as much detail as possible.
The problem is in two parts, both requires `VoiceOver` to be enabled to demonstrate.

#### Dialogs
Any dialog created from `tns-core-modules/ui/dialogs` cannot be dismissed with the z/escape-gesture. NativeScript doesn't implement the `accessibiliityPerformEscape()`-function on `UIAlertController`.

#### Modal pages without a `NavigationButton`.
Create a new page and hide the `NavigationButton`. This disables the z-gesture for this page.

Our accessibility users expects the z-gesture working in both these cases.

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.

I've create a simple example here:
https://play.nativescript.org/?template=play-ng&id=vuchBL&v=10

You need to enable VoiceOver to see the problem.

1. Start by tapping "Show Alert", this will create an alert.
Try to dismiss it by drawing a Z on the screen with two fingers.
The device will make a noise but nothing else will happen.
Our accessibility users expects the alert to be dismissed and complains about it.

2. Close the alert manually.

3. Tap the "Show modal page without the NavigationButton"
This page will not have the `NavigationButton` to the left but will have a "done"-button to the right. (This is a common design for settings pages in our app).
Once again try to dismiss the page with the z-gesture.
This won't work.

4. Go back to the home component again.

5. Now tap the last button: "Show modal page with the NavigationButton"
This is the same page as before but with the NavigationButton.
Once again try to dismiss the page with the z-gesture.
This time you end up on the previous page, as expected.

I've previous submitted #4349 but back when I didn't understand the cause and ended up closing the issue, because it looked like an error in our end.

NickIliev:
@m-abs marking this one as a feature request.

If you have an idea on how the above can be implemented, we would be more than happy to accept PRs (or you could share the ideas so the community can take a look and implement these gestures).

[...]

PR #8909, opened 2020-09-27
feat(core): a11y support

m-abs:
## PR Checklist

NativeScript doesn't support native accessibility features without our plugin `@nota/nativescript-accessibility-ext`

Unfortunately that plugin will soon go unmaintained.
I'd hate for the community to loose this feature, so here I'll contribute the feature to upstream NativeScript. 

## What is the new behavior?

- Adds new properties to `View`-class.
- CSS-helpers.
- Regain accessibility focus on back navigation. 
etc.

### TODO:
- [x] Font-size scaling for `iOS`
- [ ] GridLayout sizing. (current implementation was angular specific: https://github.com/Notalib/nativescript-accessibility-ext/blob/master/packages/nativescript-accessibility-ext/angular/directives/a11y-grid-layout.directive.ts)
- [ ] Add helper classes to theme (help needed) (current: https://github.com/Notalib/nativescript-accessibility-ext/tree/master/packages/nativescript-accessibility-ext/scss) - some of this can be done programmatically in core now.
- [x] ListView scroll to item on swipe to item. (user will loose focus if trying to swipe to an item out of view in the list-view).
- [ ] Testing.
- [ ] Update API Documentation (help needed)
- [ ] Magic Tap (iOS) (Must be registered on AppDelegate: https://developer.apple.com/documentation/objectivec/nsobject/1615137-accessibilityperformmagictap). Provide a `Application.accessibilityMagicTapEvent` which users can wire into.
- [ ] Fix escape gesture issue see: #6123 (`accessibilityPerformEscape` https://developer.apple.com/documentation/objectivec/nsobject/1615091-accessibilityperformescape). Allow default to dismiss the active dialog however also emit a new `Application.accessibilityPerformEscapeEvent` which allows an override for custom behavior - most likely via `args.cancel = true` from the event emission.

Design document here:
https://docs.google.com/document/d/1gybrT1IQh8Fb5r0v_tZTMS06yP6xYmr9pDQIBRbokZc/edit

Implements #6993
Fixes: #6125

BREAKING CHANGES:
Remove the property `automationText` as it conflicts with the new property `accessibilityId` the implements the behavior correctly.

[Describe the impact of the changes here.]

Migration steps:
[Provide a migration path for existing applications.]


m-abs:
@cla-bot check

NathanaelA:
Can you create a property `automationText` that points to `automationId` so that it is no longer a breaking change?

m-abs:
> Can you create a property `automationText` that points to `automationId` so that it is no longer a breaking change?

I can but it won't be exactly the same as the current behavior. 

 `automationId`  sets `automationIndentifier` on `iOS` and doesn't do anything on `android`.

Currently `automationText` sets both `accessibilityLabel` and `automationIndentifier` on `iOS` and overrides `contentDescription` on `android`.

Hmm... I should probably rename  `automationId`  to `automationIndentifier`, to make it more consistent with the native platform.

rigor789:
> `automationId` sets `automationIndentifier` on `iOS` and doesn't do anything on `android`.

@m-abs I think for android, we can do something like this: 

```js
const id = utils.ad.resources.getId(':id/nativescript_accessibility_id')
this.nativeViewProtected.setTag(id, value)
this.nativeViewProtected.setTag(value)
```

Tools like [wix/Detox](https://github.com/wix/Detox) use that when getting views by id on android, and `accessibilityIdentifier` on ios.


I don't 100% recall why I had to set the tag with the id and without, I took that from a sample app I made with detox: 
https://github.com/rigor789/ui-tested-nsvue-app/blob/fecd9ed0392b8903d20a0967d87d60201082bc32/app/app.js#L18-L25


In my case I named the property `testID` as that's the convention in react-native, but whatever is the most consistent should be fine!

m-abs:
@rigor789 

Thanks for working that out. :+1: I'll add it right away.

farfromrefug:
A few comments. The accessibility/index.android.ts is not snapshot compatible. It accesses native namespace at its root which will fail in snapshot generation.
I am a bit worried about the whole `initA11YView`. Storing all views from an app seems "big". Can't we find another way to get the N view from a native view ? 
Can we quantify everything which is added to an app not using accessibility?

NathanaelA:
>A few comments.

Awesome, Thanks!

> The accessibility/index.android.ts is not snapshot compatible. It accesses native namespace at its root which will fail in snapshot generation.

(Responding to this just so others seeing this issue understand, since you already know this) Snapshots are depreciated and unsupported in 7.x; we are not going to continue to mangle the code because of something that is **not** going to be supported anymore, code cache is the future.    

---
> I am a bit worried about the whole initA11YView. Storing all views from an app seems "big". Can't we find another way to get the N view from a native view ?

That is an interesting point, and a very valid point!    Thanks for noticing it!    It is using a WeakMap w/ WeakRef; the memory usage will grow some; because of the tracking (i.e. a weakmap); but GC should be able to keep the memory usage to a minimum as when the native object is GC'd the ref & map should be cleaned up.   However, we should run some tests to double check that the keys will be cleared when the ref is cleared, if not, we might have to manually .delete(key) on `unloaded`

However, the memory tradeoff for getting real A11Y support is well worth it even if it was double the memory was being used.  A11Y has become a much bigger item in multiple countries (as it should have a long time ago).  In addition in the USA multiple law suits have been filed against companies do to legislation.   So this is something we need to be on top of...
 
---

> Can we quantify everything which is added to an app not using accessibility?

I am not understanding the question, can you clarify?

m-abs:
> I am a bit worried about the whole `initA11YView`. Storing all views from an app seems "big". Can't we find another way to get the N view from a native view ?

We initial made this as a trade off.
In earlier versions of our plugin, we added a `NotificationObserver` per view, this was as you an imagine very expensive.
The `WeakMap` from a `UIView` to a `WeakRef` was the least expensive and safe way we found to do it. In our experience it doesn't cause too much of a memory overhead.

I've considered binding the `View`-instance to the native `UIView`, but I'm not sure if that reference survives when we get it inside the `NotificationObserver` callback.

m-abs:
> (Responding to this just so others seeing this issue understand, since you already know this) Snapshots are depreciated and unsupported in 7.x; we are not going to continue to mangle the code because of something that is **not** going to be supported anymore, code cache is the future.

Does this mean, I can ship the `ensureNativeClasses`-functions? They tend to make the code less readable.

farfromrefug:
Please I know you deprecate this but people still use it ! If this PRs gets in as it is it will break snapshots.  I would gladly update the code to make it snapshot compatible. It does not represent much work.
And I should point out that the way we write thing for snapshot is actually better performance wise! No no need to Marshal classes if they are not used at runtime !!! (Even If the file is actually imported).
I think N should continue to write android classes that way. Not so much drawback code wise.

As an example here all namespace / classes you added at the top of the file for easy access means they are marshalled (which can be dead slow, look at android R namespace as an example) even though they might never be used.

@NathanaelA I meant I would like to know the consequence of that PR (perfs, men, but mostly perfs) on an app which actually does not "use" or implement accessibility.

m-abs:
@farfromrefug 
Is it better now?

farfromrefug:
@m-abs thanks a lot it is perfect now!

m-abs:
I think, I found a better way to get from the `UIView` to the nativescript `View`-instance.

I use the `UIView.tag` and `view._domId`.
This adds a overhead for looking up the focused view, but removes the need for the WeakMap.

NathanaelA:
> Please I know you deprecate this but people still use it ! If this PRs gets in as it is it will break snapshots. I would gladly update the code to make it snapshot compatible. It does not represent much work.

Isn't so much the time spent making the code work; their is a wide number of factors for why it is being depreciated and will become broken in the future.    **However, this is NOT the issue to argue this in.**   This issue is for A11y and as such @m-abs is 100% free to implement it with or without a wrapper function is the only point I was trying to make.   

So at least for this issue; lets keep things on topic regarding A11y and not snapshot vs code cache.   I will be trying to create a specific issue for Snapshot depreciation in the next couple days that you can weigh in and argue any snapshot vs code cache and that we can put real benchmarks in to show the differences.

---

>Does this mean, I can ship the ensureNativeClasses-functions? They tend to make the code less readable.
Please feel free to ship it either way; we are still potentially _looking_ & _considering_ at making 7.0 be usable for snapshots; and formally announcing 7.1 as it is not supported (& breakable at that point).     We did forget to announce it wasn't supported in 7.0; but do not feel constrained by our mistake.  :grinning:    

We also want the code to be the most readable/maintainable; and if you believe that is without the wrapper functions, we are fine with whatever your choice is as a developer.  :grinning: 

---

>  I meant I would like to know the consequence of that PR (perfs, men, but mostly perfs) on an app which actually does not "use" or implement accessibility.

This is an excellent question; I'd love to know that myself.  :grinning:    However in this specific case A11y has a significant priority over speed just because of the reach of a feature like this adds to the engine with a large number of companies and developers.   In addition the legal climate in multiple countries over A11y support also plays a factor.   So, I'd love to see benchmarks on if this has any effect; but as much as I love speed and prefer speed over almost anything, This is one area that actually, imho, trumps any speed issues...

NathanaelA:
@m-abs - Pardon what might be silly (& stupid) questions, trying to understand the changes as I have no real experience with A11y.  :grinning:   

1. I don't see anything that calls the `setContentDescription` https://developer.android.com/reference/android/view/View.html#setContentDescription 
I was under the impression that was needed for accessibility support for screen readers.

2. Are you aware that Appium for android uses https://github.com/NativeScript/nativescript-dev-appium (findElementByAccessibilityId) Find by Accessibility Id -- which is the primary method used by most test and why `automationText` exists in android.  This uses the `setContentDescription` value.    It is possible that Igor's fix will fix that??? or perhaps now Appium can actually use FindById via Igor's fix, which has never worked before...     Don't know on this, just asking anyone who might now...     Before we can merge this; we need to make sure Appium will continue to work and still has a easy to use "byid" element retrieval even if we have to change the internals of our Appium driver ..

packages/core/ui/core/view/index.android.ts
```diff
@@ -1029,6 +1103,30 @@ export class View extends ViewCommon {
 			(<any>nativeView).background = undefined;
 		}
 	}
+
+	public androidSendAccessibilityEvent(eventName: AndroidAccessibilityEvent, msg?: string): void {
+		if (!isAccessibilityServiceEnabled()) {
```
facetious, line 1108: Checking the value here (and in the following two functions) seems superfluous to me. The final destination of all three ends up within `sendAccessibilityEvent`, which checks this condition anyway.

packages/core/ui/core/view/index.d.ts
```diff
@@ -225,6 +241,68 @@ export abstract class View extends ViewBase {
...
+
+	/**
+	 * This view starts a media session. Equivalent to trait = startsMedia
+	 */
+	accessibilityMediaSession: boolean;
+
+	/**
+	 * Internal use only. This is used to limit the number of updates to android.view.View.setContentDescription()
+	 */
+	_androidContentDescriptionUpdated?: boolean;
```
facetious, line 302: If it's internal use only, it could be left out of the `.d.ts` and just locally cast in order for TypeScript to continue validating the type.

packages/core/ui/core/view/index.d.ts
```diff
@@ -662,6 +735,29 @@ export abstract class View extends ViewBase {
...
+	 */
+	public androidSendAccessibilityEvent(eventName: AndroidAccessibilityEvent, msg?: string): void;
+
+	/**
+	 * iOS: post accessibility notification.
+	 * type = 'announcement' will announce `args` via VoiceOver. If no args element will be announced instead.
+	 * type = 'layout' used when the layout of a screen changes.
+	 * type = 'screen' large change made to the screen.
+	 */
+	public iosPostAccessibilityNotification(notificationType: IOSPostAccessibilityNotificationType, msg?: string): void;
```
facetious, line 749: I think these two methods would be best combined together.

```typescript
interface AccessibilityEventOptions {
  androidAccessibilityEvent: AndroidAccessibilityEvent;
  iosNotificationType: IOSPostAccessibilityNotificationType;
  message: string;
}

public sendAccessibilityEvent(options: Partial<AccessibilityEventOptions>): void;
```

packages/core/ui/page/index.d.ts
```diff
@@ -161,6 +166,11 @@ export declare class Page extends PageBase {
 	 */
 	public onNavigatedFrom(isBackNavigation: boolean): void;
 	//@endprivate
+
+	/**
+	 * Announce screen changed
+	 */
+	public accessibilityScreenChanged(refocus?: boolean): void;
```
facetious, line 173: I think you mean for this to be within the `//@endprivate`. It should also follow the pattern: `onAccessibilityScreenChanged`.

packages/core/ui/page/page-common.ts
```diff
@@ -32,6 +32,7 @@ export class PageBase extends ContentView {
 	public enableSwipeBackNavigation: boolean;
 	public backgroundSpanUnderStatusBar: boolean;
 	public hasActionBar: boolean;
+	public accessibilityAnnouncePageEnabled = true;
```
facetious, line 35: This property should mirror the definition of the other boolean properties, including specifying the default value elsewhere.

facetious (commented):
There are a few files that are misspelled as acessibility with one c; most notably the folder that they all live within.

farfromrefug:
> We also want the code to be the most readable/maintainable; and if you believe that is without the wrapper functions, we are fine with whatever your choice is as a developer. 😀
>  you are not saying you prefer code readability over performances right ? 
You know materialisation (I said marshalling before but it was not the right term) has a very heavy cost, right ? I wonder if it is actually the real reason why the Previous N team  was coding that way.

packages/core/acessibility/accessibility-service-common.ts
```diff
@@ -0,0 +1,35 @@
+import { Observable } from '../data/observable';
+
+export class SharedA11YObservable extends Observable {
+	accessibilityServiceEnabled?: boolean;
+}
+
+export const AccessibilityServiceEnabledPropName = 'accessibilityServiceEnabled';
+
+export class CommonA11YServiceEnabledObservable extends SharedA11YObservable {
```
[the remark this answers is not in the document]
m-abs: I'm considering just writing it out as Accessibility :)

packages/core/acessibility/accessibility-service.android.ts
```diff
@@ -0,0 +1,130 @@
...
+
+const accessibilityStateEnabledPropName = 'accessibilityStateEnabled';
+const touchExplorationStateEnabledPropName = 'touchExplorationStateEnabled';
+
+class AndroidSharedA11YObservable extends SharedA11YObservable {
+	[accessibilityStateEnabledPropName]: boolean;
+	[touchExplorationStateEnabledPropName]: boolean;
+
+	get accessibilityServiceEnabled(): boolean {
+		return !!this.accessibilityServiceEnabled && !!this.touchExplorationStateEnabled;
```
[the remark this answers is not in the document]
m-abs: Thanks, that is clearly an error.

packages/core/acessibility/index.android.ts
```diff
@@ -0,0 +1,667 @@
...
+			}
+
+			notifyAccessibilityFocusState(view, false, true);
+
+			return;
+		}
+	}
+}
+
+let TNSAccessibilityDelegate: android.view.View.androidviewViewAccessibilityDelegate;
```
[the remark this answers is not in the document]
m-abs: It is due to an typing error.
In the d.ts the class is called `android.view.View.androidviewViewAccessibilityDelegate` but it should be called android.view.View.AccessibilityDelegate;

I think @NathanWalker is fixing it

packages/core/acessibility/accessibility-css-helper.ts
```diff
@@ -0,0 +1,106 @@
...
+		currentFontScaleClass = fontScaleCssClasses.get(1);
+	}
+
+	if (oldFontScaleClass !== currentFontScaleClass) {
+		applyRootCssClass([...fontScaleCssClasses.values()], currentFontScaleClass);
+	}
+
+	const oldActiveFontScaleCategory = currentFontScaleCategory;
+	if (global.isAndroid) {
+		currentFontScaleCategory = fontExtraMediumClass;
```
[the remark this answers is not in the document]
m-abs: Unfortunately there is a huge difference between the font-size the user can set on the two platforms, I'll try to explain:

**Android:**

The font-size is an integer between 0.85 and 1.3 (with some rounding errors).

**iOS:**

We have the setting `preferredContentSizeCategory` that maps to a global string constant, they are mapped here:
https://github.com/NativeScript/NativeScript/blob/6b7cbe63118e425a79c11d0d92b0f54b52c8a740/packages/core/acessibility/fontscale-observable.ios.ts#L14-L27

Range `0.5, 0.7, 0.85, 1, 1.15, 1.3, 1.5, 2, 2.5, 3, 3.5, 4`

Values <0.85 we call extra small
Values >1.5 we call extra small

The (almost) shared range on iOS and Android 0.85 to 1.5 we call medium size.

We introduced this, because large increases in the font-sizes like 400% breaks most layouts.

I hope that makes sense.

packages/core/ui/core/view/index.d.ts
m-abs: Good idea.

packages/core/ui/page/index.d.ts
m-abs: The developer should be able to call this function for large updates on the page.

packages/core/ui/core/view/index.d.ts
m-abs: I get a typing error in `packages/core/acessibility/index.android.ts` if I remove it.

packages/core/ui/core/view/index.android.ts
```diff
@@ -726,13 +736,6 @@ export class View extends ViewCommon {
 		org.nativescript.widgets.OriginPoint.setY(this.nativeViewProtected, value);
 	}
 
-	[automationTextProperty.getDefault](): string {
```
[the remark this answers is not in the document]
m-abs: After @NathanaelA suggestion, I added an setter/getter pointing to the new `accessibilityIdentifier` property, which is probably what they'd want.

packages/core/application/index.d.ts
```diff
@@ -183,6 +183,11 @@ export function setCssFileName(cssFile: string): void;
  */
 export function getCssFileName(): string;
 
+/**
+ * Ensure css-class is set on rootView
+ */
+export function applyCssClass(rootView: View, cssClasses: string[], newCssClass: string): void;
```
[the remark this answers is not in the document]
m-abs: Maybe I should just import it from `application-common` instead.

Although it might be a good idea to make it available for app developers?

packages/core/acessibility/index.ios.ts
```diff
@@ -0,0 +1,264 @@
...
+ * @param values string values
+ * @param map    map lower-case name to integer value.
+ */
+function inputArrayToBitMask(values: string | string[], map: Map<string, number>): number {
+	return (
+		enforceArray(values)
+			.filter((value) => !!value)
+			.map((value) => `${value}`.toLocaleLowerCase())
+			.filter((value) => map.has(value))
+			.reduce((res, value) => res | map.get(value), 0) || 0
```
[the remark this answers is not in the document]
m-abs: If you have a better suggestion, I'd love to hear it.

`UiView.accessibilityTraits` is a bitmask for all the traits you want to set on the `UIView`.

While I'd prefer for developers to use `accessibilityRole` + `accessibilityState` sometimes it might be a nice to have more direct control.

packages/core/ui/page/page-common.ts
m-abs: I'm not sure I understand.

m-abs:
Hi @NathanaelA 

Sorry your reply got lost in all the other replies.

> @m-abs - Pardon what might be silly (& stupid) questions, trying to understand the changes as I have no real experience with A11y. 
> 
> 1. I don't see anything that calls the `setContentDescription` https://developer.android.com/reference/android/view/View.html#setContentDescription
>    I was under the impression that was needed for accessibility support for screen readers.

It is set here: https://github.com/m-abs/NativeScript/blob/feat/accessibility/packages/core/accessibility/index.android.ts#L563-L667

I attempt to recreate the behavior from iOS by concatenating `accessibilityLabel`, `accessibilityValue` and `accessibilityHint here`.

> 2. Are you aware that Appium for android uses https://github.com/NativeScript/nativescript-dev-appium (findElementByAccessibilityId) Find by Accessibility Id -- which is the primary method used by most test and why `automationText` exists in android.  This uses the `setContentDescription` value.    It is possible that Igor's fix will fix that??? or perhaps now Appium can actually use FindById via Igor's fix, which has never worked before...     Don't know on this, just asking anyone who might now...     Before we can merge this; we need to make sure Appium will continue to work and still has a easy to use "byid" element retrieval even if we have to change the internals of our Appium driver ..

We definitely need to have Appium working with this, but my experience with it is very limited.

@rigor789 Do you have some input here? Can `FindById` find views tagged via `setTag`?

vigdora:
Hi 
@NathanaelA, @rigor789 @m-abs 
the change has broken our automation e2e test using **Appium**,

on android, we used to benefit from using automationText prop
since it added the necessary  `content-desc` attribute to android's elements. 
this way Appium managed to find those elements using the _accessibility id_ selector.

for reference - look at the NS7 code:

` [automationTextProperty.setNative](value) {
        this.nativeViewProtected.setContentDescription(value);
    }`
    
in NS8 it looks now like this (probably for solving the issue for **Detox** platform):

`  [accessibilityIdentifierProperty.setNative](value) {
        this.nativeViewProtected.setContentDescription(value)
        const id = Utils.ad.resources.getId(':id/nativescript_accessibility_id');
        if (id) {
            this.nativeViewProtected.setTag(id, value);
            this.nativeViewProtected.setTag(value);
        }
    }`

**(where automationText points to accessibilityIdentifierProperty.)**

as you can see,  setContentDescription was removed, thus all our automation tests on android are broken.
can the previous (NS7) solution can be combined to the new one?

m-abs:
hi @vigdora,

I'm sorry this causes you problems.

> can the previous (NS7) solution can be combined to the new one?

No, that would make the app inaccessible for accessibility users on android and defeat the purpose of this PR.

The old implementation was wrong, Android's `setContentDescription" is not equivalent of iOS's `accessibilityIdentifier`, they serve very different purposes.

Unfortunately you have to update your tests to make them work with this change.

I haven't worked with `appium` myself, but maybe this can help you:
https://github.com/rigor789/ui-tested-nsvue-app/blob/fecd9ed0392b8903d20a0967d87d60201082bc32/e2e/firstTest.spec.js#L24

Using `setTag` was suggested by @rigor789 who linked to that repo.

vigdora:
@m-abs
this is tough...
as for as I know,  Appium doesn't  let us find elements easily cross app except using the Accessibility ID selector.
without this selector, I will need to use xpath or other strategies which will make testing way inconvenient.

The link you referred to doesn't help much since  it uses Detox or still uses the Accessibility ID with  Appium.

farfromrefug:
@vigdora there is a quick workaround. Create a plugin which adds a mixin N. In that mixin override the `[accessibilityIdentifierProperty.setNative]` function. In there call `setContentDescription` as well as the original method.
I could even be part of the nativescript appium plugin.

vigdora:
@farfromrefug 
I already tried it - but it doesnt work since the function _applyContentDescription_ (found in @nativescript/core/accessibility/index.android.js)
is clearing the contentdescription if no accessibility attribute (i.e. accessibilityLabel)  is presented on the element:

image

m-abs:
@farfromrefug and @vigdora
I don't think this discussion belongs here on a merged PR.

You should probably create a new issue about it and maybe take the discussion on slack.

I'm not very involved with {N} anymore and don't have much time to help.

farfromrefug:
@m-abs thanks we all take it along from here. 
@vigdora I think we might be able to remove it. could you try ?

vigdora:
I gave it another thought and removing it still wont help as text attribute will populate content-description 
any way.
image


I honestly have no solution for it,
I would have liked to open a PR for fixing it but the only solution I see right now is to allow disabling the accessibility once a developer choses to use automation text.
what are the chances for it to be approved?

Issue #8925, opened 2020-10-01
Still getting __metadata is not defined using --env.roduction on Android


kryptus36:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):

✔ Getting NativeScript components versions information...
✔ Component nativescript has 7.0.8 version and is up to date.
✔ Component @nativescript/core has 7.0.9 version and is up to date.
✖ Component @nativescript/ios is not installed.
✔ Component @nativescript/android has 7.0.0 version and is up to date.

"dependencies": {
    "@nativescript-community/ui-pager": "^13.0.0-alpha.10",
    "@nativescript/core": "~7.0.9",
    "@nativescript/firebase": "^11.0.0",
    "@nativescript/iqkeyboardmanager": "^2.0.0",
    "@types/node": "^12.12.7",
    "moment": "^2.27.0",
    "moment-timezone": "^0.5.28",
    "@nativescript/appversion": "^2.0.0",
    "nativescript-audio-player": "^1.0.3",
    "@nativescript/geolocation": "^7.0.0",
    "nativescript-insomnia": "^2.0.0",
    "nativescript-mqtt": "^1.2.0",
    "@nativescript/secure-storage": "^3.0.0",
    "@nativescript/theme": "^3.0.0",
    "nativescript-ui-listview": "^9.0.2",
    "nativescript-ui-sidedrawer": "^9.0.3"
  },
  "devDependencies": {
    "@nativescript/android": "7.0.0",
    "@nativescript/types": "7.0.0",
    "@nativescript/webpack": "~3.0.0",
    "node-sass": "^4.14.1",
    "tslint": "^6.1.1",
    "typescript": "3.9.7"
  },

**Describe the bug**
The app runs fine until I add --env.production. I then get:
System.err: java.lang.RuntimeException: Unable to create application com.tns.NativeScriptApplication: com.tns.NativeScriptException: Error calling module function 
System.err: ReferenceError: __metadata is not defined

May be related to the bugs referenced in #8903 but that was apparently fixed. 

**To Reproduce**
ns run android --emulator --env.production

**Expected behavior**
app runs

I did also make sure to run ns clean to make sure my build was fresh.

farfromrefug:
@kryptus36 by any chance can you share your project ? If not maybe your app/main.ts ? It would help in figuring out why it still crashes

kryptus36:
[...]

PR #8931, opened 2020-10-03
fix: ensure globals do not get tree shaked

farfromrefug:
For some reason we need to have the three of those for it to work correctly
Fixes https://github.com/NativeScript/NativeScript/issues/8925

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -10,7 +10,9 @@
       "url": "https://github.com/NativeScript/NativeScript"
   },
   "sideEffects": [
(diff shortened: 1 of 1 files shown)


NathanWalker:
Interesting thanks @farfromrefug - makes sense as due to global handling sideEffects probably does need to be expanded.

NathanWalker (approved):

NathanWalker:
Release in https://github.com/NativeScript/NativeScript/releases/tag/7.0.10-core now.

PR #8954, opened 2020-10-10
Fix: Fix iOS 10 Issue with UILayoutGuide

NathanaelA:
iOS 10 seems to have an issue with reallocing UILayoutGuide after a init.  So on the second pass thru this function that app will crash on iOS 10 with an error about `init` not being a function. 

I suspect there might be a issue with Allocation size of the object on iOS 10, and bears some detailed debugging.  But at this point using the new() works around the issue so that the app won't crash on iOS 10.

### DIFF
--- a/packages/core/ui/core/view/view-helper/index.ios.ts
+++ b/packages/core/ui/core/view/view-helper/index.ios.ts
@@ -201,7 +201,7 @@ export class IOSHelper {
 
 	static initLayoutGuide(controller: UIViewController) {
 		const rootView = controller.view;
-		const layoutGuide = UILayoutGuide.alloc().init();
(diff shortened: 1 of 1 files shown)


NathanWalker:
Thanks @NathanaelA - we'll need to use different constructor as fails to build:
```
Compiling TypeScript files for library core...
packages/core/ui/core/view/view-helper/index.ios.ts(204,23): error TS2554: Expected 1 arguments, but got 0.
```

NathanWalker:
Pushed the fix, thanks for kicking this off, ios 10 is fixed 👍 pushing a patch on this.

NathanWalker (approved):

triniwiz (commented):
LGTM

PR #8967, opened 2020-10-17
feat(core): TypeScript 4 + cleanup

NathanWalker:
* [x] TypeScript 4.x
* [x] Updates workspace to Nx 10.3.x
* [x] Adds various graphics from old repo which were used in e2e ui testing app fixing previously failing test pages
* [x] all tests passing
* [x] removed unused core-compat 
* [x] fixes core build to ensure TypeScript errors output to console if encountered rather than swallowed


packages/core/application/index.ios.ts
```diff
@@ -528,3 +529,6 @@ global.__onLiveSync = function __onLiveSync(context?: ModuleContext) {
 	const rootView = getRootView();
 	livesync(rootView, context);
 };
+
+// consistent symbol access
+export class AndroidApplication extends Observable {}
```
rigor789, line 534: Could we export a null/undefined const instead?

Something like
```ts
export const AndroidApplication = undefined;
```
Semantically that'd make sense, since AndroidApplication is undefined in an iOS context.

Another note, perhaps reword the comment to explain why this is done - so it's clearer when we stumble across this line in the future.

packages/core/application/index.android.ts
```diff
@@ -504,3 +504,6 @@ declare namespace com {
 		}
 	}
 }
+
+// consistent symbol access
+export class iOSApplication {}
```
rigor789, line 509: Can we export an undefined const instead? (more info in the android export comment)

packages/core/image-asset/index.android.ts
```diff
@@ -15,6 +15,7 @@ export class ImageAsset extends ImageAssetBase {
 		this.android = fileName;
 	}
 
+	// @ts-ignore
```
rigor789, line 18: Why do we need to `ts-ignore` on these getters (all of the ones in the pr)?

Curious if we could change the types a bit instead to not require a ts-ignore if possible!

packages/core/ui/proxy-view-container/index.ts
```diff
@@ -1,4 +1,3 @@
-import { ProxyViewContainer as ProxyViewContainerDefinition } from '.';
```
rigor789, line 1: That's strange - it used to implement itself, or was `ProxyViewContainer` both a class name and an interface name? How did this even compile before, without getting into an infinite loop. 

The change LGTM!

packages/core/application/index.ios.ts
NathanWalker: Good question - this is a practice that does emit warnings when live syncing projects against source ( not as much issue when using npm built core ) - it's part of a practice that is emerged in core over the years which lends itself to a little misdirection since `AndroidApplication` is a symbol exported by core but yet only available when running android app.

packages/core/application/index.android.ts
NathanWalker: I'll try a `const` to see if suffices.

packages/core/image-asset/index.android.ts
NathanWalker: This is related to a new breaking change in TS 4.x which may become a compilerOption flag in future to ignore/silence but does not seem a flag is available yet.
https://github.com/microsoft/TypeScript/issues/39401

packages/core/ui/proxy-view-container/index.ts
NathanWalker: Indeed yes few places in core had also been doing this unnecessary practice. Just cleanup here 👍

packages/core/application/index.android.ts
NathanWalker: @rigor789 nice - `const` works perfect as well - updated now.

packages/core/application/index.ios.ts
NathanWalker: `const` updated now 👍

packages/core/application/index.ios.ts
```diff
@@ -11,6 +11,7 @@ import { View } from '../ui/core/view';
 import { NavigationEntry } from '../ui/frame/frame-interfaces';
 // TODO: Remove this and get it from global to decouple builder for angular
 import { Builder } from '../ui/builder';
+import { Observable } from '../data/observable';
```
rigor789, line 14: Probably unused after switching to the undefined const?

rigor789 (dismissed):

packages/core/application/index.ios.ts
NathanWalker: good catch! removed now 👍

rigor789 (approved):

PR #8921, opened 2020-09-30
Fix to ensure index remains > 0 on splice

farfromrefug:
Otherwise it will crash in ListView for iOS for example as it tries to animate an item with a negative row index. See issue here https://github.com/NativeScript/NativeScript/pull/8900#issuecomment-701539011

PR #8968, opened 2020-10-17
fix: fix Observable-Array, its tests and a typo

NathanaelA:
Fixes the automated tests from the changes done in PR #8900 & #8921 which broke two of the tests.
- Those PR's also created a couple corner cases on deletion, which are now fixed.
- Added `Length` growing array (never worked, but now does, so now more like a real array)
- Added additional tests to catch the corner cases in the future.

### DIFF
--- a/apps/automated/src/data/observable-array-tests.ts
+++ b/apps/automated/src/data/observable-array-tests.ts
@@ -361,7 +361,7 @@ export const test_ObservableArray_spliceShouldRemoveSpecifiedNumberOfElementsSta
 	let result: ChangedData<number>;
 
 	// >> observable-array-splice-change
-	const array = new ObservableArray([1, 2, 3]);
+	const array = new ObservableArray([1, 2, 3, 4]);
 
 	array.on(ObservableArray.changeEvent, (args: ChangedData<number>) => {
 		// Argument (args) is ChangedData<T>.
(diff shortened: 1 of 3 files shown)


apps/automated/src/data/observable-array-tests.ts
```diff
@@ -360,6 +360,85 @@ export const test_ObservableArray_spliceShouldRemoveSpecifiedNumberOfElementsSta
...
+
+		// >> (hide)
+		result = args;
+		// << (hide)
+	});
+
+	array.splice(1, 2);
+	// << observable-array-splice-change
+
+	TKUnit.assert(result.eventName === ObservableArray.changeEvent && result.action === ChangeType.Splice && result.removed.length === 2 && result.index === 1 && result.addedCount === 0, "ObservableArray splice() should raise 'change' event with correct args!");
```
rigor789, line 382: Would it make sense to break these into multiple asserts, or at least format it to be a bit more readable?

maybe just break into multiple lines?
```ts
TKUnit.assert(
  result.eventName === ObservableArray.changeEvent &&
  result.action === ChangeType.Splice && 
  result.removed.length === 2 &&
  result.index === 1 && result.addedCount === 0,
  "ObservableArray splice() should raise 'change' event with correct args!"
)

File: apps/automated/src/data/observable-array-tests.ts [hunk 2]
```diff
@@ -360,6 +360,85 @@ export const test_ObservableArray_spliceShouldRemoveSpecifiedNumberOfElementsSta
...
+		result = args;
+		// << (hide)
+	});
+
+	// Because their is only one item in the above array the item index should be
+	// normalized to Index 1.
+	array.splice(2, 0, 1);
+	// << observable-array-splice-change
+
+	TKUnit.assert(result.eventName === ObservableArray.changeEvent && result.action === ChangeType.Splice && result.removed.length === 0 && result.index === 1 && result.addedCount === 1, "ObservableArray splice() should raise 'change' event with correct args!");
```
rigor789, line 409: Same here?

apps/automated/src/data/observable-array-tests.ts
```diff
@@ -360,6 +360,85 @@ export const test_ObservableArray_spliceShouldRemoveSpecifiedNumberOfElementsSta
...
+		result = args;
+		// << (hide)
+	});
+
+	// Because we are starting at index 2, their is NOTHING to delete
+	// So the Starting index should actually be normalized to Index 1
+	array.splice(2, 2, 1);
+	// << observable-array-splice-change
+
+	TKUnit.assert(result.eventName === ObservableArray.changeEvent && result.action === ChangeType.Splice && result.removed.length === 0 && result.index === 1 && result.addedCount === 1, "ObservableArray splice() should raise 'change' event with correct args!");
```
rigor789, line 436: Same here?

packages/core/data/observable-array/index.ts
```diff
@@ -246,7 +251,15 @@ export class ObservableArray<T> extends Observable {
 			eventName: CHANGE,
 			object: this,
 			action: ChangeType.Splice,
-			index: Math.max(Math.min(start, this._array.length-1), 0),
+
+			// The logic here is a bit weird; so lets explain why it is written this way
+			// First of all, if you ADD any items to the array, we want the index to point to
+			//   the first value of the index, so this fixes it when you put a value to high in
```
rigor789, line 257:
```suggestion
			//   the first value of the index, so this fixes it when you put a value too high in
```

NathanWalker (approved):

packages/core/data/observable-array/index.ts
```diff
@@ -113,7 +114,11 @@ export class ObservableArray<T> extends Observable {
 
 	set length(value: number) {
 		if (types.isNumber(value) && this._array && this._array.length !== value) {
-			this.splice(value, this._array.length - value);
+			let added=[];
+			for (let i=this._array.length;i < value;++i) {
+				added.push(undefined);
+			}
```
rigor789, line 120: Would this be cleaner/faster? (According to a quick benchmark, it is 50% faster, but I don't trust benchmarks that much)

```suggestion
			let added = Array(Math.max(value - this._array.length, 0)).fill(undefined);
```

apps/automated/src/data/observable-array-tests.ts
NathanaelA: These are direct copies of the existing Tests.    Not opposed to making changes; but probably should do it to all of them so that they all stay consistent...

packages/core/data/observable-array/index.ts
NathanaelA: Not sure on that one.   Interesting change...  I'm not opposed to it if you want to do another PR to change this...

The only thing is that we would have to probably put a comment on it; each time I look at it -- I have to read it several time to understand what it is doing.  :grinning:

farfromrefug: kind of agree with @NathanaelA. Also i am not even sure `Array.fill` is faster than `for`

farfromrefug:
@NathanaelA awesome and sorry for the first PR not working correctly!

NathanaelA:
@farfromrefug - Not a problem at all; good thing the tests were setup for it; I had to really sit down and think through the logic of what was expected of the index value.   

So all in all your PR was the catalyst to fix a couple issues in this area including adding length growing so that it acted more like a normal array. ;-)

Issue #5268, opened 2018-01-14
More CSS3 styles needed


jamesharvey2:
Hello NativeScript,

Please implement missing CSS3 styles such as box shadow, text shadow, psuedo classes, CSS3 shapes (https://hackernoon.com/mastering-css-series-shape-outside-44d626270b25), CSS3 filters (https://www.w3schools.com/cssref/css3_pr_filter.asp) and so on. 

Also, please increase number of built-in components we can use. More button styles, sliders, tab menus are all welcome.

tsonevn:
Hi @jamesharvey2,
Thank you for the provided feature request,
I reviewed your suggestion, however, the box shadow functionality is already logged here.
About the third suggestion, we have already implemented some pseudo-selectors. For example [button :highlighted](https://docs.nativescript.org/ui/styling#pseudo-selector). 

Regarding that, it would help if you could provide some more info about the pseudo-classes, which you think it is needed to be added in NativeScript.

Also, keep in mind that there are some limitations on iOS and Android, which could prevent us from including all CSS styling features in NativeScript.

Best Regards,

jamesharvey2:
Hi Tsonevn,

Thanks for responding. 

Text-shadow property for instance, can do amazing tricks in terms of styling. 
Check these youtube video examples and see how you can be creative with text shadow: 

https://www.youtube.com/watch?v=zM2PGapc_vM

https://www.youtube.com/watch?v=3wKE8JZzA5Y&t=20s

I can't understand why text shadow property is missing in NativeScript.  Currently, NativeScript ships with only the minimum CSS3 styles which I find to be not suitable enough for complicated UI design. 

Check React Native regarding implementing text-shadow on native layer. React Native ships with box-shadow, text-shadow.  https://facebook.github.io/react-native/docs/stylesheet.html

Also, I would like to recommend CSS3 3D properties. 
https://www.youtube.com/watch?v=HoFsAyHQLeU

and CSS3 filters! 
https://www.youtube.com/watch?v=OHX8GdpPcRA
https://www.w3schools.com/cssref/css3_pr_filter.asp

I would love get :nth-child psuedo selectors, :after, :before psuedo selectors. These selectors are essential for designing complicated user interface. Check this youtube tutorial to learn how they can be used on CSS3 UI: 

https://www.youtube.com/watch?v=GfbuVM5neds&t=33s

jamesharvey2:
Here's an example of JavaScript & CSS3 calculator using box shadow property:

[...]

Issue #3597, opened 2017-02-07
[Feature] add outline to label and button text


kazemihabib:
### Tell us about the problem

Currently there is no way to add `outline` for texts in nativescript.
but we need that in cases the color of `Label` ( or `Button` with transparent background) and parent  are same so if we don't add an outline to text of button and label the element will be invisible.

I found that for android this can be done using `shadaw for Label` and `stroke for button`.

It would be really good if we had CSS property to set outline of text.

NordlingArt:
I actually tried to do this yesterday with native code but couldn't solve it. However, in OS you can at least create a shadow effect using this:

```js
var Color = require("color").Color;

function textShadow(view, hex, props = {}) {
  var defaults = {
    alpha: 1,
    x: 0, y: 0,
    radius: 1,
  };
  for(var key in defaults) if(!props.hasOwnProperty(key)) props[key] = defaults[key];
  
  if(view.ios) {
    nativeView = view.ios;
    nativeView.layer.masksToBounds = false;
    nativeView.layer.shadowOpacity = props.alpha;
    nativeView.layer.shadowRadius = props.radius;
    nativeView.layer.shadowColor = new Color(hex).ios.CGColor;
    nativeView.layer.shadowOffset = CGSizeMake(props.x, props.y);
  }
};
```

This won't create a perfect outline though because you cannot control the intensity. What I tried to do after some googling was this:

```js
function textOutline(view, color = "#ff0000", width = -1.0) {
  color = new Color(color);
  width = parseFloat(width);
  
  if(view.ios) {
    let nativeView = view.ios;
    let attributes = NSDictionary.dictionaryWithDictionary({
      NSStrokeColorAttributeName: color.ios,
      NSForegroundColorAttributeName: view.color.ios,
      NSStrokeWidthAttributeName: width
    });
    let attributedString = NSAttributedString.alloc().initWithStringAttributes(view.text || "", attributes);
    nativeView.attributedText = attributedString;
  }
}
```

But it doesn't work. It should, but nothing happens.

tsonevn:
Hi @kazemihabib,

This issue could be related to [#550](https://github.com/NativeScript/NativeScript/issues/550), where has been logged a feature request for implementing shadow CSS property for the components. Could you verify the attached issue and to vote for it if this is your case?

Regards,
@tsonevn

kazemihabib:
Hi, @tsonevn 

that issue is about box-shadaw but my requested feature is text-shadaw.
[...]

PR #8991, opened 2020-10-28
feat(core): add new text-shadow property

tralves:
## PR Checklist

NativeScript does not support `text-shadow`. This has been a requested feature (#3597, #5268).

## What is the new behavior?
Adds `text-shadow` css property (and `textShadow` attribute).

It implements a subset of the [text-shadow web syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/text-shadow). The expected format is:
```
<offset-x> <offset-y> <blur-radius> <color>
e.g.
2 10 4 rgb(255, 100, 100)
2px 10px 2px rgba(10, 10, 10, 0.5)
1 1 1 #55a
2 2 2 #aaa
``` 
The reasons for this limitation are:
- some of the default values in the web syntax are not clear (i.e., people shouldn't use them anyway);
- It is impossible to implement multiple shadows (e.g. `1px 1px 2px red, 0 0 1em blue, 0 0 0.2em blue;`);
- parsing/tokenisation would not be trivial because there are multiple formats for colors (`rba(0, 0, 0)`, `#fff`, etc.). It would be a lot of work;

I thought I should have more feedback on this PR before adding more work.

I think it PR would still need:
- Unit tests (I didn't add them yet, but I added a demo for this feature in the `apps.ui`);
- Add more syntax options (maybe? which ones?);
- Validation/syntax errors;
- Support for formatted strings and spans (not sure if it's even possible);
- Support for tab headers (not sure if it's even possible);

Demo (apps.ui):

ns-text-shadows

#3597 and part of #5268.


tralves:
Hi!
I see that this is on the roadmap :D. As I said this PR needs some polish. Do you have pointers or tips to what I should do to make your lives easier?

packages/core/ui/text-base/text-base-common.ts
```diff
@@ -26,26 +27,6 @@ export abstract class TextBaseCommon extends View implements TextBaseDefinition
...
-	 * NS Core uses nativeViewProtected for everything related to layout, so that it can be measured, added to the parent view as a child, ect.
-	 *
-	 * However, its internal view would be the actual TextView/TextInput and to allow that sub-view to have the normal TextView/TextInput
-	 * class features, which we expose and to allow them to work on it, the internal TextView/TextInput is what the needs to have the class values applied to it.
-	 *
-	 * So all code that works on what is expected to be a TextView/TextInput should use `nativeTextViewProtected` so that any third party
-	 * components that need to have two separate components can work properly without them having to duplicate all the TextBase (and decendants) functionality
-	 * by just overriding the nativeTextViewProtected getter.
-	 **/
 	get nativeTextViewProtected() {
```
farfromrefug, line 30: Why did you remove that comment ? It is actually an important comment

packages/core/ui/text-base/index.android.ts
```diff
@@ -378,6 +378,19 @@ export class TextBase extends TextBaseCommon {
...
 	}
 
+	[textShadowProperty.getDefault](value: number) {
+		return {
+			radius: this.nativeTextViewProtected.getShadowRadius(),
+			offsetX: this.nativeTextViewProtected.getShadowDx(),
+			offsetY: this.nativeTextViewProtected.getShadowDy(),
+			color: this.nativeTextViewProtected.getShadowColor(),
+		};
+	}
```
farfromrefug, line 388: First I dont think we need a default here. We have to be careful. Get default are slow because they make native calls. And yours even make 4 of them.
We now a default view as no shadow so not having a default value should be fine and faster.
Also you should use a const value to access native view protected instead of accessing it 4 times. If a class define it as a getter your would unnecessarly call it 4 times

packages/core/ui/text-base/index.ios.ts
```diff
@@ -188,6 +188,10 @@ export class TextBase extends TextBaseCommon {
 		this._setNativeText();
 	}
 
+	[textShadowProperty.setNative](value: TextShadow) {
+		this._setShadow(value);
```
farfromrefug, line 192: Why create a _setShadow method If you call it only once?
It makes the class."bigger" for now reason

NathanaelA: Doesn't really make the class bigger (two lines?), that code has to live somewhere.     I personally prefer it to be a separate function as it makes it easier to work on shadow stuff if we detect stuff broken.  In addition we might decide later to push it up the stack in a base class for other reasons.   Finally v8 engine typically optimizes smaller functions better than larger functions (& there used to be a limit to function sizes v8 will optimize, not sure if it is still present).  So having this in a separate function makes a lot of sense to me...

packages/core/ui/text-base/text-base-common.ts
NathanaelA: Agreed, this comment should remain, it is important to understand why their is two separate getters that normally point to the same property.

farfromrefug:
@NathanWalker you merged this without review comments being fixed ? that PR removes things which should not have been touched here !

rigor789:
@farfromrefug it's on the `release/8.0.0` branch where we are still going to do cleanups & changes. Just helps manage things together - and mainly helps resolving any conflicts early.

farfromrefug:
I get it. just that PR removes comments which should remain there for everyone to understand things. 
I don't see how you are going to bring them back now (except manually obviously) now that it is merged.
thanks for your work guys !

PR #9001, opened 2020-11-01
perf: faster color parsing

farfromrefug:
Also know the rework parser shares the same known colors  (faster, smaller bundle)
There are some eslint fixes in there hope it is fine


packages/core/css/parser.ts
```diff
@@ -1,3 +1,5 @@
+import { getKnownColor } from "color/known-colors";
```
NathanWalker, line 1: Thanks for this @farfromrefug - Let's make this relative import instead of barrel.

NathanWalker (changes requested):

packages/core/css/parser.ts
farfromrefug: Sorry did not see that. Vscode did that ;)

NathanWalker (approved):
Excellent thank you 💯

PR #9002, opened 2020-11-01
fix: only require parsers if need be

farfromrefug:
this also allows to remove them from bundle.
However this is not the best way. We should use something like `global.cssParser` so that weback automatically removes the code

### DIFF
--- a/packages/core/ui/styling/style-scope.ts
+++ b/packages/core/ui/styling/style-scope.ts
@@ -2,9 +2,7 @@ import { Keyframes } from '../animation/keyframe-animation';
 import { ViewBase } from '../core/view-base';
 import { View } from '../core/view';
 import { unsetValue, _evaluateCssVariableExpression, _evaluateCssCalcExpression, isCssVariable, isCssVariableExpression, isCssCalcExpression } from '../core/properties';
-import { SyntaxTree, Keyframes as KeyframesDefinition, parse as parseCss, Node as CssNode } from '../../css';
-import { CSS3Parser, CSSNativeScript } from '../../css/parser';
-import { cssTreeParse } from '../../css/css-tree-parser';
+import { SyntaxTree, Keyframes as KeyframesDefinition, Node as CssNode } from '../../css';
 
 import { RuleSet, SelectorsMap, SelectorCore, SelectorsMatch, ChangeMap, fromAstNodes, Node } from './css-selector';
 import { Trace } from '../../trace';
@@ -214,10 +212,12 @@ class CSSSource {
 		if (this._source) {
 			switch (parser) {
 				case 'css-tree':
+					const cssTreeParse = require('../../css/css-tree-parser').pacssTreeParserse;
 					this._ast = cssTreeParse(this._source, this._file);
-
 					return;
 				case 'nativescript':
+					const CSS3Parser = require('../../css/parser').CSS3Parser;
+					const CSSNativeScript = require('../../css/parser').CSSNativeScript;
 					const cssparser = new CSS3Parser(this._source);
 					const stylesheet = cssparser.parseAStylesheet();
 					const cssNS = new CSSNativeScript();
@@ -225,6 +225,7 @@ class CSSSource {
 
 					return;
 				case 'rework':
+					const parseCss = require('../../css').parse;
 					this._ast = parseCss(this._source, { source: this._file });
 
 					return;
--- a/packages/core/ui/styling/style-scope.ts
+++ b/packages/core/ui/styling/style-scope.ts
@@ -216,8 +216,7 @@ class CSSSource {
 					this._ast = cssTreeParse(this._source, this._file);
 					return;
 				case 'nativescript':
-					const CSS3Parser = require('../../css/parser').CSS3Parser;
-					const CSSNativeScript = require('../../css/parser').CSSNativeScript;
+					const { CSS3Parser, CSSNativeScript } = require('../../css/parser');
 					const cssparser = new CSS3Parser(this._source);
 					const stylesheet = cssparser.parseAStylesheet();
 					const cssNS = new CSSNativeScript();
--- a/packages/core/ui/styling/style-scope.ts
+++ b/packages/core/ui/styling/style-scope.ts
@@ -212,7 +212,7 @@ class CSSSource {
 		if (this._source) {
 			switch (parser) {
 				case 'css-tree':
-					const cssTreeParse = require('../../css/css-tree-parser').pacssTreeParserse;
+					const cssTreeParse = require('../../css/css-tree-parser').cssTreeParse;
 					this._ast = cssTreeParse(this._source, this._file);
 					return;
 				case 'nativescript':


packages/core/ui/styling/style-scope.ts
```diff
@@ -214,17 +212,20 @@ class CSSSource {
 		if (this._source) {
 			switch (parser) {
 				case 'css-tree':
+					const cssTreeParse = require('../../css/css-tree-parser').cssTreeParse;
```
NathanWalker, line 215: I'd wonder too if this would be subject to `sideEffects` when tree shaking is utilized? - perhaps the global.cssParser option in webpack configs may be a nice way to remove these blocks for those not used.

farfromrefug: It can't webpack has no way of knowing we use one or the other. So it will add the 3. That s why I talked about global var for cssParder so that webpack can remove what we don't use

rigor789: Instead of adding yet another global, we can add "magic" strings

In this case, I would do
```js
switch(__CSS_PARSER__) {
```
and just make ts happy with
```ts
declare readonly var __CSS_PARSER__: string
```
in `global-types.d.ts`

We would inject the variable with the `DefinePlugin` in webpack.

farfromrefug: I dont think it would work Igor. Dont know if the tree shaker will clean the switch . I think we have to something like `if (__CSS_PARSER__ === 'rework')...`

rigor789: Yep, we can convert to if/else if's

[This rollup repl shows it working!](https://rollupjs.org/repl/?version=2.33.3&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCU3QiUyMGN1YmUlMkMlMjBzcXVhcmUlMjAlN0QlMjBmcm9tJTIwJy4lMkZtYXRocy5qcyclM0IlNUNuJTVDbmNvbnN0JTIwX19DU1NfUEFSU0VSX18lMjAlM0QlMjAnY3NzLXRyZWUnJTVDbiU1Q25hc3luYyUyMGZ1bmN0aW9uJTIwcnVuKCklMjAlN0IlNUNuJTVDdCUyRiUyRiUyMG5vdCUyMHRyZWUtc2hha2VuISU1Q24lNUN0c3dpdGNoKF9fQ1NTX1BBUlNFUl9fKSUyMCU3QiU1Q24lNUN0JTVDdGNhc2UlMjAnY3NzLXRyZWUnJTNBJTVDbiU1Q3QlNUN0JTVDdGNvbnNvbGUubG9nKCUyMGN1YmUoJTIwNSUyMCklMjApJTNCJTIwJTJGJTJGJTIwdmlzaWJsZSU1Q24lNUN0JTVDdCU1Q3RicmVhayUzQiU1Q24lNUN0JTVDdGNhc2UlMjAnb3RoZXInJTNBJTVDbiU1Q3QlNUN0JTVDdGNvbnNvbGUubG9nKCUyMHNxdWFyZSglMjA1JTIwKSUyMCklM0IlMjAlMkYlMkYlMjB2aXNpYmxlJTVDbiU1Q3QlNUN0JTVDdGJyZWFrJTNCJTVDbiU1Q3QlN0QlNUNuJTdEJTVDbiU1Q25hc3luYyUyMGZ1bmN0aW9uJTIwcnVuMigpJTIwJTdCJTVDbiU1Q3RpZihfX0NTU19QQVJTRVJfXyUyMCUzRCUzRCUzRCUyMCdjc3MtdHJlZScpJTIwJTdCJTVDbiU1Q3QlNUN0JTVDdGNvbnNvbGUubG9nKCUyMGN1YmUoJTIwNSUyMCklMjApJTNCJTIwJTJGJTJGJTIwdmlzaWJsZSU1Q24lNUN0JTdEJTIwZWxzZSUyMGlmKF9fQ1NTX1BBUlNFUl9fJTIwJTNEJTNEJTNEJTIwJ290aGVyJyklMjAlN0IlNUNuJTVDdCU1Q3QlNUN0Y29uc29sZS5sb2coJTIwc3F1YXJlKCUyMDUlMjApJTIwKSUzQiUyMCUyRiUyRiUyMHRyZWUtc2hha2VuJTVDbiU1Q3QlN0QlNUNuJTdEJTVDbiU1Q24lNUNucnVuKCklM0IlNUNucnVuMigpJTNCJTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlN0QlMkMlN0IlMjJuYW1lJTIyJTNBJTIybWF0aHMuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyJTJGJTJGJTIwbWF0aHMuanMlNUNuJTVDbiUyRiUyRiUyMFRoaXMlMjBmdW5jdGlvbiUyMGlzbid0JTIwdXNlZCUyMGFueXdoZXJlJTJDJTIwc28lNUNuJTJGJTJGJTIwUm9sbHVwJTIwZXhjbHVkZXMlMjBpdCUyMGZyb20lMjB0aGUlMjBidW5kbGUuLi4lNUNuZXhwb3J0JTIwZnVuY3Rpb24lMjBzcXVhcmUlMjAoJTIweCUyMCklMjAlN0IlNUNuJTVDdHJldHVybiUyMHglMjAqJTIweCUzQiU1Q24lN0QlNUNuJTVDbiUyRiUyRiUyMFRoaXMlMjBmdW5jdGlvbiUyMGdldHMlMjBpbmNsdWRlZCU1Q25leHBvcnQlMjBmdW5jdGlvbiUyMGN1YmUlMjAoJTIweCUyMCklMjAlN0IlNUNuJTVDdCUyRiUyRiUyMHJld3JpdGUlMjB0aGlzJTIwYXMlMjAlNjBzcXVhcmUoJTIweCUyMCklMjAqJTIweCU2MCU1Q24lNUN0JTJGJTJGJTIwYW5kJTIwc2VlJTIwd2hhdCUyMGhhcHBlbnMhJTVDbiU1Q3RyZXR1cm4lMjB4JTIwKiUyMHglMjAqJTIweCUzQiU1Q24lN0QlMjIlMkMlMjJpc0VudHJ5JTIyJTNBZmFsc2UlN0QlNUQlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTIyZm9ybWF0JTIyJTNBJTIyZXMlMjIlMkMlMjJuYW1lJTIyJTNBJTIybXlCdW5kbGUlMjIlMkMlMjJhbWQlMjIlM0ElN0IlMjJpZCUyMiUzQSUyMiUyMiU3RCUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJleGFtcGxlJTIyJTNBbnVsbCU3RA==)

I expect it to work the same through webpack as well!

packages/core/ui/styling/style-scope.ts
```diff
@@ -214,17 +212,20 @@ class CSSSource {
...
 					this._ast = cssTreeParse(this._source, this._file);
-
 					return;
 				case 'nativescript':
+					const CSS3Parser = require('../../css/parser').CSS3Parser;
+					const CSSNativeScript = require('../../css/parser').CSSNativeScript;
```
rigor789, line 220: I think this should also work with tree-shaking?
```suggestion
					const { CSS3Parser, CSSNativeScript } = require('../../css/parser');
```

packages/core/ui/styling/style-scope.ts
shirakaba: Various React ecosystem packages use `if(__DEV__){}` blocks explicitly because Webpack does treeshake paths based on the value of the constant defined by WebpackDefinePlugin, if it’s any help.

NathanWalker:
@farfromrefug I believe this is obsolete since we merged this https://github.com/NativeScript/NativeScript/commit/dce7408c05e35cbcfc1c066e9c9aca9f02da17ed in with the release branch, is that correct?
If that commit handles what this PR would, feel free to close this PR as changes from a different pr were already brought in which we believe covers this.
/cc @rigor789

farfromrefug:
Yes it is !

PR #9018, opened 2020-11-09
fix: ios10 tab crash

NathanaelA:
## PR Checklist

iOS 10.3 (and possibly other version) the hack for Animated sliding is crashing out on v8 engine.   Swapped hack for a different hack that passed all the same tests.

## What is the new behavior?
No Crashes.  :-)

### DIFF
--- a/packages/core/ui/tabs/index.ios.ts
+++ b/packages/core/ui/tabs/index.ios.ts
@@ -34,6 +34,10 @@ const invokeOnRunLoop = (function () {
 	};
 })();
 
+function NSRunOnLoop(fun) {
+	NSRunLoop.mainRunLoop.performBlock(fun);
+}
+
 @NativeClass
 class MDCTabBarDelegateImpl extends NSObject implements MDCTabBarDelegate {
 	public static ObjCProtocols = [MDCTabBarDelegate];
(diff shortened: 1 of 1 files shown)


packages/core/ui/tabs/index.ios.ts
```diff
@@ -34,6 +34,10 @@ const invokeOnRunLoop = (function () {
 	};
 })();
 
+function NSRunOnLoop(fun) {
+	NSRunLoop.mainRunLoop.performBlock(fun);
+}
+
```
rigor789, line 40: Is this intentional? I don't see it being used, unless it's used to override a global?

packages/core/ui/tabs/index.ios.ts
```diff
@@ -1121,14 +1125,30 @@ export class Tabs extends TabsBase {
...
+							// HACK: UIPageViewController fix; see https://stackoverflow.com/questions/15325891
+							invokeOnRunLoop(() => {
+								this.viewController.dataSource = null;
+								(<any>this.viewController).dataSource = this.viewController;
+
+								this.visitFrames(item, (frame) => (frame._animationInProgress = false));
+
+								this._canSelectItem = true;
+								this._setCanBeLoaded(value);
+								this._loadUnloadTabItems(value);
```
rigor789, line 1143: would it make sense to extract this to a function since it's the same as 1146-1150? A local arrow fn maybe?

packages/core/ui/tabs/index.ios.ts
NathanaelA: Nope, that was supposed to have been deleted.   Was another test...

packages/core/ui/tabs/index.ios.ts
NathanaelA: Not sure it is worth the overhead for two places...   But I have no issue creating a const func that is ran under either path.

NathanWalker (approved):

rigor789 (approved):

PR #9019, opened 2020-11-09
fix: color date/time ios14 tests

NathanaelA:

iOS 14 removed support for text color on Date/Time except for the Wheel.   Tests were trying to read the color attribute directly on iOS 14 to compare it to what was set; which of course then failed as the Date/Time items are not using the wheel.

## What is the new behavior?
Fixes both tests to detect if on iOS 14, and then ignore that test.

### DIFF
--- a/apps/automated/src/ui/date-picker/date-picker-tests.ts
+++ b/apps/automated/src/ui/date-picker/date-picker-tests.ts
@@ -1,15 +1,14 @@
 import * as TKUnit from '../../tk-unit';
 import * as testModule from '../../ui-test';
 import * as datePickerTestsNative from './date-picker-tests-native';
-import * as color from '@nativescript/core/color';
 import * as platform from '@nativescript/core/platform';
+import { Color } from '@nativescript/core';
 
 // >> date-picker-require
 import * as datePickerModule from '@nativescript/core/ui/date-picker';
 // << date-picker-require
 
 import * as helper from '../../ui-helper';
-import {Device} from "../../../../../packages/core";
 
 export function test_recycling() {
 	helper.nativeView_recycling_test(() => new datePickerModule.DatePicker());
@@ -80,7 +79,7 @@ export class DatePickerTest extends testModule.UITest<datePickerModule.DatePicke
 		if (platform.Device.os === platform.platformNames.ios) {
 			const SUPPORT_TEXT_COLOR = parseFloat(platform.Device.osVersion) < 14.0;
 			if (SUPPORT_TEXT_COLOR) {
-				this.testView.color = new color.Color('red');
+				this.testView.color = new Color('red');
 				TKUnit.assertEqual(this.testView.color.ios.CGColor, this.testView.ios.valueForKey('textColor').CGColor, 'datePicker.color');
 			}
 		}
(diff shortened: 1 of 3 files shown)


apps/automated/src/ui/date-picker/date-picker-tests.ts
```diff
@@ -9,6 +9,7 @@ import * as datePickerModule from '@nativescript/core/ui/date-picker';
 // << date-picker-require
 
 import * as helper from '../../ui-helper';
+import {Device} from "../../../../../packages/core";
```
NathanWalker, line 12: Change this to: `import { Device } from '@nativescript/core';`

apps/automated/src/ui/date-picker/date-picker-tests.ts
```diff
@@ -77,8 +78,11 @@ export class DatePickerTest extends testModule.UITest<datePickerModule.DatePicke
 
 	public test_set_color() {
 		if (platform.Device.os === platform.platformNames.ios) {
-			this.testView.color = new color.Color('red');
-			TKUnit.assertEqual(this.testView.color.ios.CGColor, this.testView.ios.valueForKey('textColor').CGColor, 'datePicker.color');
+			const SUPPORT_TEXT_COLOR = parseFloat(platform.Device.osVersion) < 14.0;
```
NathanWalker, line 81: Just use `Device` import here while in here and since imported above. `platform.Device` > `Device`

apps/automated/src/ui/date-picker/date-picker-tests.ts
```diff
@@ -77,8 +78,11 @@ export class DatePickerTest extends testModule.UITest<datePickerModule.DatePicke
...
+			if (SUPPORT_TEXT_COLOR) {
+				this.testView.color = new color.Color('red');
```
NathanWalker, line 83: While in here perhaps update `new color.Color` > `new Color` and just update to import from the barrel as well since in same file here. Cleanup as we go.

NathanWalker (changes requested):

apps/automated/src/ui/date-picker/date-picker-tests.ts
NathanaelA: LOL, this never should have been added; my editor must have been trying to be smart...

apps/automated/src/ui/date-picker/date-picker-tests.ts
NathanaelA: Actually `Device` should never have been added, `platform.Device` should be correct as `platform` is already imported because of other checks needed in this file.

apps/automated/src/ui/date-picker/date-picker-tests.ts
NathanaelA: This was removed in latest commit.

apps/automated/src/ui/date-picker/date-picker-tests.ts
NathanaelA: Updated Color to use Barrel in both date & time picker tests.

NathanWalker (approved):

PR #9025, opened 2020-11-12
add frame navigation events

farfromrefug:
very practical when you want to know the state of the navigation at  higher level.
Like a drawer menu


rigor789 (approved):
LGTM.

I've had this (similar) in NSVue for a while, but nice to have it in core:
https://github.com/nativescript-vue/nativescript-vue/blob/f1110f14f3dca664c6789e4d2cbb61dbaba3ea5d/platform/nativescript/runtime/components/frame.js#L128
https://github.com/nativescript-vue/nativescript-vue/blob/f1110f14f3dca664c6789e4d2cbb61dbaba3ea5d/platform/nativescript/runtime/components/frame.js#L135

PR #9026, opened 2020-11-12
fix(core-android): ActivityIndicator changed transferred its color to other ActivityIndicators of the same page

### LINKED ISSUE #9021: bug(core-android): Setting ActivityIndicator color changes the color of other ActivityIndicators
**Environment**
 - CLI: 7.0.10
 - Cross-platform modules:
 - Android Runtime: 7.0.1
 - iOS Runtime:
 - Plugin(s):

**Describe the bug**
If one happens to nest multiple ActivityIndicators inside the same page and happen to change the color of the last one, this change will be applied to all ActivityIndicators.

**To Reproduce**
Add multiple ActivityIndicators inside a view.

**Expected behavior**
Each indicator should use its own color.

DimitrisRK:
## PR Checklist

Changing color to an ActivityIndicator applies the new color to all other ActivityIndicators in the same page.
That only applies to Android.

## What is the new behavior?
Each indicator has independent color by calling `mutate()` on indicator drawable instance at the creation of view.

Fixes/Closes #9021 .

### DIFF
--- a/apps/ui/src/progress-bar/activity-indicator-page.xml
+++ b/apps/ui/src/progress-bar/activity-indicator-page.xml
@@ -1,4 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <Page>
-  <ActivityIndicator busy="true" verticalAlignment="center" horizontalAlignment="center"/>
+  <StackLayout>
+    <ActivityIndicator busy="true" color="red" verticalAlignment="center" horizontalAlignment="center"/>
+    <ActivityIndicator busy="true" color="green" verticalAlignment="center" horizontalAlignment="center"/>
+    <ActivityIndicator busy="true" color="blue" verticalAlignment="center" horizontalAlignment="center"/>
+  </StackLayout>
 </Page>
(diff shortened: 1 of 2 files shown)


farfromrefug:
Wouldn t it be better to mutate only when setting the color? If all your activity indicators use the theme color no need to mutate anythin

DimitrisRK:
> Wouldn t it be better to mutate only when setting the color? If all your activity indicators use the theme color no need to mutate anythin

That is because an indicator's color can be changed multiple times depending on one's needs, resulting in pointless `mutate()` calls even though they have no appliance after first call. Initializing it once seemed good to me.

farfromrefug:
Then mark the indicator as mutated. Oersonnaly as I use theme color for indicators I see that PR as a perf lost. I understand the need but we should do it right and mutate only when needed

DimitrisRK:
> Then mark the indicator as mutated. Oersonnaly as I use theme color for indicators I see that PR as a perf lost. I understand the need but we should do it right and mutate only when needed

You are right! To be honest, I didn't consider theme-based indicators. I just committed few more changes that make things better.

rigor789 (approved):

PR #9017, opened 2020-11-08
lint: fix most eslint issues

farfromrefug:
This PR fixes most eslint issues. For some it adds a TODO because the why is not clear.
This is really important in the sense that it makes working with eslint/vscode much smoother.
Before this i was flooded with warning and errors (more that 3000).

PR #9028, opened 2020-11-12
fix: rollback regexp eslint changes

farfromrefug:
this fixes the regression discovered here https://github.com/NativeScript/NativeScript/pull/9017#issuecomment-726242266

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,6 +25,7 @@
 		"no-prototype-builtins": "off",
 		"no-inner-declarations": "off",
 		"no-constant-condition": "off",
+		"no-useless-escape": "off",
 		"@nrwl/nx/enforce-module-boundaries": [
 			"error",
 			{
--- a/packages/core/css/parser.ts
+++ b/packages/core/css/parser.ts
@@ -237,7 +237,7 @@ export function parseRepeat(value: string, start = 0, keyword = parseKeyword(val
 	return null;
 }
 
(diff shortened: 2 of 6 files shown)


DimitrisRK:
@farfromrefug There had been test failures in automate app, so I took a deep look.
It seems that few promise changes have also broken few things.
What I discovered so far is these 2 lines:
https://github.com/NativeScript/NativeScript/blob/master/packages/core/ui/animation/animation-common.ts#L103
https://github.com/NativeScript/NativeScript/blob/master/packages/core/ui/animation/animation-common.ts#L110

Reverting those got rid of about 20 test failures.

farfromrefug:
damn let me fix this

farfromrefug:
automated app tests are running (still issues but not related to this)

NathanaelA (approved):

NathanaelA:
Thank you @farfromrefug - I can confirm that the automated tests are fully working again on iOS and has one broken on Android that appears to be unrelated to this set of PR's.

farfromrefug:
Thanks @NathanaelA sorry I broke it in the first place. Would nt be a good idea to have automated apps ran before merging any PR ? I know it is heavy but it really seems necessary

NathanaelA:
Totally agree, I thought Automated was already was being ran (in fact I went and looked to see how the prior PR passed the tests because I was surprised to see these two follow up PRs in my inbox, and can see it isn't running.  Ugh, something else to get added/fixed, LOL -- it never ends  ...)

PR #9032, opened 2020-11-13
fix: make object optional in notify

farfromrefug:
if not set it will be set to this.

This always bugged me to have to set this all the time. It seems better that way to me


NathanWalker (approved):
Totally agree thank you!

Issue #8712, opened 2020-07-16
TabStrip Switch Item Issue


phoenix503:
**Environment**
I am using it in Android v10.0 

**Describe the bug**
Try switching between the tabs by sliding on the screen not on the tabs and see the color it is not as specified which is given. The colors are not changing for selectedtab.

**To Reproduce**
https://play.nativescript.org/?template=play-ng&id=TStMfT

**Expected behavior**
The unselected Tab color should change accordingly 

**Sample project**
https://play.nativescript.org/?template=play-ng&id=TStMfT

felixkrautschuk:
@phoenix503 seems to be a duplictate of the issue I created some months ago:

https://github.com/NativeScript/NativeScript/issues/8552

phoenix503:
@felixkrautschuk  it is still Open , no solution is given or solved by NS

felixkrautschuk:
@phoenix503 that's right, but it is not useful to open a new discussion about an already existing issue.

phoenix503:
okay @felixkrautschuk

CL-Guzman-Alejandro:
This bug still present with no way to solve this.

DimitrisRK:
@CL-Guzman-Alejandro Seems it's a rule priority issue as it works pretty good if you do it with CSS rules that override theme rules (using a stronger rule).
Example:
```
.my-parent-layout TabStrip {
    selected-item-color: red;
}
```

Issue #8552, opened 2020-04-24
Android Tabs-component styling issues when swiping tabs fast


felixkrautschuk:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 6.5.0
 - Cross-platform modules: 6.51
 - Android Runtime: 6.5.0
 - iOS Runtime: 6.5.0
 - Plugin(s): -

**Describe the bug**
When swiping through the tabs fast, the _selected-item-color_ is not correctly updated. Sometimes multiple tabs are marked as "currently selected".

![android_tabs_crash_backnav](https://user-images.githubusercontent.com/6443021/80239172-05c4fe00-8660-11ea-9dc7-ac2ba66f01cd.gif)

**To Reproduce**
1.) Open the sample app
2.) navigate to the ListView page (or the RadListView page)
3.) swipe very fast between the tabs

**Expected behavior**
Only the currently selected item should be marked as "currently selected", even when swiping very fast or cancel swiping process.

**Sample project**
[ns-tabs-issues.zip](https://github.com/NativeScript/NativeScript/files/4530416/ns-tabs-issues.zip)

schroedan:
This also affects iOS not only Android.

rigor789:
Closing because BottomNavigation and Tabs are no longer part of core since NativeScript 8.0:

https://blog.nativescript.org/nativescript-8-announcement/#important-note-about-bottomnavigation-and-tabs

Issue #8551, opened 2020-04-23
Tabs highlight/selected color not applied when using gestures


cdmistman:
**Environment**
 - CLI: 6.5.0
 - Cross-platform modules: 6.5.1
 - Android Runtime: 6.5.0
 - iOS Runtime: N/A (not using right now)
 - Plugin(s): `nativescript-angular` ("^8.21.0"), `@angular/core` ("^8.2.14")

**Describe the bug**
When using custom highlight colors for `TabStrip`, the specified color is only used for tapping to select the tab. When using the swipe gesture to navigate between tabs, the default blue color is used for the tab strip item highlight color.

**To Reproduce**
1. Have a `Tabs` component with at least 2 tabs
2. Assign a CSS class to the `TabStrip`
3. In the CSS class, assign the `unselected-item-color`, `highlight-color`, and `selected-item-color` values to easy-to-differentiate colors
4. In the app, try switching between the tabs, both by tapping on the tab to navigate to, as well as swiping.

**Expected behavior**
There should be no coloring difference between tapping and selecting tabs for navigation

**Sample project**
[Playground](https://play.nativescript.org/?template=play-ng&id=mpAVJ6)

**Additional context**
So far, every `@nativescript/core` version I've tried is affected by this bug, including `6.6.0-next-2020-04-20-155205-01`
I also haven't been able to test on iOS, so I can only confirm this bug affects Android

NickIliev:
@cdmistman you have a syntax error in the CSS -  the JavaScript property is `unSelectedItemColor` which means that it translates to the following CSS property`un-selected-item-color` (note the first dash)


[Here is the modified demo](https://play.nativescript.org/?template=play-ng&id=mpAVJ6&v=3) (I've used some more "lively" colors) that demonstrates that everything is working as expected with taps and with gestures. I've tested on Android 10 and on iOS 13 and both are working as expected - do let me know if you have an issue with a specific API level.

cdmistman:
Interesting... I believe your code actually raises a different issue.

[...]

PR #9040, opened 2020-11-15
fix(core-android): Tabs selected item color gets applied to wrong Tabs or not applied at all

### LINKED ISSUE #8543: Tab label is not changing its color when swipe isdisabled
When swiping to another tab label is highlighted. When swipe is disabled label is NOT highlighted

https://play.nativescript.org/?template=play-ng&id=WxSW9I

this is very annoying issue because i cannot use tabs with RadViewList swipe actions :(
Can you fix it or provide workaround. I have tried to add selectedIndexChange event and assign it to property but this does not help either.

DimitrisRK:
## PR Checklist

There are couple of issues regarding Tabs navigation UX for some time now.
1) TabStrip selected color is not applied to the right tabstrip item if one changes tabs using swipe.
2) TabStrip selected color is not applied to any tabstrip items after tab change, if swipe is disabled.
3) Property `isAnimationEnabled` is not working for Tabs with swipeEnabled set to false.

## What is the new behavior?
This PR ensures colors are properly changed in all cases and animation to be configurable in all cases.
In terms of performance, this method won't try to update colors when not needed.
Additionally, it prevents the pointless calls of `setCurrentItem` every time `selectedIndexProperty` changes.

Until now, the following performance issue had been occuring:
`setCurrentItem` calls a native method which by its turn calls `setPrimaryItem` method.
`setPrimaryItem` is overridden inside Tabs UI component and changes `selectedIndex` property.
Upon change, `selectedIndex` property will call `setCurrentItem` again (called using `setNative` symbol)

Luckily, since `setCurrentItem` tried to set the same index there was no recursion but still executed tasks that shouldn't.

I'm positive this is a very good solution to get rid of some `TabStrip` UX bugs.
People are desperate with Tab components and that may give a small relief for Android platform at least.

There are lots of Tabs examples in UI app where this can be tested, but I also made sure to add selected/unselected colors to swipe-disabled-page and tabs-position-page examples.

Hope you consider this as a good alternative.

Fixes/Closes #8543 #8551 #8552 #8712


packages/core/ui/tabs/index.android.ts
```diff
@@ -952,11 +952,9 @@ export class Tabs extends TabsBase {
 	}
 
 	[selectedIndexProperty.setNative](value: number) {
-		// TODO
-		// if (Trace.isEnabled()) {
-		//     Trace.write("TabView this._viewPager.setCurrentItem(" + value + ", " + smoothScroll + ");", traceCategory);
-		// }
-		this._viewPager.setCurrentItem(value, this.animationEnabled);
+		if (this._viewPager.getCurrentItem() != value) {
+			this._viewPager.setCurrentItem(value);
```
rigor789, line 956: Shouldn't this pass the `animationEnabled` boolean?

rigor789 (commented):
Not sure if intentional but `packages/core/platforms/android/widgets-release.aar` has been removed, I believe it just needs to be rebuilt with the native changes.

packages/core/ui/tabs/index.android.ts
DimitrisRK: @rigor789 This is handled in ui-mobile-base android class:
https://github.com/DimitrisRK/NativeScript/blob/tabs-selected/packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/TabViewPager.java#L58

Note: This java code is not an addition of mine. `animationEnabled` check had always been inside overridden method but it had an extra `swipeEnabled` check which was breaking proper functionality for all cases.

DimitrisRK:
> Not sure if intentional but `packages/core/platforms/android/widgets-release.aar` has been removed, I believe it just needs to be rebuilt with the native changes.

Sorry, just noticed first comment. It's not removed, just looks like this because it's modified (few lines were left unused, so they got removed).
I modified and rebuilt it as expected.

rigor789 (approved):

PR #9056, opened 2020-11-19
feat: allow app to start without a root view

edusperoni:
## PR Checklist

When you bootstrap the app, the root view MUST be created synchronously.

## What is the new behavior?
When bootstrapping the app, you can set the root view to `null` which will ensure the app will start without a root view (blank screen). It's then up to the developer (usually flavor) to set this root view.

In the Angular context, this is now possible:

1. on launch, begin AppModule bootstrap
2. set the rootview to null
3. schedule a macroTask to check if bootstrap is done
3.1. if it is, set the rootView (no flashes)
3.2. if it isn't, set the rootView as a temporary view (user provided splash screen)
4. on bootstrap complete, set root view from AppModule's rootView

### DIFF
--- a/packages/core/ui/frame/index.android.ts
+++ b/packages/core/ui/frame/index.android.ts
@@ -1301,6 +1301,7 @@ class ActivityCallbacksImplementation implements AndroidActivityCallbacks {
 		if (!rootView) {
 			const mainEntry = application.getMainEntry();
 			const intent = activity.getIntent();
+			// useful for integrations that would like to set rootView asynchronously after app launch
 			let shouldRootViewBeEmpty = false;
 
 			if (fireLaunchEvent) {
--- a/packages/core/ui/frame/index.android.ts
+++ b/packages/core/ui/frame/index.android.ts
@@ -1301,16 +1301,16 @@ class ActivityCallbacksImplementation implements AndroidActivityCallbacks {
 		if (!rootView) {
 			const mainEntry = application.getMainEntry();
 			const intent = activity.getIntent();
-			let ensureNoRootView = false;
+			let shouldRootViewBeEmpty = false;
 
 			if (fireLaunchEvent) {
 				// entry point for Angular and Vue frameworks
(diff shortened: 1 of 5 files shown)


packages/core/ui/frame/index.android.ts
```diff
@@ -1301,13 +1301,18 @@ class ActivityCallbacksImplementation implements AndroidActivityCallbacks {
...
+			let ensureNoRootView = false;
 
 			if (fireLaunchEvent) {
 				// entry point for Angular and Vue frameworks
 				rootView = notifyLaunch(intent, <any>savedInstanceState, null);
+				ensureNoRootView = rootView === null;
 			}
 
 			if (!rootView) {
+				if(ensureNoRootView) {
```
rigor789, line 1313:
```suggestion
				if(rootView === null) {
				  // root view is explicitly set to null - bail
```
The `ensureNoRootView` variable seems redundant (and confusing a bit)

NathanWalker (approved):

packages/core/ui/frame/index.android.ts
edusperoni: It's mainly for consistency. In iOS if you try to `resetRootView(null)` it'll throw, so this is to ensure that rootView can only be null during the launch event, and that specific function is called both on launch (`fireLaunchEvent = true`) and on `resetRootView`

rigor789: Oh ok - thanks for clarifying! I still think the name could be changed to something more "self-explanatory"

rigor789 (approved):
LGTM

NathanWalker (approved):

packages/core/ui/frame/index.android.ts
NathanWalker: Went with this more clear name and added comment:
```
// useful for integrations that would like to set rootView asynchronously after app launch
let shouldRootViewBeEmpty = false;
```

PR #8495, opened 2020-04-02
feat(utils): work openFile in simulator

tarunama:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/DevelopmentWorkflow.md#running-unit-tests.
- [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/WritingUnitTests.md.

## What is the current behavior?
can not open file in iOS simulator. 

## What is the new behavior?
can open file in iOS simulator

Implements #8410

PR #9065, opened 2020-11-25
feat(core): Utils.isSimulatorOrEmulator

### LINKED ISSUE #9064: There is no Utility to determine whether current execution is on a simulator or emulator
**Is your feature request related to a problem? Please describe.**
The current function `isRealDevice` does not work properly on Android, and returns data on iOS instead of a boolean.
There are other relying functions that call `isRealDevice`. 

**Describe the solution you'd like**
Instead of introducing a breaking change, create a new `isSimulatorOrEmulator` function that actually returns a boolean.

**Describe alternatives you've considered**
Changing the `isRealDevice` function. This would cause breaking changes.

alexziskind1:
## PR Checklist

- [X] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it. #9064 

No function currently exists to allow a check for a simulator or emulator

## What is the new behavior?
The new Util function `isSimulatorOrEmulator` handles this

Fixes/Implements/Closes #9064 .

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -93,7 +93,7 @@ export type { InstrumentationMode, TimerInfo } from './profiling';
 export { encoding } from './text';
 export * from './trace';
 export * from './ui';
-import { GC, isFontIconURI, isDataURI, isFileOrResourcePath, executeOnMainThread, mainThreadify, isMainThread, dispatchToMainThread, releaseNativeObject, getModuleName, openFile, openUrl, isRealDevice, layout, ad as androidUtils, iOSNativeHelper as iosUtils, Source } from './utils';
+import { GC, isFontIconURI, isDataURI, isFileOrResourcePath, executeOnMainThread, mainThreadify, isMainThread, dispatchToMainThread, releaseNativeObject, getModuleName, openFile, openUrl, isRealDevice, isSimulatorOrEmulator, layout, ad as androidUtils, iOSNativeHelper as iosUtils, Source } from './utils';
 import { ClassInfo, getClass, getBaseClasses, getClassInfo, isBoolean, isDefined, isFunction, isNullOrUndefined, isNumber, isObject, isString, isUndefined, toUIString, verifyCallback } from './utils/types';
 export declare const Utils: {
 	GC: typeof GC;
@@ -111,6 +111,7 @@ export declare const Utils: {
 	openFile: typeof openFile;
 	openUrl: typeof openUrl;
 	isRealDevice: typeof isRealDevice;
+	isSimulatorOrEmulator: typeof isSimulatorOrEmulator;
 	layout: typeof layout;
 	android: typeof androidUtils;
 	ad: typeof androidUtils;
--- a/packages/core/index.ts
+++ b/packages/core/index.ts
@@ -110,7 +110,7 @@ export * from './trace';
 
 export * from './ui';
 
-import { GC, isFontIconURI, isDataURI, isFileOrResourcePath, executeOnMainThread, mainThreadify, isMainThread, dispatchToMainThread, queueMacrotask, releaseNativeObject, getModuleName, openFile, openUrl, isRealDevice, layout, ad as androidUtils, iOSNativeHelper as iosUtils, Source, RESOURCE_PREFIX, FILE_PREFIX } from './utils';
+import { GC, isFontIconURI, isDataURI, isFileOrResourcePath, executeOnMainThread, mainThreadify, isMainThread, dispatchToMainThread, queueMacrotask, releaseNativeObject, getModuleName, openFile, openUrl, isRealDevice, isSimulatorOrEmulator, layout, ad as androidUtils, iOSNativeHelper as iosUtils, Source, RESOURCE_PREFIX, FILE_PREFIX } from './utils';
 import { ClassInfo, getClass, getBaseClasses, getClassInfo, isBoolean, isDefined, isFunction, isNullOrUndefined, isNumber, isObject, isString, isUndefined, toUIString, verifyCallback } from './utils/types';
 
 export const Utils = {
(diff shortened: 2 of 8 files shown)


NathanWalker (dismissed):
@triniwiz and I will add some extra security detections in this as well. Thanks @alexziskind1 👏

NathanaelA (commented):
Why add this?  Shouldn't simulator be just `!isRealDevice()` why add extra bloat to the framework for a unneeded helper function?  

Second, even if we really wanted the "helper" function `isSimulatorOrEmulator` shouldn't it just be `return !isRealDevice()` so all actual detection code is contained in one function rather than split into two places that could need changes in the future?

farfromrefug:
Are you sure that test is enough on android ? I have something like more complex https://github.com/nativescript-community/extendedinfo/blob/master/src/extendedinfo.android.ts#L4
Also you should cache the value
It is a simple Boolean  when cached while the test requires a lot more operations 

And to.me that name is far too long :)  we don't really about the fact that iios and android don't use the se name convention

mapo80:
Hi, I think that @farfromrefug is a better solution, in fact I've adopted solution from flutter:

[https://github.com/flutter/plugins/blob/master/packages/device_info/android/src/main/java/io/flutter/plugins/deviceinfo/MethodCallHandlerImpl.java](https://github.com/flutter/plugins/blob/master/packages/device_info/android/src/main/java/io/flutter/plugins/deviceinfo/MethodCallHandlerImpl.java
)

farfromrefug:
@mapo80 your link was wrong here is the good link https://github.com/flutter/plugins/blob/master/packages/device_info/device_info/android/src/main/java/io/flutter/plugins/deviceinfo/MethodCallHandlerImpl.java

mapo80:
> @mapo80 your link was wrong here is the good link https://github.com/flutter/plugins/blob/master/packages/device_info/device_info/android/src/main/java/io/flutter/plugins/deviceinfo/MethodCallHandlerImpl.java

Fixed, thanks!

NathanaelA:
Instead we should add the additional checks that are used by other frameworks/plugins to beef up the detection of a simulator vs real device so the original `export function isRealDevice(): boolean;` function is correct on both platforms again.

alexziskind1:
> Instead we should add the additional checks that are used by other frameworks/plugins to beef up the detection of a simulator vs real device so the original `export function isRealDevice(): boolean;` function is correct on both platforms again.

@NathanaelA The reason behind the new function implementation is that it does not present a breaking change. If isRealDevice is fixed then it can lead to breaking changes, since other framework functions (and potentially many client apps) use the result of this call incorrectly. In other words, on iOS when the device is real, it returns paths, which are being used elsewhere - removing this would result in a breaking change.

alexziskind1:
> Are you sure that test is enough on android ? I have something like more complex https://github.com/nativescript-community/extendedinfo/blob/master/src/extendedinfo.android.ts#L4
> Also you should cache the value
> It is a simple Boolean when cached while the test requires a lot more operations
> 
> And to.me that name is far too long :) we don't really care about the fact that iios and android don't use the se name convention

@farfromrefug perhaps there may be a style guide that states how long function names can be and how specific they need to be, but for now, I don't think there is one. Renaming this is fine as long as it is clear what the function does. Calling the function `isEmulator` hints at being specific to Android. Calling it `isSimulator` hints at being specific to iOS. This function name is clear and it relates that it is a cross platform function.

NathanaelA:
Fixing `isRealDevice` is not a breaking change as it is only used in the Framework as a boolean value here:

image
and here:
image

According to the Android side, the iOS utils/index and the `.d.ts` file it is supposed to be boolean in the first place, so this should be fixed since the return isn't boolean which is what is causing you your first set of issues.    As you can see in this image; everything EXCEPT the one nativeHelpers ios file side is marked boolean.  We should NEVER have Android & iOS versions of the function return different types, that is a bug for sure.   

image

---

Second, If `isRealDevice` is not reporting correctly,  beefing up the support on it would be helpful to all the plugins as if a plugin is trusting our code to be correct and it isn't, then it should again be fixed. 

---

Adding extra unneeded code that has to be maintained (so if we accepted this, when we go to beef up simulator detection we have to change TWO sets of code, that is bad imho) so that is something we should try to avoid.    Using `!isRealDevice()` tells you the exact same thing.    And if others maintainers agree with you and they really do want a `isSimulatorOrEmulator()` function, then it really should be a simple `return !isRealDevice();` so that we only have one set of code to maintain.

alexziskind1:
@NathanaelA totally agree with you that `isRealDevice` should be fixed, but disagree with you that it is not a breaking change. This is a public API and plugins as well as other client apps could be using `isRealDevice` as it is. If the fix is implemented, then it will break those. This PR is added functionality that will allow `isRealDevice` usage to continue as is, without breaking anything, but also provide a good step forward for fixing it in the future, as well as providing the needed simulator check right now.

NathanaelA:
Except if anything external is using isRealDevice() they are already expecting it to be boolean.  The actual exposed `.d.ts` says it is BOOL.   So no, not a breaking change as TS will have told them it is a boolean, and their code already expects iOS and Android to be returning a Boolean....  :grinning:

alexziskind1:
@NathanaelA fair point. However, `isRealDevice` also returns the opposite of the truth on Android. So therefore it return true when run on an Emulator. This would be a breaking change.

NathanaelA:
Well then it is a bug that should be fixed, as we shouldn't be getting the **opposite** values from different platforms.    Just like any other bug should be fixed.   

1. This feature was added in April and released at the end of very end of May, and apparently not tested very well.  https://github.com/NativeScript/NativeScript/pull/8495
2.  In addition, I'm going to guess nobody knew about this function as out of 1,093 plugins I have indexed, not single one use this "feature".   Nor was it in the release notes.   :grinning:

farfromrefug:
I for one did not even know it was there! I doubt many apps/ plugins use it today

alexziskind1:
testing

PR #9067, opened 2020-11-25
feat: rewrite the ObservableArray

farfromrefug:
This is now an ArrayLike object. Meaning we can do `test[4]= 4;` and the array will be changed as well as a change event trigger.
It should work exactly as the current implementation.
One nice feature i would have liked which does not work yet is ` test instanceof Array`.
If that was to work it means an `ObservableArray` could be used anywhere and even passed to native methods (i think)

### DIFF
--- a/packages/core/data/observable-array/index.ts
+++ b/packages/core/data/observable-array/index.ts
@@ -1,6 +1,4 @@
-/* eslint-disable prefer-rest-params */
 import { Observable, EventData } from '../observable';
-import * as types from '../../utils/types';
 
 export class ChangeType {
 	static Add = 'add';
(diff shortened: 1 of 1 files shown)


farfromrefug:
I found some issues with the WeakMap and that new implementation. That PR is not ready yet

NathanWalker:
@farfromrefug If this were ready soon we could likely include in 8.6, would just need to check any breaking changes to mention if so. Otherwise can target 9.0.

farfromrefug:
@NathanWalker that one is really old. I will check if I use it in my fork. Doesn't Dimitris created one doing a better job?

PR #9083, opened 2020-12-02
faster  style-scope setPropertyValues

farfromrefug:
That Pr makes the style-scope `setPropertyValues` much faster. It tries not to do so many loops.
The only real change is the fact that i dont freeze the `_appliedPropertyValues`  object anymore. I need that to go faster and as the object is not used outside of that function it does not seem to make a difference.

Now do we have tests which would ensure that is not breaking anything?


NathanaelA (changes requested):
I would leave this alone as the idea here is to let developers know that even though this is available on the Prototype scope it is read only!

packages/core/ui/styling/style-scope.ts
```diff
@@ -552,60 +552,47 @@ export class CssState {
 		const newPropertyValues = new view.style.PropertyBag();
 		matchingSelectors.forEach((selector) => selector.ruleset.declarations.forEach((declaration) => (newPropertyValues[declaration.property] = declaration.value)));
 
-		const oldProperties = this._appliedPropertyValues;
-
-		let isCssExpressionInUse = false;
-
+		const oldProperties = this._appliedPropertyValues || {};
```
NathanaelA, line 555: This can be left the way it was, no need to add the `|| {}` as _appliedPropertyValues ALWAYS has a value.  It is defaulted to EmptyPropertyBag  on construction.

packages/core/ui/styling/style-scope.ts
```diff
@@ -398,7 +398,7 @@ export class CssState {
 
 	_onDynamicStateChangeHandler: () => void;
 	_appliedChangeMap: Readonly<ChangeMap<ViewBase>>;
-	_appliedPropertyValues: Readonly<Record<string, unknown>>;
+	_appliedPropertyValues: Record<string, unknown>;
```
NathanaelA, line 401: This should be reverted back, it is letting other developers know that if they access it -- it is read only.

packages/core/ui/styling/style-scope.ts
```diff
@@ -552,60 +552,47 @@ export class CssState {
...
+			}
+			valuesToApply[property] = value;
 		}
 
-		// Property values are fully updated, freeze the object to be used for next update.
-		Object.freeze(newPropertyValues);
-
 		// Unset removed values
 		for (const property in oldProperties) {
-			if (!(property in newPropertyValues)) {
```
NathanaelA, line 592: If I read your code correctly You do realize that without this check, you can end up doing double the work on all css properties.   Example: Say the property is `background` then Line 587 will be triggered, then later 598 causing the new background to be applied twice; if the background property is long running you created extra runtime overhead for the same set value.   The purpose of this check is to eliminate double setting of the value since it will be set below at line 598.

NathanaelA (changes requested):
Have you benchmarked this against the current code, I'd be interested to know what savings (if any) you create with this refactor.

farfromrefug:
@NathanaelA no benchmark no. Just globally faster as you loop through new props only once and loop through less old properties.
Would be nice to have benchmarks around all that properties/CSS thing

packages/core/ui/styling/style-scope.ts
farfromrefug: @NathanaelA then we should make it private instead ? the readonly think makes tsc not happy when i delete properties on `oldProperties`

packages/core/ui/styling/style-scope.ts
farfromrefug: @NathanaelA good catch. the issue is up. I mean in fact any value from `newPropertyValues` can be removed from `oldProperties`, right? So i should simply add the `delete oldProperties[property];` for all values in the loop of  `newPropertyValues`

packages/core/ui/styling/style-scope.ts
NathanaelA: Yeah, you are right, I don't see any reason at this point this needs to be public.     Good idea!   Mark it as private.

packages/core/ui/styling/style-scope.ts
NathanaelA: I believe so...  The tests will tell us for certain as the Styling code is quite complicated in some areas on what it expects/needs.

NathanaelA:
You can add some bench marks into the code; rename the existing function to `_setPropertyValues` and create a new `setPropertyValues` and add `const x = Date.now(); this._setPropertyValues(...); y = Date.now(); console.log(y-x);` it isn't totally precise but it will give a rough idea  of the speed.    

Then run it using the prior code with a heavy css app (or maybe even the automated app) a couple times and copy the numbers.   Then use your new code with the same apps.

This will give you a pretty good idea if this is faster, slower or even majorly faster/slower.

rigor789:
> You can add some bench marks into the code; rename the existing function to _setPropertyValues and create a new setPropertyValues and add const x = Date.now(); this._setPropertyValues(...); y = Date.now(); console.log(y-x); it isn't totally precise but it will give a rough idea of the speed.

@farfromrefug @NathanaelA even simpler: 

```js
console.time('setPropertyValues')
this._setPropertyValues(...);
console.timeEnd('setPropertyValues')
```
Will print the time it has taken :D

NathanWalker:
@farfromrefug I've updated this PR with latest main and looks like just 4 tests are affected by these changes:
```
  === ALL TESTS COMPLETE ===
  1632 OK, 4 failed
  DURATION: 22370.61 ms
  === END OF TESTS ===
  Test: STYLE.test_css_calc_and_variables FAILED: Stack - width === 10% At /value actual: '1' and expected: '0.1' differ.
  Test: STYLE.test_css_variable_fallback FAILED: undefined-css-variable-without-fallback Actual: <#0000FF>(string). Expected: <undefined>(undefined)
  Test: STYLE.test_css_variables FAILED: Cannot read property 'hex' of undefined
  Test: STYLE.test_nested_css_calc_and_variables FAILED: Stack - width === 200% At /value actual: '1' and expected: '2' differ.
```
Would you be able to peek sometime? We can queue this up to include in 8.4. It does appear that these changes will make the tests finish in faster time which leads me to believe there are some perf gains here.

farfromrefug:
@NathanWalker all tests are now passing. Hope i got it all right. The issue was that css vars/exps need to be handled separately in the sense that all css vars must be evaluated before we can evaluate css expressions. I added a comment so we dont forget about it anymore.
It should not make things any slower

PR #9092, opened 2020-12-07
feat: ios added ignoreSafeArea property

farfromrefug:
I added a new property to completely ignore safearea computation. It gets important with components like bottomsheets.
Without it i can only get unwanted view "padding" as seen in the next screenshot.
And there is no real way right now to simply "ignore" that top safe area (because the view controller has a  safe area).
Simulator Screen Shot - iPhone 11 Pro Max - 2020-12-07 at 13 33 04

By setting that new property (which has no consequence when not set) i get correct "p
Simulator Screen Shot - iPhone 11 Pro Max - 2020-12-07 at 13 36 10
adding"

I also think this is the property which should be set on all ListView cells to totally ignore safearea inside listviews

### DIFF
--- a/packages/core/ui/core/view/index.ios.ts
+++ b/packages/core/ui/core/view/index.ios.ts
@@ -262,7 +262,9 @@ export class View extends ViewCommon implements ViewDefinition {
 		if (majorVersion <= 10) {
 			return null;
 		}
-
+		if (this.iosIgnoreSafeArea) {
+            return frame;
+        }
 		if (!this.iosOverflowSafeArea || !this.iosOverflowSafeAreaEnabled) {
 			return IOSHelper.shrinkToSafeArea(this, frame);
 		} else if (this.nativeViewProtected && this.nativeViewProtected.window) {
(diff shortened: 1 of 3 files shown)


farfromrefug:
Forgot to mention that in the same "vibe" that PR contains a fix for applying safe area in view controller where the top of the view controller is not the top of the window (like the bottom sheet).
https://github.com/NativeScript/NativeScript/blob/576a0dcb632b84c94d2280e1d324e10856797ee0/packages/core/ui/core/view/view-helper/index.ios.ts#L355
Again should have no consequence for "normal" view controllers

DimitrisRK:
Since there is a need for multiple safe area modifications I came to this thought.
Would it make sense if there was a single safeArea property that accepts types?

Here is an example in the form of xml:
```
<GridLayout iosSafeArea="normal"/>
<GridLayout iosSafeArea="overflow"/>
<GridLayout iosSafeArea="ignore"/>
```

rigor789:
@DimitrisRK I like it - however I wonder if we should just use `safeArea="..."` opening up the possibility to implement on Android in the future (if required).

Either way it'd be a breaking change - but the migration would be easy - so I'd say we could go for it in 8.0

farfromrefug:
Good idea. Could make things simpler. In the meantime would be nice to still look at this PR. In a sense a bottom sheet is not usable without that property . well it will break as soon as you start using a listview or a pager.

rigor789 (approved):

PR #9099, opened 2020-12-16
fix: roll back CopyWebpackPlugin to v4.6.0

### LINKED ISSUE #8849: Nativescript 7 HMR issue, update on every change in code
**Environment**
 - CLI: 7.0.8
 - Cross-platform modules:
 - Android Runtime: 7.0.
 - iOS Runtime:
 - Plugin(s):

**Describe the bug**
Whenever you type to change the code even if you press the spacebar, HMR is fired and the app is restarted.
**To Reproduce**
tns run android

**Expected behavior**
HMR is fired only after you save the changes and the app should not restart.
**Sample project**
-none

**Additional context**
-none

shirakaba:
## PR Checklist

- [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/WritingUnitTests.md

HMR non-functional

## What is the new behavior?
HMR functional

Closes https://github.com/shirakaba/react-nativescript/issues/65
Closes #8849

Not aware of any breaking changes; this is fixing a regression more than anything.

### DIFF
--- a/packages/webpack/package.json
+++ b/packages/webpack/package.json
@@ -52,7 +52,7 @@
   "dependencies": {
     "@angular-devkit/core": "~10.0.0",
     "clean-webpack-plugin": "~3.0.0",
-    "copy-webpack-plugin": "~6.0.2",
+    "copy-webpack-plugin": "4.6.0",
     "css": "~3.0.0",
     "css-loader": "~4.2.0",
     "escape-string-regexp": "~4.0.0",
--- a/packages/webpack/templates/webpack.angular.js
+++ b/packages/webpack/templates/webpack.angular.js
@@ -116,9 +116,9 @@ module.exports = env => {
 
   const copyIgnore = { ignore: [`${relative(appPath, appResourcesFullPath)}/**`] };
   const copyTargets = [
-    { from: 'assets/**', noErrorOnMissing: true, globOptions: { dot: false, ...copyIgnore } },
-    { from: 'fonts/**', noErrorOnMissing: true, globOptions: { dot: false, ...copyIgnore } },
-    ...copyReplacements
+    { from: { glob: 'assets/**', dot: false } },
+    { from: { glob: 'fonts/**', dot: false } },
+    ...copyReplacements,
   ];
 
   if (!production) {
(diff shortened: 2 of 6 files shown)


shirakaba:
Note: All the links in the PR checklist are broken, so I can't read up on those.

shirakaba:
See https://github.com/NativeScript/nativescript-dev-webpack/pull/700/files for an earlier commit showing what the CopyWebpackPlugin configuration looked like back when we were using v4.6.0.

rigor789 (approved):
LGTM - I tested locally and it's working. 

However, I'm a bit unsure on the release path:
* Should not be an in-range update (ie a patch/minor version) - it would break projects unless they revert the options syntax in their configs
* Should not be a major bump - it's a downgrade, doesn't really make sense as a major imo
* Should be opt-in by projects

My idea:
Publish under `3.1.0-hmr.0` and the `hmr` tag on npm.

Another issue: do we merge it? Do we keep it around as a PR, or perhaps a new branch? Since it's going to be tagged as `hmr` and not `latest` - it shouldn't really live on `master`.

**EDIT:**
Published as `@nativescript/webpack@3.1.0-hmr.0` (tagged as `hmr`) - we will keep the source in this PR without merging it (may merge it down to a new branch if necessary in the future). A short guide to use will be posted early next week, but a quick overview:
```bash
npm i --save-dev @nativescript/webpack@3.1.0-hmr.0
rm -rf webpack.config.js node_modules
npm i
```

This should re-create the `webpack.config.js` with the downgraded CopyWebpackPlugin syntax.

---
Converted PR to draft to avoid accidental merging.

shirakaba:
@rigor789 Thanks so much for testing this out.

I agree that versioning it in-range would be problematic for existing projects.

I think major would actually be acceptable, though – we’re not aware of any merit that was gained by updating the plugin (so I wouldn’t consider it an upgrade), and yet we are sorely aware of some functionality having been lost (HMR). I can guarantee that users will continue to come asking why HMR isn’t working and I don’t think they should have to opt-in to a special version name to restore it – it should be considered a stable and desired part of NativeScript functionality, not a “nice to have”. By marking it simply as a major version update, we’d keep things correct by semver, easy to explain, and most importantly, easy to maintain (by virtue of not having to maintain a separate branch with this commit).

rigor789:
@shirakaba you are absolutely right - will get this merged and bumped as 4.0! 🎉

PR #9103, opened 2020-12-18
feat(webpack): support workspace style configs

NathanWalker:
* Support workspace solution style tsconfig
* Works with all angular integration options (default templates, schematics or workspaces)
* Will be simplified when {N} 8 and webpack 5 integrations
* Published as `3.1.0-rc.0` to allow workspace updates to occur now


NathanWalker:
Going to embed this in schematics for now - When {N} 8 arrives with webpack 5 we can handle this holistically.

PR #9110, opened 2020-12-24
feat(color): added utilities and improved color parsing performance

farfromrefug:
This is second part of my last PR about adding features to Color. It actually can replace the last PR. And it does more:
* add new utilities function taken from tinycolor
* make `Color` much faster. Running profiling in my apps i realized `Color` was pretty slow. 
    - `hex` method was dead slow. `_argb` is now used to create the `hex` string.
    - all the regexp tests were actually unnecessary as it would throw an error anyway if the color was not formatted correctly.


NathanWalker:
@farfromrefug when get a chance possible to resolve the conflicts and we could get this into 8.1?

farfromrefug:
@NathanWalker i want to but i don't understand some changes you made when you merged the first one.
like for example your brought back `_normalizeHex` when it is not necessary and takes time.
```
const hex = this._normalizeHex(arg);
this._argb = this._argbFromString(hex)
```
Same thing with `hex` getter. You splitted the components instead of simply using `toString(16)`. Yes we could end up with `#fff` but do we really care? It is faster.

I ll fix the rest when we ve decided on this.

farfromrefug:
@NathanWalker i just realized that it was a bit worse that i thought. The changes you made when you rewrote my original PR actually broke most of the colors functions (released in 8.x). 
That PR (once i force the resolves) fixes them back. I will check the linting but please accept it as it is or test it if you rewrite it.

farfromrefug:
@NathanWalker i updated the PR and fixed everything. AS a note currently in N all color functions based on hsl/hsv are broken

NathanWalker (approved):

farfromrefug:
@NathanWalker I still need to push one last fix on the hex getter

farfromrefug:
@NathanWalker this PR really fixes a lot. I just added a fix for `linear-gradient` parsing of hex colors with the new `#rrggbbaa` format

NathanWalker (approved):

farfromrefug:
@NathanWalker do we have tests for linear-gradient parsing ? (Not in front of my computer). I changed lot to refactor and handle all colors parsing in Color class. Would be good to ensure all that is good

PR #9111, opened 2020-12-24
fix: prevent circular reference && Cannot access … before initialization

farfromrefug:
I am not sure why i dont see the issue in all apps but the issue is definitely there:
* `Page` imports `Frame`
*  `Frame` imports `Page`
I ended up in a case  where i got `Cannot access … before initialization` because of that.
By using `frame.constructor.name === 'Frame'` now `Frame` is just used as a type in `page-common` and thus we dont get a circular referecence anymore


NathanWalker:
Thanks @farfromrefug we can probably further enforce/ensure only the `type` is imported by also doing this at the top:
```
import type { Frame } from '../frame';
```

If you'd like to make the one last improvement we'll get this merged and out in a patch.

NathanWalker (approved):

PR #9119, opened 2020-12-30
fix(android): refactor to make less calls to native

farfromrefug:
This PR simplifies the android code a bit to make less calls to android native constants. Thus less JNI bridge calls =>faster


packages/core/ui/core/view/index.android.ts
```diff
@@ -63,6 +63,9 @@ const statePressed = 16842919; // android.R.attr.state_pressed
 const stateEnabled = 16842910; // android.R.attr.state_enabled
 const styleAnimationDialog = 16973826; // android.R.style.Animation_Dialog
 
+const VERTICAL_GRAVITY_MASK = 112; // android.view.Gravity.VERTICAL_GRAVITY_MASK
+const HORIZONTAL_GRAVITY_MASK = 7; // android.view.Gravity.HORIZONTAL_GRAVITY_MASK
+
```
NathanaelA, line 68: I would also add the `LEFT` and `TOP` up here so they are no longer magic numbers.   Your refactor leaves `.LEFT` and `.TOP` in the source as magic numbers so people have no idea why they are these values.

packages/core/ui/core/view/index.android.ts
```diff
@@ -860,30 +863,32 @@ export class View extends ViewCommon {
 	[horizontalAlignmentProperty.setNative](value: HorizontalAlignment) {
 		const nativeView = this.nativeViewProtected;
 		const lp: any = nativeView.getLayoutParams() || new org.nativescript.widgets.CommonLayoutParams();
+		const gravity = lp.gravity;
+		const weight = lp.weight;
```
NathanaelA, line 867: I would eliminate these two extra variables.   And revert the rest of the statements back to using `lp.gravity` and `lp.weight` as all you have done is added additional memory usage (which has to be GC'd) for no purpose and now have mixed values throughout below.  You assign `lp.gravity` from `gravity` on line 872.   This just confuses the source for people who later read it.

packages/core/ui/core/view/index.android.ts
```diff
@@ -898,30 +903,32 @@ export class View extends ViewCommon {
 	[verticalAlignmentProperty.setNative](value: VerticalAlignment) {
 		const nativeView = this.nativeViewProtected;
 		const lp: any = nativeView.getLayoutParams() || new org.nativescript.widgets.CommonLayoutParams();
+		const gravity = lp.gravity;
+		const height = lp.height;
```
NathanaelA, line 907: Same comment as above.

packages/core/ui/core/view/index.android.ts
```diff
@@ -860,30 +863,32 @@ export class View extends ViewCommon {
...
 		const lp: any = nativeView.getLayoutParams() || new org.nativescript.widgets.CommonLayoutParams();
+		const gravity = lp.gravity;
+		const weight = lp.weight;
 		// Set only if params gravity exists.
 		if (lp.gravity !== undefined) {
 			switch (value) {
 				case 'left':
-					lp.gravity = android.view.Gravity.LEFT | (lp.gravity & android.view.Gravity.VERTICAL_GRAVITY_MASK);
-					if (lp.weight < 0) {
+					lp.gravity = 3 | (gravity & VERTICAL_GRAVITY_MASK); // android.view.Gravity.LEFT
```
NathanaelA, line 872: What is `3`, `1`, `5`, `7`.   Definitely would be better to create the `const` at the top so that the code is easier to read.

NathanaelA (changes requested):

packages/core/ui/core/view/index.android.ts
farfromrefug: In fact I would not. I forgot a if lp.gravity which makes 2 calls to native accesser. Need to fix this. The idea is to make sure we don't access those native variables multiple times => multiple jni calls

packages/core/ui/core/view/index.android.ts
farfromrefug: I remember  i copied the way it was done somewhere else. And I didn't want to unecessarly create global module variables. 
But sure I ll change it

NathanWalker (approved):

PR #9120, opened 2020-12-30
Faster background color

farfromrefug:
When not using border radius or border colors it is much faster, it is much faster to set only backgroundColor.
In my profiling tests the creation of the border drawable was really slow.


NathanaelA:
This sounds like a awesome change, Thanks!   Have you run it through the app tests to verify it doesn't break anything?

farfromrefug:
Might need to run it again on that branch. Did pass on my master branch

NathanWalker (approved):

PR #9121, opened 2020-12-30
fix: no need for batchUdated if no selector

farfromrefug:
very simple change preventing unecessary batchUpdate


NathanaelA (approved):
Awesome, thanks!

PR #9126, opened 2021-01-03
Webpack5 android appcomponents

farfromrefug:
* move `default.config.stub.js` to `default.config.stub.ts` so that it is copied by the build
* bring back `android-app-components-loader` until we find a better solution


packages/webpack5/src/loaders/android-app-components-loader/index.ts
```diff
@@ -0,0 +1,18 @@
...
+    const augmentedSource = `
+        if (global.isAndroid) {
+            ${imports}
+        }
+
+        ${content}
+    `;
+
+    this.callback(null, augmentedSource, map);
+};
```
rigor789, line 18: I don't think we need to do it like this (I will confirm this, just sharing thoughts) - essentially what we care about is that the app modules are in the bundle for SBG to pick them up & generate the java classes.

With the new config setup - we can easily append these to the main entry instead I believe (to be confirmed).

rigor789 (commented):
The reason I kept the config stub as plain `.js` and copied by the build script (when you run npm pack) is because the `ts->js` conversion could potentially change the syntax/formatting - this could change from version to version - I'd rather maintain the final version, which we can easily diff in case it needs to change etc.

On the `android-app-component-loader` - I will play around with it to see if we need it.

farfromrefug:
@rigor789 OK was not copied for me. Maybe because the folder was not existing yet ?

rigor789:
@farfromrefug oh that could be the case
```
"prepack": "npm run build && cp -R src/stubs dist/stubs && chmod +x dist/bin/index.js"
```
I never created that dir 😆 - will probably change this up anyways - don't like complex npm scripts too much...

rigor789:
Implemented in a simpler approach here: https://github.com/NativeScript/NativeScript/commit/73e97116d50ccc202059488cf8f274162f4b12fb

PR #9133, opened 2021-01-05
chore: refactor to allow Webpack to bundle only what s necessary.

farfromrefug:
* added define variable for UI custom flavors allowing flavors to remove unecessary code (builder)
* added define variable to disable XML parser (enabled by default to not break anything)

This is kind of a big one. The refactoring / split into files should not have any consequence though.


shirakaba:
@farfromrefug Would it be possible to expose an option allow users to opt-in to including the XML parser?

Use-case: Users have an existing XML-based app that they want to incrementally port to a custom UI flavour such as React. I've actually had to do this for both of my apps [NSIDE](https://github.com/shirakaba/nside) and [RPS Project Tracker 2.0](https://github.com/shirakaba/rpstrackerrns).

Even if the use-case does not relate to incremental porting, I wouldn't want to stop teams from using a mix of different flavours (e.g. some corners in Svelte, other corners in XML, other corners in React) in their project, if possible.

packages/webpack5/src/helpers/flavor.ts
```diff
@@ -2,6 +2,23 @@ import { defaultConfigs } from '@nativescript/webpack';
 import { getAllDependencies } from './dependencies';
 import { error } from './log';
 
+/**
+ * Utility to determine the project flavor based on installed dependencies
+ * (vue, angular, react, svelete, typescript, javascript...)
+ */
+export function projectUsesCustomFlavor(): boolean {
+	const dependencies = getAllDependencies();
+
```
rigor789, line 11: Could be simplified to something like:
```suggestion
	return [
	  'vue', 
	  'angular',
	  'react',
	  'svelte'
	].includes(determineProjectFlavor())
```

packages/core/ui/styling/style-scope.ts
```diff
@@ -590,13 +586,13 @@ export class CssState {
...
 		for (const property in oldProperties) {
 			if (!(property in newPropertyValues)) {
-				if (property in view.style) {
-					view.style[`css:${property}`] = unsetValue;
+			if (property in view.style) {
+				view.style[`css:${property}`] = unsetValue;
 				} else {
-					// TRICKY: How do we unset local value?
-				}
+				// TRICKY: How do we unset local value?
```
shirakaba, line 592: I'm not sure whether we're discussing the same thing here, but in React NativeScript, I have a trick for resetting values such as the `spans` property on `FormattedString`. I'd definitely advise putting it in a try-catch, though:

https://github.com/shirakaba/react-nativescript/blob/6561b6821d8b3f0ac6bd0792dede6e7efeb5a310/react-nativescript/src/nativescript-vue-next/runtime/nodes.ts#L531-L546

packages/core/ui/repeater/index.ts
```diff
@@ -129,7 +129,7 @@ export class Repeater extends CustomLayoutView {
 			}
 
 			if (!viewToAdd) {
-				viewToAdd = this.itemTemplate ? Builder.parse(this.itemTemplate, this) : this._getDefaultItemContent(i);
+				viewToAdd = (!__UI_CUSTOM_FLAVOR__ && this.itemTemplate) ? Builder.parse(this.itemTemplate, this) : this._getDefaultItemContent(i);
```
rigor789, line 132: For global "magic" variables, I think we should settle on a convention & use full statements like 
```ts
if(__UI_CUSTOM_FLAVOR__) { 
  // ...
} else {
  // ...
}
```
 
This makes it simpler to scan & understand what happens with/without a flag.

packages/core/ui/styling/style-scope.ts
```diff
@@ -590,13 +586,13 @@ export class CssState {
 		// Unset removed values
 		for (const property in oldProperties) {
 			if (!(property in newPropertyValues)) {
-				if (property in view.style) {
-					view.style[`css:${property}`] = unsetValue;
+			if (property in view.style) {
+				view.style[`css:${property}`] = unsetValue;
```
shirakaba, line 590: Careful here. `unsetValue` isn't universally safe to set. I've found four properties for which the Style validator doesn't accept it (and therefore the app crashes if attempted), even though it may be technically safe to set if the validator would simply allow the value through:

https://github.com/shirakaba/react-nativescript/blob/08f92e7a19f0eec471d2ff8c3614e494b5872099/react-nativescript/src/shared/CSSPropertyOperations.ts#L20-L28

There may be more such properties; these are merely the four that I managed to determine.

The root problem is that Property doesn't strictly require the specification of a default value, and doesn't implicitly accept `unsetValue`; the fact that these aspects are both optional is a major reason that HMR is so buggy.

packages/webpack5/src/configuration/base.ts
```diff
@@ -240,7 +242,9 @@ export default function (config: Config, env: IWebpackEnv): Config {
 		{
 			__DEV__: mode === 'development',
 			__NS_WEBPACK__: true,
-			__CSS_PARSER__: JSON.stringify('css-tree'), // todo: replace from config value
+			__UI_XML_PARSER__: true,
+			__UI_CUSTOM_FLAVOR__: projectUsesCustomFlavor(),
```
rigor789, line 246: Can we think of a better name for `__UI_CUSTOM_FLAVOR__`?

Should boolean flags like `__UI_XML_PARSER__` use a name like `__UI_USE_XML_PARSER__` to imply it's a boolean value? I guess in this case - it follows the same logic as `__CSS_PARSER__` - which can be a string or a boolean - probably doesn't make sense in this case to prefix with `USE`.

packages/webpack5/src/helpers/flavor.ts
```diff
@@ -2,6 +2,23 @@ import { defaultConfigs } from '@nativescript/webpack';
 import { getAllDependencies } from './dependencies';
 import { error } from './log';
 
+/**
+ * Utility to determine the project flavor based on installed dependencies
+ * (vue, angular, react, svelete, typescript, javascript...)
+ */
+export function projectUsesCustomFlavor(): boolean {
```
rigor789, line 9: Same question as with `__UI_CUSTOM_FLAVOR__` - can we think of a better name/terminology here?

packages/webpack5/src/helpers/flavor.ts
```diff
@@ -2,6 +2,23 @@ import { defaultConfigs } from '@nativescript/webpack';
...
+	const dependencies = getAllDependencies();
+	
+	if (dependencies.includes('nativescript-vue') ||
+		dependencies.includes('@nativescript/angular') ||
+		dependencies.includes('react-nativescript') ||
+		dependencies.includes('svelte-native')
```
shirakaba, line 15: If possible, renderer developers should also be allowed to explicitly inform the webpack config that their project uses a custom flavour (would this be possible? I'm just skimming this); otherwise, we'll cause friction to some future flavours. There are also some lesser flavours out there like crank-native (which admittedly I've retired) and the unfinished flavours for Glimmer and Preact, etc., so it would be best not to make hard assumptions.

shirakaba (changes requested):

packages/webpack5/src/configuration/base.ts
shirakaba: Something like `__USES_EXTERNAL_UI_RENDERER__` jumps to mind. Something along those lines.

I agree that in both cases, the variable name should clearly suggest the boolean nature, at least.

packages/webpack5/src/helpers/flavor.ts
farfromrefug: @rigor789 absolutely ! It is up for debate :D

packages/webpack5/src/configuration/base.ts
farfromrefug: @rigor789 i like the `__UI_USE_XML_PARSER__`. I vote for this

packages/core/ui/repeater/index.ts
farfromrefug: @rigor789 you are right! will change that.

packages/webpack5/src/helpers/flavor.ts
rigor789: I've implemented this for custom targets (like desktop in the future) - we can definitely apply the same/similar approach for flavors/these helpers (outside of the scope of this PR)

farfromrefug:
> @farfromrefug Would it be possible to expose an option allow users to opt-in to including the XML parser?
> 
> Use-case: Users have an existing XML-based app that they want to incrementally port to a custom UI flavour such as React. I've actually had to do this for both of my apps [NSIDE](https://github.com/shirakaba/nside) and [RPS Project Tracker 2.0](https://github.com/shirakaba/rpstrackerrns).
> 
> Even if the use-case does not relate to incremental porting, I wouldn't want to stop teams from using a mix of different flavours (e.g. some corners in Svelte, other corners in XML, other corners in React) in their project, if possible.

@shirakaba it is already there and on by default. `__UI_XML_PARSER__` is it. I personally already turn it off

packages/core/ui/repeater/index.ts
farfromrefug: @rigor789 commited the change. Seems a bit weird in some case as it creates an empty scope for the `if(__UI_CUSTOM_FLAVOR__)`. Let me know if that is good with you

packages/webpack5/src/configuration/base.ts
farfromrefug: about `__UI_CUSTOM_FLAVOR__` should we go then for `__UI_USE_EXTERNAL_RENDERER__` ?

packages/core/ui/repeater/index.ts
rigor789: Totally fine with me, minification should drop it anyway.

packages/webpack5/src/helpers/flavor.ts
farfromrefug: Should be good now i resolve that conversation

packages/webpack5/src/configuration/base.ts
farfromrefug: Should be good now i resolve that conversation

packages/core/ui/repeater/index.ts
farfromrefug: Should be good now i resolve that conversation

PR #9154, opened 2021-01-20
Ios modal status bar

farfromrefug:
I have a plugin `@nativescript-community/system-ui` which allows to change status bar background color.
without `updateWithWillAppear` there was no way for my plugin to rollback the status bar color once i closed a modal page which was changing it.
the `updateWithWillDisappear` allows to have smoother modal close by changing the status bar color as soon as the modal close animation starts
My plugin can hook those methods through class mixins
This is mostly used by fullscreen modals.

This PR mostly dont change behavior. I just added `updateStatusBar` from `updateWithWillAppear` to ensure the status bar style is also updated in the modal close case.

Working on that i saw that N is not firing page navigation events for modals. I would raise the question should we not bring it back? I feel it weird for a page that i show with modal not to get the events. Same for the page being navigating from.


NathanWalker (approved):

farfromrefug:
@NathanWalker thanks !!!

PR #9175, opened 2021-01-26
feat: setSelection method for editable text components

farfromrefug:
this add support for `setSection` for all editable text components

### DIFF
--- a/packages/core/ui/editable-text-base/index.ios.ts
+++ b/packages/core/ui/editable-text-base/index.ios.ts
@@ -194,7 +194,10 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
 		const view = this.nativeTextViewProtected;
 		if (stop !== undefined) {
 			const begin = view.beginningOfDocument;
-			view.selectedTextRange = view.textRangeFromPositionToPosition(view.positionFromPositionOffset(begin, start), view.positionFromPositionOffset(begin, stop));
+			view.selectedTextRange = view.textRangeFromPositionToPosition(
+				view.positionFromPositionOffset(begin, start), 
+				view.positionFromPositionOffset(begin, stop)
+			);
 		} else {
 			const begin = view.beginningOfDocument;
 			const pos = view.positionFromPositionOffset(begin, start);
--- a/packages/core/ui/editable-text-base/index.ios.ts
+++ b/packages/core/ui/editable-text-base/index.ios.ts
@@ -192,13 +192,12 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
 	}
 	public setSelection(start: number, stop?: number) {
 		const view = this.nativeTextViewProtected;
-			if (view) {
-				if (stop !== undefined) {
+		if (view) {
+			if (stop !== undefined) {
 				const begin = view.beginningOfDocument;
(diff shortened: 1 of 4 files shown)


packages/core/ui/editable-text-base/index.ios.ts
```diff
@@ -189,6 +190,17 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
 
 		this.nativeTextViewProtected.autocorrectionType = newValue;
 	}
+	public setSelection(start: number, stop?: number) {
+		const view = this.nativeTextViewProtected;
+		if (stop !== undefined) {
+			const begin = view.beginningOfDocument;
+			view.selectedTextRange = view.textRangeFromPositionToPosition(view.positionFromPositionOffset(begin, start), view.positionFromPositionOffset(begin, stop));
```
rigor789, line 197: Perhaps this makes this line a bit more readable
```suggestion
			view.selectedTextRange = view.textRangeFromPositionToPosition(
						view.positionFromPositionOffset(begin, start),
						view.positionFromPositionOffset(begin, stop)
			);
```

rigor789 (approved):

packages/core/ui/editable-text-base/index.ios.ts
```diff
@@ -189,6 +190,20 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
 
 		this.nativeTextViewProtected.autocorrectionType = newValue;
 	}
+	public setSelection(start: number, stop?: number) {
+		const view = this.nativeTextViewProtected;
+		if (stop !== undefined) {
```
NathanWalker, line 195: Might be good to wrap all this in conditional to protect against view (android side does it already but always best to be careful with ios as well):
```
if (view) {
```

NathanWalker (changes requested):

farfromrefug:
@NathanWalker not sure why it says changes are requested still. I did the asked changes

packages/core/ui/editable-text-base/index.ios.ts
```diff
@@ -189,6 +190,22 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
...
+				const begin = view.beginningOfDocument;
+				view.selectedTextRange = view.textRangeFromPositionToPosition(
+					view.positionFromPositionOffset(begin, start), 
+					view.positionFromPositionOffset(begin, stop)
+				);
+			} else {
+				const begin = view.beginningOfDocument;
+				const pos = view.positionFromPositionOffset(begin, start);
+				view.selectedTextRange = view.textRangeFromPositionToPosition(pos, pos);
+			}
```
rigor789, line 206: Formatting seems to be off here - did you run `npm run setup` after cloning the repo to ensure husky & git-hooks are installed? These should then auto-format changed files on commit.

rigor789 (dismissed):
LGTM - couple formatting issues only.

packages/core/ui/editable-text-base/index.ios.ts
farfromrefug: yes I do and I use eslint. wonder why it does not work. shouldn't it prevent the commit or auto format ?

rigor789 (approved):

NathanWalker (approved):

PR #9190, opened 2021-02-03
fix(android): prevent “flashing” activity on app start

farfromrefug:
This was due to a change made a while back to fix an issue with resetRootView.
The consequence was that we were waiting for the frame to be “attachedToWindow” to load the first frame page.
That PR make it so that we apply that check/fix only when the activity was reset. Thus it wont happen on app start

The flash issue is currently seeable on ALL apps. Though it depends on the complexity of the first frame page. the more complex the longer you lll see the "white" screen issue (white being your activity background color)


rigor789:
We need to make sure the are no cases where `_onRootViewReset` is not called at all - as those cases would fail with this change. Don't know without a deeper look if a scenario like that is even possible, just leaving it here as a note.

NathanWalker (approved):

PR #9196, opened 2021-02-05
ci: run apps/automated on CI

rigor789:
TIMER tests are flaky, we may want to try a different strategy to test timers, perhaps with https://github.com/sinonjs/fake-timers or something similar.


farfromrefug:
pretty awesome. I really think it will help a lot !

NathanWalker (approved):

PR #9210, opened 2021-02-12
chore: background color tests and refactoring

farfromrefug:
Following the issues about background color i added some tests to check for setting a color then removing it.
I also refactored a bit to try and make it more understandable. i hope we have enough tests and that i did not break anything :s

### DIFF
--- a/apps/automated/src/main-page.ts
+++ b/apps/automated/src/main-page.ts
@@ -18,7 +18,7 @@ Trace.addCategories(Trace.categories.Test + ',' + Trace.categories.Error);
 // ));
 
 function runTests() {
-	setTimeout(() => tests.runAll(''), 10);
+	setTimeout(() => tests.runAll('VIEW'), 10);
 }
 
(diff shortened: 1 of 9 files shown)


apps/automated/src/main-page.ts
```diff
@@ -18,7 +18,7 @@ Trace.addCategories(Trace.categories.Test + ',' + Trace.categories.Error);
 // ));
 
 function runTests() {
-	setTimeout(() => tests.runAll(''), 10);
+	setTimeout(() => tests.runAll('VIEW'), 10);
```
rigor789, line 21:
```suggestion
	setTimeout(() => tests.runAll(''), 10);
```

NathanWalker:
Thanks @farfromrefug for the extra tests, added them from this PR directly to main here:
https://github.com/NativeScript/NativeScript/commit/5a86a60b61d35b2c95045bca64b29a46d8b38308
all pass 🆗

PR #9223, opened 2021-02-17
fix: refactor background handling.

farfromrefug:
The idea is for views to handle special case themselves.
I realised that android.widget.Button was materialized even if you were not using it because there was a test for instanceof in the background handling. Now the special background handling is done in Button and ActionBar

Also we dont create the `_cachedDrawable` again. Instead we use the one we got from `backgroundInternal.getDefault`.

It also makes the code much easier to understand.


NathanWalker (approved):

rigor789:
@NathanWalker we actually can't merge this with the 8.0 changes, will need to redo on top of the 8.0 changes.

farfromrefug:
@rigor789 i think i merged it with 8.0.0 and boxShadows. All tests are passing. Should also fix the issue with actionbar background color

packages/core/ui/action-bar/index.android.ts
```diff
@@ -7,13 +7,18 @@ import { colorProperty } from '../styling/style-properties';
 import { ImageSource } from '../../image-source';
 import * as application from '../../application';
 import { isAccessibilityServiceEnabled, updateContentDescription } from '../../accessibility';
+import type { Background } from 'ui/styling/background';
```
rigor789, line 10:
```suggestion
import type { Background } from '../styling/background';
```

Believe that should be a relative import

packages/core/ui/styling/background.android.ts
```diff
@@ -1,139 +1,14 @@
...
-import { BackgroundClearFlags } from './background-common';
 export * from './background-common';
 
-interface AndroidView {
-	_cachedDrawable: android.graphics.drawable.Drawable.ConstantState | android.graphics.drawable.Drawable;
-}
-
 // TODO: Change this implementation to use
 // We are using "ad" here to avoid namespace collision with the global android object
 export namespace ad {
```
rigor789, line 11: Can we remove the empty namespace?

packages/core/ui/core/view/index.android.ts
```diff
@@ -23,6 +24,7 @@ import lazy from '../../../utils/lazy';
 import { accessibilityEnabledProperty, accessibilityHiddenProperty, accessibilityHintProperty, accessibilityIdentifierProperty, accessibilityLabelProperty, accessibilityLanguageProperty, accessibilityLiveRegionProperty, accessibilityMediaSessionProperty, accessibilityRoleProperty, accessibilityStateProperty, accessibilityValueProperty } from '../../../accessibility/accessibility-properties';
 import { AccessibilityLiveRegion, AccessibilityRole, AndroidAccessibilityEvent, setupAccessibleView, isAccessibilityServiceEnabled, sendAccessibilityEvent, updateAccessibilityProperties, updateContentDescription, AccessibilityState } from '../../../accessibility';
 import * as Utils from '../../../utils';
+import { CSSShadow } from 'ui/styling/css-shadow';
```
rigor789, line 27:
```suggestion
import { CSSShadow } from '../../styling/css-shadow';
```

packages/core/ui/styling/background.android.ts
farfromrefug: @rigor789 i did not remove it as it was exported. But i guess we can

rigor789 (approved):

farfromrefug:
@rigor789 could we merge this in 8.1.0 ? i confirmed it fixes this issue (it works in my fork) https://github.com/NativeScript/NativeScript/issues/9263

PR #9240, opened 2021-02-23
fix(android): KeyboardType now respects numbers

NathanaelA:
## PR Checklist

Passing in a number value for the Android Keyboard type is supposed to be supported, however it wasn't converting the "string" value to a number.

## What is the new behavior?
It will now convert the number to a string.

### DIFF
--- a/packages/core/ui/editable-text-base/index.android.ts
+++ b/packages/core/ui/editable-text-base/index.android.ts
@@ -472,4 +472,15 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
 			this.nativeTextViewProtected.setFilters(newFilters);
 		}
 	}
+
+	public setSelection(start: number, stop?: number) {
+		const view = this.nativeTextViewProtected;
+		if (view) {
+			if (stop !== undefined) {
+				view.setSelection(start, stop);
+			} else {
+				view.setSelection(start);
+			}
+		}
+	}
 }
--- a/packages/core/ui/editable-text-base/index.android.ts
+++ b/packages/core/ui/editable-text-base/index.android.ts
@@ -285,7 +285,7 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
 				if (!isNaN(inputType)) {
 					newInputType = inputType;
 				} else {
-					newInputType = android.text.InputType.TYPE_DATETIME_VARIATION_NORMAL;
+					newInputType = android.text.InputType.TYPE_CLASS_TEXT;
 				}
 				break;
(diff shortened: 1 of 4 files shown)


packages/core/ui/editable-text-base/index.android.ts
```diff
@@ -281,7 +281,12 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
 				break;
 
 			default:
-				newInputType = value;
+				const inputType = +value;
+				if (!isNaN(inputType)) {
```
rigor789, line 285: Since we have an if/else, should we switch to just `isNaN` rather than `!isNaN` - just makes it a bit easier to read?

packages/core/ui/editable-text-base/index.android.ts
```diff
@@ -281,7 +281,12 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
...
+					newInputType = inputType;
+				} else {
+					newInputType = android.text.InputType.TYPE_DATETIME_VARIATION_NORMAL;
```
rigor789, line 288: What's the reason for using `TYPE_DATETIME_VARIATION_NORMAL` and not `TYPE_CLASS_TEXT`?

packages/core/ui/editable-text-base/index.android.ts
NathanaelA: Not in disagreement, but I felt that keeping the style the same as all the other code below was more important...   (See the other places like line 333, 389.)

packages/core/ui/editable-text-base/index.android.ts
NathanaelA: Opps that is a typo, that was supposed to be VARIATION_NORMAL, good catch.  I'll fix.

NathanaelA: It has been updated to TYPE_CLASS_TEXT as all the *_variation_normal flags are `0` so they don't actually do anything, so using TEXT_CLASS_TYPE seems like a better default in this case.

packages/core/ui/editable-text-base/index.android.ts
rigor789: I guess the approach was to have the "fallback" in the `else` case - which does make sense, so guess we should just leave it - probably not worth adding a new utility for `!isNaN` (like `isNumber`, since the `isNaN` version kinda reads weird: `not is not a number`).

NathanWalker (approved):

PR #9243, opened 2021-02-24
fix(bottom-navigation): fragment handling

rigor789:
## PR Checklist

## What is the new behavior?

### DIFF
--- a/nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts
+++ b/nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts
@@ -540,7 +540,9 @@ export class BottomNavigation extends TabNavigationBase {
 
             if (fragment != null) {
                 fragment.setMenuVisibility(true);
-                fragment.setUserVisibleHint(true);
+                // commenting out as it fixes rare crash when going
+                // back from deeply nested/modally shown fragments
+                // fragment.setUserVisibleHint(true);
             }
 
             this._currentFragment = fragment;
--- a/nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts
+++ b/nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts
@@ -582,8 +582,8 @@ export class BottomNavigation extends TabNavigationBase {
                     fragmentExitTransition.setResetOnTransitionEnd(true);
                 }
                 if (fragment && fragment.isAdded() && !fragment.isRemoving()) {
-                    const pfm = (<any>fragment).getParentFragmentManager ? (<any>fragment).getParentFragmentManager() : null;
-                    if (pfm && !pfm.isDestroyed()) {
+                    const pfm = (<any>fragment).getParentFragmentManager ? (<any>fragment).getParentFragmentManager() : fragmentManager;
+                    if (pfm) {
                         try {
                             if (pfm.isStateSaved()) {
                                 pfm.beginTransaction().remove(fragment).commitNowAllowingStateLoss();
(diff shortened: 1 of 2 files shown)


rigor789:
Released as [`tns-core-modules@6.5.25-tns-next-02-25-2021-599428343`](https://github.com/NativeScript/NativeScript/runs/1979015841?check_suite_focus=true) to evaluate.

nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts
```diff
@@ -540,7 +540,9 @@ export class BottomNavigation extends TabNavigationBase {
 
             if (fragment != null) {
                 fragment.setMenuVisibility(true);
-                fragment.setUserVisibleHint(true);
+                // commenting out as it fixes rare crash when going
+                // back from deeply nested/modally shown fragments
+                // fragment.setUserVisibleHint(true);
```
NathanaelA, line 545: This looks good to me, apparently, this call is depreciated in AndroidX.

nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts
```diff
@@ -582,8 +584,8 @@ export class BottomNavigation extends TabNavigationBase {
                     fragmentExitTransition.setResetOnTransitionEnd(true);
                 }
                 if (fragment && fragment.isAdded() && !fragment.isRemoving()) {
-                    const pfm = (<any>fragment).getParentFragmentManager ? (<any>fragment).getParentFragmentManager() : null;
-                    if (pfm && !pfm.isDestroyed()) {
```
NathanaelA, line 586: This change concerns me, Why would we NOT be checking if the fragment is destroyed?

nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts
```diff
@@ -582,8 +584,8 @@ export class BottomNavigation extends TabNavigationBase {
...
+                    const pfm = (<any>fragment).getParentFragmentManager ? (<any>fragment).getParentFragmentManager() : fragmentManager;
+                    if (pfm) {
```
NathanaelA, line 588: What happens if the .getParentFramentManger() returns us a isDestroyed() framement here?

rigor789: While testing I could not trigger a case where a destroyed one is returned - but even if it were to happen - I believe FragmentManager handles that case internally, otherwise the try/catch should catch it.

NathanaelA: Was there some reason why it was removed though?   Having a Exception occur is a LOT LOT LOT more costly than doing a check...

rigor789: I'll need to pull up the test project to confirm, but I suspect it had to do with the fragment manager actually being destroyed, but the fragment still being "attached" to the destroyed manager - this would then cleanly detach it, so it can be re-attached without crashing with "Fragment already attached".

rigor789:
_published as 6.5.26_

nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts
NathanWalker: In this case as found in testing from multiple angles, this is in context of destroying a fragment and the check above it (`if (fragment && fragment.isAdded() && !fragment.isRemoving()) {`) is only check needed for protection. If it gets inside calling remove on it is desired and doesn't hurt if was already destroyed anyway.

NathanWalker (approved):

PR #9289, opened 2021-03-31
fix: crash on slide transition

farfromrefug:
the N android `CustomTransition` widget takes an `AnimatorSet` and not an `ObjectAnimator`


NathanWalker (approved):

NathanWalker:
@farfromrefug would this adversely affect prior versions?

farfromrefug:
@NathanWalker no it should not affect prior version as I don't think right now N uses that class. only my fork (pending PRs for android fragment changes) uses it right now

PR #9320, opened 2021-04-09
fix: webpack5 ES incompatibilities using JS

NathanaelA:
## PR Checklist

- [x] The PR title follows our guidelines: 

The current webpack 5 & babel loader does not actually support things the NativeScript v8 engine actually supports. (see notes).   Certain ES code will break webpack and/or babel:
```
// Android & iOS
class C {  
   static x = 'x'; 
};

// ios only (as the android runtime isn't new enough yet in NS 8.0)
class C {
  #x() { return 42; }
  x() {
    return this.#x();
  }
}
```

## What is the new behavior?
Both of these class types will now be accepted by webpack and run on the actual devices since the actual v8 engine supports them.

## Notes:
1. Please note; if you look at the official https://kangax.github.io/compat-table/es6/ you will see other things where Babel 7 doesn't fully support some of the later ES syntax as the v8 engine we are using as shown here  http://nativescript.rocks/compat/esnext/ 

2. This updates the master template to remove both the Babel loader from JS, as it will crash during compiling with errors if you attempt to use things like the above classes (and some of the other JS code as seen in the kangax tables) -- pure ES/JS code should not be run thru Babel as the v8 engine we are using already supports anything it does, no need to slow things down nor add additional compatibility issues.
 
2. This also fixes Webpack 5 so that the Acorn parser allows private & static values in pure ES code without throwing any errors, as the Acorn parser has a extension to enable this support.

### DIFF
--- a/packages/webpack5/__tests__/configuration/__snapshots__/angular.spec.ts.snap
+++ b/packages/webpack5/__tests__/configuration/__snapshots__/angular.spec.ts.snap
@@ -90,17 +90,6 @@ exports[`angular configuration for android 1`] = `
         test: /\\\\.js$/,
         exclude: [
           /node_modules/
-        ],
-        use: [
-          /* config.module.rule('js').use('babel-loader') */
-          {
-            loader: 'babel-loader',
-            options: {
-              generatorOpts: {
-                compact: false
-              }
-            }
-          }
         ]
       },
       /* config.module.rule('workers') */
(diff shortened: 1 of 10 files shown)


farfromrefug:
@NathanaelA on that note is the process to upgrade v8 on android is documented somewhere? @darind do you know?
I would like to upgrade the android v8 runtime to be in sync with iOS but also because i really want to see what [this](https://v8.dev/blog/v8-release-87#unsafe-fast-js-calls) can do for us.

NathanaelA:
@farfromrefug - Martin, it isn't always a simple update; but basically you clone the android runtime then change the v8 checkout number to a later version re-run ninja, and then rebuild.   However, one thing is that the newer v8's now also supports a native implementation of `WeakRef`, this is one of those items we should make sure is switched.   Instead of our implementation as the new finalizer system actually looks for WeakRef classes and our "WeakRef" class is not the class it knows.   So this is one thing that we do know will need to be revamped in the next version of the Android runtime.

farfromrefug:
@NathanaelA Thanks Nathanael, it makes sense. Also means it will requires some work. 
I saw that @rigor789 is doing  it for the iOS runtime (WeakRef switch ). Not sure i am the right one to do this :P Will think about it and maybe try it.
thanks again!

PR #9347, opened 2021-04-20
fix(android): prevent possible crash when app goes to background

farfromrefug:
We faced the same issue with the material components. The old N team knew about it and try to prevent it by testing the fragment manager. However that test is not enough and we can see that we still go into `onAttachedWindow` after `onStop` while putting the app to background.
This new check ensure we dont go there in that case.
It might solve some reported issues not sure.


packages/core/ui/frame/index.android.ts
```diff
@@ -144,8 +145,9 @@ export class Frame extends FrameBase {
 		super._onAttachedToWindow();
 
 		// _onAttachedToWindow called from OS again after it was detach
-		// TODO: Consider testing and removing it when update to androidx.fragment:1.2.0
-		if (this._manager && this._manager.isDestroyed()) {
+		// still happens with androidx.fragment:1.3.2
+		const activity = androidApplication.foregroundActivity;
+		if ((this._manager && this._manager.isDestroyed()) || !activity.getLifecycle().getCurrentState().isAtLeast(androidx.lifecycle.Lifecycle.State.STARTED)) {
```
NathanWalker, line 150: Do you think it’s possible that getLifecycle() or getCurrentState() could ever return null under some unexpected conditions? Might be safest to just guard those calls as well?

farfromrefug: They should not and I would prefer to not have the code let you believe they can. 
androidx appcompatacticity have a life cycle and getCurrentState returns an enum

NathanWalker (approved):

PR #9348, opened 2021-04-20
chore: combine prettier with eslint

farfromrefug:
This simply plugs eslint and prettier together. That way when you lint with eslint it also fixes indentations and such


NathanWalker:
We'll do a `nx migrate latest` first before merging as Nx 12 did provide full eslint migrations and can mesh this up against those.

PR #9369, opened 2021-04-30
fix(webpack): compilation errors should no longer ignored

pfumagalli:
## PR Checklist

- [X] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/tools/notes/DevelopmentWorkflow.md#running-unit-tests-application.

When something is broken within WebPack compilation, the NativeScript WebPack CLI will simply ignore those and let the NativeScript CLI continue with the build as if nothing happened...

## What is the new behavior?

When the WebPack build fails, also the NativeScript build should fail.

### DIFF
--- a/packages/webpack5/src/configuration/base.ts
+++ b/packages/webpack5/src/configuration/base.ts
@@ -231,6 +231,11 @@ export default function (config: Config, env: IWebpackEnv = _env): Config {
 			.plugin('ForkTsCheckerWebpackPlugin')
 			.use(ForkTsCheckerWebpackPlugin, [
 				{
+					// If we use "async" errors compiling typescript will be ignored by
+					// WebPack (we will send the "compilation" message back to the CLI,
+					// and the process exit code will be zero), therefore we will end
+					// up with a broken build
+					async: false,
 					typescript: {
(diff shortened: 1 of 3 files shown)


packages/webpack5/src/configuration/base.ts
```diff
@@ -231,6 +231,11 @@ export default function (config: Config, env: IWebpackEnv = _env): Config {
 			.plugin('ForkTsCheckerWebpackPlugin')
 			.use(ForkTsCheckerWebpackPlugin, [
 				{
+					// If we use "async" errors compiling typescript will be ignored by
+					// WebPack (we will send the "compilation" message back to the CLI,
+					// and the process exit code will be zero), therefore we will end
+					// up with a broken build
+					async: false,
```
rigor789, line 238: This should be `true` in watch/dev mode, and false in release mode to speed up the builds during dev. Just noting here, will merge as-as and then refactor.

The CLI won't restart the app if there are build (synchronous) errors, but will still restart if there are async ts errors - which is a fair trade-off for the speed boost it gives...

rigor789 (approved):

PR #9384, opened 2021-05-08
Webpack 5 improvements

edusperoni:
Please do not squash this PR

## PR Checklist

---

## What is the new behavior?
1. Move xml rules to base
2. Allow virtual modules to work with angular 11
3. Add support for the new angular webpack plugin (12)


farfromrefug:
@edusperoni please could you exclude assets folder for xml parsing? When you have xml / html or any of those files in the assets folder which are needed by third party libs it is quite annoying to see those constant errors from webpack as it tries to parse them.

edusperoni:
@farfromrefug I've not added any code to bundle those assets with webpack, just to allow any XML that has already been required with require.context.

Assets should be copied with the copy plugin so no loaders would parse them. To my understanding only core projects include all files under `~/` so those projects would have to use the webpack chain to exclude the assets folder.

Can you try this PR on a project to see if this is an issue still?

farfromrefug:
@edusperoni i am not 100% sure now but I think the assets folder is not ignore by webpack loaders so anything in it would be loaded by webpack ( local HTML js XML files for eaxmple).
@rigor789 do you know if we exclude the assets folder ?

edusperoni:
@farfromrefug assets are copied with this rule:
https://github.com/NativeScript/NativeScript/blob/87c83fe0e668b2508a779a42f7fa5e4bbeaac4e8/packages/webpack5/src/configuration/base.ts#L372

and **core flavor** files (ts/js) are imported like this:

https://github.com/NativeScript/NativeScript/blob/87c83fe0e668b2508a779a42f7fa5e4bbeaac4e8/packages/webpack5/src/configuration/typescript.ts#L11-L22

So it goes through everything under `~/` that matches `/\\.(xml|js|(?<!d\\.)ts|s?css)$/`. If the developer wants to ignore files, they can add IgnorePlugin and ignore those files from the compilation (https://webpack.js.org/plugins/ignore-plugin/). Also, `~/` maps to `src`, so I'm pretty sure it already ignores assets.

Any flavor other than core js/ts don't use `require.context` at all, so that's not an issue.

edusperoni:
@cla-bot check

PR #9392, opened 2021-05-12
feat: add dismiss keyboard helper

triniwiz:


bradmartin:
Curious if there are edge cases with Android where the following is more useful:

```ts
 const activity = Application.android.foregroundActivity;
      const context = Utils.ad.getApplicationContext();
      const inputManager = context.getSystemService(
        android.content.Context.INPUT_METHOD_SERVICE
      );
      inputManager.hideSoftInputFromWindow(
        activity.getCurrentFocus().getWindowToken(),
        android.view.inputmethod.InputMethodManager.HIDE_NOT_ALWAYS
      );
```

triniwiz:
@bradmartin anything to make as stable a possible I reused the existing function 😁

bradmartin:
Definitely. I have always used the inputmanager approach. I'm not sure of any negative side effects of doing so but it seems to work fine.

NathanWalker (approved):

PR #9494, opened 2021-08-03
perf: allow tree-shaking xml and css parsers

rigor789:
Based on https://github.com/NativeScript/NativeScript/pull/9133


farfromrefug:
@rigor789 awesome.Are all those changes normal? seems weird

rigor789:
@farfromrefug yeah, I'm not sure why they show up like that - I tried merging master into the branch to only show differences, but that didn't work. I suspect it has to do with a rebase somewhere, so these commits are under a different commit-hash & they don't match the ones on master...

Will most likely re-apply the changes manually on top of `master` to make sure we don't include something that we shouldn't.

farfromrefug:
Damn too bad let me know if i can help

PR #9495, opened 2021-08-03
fix(android): nested dialogs

NathanWalker:
## PR Checklist

Android dialogs/fragments handled inside nested views can crash.

## What is the new behavior?

Android dialogs/fragments handled inside nested views no longer crash.

### DIFF
--- a/apps/ui/src/modal-view/nested-modal-tab.xml
+++ b/apps/ui/src/modal-view/nested-modal-tab.xml
@@ -1,0 +1,24 @@
+<Page xmlns="http://schemas.nativescript.org/tns.xsd">
+
+    <ActionBar>
+        <Label text="Nested Modal Tab"></Label>
+    </ActionBar>
+
+    <TabView selectedTabTextColor="green">
+         <TabViewItem title="First">
+            <GridLayout>
+                <Frame id="nestedFrame" defaultPage="nested-frames/nested-page" actionBarVisibility="always"></Frame>
+            </GridLayout>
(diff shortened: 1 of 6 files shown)


triniwiz (approved):
LGTM

farfromrefug:
@NathanWalker @triniwiz i commented on this commit. I think this change is very dangerous and i think any fix you are trying to apply should be done differently. Please look closely at this.

NathanWalker:
@farfromrefug when you can switch to using core “next” tag on any developments today/tomorrow and let us know if you encounter anything.

farfromrefug:
@NathanWalker i dont have any app doing that right now. And i will revert that commit in my fork. I really would advise against that change!

edusperoni:
@NathanWalker this PR introduced a crash for me. When closing the app via the back button and then reopening, the app crashes. I believe it's because you're saving the fragment manager reference to global and only overwritting on modal hide, so things may get messy if an app is closed and the modal was properly disposed in the first place.


Actually, it might be because I'm using bottom sheet from @farfromrefug and I believe it doesn't call the _hideNativeModalView at all.

A better solution might be setting a `_dialogRoot` in the modal view and iterating up the view tree until it finds a fragment or a `_dialogRoot`

farfromrefug:
@edusperoni dont have my code right now but yes i am not calling `_hideNativeModalView`. 
And it shows how that commit can crash for any plugin not relying on that method

PR #9497, opened 2021-08-04
svelte changes from halfnelson

farfromrefug:
I think this should do it from what he did


farfromrefug:
@rigor789 why is the bot asking for @halfnelson  cla ...

PR #9508, opened 2021-08-11
fix: android support for clipToBounds

farfromrefug:
It comes with a major refactoring of the BorderDrawable. Now all Paths are cached not to regenerate them on every draw pass.

Also it is important to understand that the clipping feature on android comes at a cost.
It will be done only when truly necessary. but with complex paths ((border radius > 0 && border non uniform) || border-width >0) then we must apply a costly drawing pass (though hardware accelerated) to accomplish it.


farfromrefug:
Screenshot 2021-08-11 at 13 46 03

Issue #9523, opened 2021-08-24
iOS UIButton backgroundColor does not respond to color changes


Rob4226:
I have found that after a Button is created on iOS, it is not possible to change the backgroundColor of it.  It doesn't change if you change the CSS class nor does accessing the background color like `btn.backgroundColor = new Color('blue')` The below only works on Android, where as on iOS the button background color does not change after it is first set:

```typescript
const btn1 = new Button();
const btn2 = new Button();

let toggle = true;

// Background color doesn't change after it's created
btn1.on("tap", (_args) => {
  if (toggle) {
    btn1.color = new Color("red");
    btn1.backgroundColor = new Color("blue");
    toggle = false;
  } else {
    btn2.color = new Color("yellow");
    btn2.backgroundColor = new Color("black");
    toggle = true;
  }
});

stackLayout.addChild(btn1);
stackLayout.addChild(btn1);
}
```

I have had tons of issues on iOS when changing between dark and light themes while the app is running especially with buttons not responding to color changes and I think this might be the culprit.  It appears that the backgroundColor property is not implemented on the iOS Button view.

I have made a PR to fix this in @nativescript/core.

NathanWalker:
@Rob4226 thank you, possible to post a PR of the change? Looks like just a reference to a fork at moment. If post as a PR we can merge out to the next tag and it will be included in the 8.1 release soon.

Rob4226:
@NathanWalker Yes, I was just running some more tests on it, but it looks good. It's PR #9524. 

Also, I created a new branch using @nativescript/core@7.3.0 and made the necessary changes to fix v7.  I have a couple projects still stuck on 7 so would it be possible to also release a v7.3.1 with this fix?  If you already have a branch for v7 fixes please let me know and I will base the commit on that instead of a new branch. Thanks.

NathanWalker:
Sure thing we could do that, much appreciated 🤗

PR #9524, opened 2021-08-24
fix(ios): background color property on Button

Rob4226:
Currently Button on iOS doesn't respond to background color changes
after it is created. This commit implements the backgroundColorProperty
to access the native UIButton's backgroundColor property.

Fixes NativeScript/NativeScript#9523


NathanWalker (approved):

farfromrefug:
@NathanWalker are you this would not break background handling with gradient , border... ?

CatchABus:
> @NathanWalker are you this would not break background handling with gradient , border... ?

+1. During my background handling rework, I concluded this PR really caused problems with button background.
On top of that, it seems this change was unneeded so something else might have caused the reporter's problem.
PR #10374 will attempt to get rid of these changes.

PR #9533, opened 2021-09-03
fix(modal): persist modal through configuration changes

edusperoni:
## PR Checklist

When you change the device configuration, the activities will be destroyed and the app will recreate them. This includes dialogs.

Currently, doing this will crash because we're clearing the dialog's options and dismissing the modal.

## What is the new behavior?
We no longer dismiss the modal or clear it's config if it's happening during a configuration change.

Also, we cache the activity because `this.getActivity()` can return null if the modal has been detached from the activity.


packages/core/ui/frame/index.ios.ts
```diff
@@ -419,7 +419,7 @@ class UINavigationControllerImpl extends UINavigationController {
 	@profile
 	public viewDidDisappear(animated: boolean): void {
 		super.viewDidDisappear(animated);
-		const owner = this._owner.get();
+		const owner = this._owner.get?.();
```
NathanWalker, line 422: @edusperoni would you mind adding the optional chaining after the _owner? as well? - may sound odd but I have seen cases (especially on destroy where the owner weakref itself has already been cleaned up. We just patched a similar case where a crash was observed exactly on that case here:
https://github.com/NativeScript/NativeScript/commit/a5fd53b0168ab90d66d056973592ea64d821e016

So like this to be totally safe:

```
this._owner?.get?.();
```

NathanWalker (changes requested):

packages/core/ui/frame/index.ios.ts
NathanWalker: Nvm actually - noticed once the base was changed, I was able to update the merge conflict to resolve it.

NathanWalker (approved):

PR #9539, opened 2021-09-07
feat: clipsToBounds support

rigor789:
@farfromrefug here's a branch with the clipsToBounds changes - we had to revert them in the release branch here: https://github.com/NativeScript/NativeScript/commit/778540573a2827a9c029e7381b17b8a85104e334

We've seen some issues with this change, some with the borders and some with boxShadow, so we'd like to step back a bit and make sure everything works as expected before we release this.


farfromrefug:
@DimitrisRK  you cant use `setClipToBounds` as it is a rect. It wont work with rounded corners. `setClipToOutline` is the way to go on api >21.
@rigor789 what issue did you see? But yes clipToBounds will clip shadows. Maybe we should disable it by default (better perfs). But then what do we do on iOS?

CatchABus:
> @DimitrisRK you cant use `setClipToBounds` as it is a rect. It wont work with rounded corners. `setClipToOutline` is the way to go on api >21. @rigor789 what issue did you see? But yes clipToBounds will clip shadows. Maybe we should disable it by default (better perfs). But then what do we do on iOS?

I was wondering why this PR had flaws and It seems Outline has its own limitations.
```
As of API 33, all Outline shapes support clipping. Prior to API 33, only Outlines that could be represented as a rectangle, circle, or round rect supported clipping.
```
https://developer.android.com/reference/android/graphics/Outline#canClip()

It seems that `setClipToOutline` will not clip paths for the time being. :(

farfromrefug:
@CatchABus indeed this is why i am using clipping path (if i remember correctly)

PR #9555, opened 2021-09-20
fix: replace autofill string in textfield

### LINKED ISSUE #9486: [iOS] wrong form data value when using autocomplete (playground demo included)
**Describe the bug**
We got a lot of users complaining that they didn't receive the signup email. When checking their accounts we noticed that they often entered an email address like "infinfo@nstudio.io"  or "jeff.jeff.bezos@amazon.com". We thought they were just mistyping their addresses, until we noticed that it had to do with autocompleting emails on ios.

**To Reproduce**
- on iOS (tested with iOS 14) start typing an email  e.g. "jeff"
- now you see that "jeff.bezos@amazon.com" is suggested. click on the suggestion to use it
- the content of the Textfield is now "jeff.bezos@amazon.com"
- the content of the data model however is "jeffjeff.bezos@amazon.com"  (what you typed before + the full autosuggestion)

**Expected behavior**
The data model should contain "jeff.bezos@amazon.com" instead of "jeffjeff.bezos@amazon.com"

**Sample project**
https://play.nativescript.org/?template=play-js&id=METmUA  (Plain JS)
https://play.nativescript.org/?template=play-vue&id=eOi529  (VueJS)

**Video preview**
We're typing "et" and then using the autosuggested email "kundenzentrum@ethikbank.de". The data model then contains "etkundenzentrum@ethikbank.de"

https://user-images.githubusercontent.com/17496592/127144732-41886a13-06e5-4896-bec7-22332f793b33.MP4

**Environment**
 - CLI: 8.0.2
 - Cross-platform modules: 8.0.8
 - Android Runtime: 8.0.0
 - iOS Runtime: 8.0.0

janoshrubos:
## PR Checklist

On iOS if we type a few character and use the auto-fill function it will concat the strings instead of replacing to the suggested one.
Video example in the issue: https://github.com/NativeScript/NativeScript/issues/9486

## What is the new behavior?
After typing the the start of an e-mail and tap on auto-fill, it will replace the text.

**Note:**
The solution is based on the string's delta, which means if we copy and paste a text to a textfield, it will behave like we used auto-fill and replace the text in the view model.

Fixes #9486

### DIFF
--- a/packages/core/ui/text-field/index.ios.ts
+++ b/packages/core/ui/text-field/index.ios.ts
@@ -205,7 +205,8 @@ export class TextField extends TextFieldBase {
 		}
 
 		if (this.updateTextTrigger === 'textChanged') {
-			if (textField.secureTextEntry && this.firstEdit) {
+			const shouldReplaceString = (textField.secureTextEntry && this.firstEdit) || delta > 1;
+			if (shouldReplaceString) {
 				textProperty.nativeValueChange(this, replacementString);
 			} else {
 				if (range.location <= textField.text.length) {


NathanWalker (approved):

20051231:
seems this pull request breaks emoji keyboard input since emoji characters are longer than 1 in js
if user input multiple emojis in a row, only one emoji is kept since the whole string is replaced every time.

The below line is not enough to detect autofill or keyboard
```
const shouldReplaceString = (textField.secureTextEntry && this.firstEdit) || delta > 1;
```

farfromrefug:
@20051231 good catch. could you open an issue so we can follow this issue ?

JJVetter:
opened an issue since i didnt see one: https://github.com/NativeScript/NativeScript/issues/9816

PR #9563, opened 2021-09-22
fix(android/application): org.nativescript.widgets.Utils::getDrawable

### LINKED ISSUE #9562: Calling js method onConfigurationChanged failed - Error: java.lang.StringIndexOutOfBoundsException
**Environment**
 - CLI: 8.1.2
 - Cross-platform modules: 8.1.3
 - Android Runtime: 8.1.1
 - iOS Runtime: N/A
 - XCode Version: N/A
 - Plugin(s): Many (not relevant here)

**Describe the bug**

```
NativeScriptException

Calling js method onConfigurationChanged failed
Error: java.lang.StringIndexOutOfBoundsException: length=4; index=6
```

image

This bug probably happens because I may have a `null` src somewhere on an image. 

In my app, I only use distant images (on a server) with Image component, except two background-image with a static path:

```css
background-image: url("res://background");
background-image: url("res://watermark");
```

i don't think these static paths cause the problem. I am getting the report in Sentry so I don't really know what causes the problem.

**To Reproduce**

?

**Expected behavior**

NativeScript should not throw an exception but should show a log for this problem.

**Sample project**

I don't have a reproduction scenario actually

**Additional context**

I am actually making a PR which possibly fixes the issue.

sebj54:
Avoid StringIndexOutOfBoundsException in org.nativescript.widgets.Utils

As it is my first PR to this repo, I am not experimented that much with native development. Feel free to guide me so I'll now next time :)

For example, I don't know how to test this file in a NativeScript project :/

## PR Checklist

Calling org.nativescript.widgets.Utils::getDrawable with a uri shorter than 6 characters throws a StringIndexOutOfBoundsException.

## What is the new behavior?

An early check in this method ensures uri is longer than 6 characters so a warning is shown in the console.

Fixes #9562.

### DIFF
--- a/packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java
+++ b/packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java
@@ -41,7 +41,7 @@ public class Utils {
 		
 		if (uri.length() >= resPrefixLength) {
 			String resPath = uri.substring(resPrefixLength);
-		  resId = context.getResources().getIdentifier(resPath, "drawable", context.getPackageName());
+			resId = context.getResources().getIdentifier(resPath, "drawable", context.getPackageName());
 		}
 		
 		if (resId > 0) {
--- a/packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java
+++ b/packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java
@@ -38,7 +38,7 @@ public class Utils {
 	public static Drawable getDrawable(String uri, Context context){
 		int resId = 0;
 		int resPrefixLength = "res://".length();
-		
+
 		if (uri.length() > resPrefixLength) {
 			String resPath = uri.substring(resPrefixLength);
 			resId = context.getResources().getIdentifier(resPath, "drawable", context.getPackageName());
--- a/packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java
+++ b/packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java
@@ -36,8 +36,14 @@ import java.util.concurrent.Executors;
 
 public class Utils {
 	public static Drawable getDrawable(String uri, Context context){
-		String resPath = uri.substring("res://".length());
-		int resId = context.getResources().getIdentifier(resPath, "drawable", context.getPackageName());
+		int resId = 0;
+		int resPrefixLength = "res://".length();
+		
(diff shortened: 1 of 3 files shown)


packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java
```diff
@@ -36,8 +36,14 @@
 
 public class Utils {
 	public static Drawable getDrawable(String uri, Context context){
-		String resPath = uri.substring("res://".length());
-		int resId = context.getResources().getIdentifier(resPath, "drawable", context.getPackageName());
+		int resId = 0;
+		int resPrefixLength = "res://".length();
+		
+		if (uri.length() >= resPrefixLength) {
```
rigor789, line 42: I think we may want the check to be `>` since `>=` would make `restPath` an empty string in case the `uri` is `res://`

```suggestion
		if (uri.length() > resPrefixLength) {
```

rigor789:
@sebj54 thanks for taking the time to do a PR - the change look good to me, I left a minor suggestion in the review.

The way you can test this requires building the widgets, which can be done by running `npm start` ([after setting up the repo](https://github.com/NativeScript/NativeScript#contribute)) and selecting the `@nativescript/ui-mobile-base: Build for npm` option - this should build the widgets lib. Then you can run any of the test apps in the repo: ui, automated or toolbox.

For this case specifically, I think a new test-case in the `ui` app would suffice - adding a few views and setting their `background-image` to a few variants like `res://` `res://non-existent` `res://icon`

packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java
sebj54: You're right, it will be more performant like this. No need to try to load an empty path!

sebj54:
@rigor789 Thank you for your help! I added some views in the `ui` app. It runs correctly on my device but feel free to let me know if it's not exactly what was expected.

PS: The Android CI test has failed, is it expected?

NathanWalker (dismissed):

rigor789 (approved):

rigor789:
@sebj54 yeah, the CI has been flaky for some time now, still not sure why - it works locally... 

I added a few more test-cases - the last few make the app crash without the fix, so they verify that this fixes the IndexOutOfBoundsException. 

Awesome job 🎉

sebj54:
Thanks for the time you took to get me ready :)
So happy to have one commit in such an awesome project! 🙌

rigor789:
Thanks for taking the time @sebj54, hopefully sparked some interest in contributing again in the future 🎉

sebj54:
I hope to be able to do it soon :)

PR #9570, opened 2021-09-23
feat(datepicker): add ability to also show time

NathanWalker:
## PR Checklist

`DatePicker` did not allow the ability to show a time picker alongside the date. Although there is a distinct `TimePicker` in core, it was often not flexible enough in the sense of how one would often need to use a date picker with time picker.

## What is the new behavior?

`DatePicker` now supports a `showTime` property which will allow the date and the time to be set properly.
This also adds properties for hour, minute and second as well as a proper attribute for `iosPreferredDatePickerStyle` to allow easy dynamic view binding.

Further this can be useful in scenarios where [this plugin](https://docs.nativescript.org/plugins/datetimepicker.html) may not fit the exact use case given how it creates two separate native controls vs. just one with these changes.


rigor789:
This still needs the Android side done before we can merge.

Issue #9581, opened 2021-09-28
css calc inline break app


vallemar:
If you insert the width property: `calc(x - x);` in an element with inline styles the application breaks, it also happens if you do `object.nativeComponent.width = "calc(x - x)"`

Works fine if put as styles of a class.
```
.bar{
   width: calc(x - x);
}
```

**Sample project**
https://play.nativescript.org/?template=play-vue&id=mncNi9&v=2

PR #9582, opened 2021-09-28
fix: css calc inline break app

vallemar:
If you insert the width property: `calc(x - x);` in an element with inline styles the application breaks, it also happens if you do `object.nativeView.width = "calc(x - x)"`

Works fine if put as styles of a class.
```
.bar{
   width: calc(x - x);
}
```

**Sample project**
https://play.nativescript.org/?template=play-vue&id=mncNi9&v=2

Note: with NS 8.1.x it also breaks the app

Fix: https://github.com/NativeScript/NativeScript/issues/9581


NathanWalker (approved):
Excellent 👏

rigor789:
Thanks for the PR, `calc` is a CSS feature - so in this case, I'd say it's working as intended. 

We can make it work on properties like this - but I'm a little hesitant on the change due to the processing overhead of evaluating the expressions.

I would like us to run a few benchmarks, and hold off on merging this right away - some numbers would make this easier to reason about.

/cc @NathanWalker @farfromrefug

vallemar:
I agree with what you say.  I think that setting the property in the `nativeView.width = calc(x)` format might be part of the developer to calculate the size, but I also think that `inline styles` should work and currently does not.  Anyway I would like to see numbers.  How do you normally do this?

edusperoni:
Inline style should work with `view.setInlineStyle("width: calc(x)")`, does it not?

Maybe `view.style.width="calc(x)"` also should work?

I believe the distinction here is that views have different ways of setting their properties and CSS is an abstraction on top of them.

rigor789:
Looks like the 8.2.0 merge auto deleted and closed this, and I can't seem to reopen it. 

Not sure what the consensus is regarding the change, we can still discuss it!

PR #9597, opened 2021-10-06
Updated dead links

wenzdey:
I opted to use archive link for SpiderMonkey Parser_API page because I couldn't find an alternative.


NathanWalker (approved):
Excellent thank you @wenzdey !

Issue #9353, opened 2021-04-23
NS8 box-shadow and border-radius, z-index issues


kanclalg:
Hi, I was excited about NS8 box-shadow implementation, but looks like there are still few issues with it :(

1. Box-shadow does not work with border-radius.
 - on iOS it doesn't show shadow at all.
 - on Android it shows shadow with border-radius, but element itself is not rounded.
2. Strange behaviour when box-shadow is used with z-index. See screenshot

box-shadow + border-radius:
iOS:
![box-shadow-with-border-radius](https://user-images.githubusercontent.com/11705645/115871751-a142a780-a449-11eb-9bf2-23dc89088b9d.png)
Android:
![box-shadow-with-border-radius-android](https://user-images.githubusercontent.com/11705645/115872177-2ded6580-a44a-11eb-9080-1f46819c9e7c.png)

box-shadow + z-index:
iOS:
![box-shadow-with-z-index](https://user-images.githubusercontent.com/11705645/115871758-a30c6b00-a449-11eb-8af4-9782aa8e66a4.png)
Android:
![box-shadow-with-z-index-android](https://user-images.githubusercontent.com/11705645/115872193-3180ec80-a44a-11eb-98a1-9c8b6a08bb8a.png)

maxaqquepucho:
I have the same issue, is there a temporary solution for this?

AndreasKarlzzon:
+1

PR #9612, opened 2021-10-13
fix(iOS): box-shadow and border-radius

vallemar:
This PR fixes the bug on iOS with box-shadow and border-radius. In android it has been magically corrected (I don't know if someone has corrected it in another PR but it seems to work)
The problem was that `.clipsToBounds = true` was called after setting the shadow because [_setNativeClipToBounds](https://github.com/vallemar/NativeScript/blob/d75de63bb4f3239ce2e145967a0f3cf4ae12c328/packages/core/ui/core/view/index.ios.ts#L844) is being called after [iosBackground.createBackgroundUIColor](https://github.com/vallemar/NativeScript/blob/d75de63bb4f3239ce2e145967a0f3cf4ae12c328/packages/core/ui/core/view/index.ios.ts#L841) and  `_setNativeClipToBounds` does not control it if it has shadow

Fixes/Implements/Closes  https://github.com/NativeScript/NativeScript/issues/9353#.


vallemar:
I don't know if it would make more sense to leave the method more similar to the previous logic, but with the exception of having `box-shadow` and `border-radius` instead of just checking the `border-radius`:
```ts
_setNativeClipToBounds() {
		const backgroundInternal = this.style.backgroundInternal;
		const hasBoxShadowAndBorderRadius =  backgroundInternal.hasBoxShadow() && backgroundInternal.hasBorderRadius();
		this.nativeViewProtected.clipsToBounds = (this.nativeViewProtected instanceof UIScrollView || backgroundInternal.hasBorderWidth() || backgroundInternal.hasBorderRadius()) && !hasBoxShadowAndBorderRadius;
	}
```

vallemar:
@NathanWalker  Is there unit test associated with the view class? I can't find them

NathanWalker (approved):

PR #9715, opened 2021-12-25
fix(webpack5): Remove 'ts-loader' from webpack 'javascript' configuration

CatchABus:
## PR Checklist

If there is an attempt to use a plugin dynamic import inside a JavaScript app, one will receive errors about `tsconfig.json` being missing as `webpack` will use lazy loading and try to load typescript definition files inside plugin.

Specifically, it occurs when the require path itself is dynamic:
```
var suffix = "messaging";
import(`firebase-${suffix}`);
```

Perhaps, this can be solved by creating a `tsconfig.json` file but the point is that JavaScript apps should not bother with TS files and developers should not worry about TypeScript setup.
Still, plugins include TS definition files and that confuses webpack in NS JavaScript apps.

## What is the new behavior?
New behaviour will remove `ts-loader` if developers declare they use javascript configuration in webpack.

Old webpack plugin excluded `ts-loader` from `webpack.javascript.config` file but it seems that webpack5 makes sure most kinds of apps inherit it.

### DIFF
--- a/packages/webpack5/__tests__/configuration/__snapshots__/base.spec.ts.snap
+++ b/packages/webpack5/__tests__/configuration/__snapshots__/base.spec.ts.snap
@@ -90,9 +90,6 @@ exports[`base configuration for android 1`] = `
         test: [
           /\\\\.ts$/
         ],
-        exclude: [
-          /\\\\.d.ts$/
-        ],
         use: [
           /* config.module.rule('ts').use('ts-loader') */
           {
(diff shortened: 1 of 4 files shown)


rigor789:
https://github.com/NativeScript/NativeScript/blob/1601caff4e175024248a548a39755d98b175705a/packages/webpack5/src/configuration/base.ts#L229-L231

Maybe it would make sense to provide a default config in those cases - my thinking here is that js projects can still benefit from occasionally using a ts file here & there.

PR #9729, opened 2022-01-05
feat: Implemented new expression parser for NativeScript Core

DimitrisRK:
## PR Checklist

For years now, vanilla NativeScript has been providing us with its own data-binding system that can also be used inside XML which has been really helpful.

However, in additional to popular frameworks, one has very limited options for composing an expression in NativeScript Core.
For example, one cannot call any functions that are not direct properties of binding context or let's say use features from newest ECMA versions.

To parse expressions, NativeScript is using a certain package's sources inside repo: https://www.npmjs.com/package/polymer-expressions.
The `polymer-expressions` package has been abandoned for 5 years now, and Polymer library itself is in maintainance mode.
It's using version `2.0.0-dev` of `esprima` parser which is also quite old, to manage an AST Expression schema and all in all it has very limited features.

Here is a very small sample of what is the current state of expression parsing:
```xml
<Page actionBarHidden="true" xmlns="http://www.nativescript.org/tns.xsd">
	<StackLayout>
		<Label text="{{ funcOfContext() }}"/> <!-- Possible -->
		<Label text="{{ propertyOfContext.func() }}"/> <!-- Not possible -->
                <Label text="{{ funcOfContext(...prop) }}"/> <!-- Not Possible -->

		<Label text="{{ prop.test }}"/> <!-- Possible -->
		<Label text="{{ prop?.test }}"/> <!-- Not possible -->

		<Label text="{{ prop || prop2 }}"/> <!-- Possible -->
		<Label text="{{ prop ?? prop2 }}"/> <!-- Not possible -->

		<Label text="in Operator" textWrap="{{ prop in obj ? true : false }}"/> <!-- Not possible -->
		<Label text="typeof Operator" textWrap="{{ typeof prop == 'object' ? true : false }}"/> <!-- Not possible -->
	</StackLayout>
</Page>
```

## What is the new behavior?
This began as an experiment at holidays and I ended up composing a new AST parser that keeps existing behaviour and adds more features.
On top of that, it makes possible what is impossible in the XML example below.

There is a dozen of methods to parse AST schemas out there, with browserify's `static-eval` being a good option but none of them had an out-of-the-box proper support for NativeScript features and contained few bugs.

Just like before, expression parsing is done using AST pattern and parsed expression results are stored in custom parser's cache to avoid parsing over and over.

### Features

I have made sure to support lots of additional features like constructor calls, spread operator for both arrays and objects which can also be used as a function arguments, null coalescing operator, optional chaining, in & typeof operators, and even untagged template literal syntax!

The old approach did not support function and arrow lambda expressions and so does the new one. Adding support for such expressions will make things too complicated as there will be the need of statements support like variable declarations, assign operators and more.

Old mechanism used `esprima`. I started creating this patch using the same parser but I eventually switched to `acorn` since it's already used in NativeScript for webpack functionality and supports latest ecma features.

I have also made sure function calls and converters have their appropriate call context.

### Converters
https://v7.docs.nativescript.org/core-concepts/data-binding#using-converters-in-bindings

Regarding NativeScript converters, there is a small detail that makes new behaviour differ from the old:

```xml
<Page actionBarHidden="true" xmlns="http://www.nativescript.org/tns.xsd">
	<StackLayout>
                <!-- Before -->
		<Label text="{{ apple | convertAppleToJuice }}"/>
                <Label text="{{ date | formatDate('yyyy') }}"/>

                <!-- Now -->
		<Label text="{{ apple | convertAppleToJuice() }}"/> <!-- Converter has to look like a function call even without arguments -->
                <Label text="{{ date | formatDate('yyyy') }}"/>
	</StackLayout>
</Page>
```
I preferred not to waste time for adding further support for identifiers that might be converters to keep parser cleaner, not to mention it seems less confusing than before.

### Error Reporting
Regarding expression parsing errors, the old approach enclosed all logic inside a try catch which would print a not-so-specific message about a binding error. The new approach uses a separate try catch for `acorn` parse, in order to print parser errors regarding expression syntax and other stuff during parsing.

### Conclusion

This will have positive results for vanilla NS developers. There have been many times I have personally attempted to make something work in XML without much effort but failed. On top of that, this will add missing new ecma features in the XML part of NativeScript.


NathanWalker:
We will put this out with next alpha testing target of 8.2 - excellent work @DimitrisRK

NathanWalker (approved):

DimitrisRK:
> We will put this out with next alpha testing target of 8.2 - excellent work @DimitrisRK

Thank you very much! It would be great if this gets into alpha!

PS: My bad, I noticed 2 test fails in iOS. I didn't notice them due to already existing tests failing in master branch.
I'll run automated locally and try get rid of those as soon as possible.

DimitrisRK:
cla

DimitrisRK:
@cla-bot

DimitrisRK:
@cla-bot check

DimitrisRK:
There have been few fixes that also let tests to pass. No related tests are breaking now.

farfromrefug:
@DimitrisRK this is pretty awesome! and you also actually reduce N package significantly

DimitrisRK:
> @DimitrisRK this is pretty awesome! and you also actually reduce N package significantly

Thank you Martin!! I really hope this makes things easier.

Might also be the first step to get rid of outdated core `jslib` directory that contains 3rd-party package sources.
The only thing left in there is `easysax` XML parser whose sources are taken from here: https://www.npmjs.com/package/easysax
I'm positive its functionality may easily be covered using `sax` or `saxes`.

NathanWalker:
@DimitrisRK this is in `8.2.0-alpha.1` now.

PR #9731, opened 2022-01-06
fix(8.2-alpha): `Application.android` and `Application.ios` were undefined in startup

DimitrisRK:
## PR Checklist

Since application instances are now created during the call of `Application.run`, core `index.ts` of NativeScript will always export `Application.android` and `Application.ios` as undefined.
This was tested with a JavaScript app on android device.

I believe the reason is that these properties are exported inside a constant named `Application`, therefore imported properties are never updated when changed.

Here is the root of the problem:
```js
export const Application = {
	launchEvent,
	displayedEvent,
	uncaughtErrorEvent,
	discardedErrorEvent,
	suspendEvent,
	resumeEvent,
	exitEvent,
	lowMemoryEvent,
	orientationChangedEvent,
	systemAppearanceChangedEvent,
	systemAppearanceChanged,
	fontScaleChangedEvent,

	getMainEntry,
	getRootView,
	resetRootView: _resetRootView,
	getResources,
	setResources,
	setCssFileName,
	getCssFileName,
	loadAppCss,
	addCss,
	on,
	off,
	notify,
	hasListeners,
	run,
	orientation,
	getNativeApplication,
	hasLaunched,
	systemAppearance,
	setAutoSystemAppearanceChanged,

	android: appAndroid,
	ios: iosApp,
};

```
## What is the new behavior?
New behaviour will export everything from application module using ECMA2020 `export * as`.
This is not perfect as it exports everything, but the other alternative is to initialize application instances much earlier like before.

@NathanWalker This is related to changes here: https://github.com/NativeScript/NativeScript/commit/88e355d8144cff4d13c59ac83cdbc2c00fd2fe35
I'm not sure how this will affect things but `android` and `ios` application properties will be undefined before `Application.run` is called. There might be people who tend to use these properties before application runs.

Another alternative could perhaps be meddling with `export default` in application android and ios files.

### DIFF
--- a/packages/core/application/index.android.ts
+++ b/packages/core/application/index.android.ts
@@ -215,6 +215,7 @@ export function _resetRootView(entry?: NavigationEntry | string): void {
 	}
 	callbacks.resetActivityContent(activity);
 }
+export const resetRootView = _resetRootView;
 
 export function getMainEntry() {
 	return mainEntry;
--- a/packages/core/application/index.ios.ts
+++ b/packages/core/application/index.ios.ts
@@ -464,6 +464,7 @@ export function _resetRootView(entry?: NavigationEntry | string) {
 	mainEntry = typeof entry === 'string' ? { moduleName: entry } : entry;
 	iosApp.setWindowContent();
 }
+export const resetRootView = _resetRootView;
 
 export function getNativeApplication(): UIApplication {
 	return iosApp.nativeApp;
--- a/packages/core/index.ts
+++ b/packages/core/index.ts
@@ -5,44 +5,7 @@ import './globals';
 export { iOSApplication, AndroidApplication } from './application';
 export type { ApplicationEventData, LaunchEventData, OrientationChangedEventData, UnhandledErrorEventData, DiscardedErrorEventData, CssChangedEventData, LoadAppCSSEventData, AndroidActivityEventData, AndroidActivityBundleEventData, AndroidActivityRequestPermissionsEventData, AndroidActivityResultEventData, AndroidActivityNewIntentEventData, AndroidActivityBackPressedEventData, SystemAppearanceChangedEventData } from './application';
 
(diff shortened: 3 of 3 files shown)


NathanWalker:
Thanks @DimitrisRK - exporting * may work fine here, will test with it. The change is intentional to allow @nativescript/core usage within different contexts (outside of pure app usage- for example, in capacitor can use helpers from it). I believe with your change here will work just as well.

DimitrisRK:
> Thanks @DimitrisRK - exporting * may work fine here, will test with it. The change is intentional to allow @NativeScript/core usage within different contexts (outside of pure app usage- for example, in capacitor can use helpers from it). I believe with your change here will work just as well.

I see! I had to change exported `Application` constant because its export content was not updated when application was initialized so all imports would give access to `undefined` app instances. I also think this will do.

edusperoni:
One alternative could perhaps be turning those 2 into getters?

DimitrisRK:
> One alternative could perhaps be turning those 2 into getters?

@edusperoni Your idea is much better. Turning previous export like this solves issues:
```
export const Application = {
    launchEvent,
    displayedEvent,
    uncaughtErrorEvent,
    discardedErrorEvent,
    suspendEvent,
    resumeEvent,
    exitEvent,
    lowMemoryEvent,
    orientationChangedEvent,
    systemAppearanceChangedEvent,
    systemAppearanceChanged,
    fontScaleChangedEvent,
    getMainEntry,
    getRootView,
    resetRootView: _resetRootView,
    getResources,
    setResources,
    setCssFileName,
    getCssFileName,
    loadAppCss,
    addCss,
    on,
    off,
    notify,
    hasListeners,
    run,
    orientation,
    getNativeApplication,
    hasLaunched,
    systemAppearance,
    setAutoSystemAppearanceChanged,
    get android() {
        return appAndroid;
    },
    get ios() {
        return iosApp;
    }
};
```

I wonder if somebody can make a new patch for it to remove changes of this PR and add this, or I can add in one of my patches.

NathanWalker:
https://github.com/NativeScript/NativeScript/commit/411cdb5ff40eaf2dd2281d253066f9fe1de411b4

PR #9735, opened 2022-01-08
feat(ui): TouchManager for ease in adding interactivity

NathanWalker:
## PR Checklist

No easy way to add dynamic touch down/up animations for nice interactivity in the UI.

## What is the new behavior?

Introduces a `TouchManager` with an easy api to enable rich interactivity throughout the UI.
Also adds `touchAnimation` and `ignoreTouchAnimation` properties to views. This allows each view to opt into and define/control it's own touch down/up animations.

`TouchManager.enableGlobalTapAnimations = true` can also be used alongside `TouchManager.animations.{up|down}` animations to define consistent reusable animations for the entire UI by auto detecting and attaching to any view with a `tap` binding.

### DIFF
--- a/apps/automated/src/http/http-string-worker.ts
+++ b/apps/automated/src/http/http-string-worker.ts
@@ -1,13 +1,13 @@
 // todo: figure out why this worker is including the whole core and not just the Http module
 // ie. tree-shaking is not working as expected here. (same setup works in a separate app)
-import { initGlobal } from '@nativescript/core/globals/index';
-initGlobal();
+// import { initGlobal } from '@nativescript/core/globals/index';
+// initGlobal();
 
-import { Http } from '@nativescript/core';
+import { getString } from '@nativescript/core/http';
 
 declare var postMessage: any;
 
-Http.getString('https://httpbin.org/get').then(
+getString('https://httpbin.org/get').then(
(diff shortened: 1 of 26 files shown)


farfromrefug:
@NathanWalker very nice. So we agree it is disabled by default?  maybe `enableGlobalTapAnimations` and other props should be initiated to `false`.
Also i think this should disabled by default for buttons. Dont you think? It is already handled natively

NathanWalker:
Everything is disabled by default - completely opt-in. I’m writing a blog post which will break all this down, the why and how it unlocks Rich interactivity in ways that aren’t possible easily without quite bit of overhead (in case of iOS was not even possible given the inability to customize certain gestures by default without manually configuring natively).

farfromrefug:
ok. then i think variables should be initialized to ensure the code is clear. and to prevent errors with like `!!` because the commonly used`!!` will return true for undefined.

NathanWalker:
@farfromrefug all is safe as is regarding default values `undefined` is false with `!!` as expected:
```
> a = undefined
> !!a
false
```

farfromrefug:
you are right my bad about the !! test. still think it should be a forced rule to always initialize class properties. it does not cost anything and brings clarity

PR #9753, opened 2022-01-28
fix(core): pan gesture on Android

asharghi:
## PR Checklist

Note: I have not created an issue first, but have talked with Igor on Discord about this

When paning an element, the event.status should be 3 (GestureStateTypes.ended) when the paning is done. On iOS event.status 3 is only fired once when paning is done, but on Android you will receive event.status 3 even when you are not finished paning. It's because it triggers status 3 when Android fires android.view.MotionEvent.ACTION_CANCEL and the paning is not done on this event. The documentation for ACTION_CANCEL says: `The current gesture has been aborted. You will not receive any more points in it. You should treat this as an up event, but not perform any action that you normally would.`. So we should not use this.

## What is the new behavior?
Removing these lines gives us corresponding behavior on both iOS and Android for Paning statuses.

BREAKING CHANGES:
Pan status 3 (GestureStateTypes.ended) will not fire multiple times on Android

### DIFF
--- a/packages/core/ui/gestures/index.android.ts
+++ b/packages/core/ui/gestures/index.android.ts
@@ -455,10 +455,6 @@ class CustomPanGestureDetector {
 	public onTouchEvent(event: android.view.MotionEvent) {
 		switch (event.getActionMasked()) {
 			case android.view.MotionEvent.ACTION_UP:
-			case android.view.MotionEvent.ACTION_CANCEL:
-				this.trackStop(event, false);
-				break;
-
 			case android.view.MotionEvent.ACTION_DOWN:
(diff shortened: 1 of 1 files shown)


asharghi:
@cla-bot check

farfromrefug:
i think it is the wrong fix! i thinks this breaks panning outside of screen without lifting up. 
if i understand the correct fix should be to prevent duplicate event on up after cancel

NathanWalker:
Thanks for posting this @asharghi we'll post a RFC within month of February where we discussing using the forthcoming 8.2 `TouchManager` to manage scenarios of which you raise here.

asharghi:
Totally agree that's the way to go. Thanks @NathanWalker!

Issue #8485, opened 2020-04-01
Add maxLines Property to Label


wendt88:
**Is your feature request related to a problem? Please describe.**
It would be nice to add `maxLines` property to the `Label` view. It could be done with a angular/vue directive, but it's always buggy, especially inside a `ListView`

**Describe the solution you'd like**
```
if (view.android) {
        view.android.setSingleLine(false);
        view.android.setMaxLines(maxLines);
        view.android.setEllipsize(android.text.TextUtils.TruncateAt.END);
}
else if (view.ios) {
        viewl.ios.numberOfLines = maxLines;
        view.ios.adjustsFontSizeToFitWidth = false;
        view.ios.lineBreakMode = NSLineBreakMode.ByTruncatingTail;
}
```

**Additional context**
Inside a `ListView` the View gets initialized without native element. so the properties above must be set later. And that does not updates the View.

peppeg85:
hello, i'm also interested, sorry fo the cross-post, i've seen this solution, but hiw to apply in plain js?

https://gist.github.com/m-abs/ff9da8ed6c01c5c5115da37e4b535a2a
as for the other post, i have a list like this and want to truncate the notification-description class text
````
<lv:RadListView.itemTemplate>
                            <GridLayout class="notification-container" columns="100,*" rows="150" backgroundColor="#ffffff">
                                <StackLayout col="0" row="0" class="day-stack">
                                    <Label class="notification notification-start-month" text="{{ eventStartMonth }}"/>
                                    <Label class="notification notification-start-day" text="{{ eventStartDay }}"/>
                                    <Label class="notification notification-start-month" text="{{ eventStartHour }}"/>
                                </StackLayout>
                                <StackLayout col="1" row="0" class="description-stack">
                                    <Label class="notification-title" text="{{ title }}"/>
                                    <Label class="notification-description" text="{{ description }}"/>
                                </StackLayout>
                            </GridLayout>
                        </lv:RadListView.itemTemplate>
````

MrDrk94:
Any progress on this?

brianrclow:
I'm running into the same issue that @wendt88 talked about. His solution posted above does work well but I still think it would be a nice addition to the `Label` to have the ability to set `maxLines`.

rigor789:
Probably would make sense to look into bringing the following new CSS (and direct attribute) properties:
1. `line-break` - https://developer.mozilla.org/en-US/docs/Web/CSS/line-break
[...]

PR #9761, opened 2022-02-04
perf(ios): uifont and formatted string optimizations plus uiimage scaling

NathanWalker:
* optimize memory leaks around ios attributed string handling and font assignments
* improve memory leaks around uiimage scaling (needs further testing)


farfromrefug:
seems quite nice. a few remarks :
* you changed some class method names which i use to subclass. it will break most of my plugins. not the biggest of deals but asking to see if name changes are really necessary
* i like the fact that you go native but i would personaly feel like the setting of the attributed text would remain on the js side. it would remove the need for the UIView category which i think we should avoid as much as possible (could break pods in some cases like if they use the same method name)
* i like the fact that it is not part of the mobile ui lib and directly present as source code. we should do that for everything !
* finally why all those PRs on the 8.2.0? shouldn t they be done on master and master regularly merged back into 8.2.0 ? i feel.like this process kind of break master being the main dev branch.it also wouls bring a lot more clarity on the github website as what s being merge commited as master is the default branch

NathanWalker:
Thanks for feedback @farfromrefug 
* Definitely let me know which methods you may be overriding and we'll work together on those points
* Usage of extraneous Dictionaries in JS side contribute to unnecessary memory pressure thus the optimization
* We can and should do more optimizations along these lines
* It could be done against master and may still target master as there are no breaking changes here however likely no further patches were planned for publishing off master so most likely for inclusion in 8.2

farfromrefug:
@NathanWalker 
* you actually removed `setFormattedTextDecorationAndTransform` and `setTextDecorationAndTransform` which i am overriding in `@nativescript-community/ui-label`. Right it breaks it. But i would say keep it that way. I will change my override of `_setNativeText`
* on that note and related to the discussion here https://github.com/NativeScript/NativeScript/issues/8485#issuecomment-1030332036. I would like us to join the efforts in improving native attributed text creation on iOS/Android. I do this in `@nativescript-community/text`. What i am thinking is for us to find a way to have the AttrString created on the native side (like you do on iOS and like i do on Android) with a method we share with N core and my `@nativescript-community/text` but find a way to allow my plugin to "override" or "augment" it. That way i can have a faster dev cycle that N and then we can regularly merge when we fill like it. What do you think? if you agree we would have to "design" a pattern to allow us to do that.
* About the NSDict created on the native side i agree. What i am suggesting is that you create it on the native side return it to JS and apply it to the view (UILabel, UIButton,...) on the JS side. That way your `UIView` category is not needed anymore. What do you think?
* about the branch thing i understand that we are close to releasing 8.2. My point is could we agree on a workflow where as much as possible all PRs are made against master and merge in release branches when necessary? As i said it brings so much more visibility on what s happening in the repo. Right now 8.2 is like a hidden dev branch of N. Also would be easier for forks. Because right now the merge back from 8.2 to master is getting bigger and if a fork is based on master the merge would get tougher and tougher.

rigor789:
@farfromrefug on the NSDictionaries, the main issue right now is that there's some sort of memory leak in the v8 runtime (DictionaryAdapter) that we still need to find and fix...

re: branching - we've brought this up before, how master/main should always contain the latest code (released as next), but on the other hand, working on an isolated release branch allows a bit more flexibility to try new things, and easily revert them. The other thing it allows is publishing patch releases off of main/master until the release branch is ready to be released. In any case, everything is solvable - ie. if we need to release a patch to a prior minor version, we can always branch off of a specific commit (usually the `release: xy` one) and do the patch + merge back to master - but that can also get confusing/messy. I think the best way forward is to finish the 8.2 release as we are currently handling it, and re-think the approach for the next one. Perhaps even create a flow-chart in the repo describing the workflow (this would be useful in general for some of the processes we use: issue handling, feature request handling, PR handling, release handling etc.).

farfromrefug:
@rigor789 thanks for all the explanations. I understand better the move to native code. 
For the branch i understand what you mean. I think we should be careful on which one are targeting master/releases branches. This PR for example could be on master. Let s think all that together after 8.2 as you said.
Thanks

farfromrefug:
@NathanWalker i sadly just discovered that this PR broke `@nativescript-community/text` as explained in one of my above comments. Really sad seeing it was allowing plugins to add span features without actually needing the core to implement it :s
And seeing what you do with categories it is not easy to fix.  I note it here for every @nativescript-community/text users to be able to find this

farfromrefug:
@NathanWalker for reference this is how i reverted to fix. https://github.com/Akylas/NativeScript/commit/e3284a4a8aaebb19e44498b2c4663dcceb186969 
Not perfect but no easy way around it

PR #9777, opened 2022-02-16
fix: UIImage leaking after Image is disposed

rigor789:
## PR Checklist

Screenshot 2022-02-16 at 11 12 33 PM

## What is the new behavior?

Screenshot 2022-02-16 at 11 14 43 PM

The above screenshots are produced by running the following test code, navigating to the "Show Image" page, and back.

```js
const {
  Application,
  Frame,
  Page,
  Button,
  Image,
} = require("@nativescript/core");

let count = 0;

const makeModalPage = () => {
  const page = new Page();
  page.id = `page${count++}`;

  const image = new Image();

  image.src =
    "https://i.picsum.photos/id/669/5000/5000.jpg?hmac=VlpchW0ODhflKm0SKOYQrc8qysLWbqKmDS1MGT9apAc";

  page.content = image;

  return page;
};

Application.run({
  create() {
    const rootFrame = new Frame();
    rootFrame.navigate({
      create: () => {
        const page = new Page();
        const button = new Button();

        button.text = "Open Image";
        button.on("tap", () => {
          rootFrame.navigate({
            create() {
              return makeModalPage();
            },
          });
        });

        page.content = button;

        return page;
      },
    });

    return rootFrame;
  },
});
```

### DIFF
--- a/packages/core/ui/image/index.ios.ts
+++ b/packages/core/ui/image/index.ios.ts
@@ -24,6 +24,21 @@ export class Image extends ImageBase {
 		this._setNativeClipToBounds();
 	}
 
+	public disposeNativeView(): void {
+		super.disposeNativeView();
+
+		if (this.imageSource?.ios) {
+			// release the native UIImage
+			CFRelease(this.imageSource.ios);
+		}
+
+		this.imageSource = null;
+
+		if (this.nativeViewProtected?.image) {
+			this.nativeViewProtected.image = null;
+		}
+	}
+
 	private setTintColor(value: Color) {
 		if (value && this.nativeViewProtected.image && !this._templateImageWasCreated) {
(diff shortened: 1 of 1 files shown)


NathanWalker (approved):

Archez:
@rigor789 Should we apply the same fix when updating the image source property? Looks like we just null out any existing image source before creating the replacement.

rigor789:
@Archez good point - yes absolutely, probably makes sense to create a getter/setter setup on `imageSource` and handle it in the setter, so we can just `= null` it and it would do the release.

And for clarity, we did discover this change causes some crashes, and will be addressing them before this goes into the final release.

The crashes were related to accessing the freed memory after it's been released, and my hunch is that it's related to the async queue of the image decoding, but need to dig a bit deeper to find.

PR #9661, opened 2021-11-06
chore: draft at using new API to replace java.io.File

farfromrefug:
This is a draft PR at introducing new APIs for Android 11.
With Android 11 `java.io.File` is rendered useless because of permissions.

This PR replaces `java.io.File` in most places.
It also moves most file system logics to the Java side to improve perfomances.

The work is not finished yet:
* i implemented a new `org.nativescript.widgets.Utils.getFileStats` methods to get all stats for a `File` or a `Folder`. Still needs to be plugged on the JS side
* need to implement `enumEntities` method on the Java side (i have the code).
* need to implement removeFile/createFile/mkdirs .... For that i have no idea yet.

I also moved the mobile-ui-base android project to gradle 7.0 with latest features. Thus the changes in java files.

Any help on this is welcome. We need this ASAP

PR #9807, opened 2022-03-04
feat(android):add content uri support for File

triniwiz:
Allows the `File` class to handle `content://` uris on android


farfromrefug:
@triniwiz i feel like the Android File implementation is a bit dirty being done the way it is . Why not in [file-system-access.android.ts](https://github.com/NativeScript/NativeScript/blob/04a12337ea364bed344dbfa2c92c7391024e67a5/packages/core/file-system/file-system-access.android.ts) ? It would be much cleaner and prevent the need for the `global.isAndroid`

Also is your code working with a file picked through `Intent.ACTION_OPEN_DOCUMENT_TREE` or `Intent.ACTION_OPEN_DOCUMENT` ? Seems weird that it would work without using DocumentContract.

Also it is a bit sad that your rewrite a way to read text in FileHelper and not use the same way in https://github.com/NativeScript/NativeScript/blob/master/packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Async.java
My implementation here https://github.com/NativeScript/NativeScript/pull/9661 had the benefit of refactoring code and not have duplicated functions.
May i ask you decided not to build on top? It even had a native test app to run automatic test on the native impl

triniwiz:
@farfromrefug this is a temp what I'm really working on is much larger that this , also I just noticed that read from the async method an it's a bit cleaner now

triniwiz:
I don't see why it won't work with files picked with those , I believe the content resolver should be fine with those.

NathanWalker (approved):

PR #9808, opened 2022-03-04
feat(config): cli.additionalPathsToClean to clean other paths with 'ns clean'

Archez:
## PR Checklist

## What is the new behavior?

Add typing info for the existing `additionalPathsToClean` property that `ns clean` reads from.

### DIFF
--- a/packages/core/config/config.interface.ts
+++ b/packages/core/config/config.interface.ts
@@ -100,6 +100,11 @@ interface IConfigCLI {
 	 * Defaults to the CLI set package manager, or `npm` if not set globally
 	 */
 	packageManager: 'yarn' | 'pnpm' | 'npm';
+	
+	/**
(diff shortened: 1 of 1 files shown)


rigor789:
Thanks @Archez! I do wonder if we should add this under the `cli` key (and update cli to match)?

Archez:
@rigor789 I was thinking the same when adding it. Wasn't sure if that would be considered a breaking change, but I can make an update so the top level config isn't bloated.

rigor789:
@Archez it would be fine - this was never a documented feature (I added it experimentally, but we can totally move it to the `cli` key)

NathanWalker:
@rigor789 @Archez I moved it under `cli` key as completely agree makes sense there and super useful! Needed something like this recently. Will merge to include with 8.2 and @rigor789 we can adjust this is cli to look for it in the `cli` key prior to release.

PR #9809, opened 2022-03-04
feat(types-ios): reduced ios types to common types for optimized typescript resolution

NathanWalker:
Optimize iOS types by more than 50% from >20,000 declarations to ~8,000 vastly improving TypeScript editors.

If you are a plugin developer and/or just need the entire full set you can include all >20,000 with the following:

```
/// <reference path="@nativescript/types-ios/src/complete.d.ts" />
```


farfromrefug:
@NathanWalker awesome!

PR #9824, opened 2022-03-13
feat(core): An iterable ObservableArray

CatchABus:
## PR Checklist

ObservableArray works as an object wrapper that supports some Array methods.
Additionally, methods like `filter` or `map` will return the nested array itself which reduces the worth of wrapper itself.

## What is the new behavior?
ObservableArray becomes iterable using `Symbol.iterator` symbol. As a result, it becomes compliant with spread operator and can be used as an argument in `Array.from` and `Array.prototype.concat` methods.
Also, some changes have been made to make ObservableArray more usable.

### Negative indexing
Since we make use of `getItem` and `setItem` methods for indexing, it's a good chance to add more advanced functionality.
These methods will now support negative indexing, similar to `Array.prototype.at`.

For instance:
```js
var obsArray = new ObservableArray(1, 2, 3);
var item = obsArray.getItem(-1); // Get last item
obsArray.setItem(-1, item + 1); // Set new last item
```

### ObservableArray as iterable
There are many APIs that will accept ObservableArray as an argument. Some examples include:
[new Map([iterable])](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)
[new WeakMap([iterable])](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap)
[new Set([iterable])](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)
[new WeakSet([iterable])](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet)
[Promise.all(iterable)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all)
[Promise.race(iterable)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race)
[Array.from(iterable)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from)

### ObservableArray to JSON
One can encode an ObservableArray and receive an encoded array content as a result.

### ObservableArray to Array
One can get a plain array copy of ObservableArray using the following methods:
```js
var obsArray = new ObservableArray(1, 2, 3);
var arr1 = Array.from(obsArray);
var arr2 = [...obsArray];
var arr3 = [].concat(obsArray);
```

### Iterating ObservableArray using for...of
One can now iterate an ObservableArray using `for...of` syntax and `await` keyword is also supported out of the box.
```js
var obsArray = new ObservableArray(1, 2, 3);
for (let num of obsArray) {
    console.log(num);
}
```

### Methods
Regarding ObservableArray's methods, the following changes have been made:
- Added `includes` and `find` methods
- Methods `indexOf` and `lastIndexOf` will use the built-in methods of nested Array
- Method `push` will not spread array-like arguments anymore. This is very confusing (for example, one wants to populate an observable array with arrays but push will force-spread the array argument) and does not follow the same functionality as `Array.prototype.push`. One can use spread operator to push items of another array into `ObservableArray`
- Methods that returned a modified array (in-case concept) will return modified `ObservableArray` itself
- Methods that returned a new array will return a new `ObservableArray` containing needed elements
- Callbacks that accepted nested array as last argument will now accept ObservableArray itself as last argument (e.g. callback of forEach method).

Samples:
```js
var obsArray = new ObservableArray(1, 2, 3);
obsArray.includes(3); // true
obsArray.find(item => item.value > 1); // Will return first item that is greater than 1
obsArray.sort(); // Will return modified ObservableArray
obsArray.filter(item => item.value == 1); // Will return modified ObservableArray

obsArray.forEach((value, index, observableArray) => console.log(value)); // Third argument will be observableArray itself
```

## BREAKING CHANGES:

- Method `push` will now handle arguments just like `Array.prototype.push`
- Certain existing methods will now return `ObservableArray` instance instead.
- Callback arguments that contained an array argument themselves will now contain an ObservableArray argument.

### DIFF
--- a/apps/automated/src/data/observable-array-tests.ts
+++ b/apps/automated/src/data/observable-array-tests.ts
@@ -89,7 +89,7 @@ export const test_ObservableArray_concatShouldReturnNewArrayWithNewItemsAtTheEnd
 	const array = new ObservableArray([1, 2, 3]);
 	const result = array.concat([4, 5, 6]);
 	// << observable-array-combine
-	TKUnit.assert(result.length === 6 && result[4] === 5, 'ObservableArray concat() should add items at the end!');
+	TKUnit.assert(result.length === 6 && result.getItem(4) === 5, 'ObservableArray concat() should add items at the end!');
 };
 
 export const test_ObservableArray_joinShouldReturnStringWithAllItemsSeparatedWithComma = function () {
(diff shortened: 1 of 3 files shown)


DimitrisRK:
After few tests, I marked it as ready.
Anyone is welcome to test scenarios or suggest things.

facetious:
What is the benefit provided by this breaking change? Existing plugins are failing to operate correctly using this implementation.

CatchABus:
> What is the benefit provided by this breaking change? Existing plugins are failing to operate correctly using this implementation.

Mostly, it's to match the API of common Arrays as much as possible. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push

Also, an ObservableArray can consist of items that are actually arrays. ObservableArray.prototype.push didn't handle this very well.

Eventually, we want to make ObservableArray feel more familiar to users.

PR #9834, opened 2022-03-19
fix: Close popup views on live-sync

DimitrisRK:
## PR Checklist

RootLayout popup views will not update or close during live-sync.

## What is the new behavior?
RootLayout popup views will close during live-sync.

During my work on this, I took the liberty to do few more important improvements.

1) Method `closeAll` of `RootLayout` will now use Promise.all so that developers can await for all simultaneous popup dismissals.

2) Fixed a small issue that caused live-sync not to reload changes in current page if a `Modal` was shown atop of it. This was a minor bug that occured during early NativeScript 6 updates.

3) Regarding error handling, promise-based methods of `RootLayout` have also gotten few improvements.
The old approach was using try clauses inside promises which caused a promise never to reject in case of error.
There were few trace messages in case trace was enabled but that still didn't handle the errors at all.
These clauses were removed as errors will automatically reject the promise they are thrown into.

4) Validation check for given view was also added in case view was not a valid instance (e.g. undefined).

I believe `RootLayout` is in need of few improvements and this patch is a good start.

### DIFF
--- a/packages/core/ui/core/view/view-common.ts
+++ b/packages/core/ui/core/view/view-common.ts
@@ -191,31 +191,25 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
 			Trace.write(`${this}._onLivesync(${JSON.stringify(context)})`, Trace.categories.Livesync);
 		}
 
-		let hasPopupView = false;
+		let handled = false;
 
 		const rootLayout = global.rootLayout;
 		if (rootLayout != null && rootLayout.topmost() != null) {
 			rootLayout.closeAll();
-			hasPopupView = true;
+			handled = true;
 		}
 
 		if (this._closeAllModalViewsInternal()) {
-			hasPopupView = true;
-		}
-
-		if (hasPopupView) {
-			return true;
+			handled = true;
 		}
 
 		if (this._handleLivesync(context)) {
 			return true;
 		}
 
-		let handled = false;
 		this.eachChildView((child: ViewCommon) => {
 			if (child._onLivesync(context)) {
 				handled = true;
-
 				return false;
 			}
 		});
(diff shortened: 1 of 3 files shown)


NathanWalker:
@DimitrisRK this is awesome thank you! One small comment, do you mind keeping the rejects instead of throw?
Generally in mobile dev of this nature a throw will disrupt the flow of dev more than a reject because even though those conditions are invalid, a reject with a trace message would be better developer experience because they can see issue in the log however does not end up stopping their development right away (mobile apps don’t have a refresh button like browsers) so better to let dev know through a reject in this environment vs a throw which can cause dev to stop and in some cases dev would need to relaunch their runner to continue.

DimitrisRK:
> @DimitrisRK this is awesome thank you! One small comment, do you mind keeping the rejects instead of throw? Generally in mobile dev of this nature a throw will disrupt the flow of dev more than a reject because even though those conditions are invalid, a reject with a trace message would be better developer experience because they can see issue in the log however does not end up stopping their development right away (mobile apps don’t have a refresh button like browsers) so better to let dev know through a reject in this environment vs a throw which can cause dev to stop and in some cases dev would need to relaunch their runner to continue.

Thank you very much for pointing out!
Then, how about a reject with an error as a parameter?
Like `reject(new Error("My error message"))`?
This will more or less look like the rejection approach in HTTP client requests.

Or we could replace those throws with a trace message along with a plain `reject()`. Like:
```js
if (Trace.isEnabled()) {
    Trace.write("My error message");
}
reject();
```

NathanWalker:
Good question, I think `reject(new Error("My error message"));` is good in this case, won't need to enable Trace for that and would be more expected handling.

NathanWalker (approved):

PR #9872, opened 2022-04-14
perf(webpack): Attempt to resolve namespaces before module names

CatchABus:
## PR Checklist

XML namespace loader attempts to resolve module paths based on element name before namespaces.

## What is the new behavior?
XML namespace loader attempts to resolve module paths based on element name after namespaces.

I have a directory with an index.js that exports all my components. In some cases, loader attempted to resolve a path based on element name and this results in case-sensitivity problems.

For example, my index.js exported `Select` component from `./select.js`. XML loader attempts to load a module using `mypath/Select.js` as a module name. Method `resolve` happens to be case insensitive in MacOs, and results in conflict between Select.js and select.js.

This change in order helps get rid of such issues as loader will attempt to load namespace as a module first.

### DIFF
--- a/packages/webpack5/src/loaders/xml-namespace-loader/index.ts
+++ b/packages/webpack5/src/loaders/xml-namespace-loader/index.ts
@@ -72,13 +72,14 @@ async function parseXML(content: string): Promise<ParseResult> {
 		const localNamespacePath = join(this.rootContext, namespace);
 		const localModulePath = join(localNamespacePath, elementName);
 
+		// Check namespaces before module names as a folder may contain an index file with exports
 		const resolvePaths = [
 			localNamespacePath,
 			localModulePath,
-			moduleName,
 			namespace,
-			`~/${moduleName}`,
+			moduleName,
 			`~/${namespace}`,
+			`~/${moduleName}`,
 		];
 
 		// fallbacks for codeless namespaces


NathanWalker:
@DimitrisRK do you see any potential breaking changes with this?

DimitrisRK:
> @DimitrisRK do you see any potential breaking changes with this?

I can try few different scenarios in my app at the weekend, though it's not supposed to break existing behaviour.
I'm going to mark this as draft and let you know when I'm done!

rigor789:
I'm a bit worried about this change, do the tests pass after the changes? Can we add a test case to check for correctness in the case you are describing?

https://github.com/NativeScript/NativeScript/blob/ee58612c3c97fc4ea107893f84b3ba839449dab0/packages/webpack5/__tests__/loaders/xml-namespace-loader.spec.ts

DimitrisRK:
> I'm a bit worried about this change, do the tests pass after the changes? Can we add a test case to check for correctness in the case you are describing?
> 
> https://github.com/NativeScript/NativeScript/blob/ee58612c3c97fc4ea107893f84b3ba839449dab0/packages/webpack5/__tests__/loaders/xml-namespace-loader.spec.ts

I wanted to, but it seems unit tests won't run in webpack5. Tests are trying to resolve dependencies that are not available in package.json. They are referenced as dependencies in webpack 4.1 though, so I guess few things are still outdated.

rigor789:
@DimitrisRK hmm that's odd - been keeping the test up-to-date, if you try: 
```bash
cd packages/webpack5
yarn test --watch
```

Does it work? If not, what are the errors? Last time I ran them, they were all green

DimitrisRK:
> @DimitrisRK hmm that's odd - been keeping the test up-to-date, if you try:
> 
> ```shell
> cd packages/webpack5
> yarn test --watch
> ```
> 
> Does it work? If not, what are the errors? Last time I ran them, they were all green

I'm sorry, it seems the problem was on my side after all.
It looks like all existing tests pass. I'll try to think of a good way to test this case!

CatchABus:
> @DimitrisRK hmm that's odd - been keeping the test up-to-date, if you try:
> 
> ```shell
> cd packages/webpack5
> yarn test --watch
> ```
> 
> Does it work? If not, what are the errors? Last time I ran them, they were all green

I'm not sure how to create a test for this since the problem is related to case-sensitivity of `require` in mac OS.
Everything is good on linux.

I receive dozens of warnings similar to this:
```
WARNING in ./app/components/script/Select.js
There are multiple modules with names that only differ in casing.
This can lead to unexpected behavior when compiling on a filesystem with other case-semantic.
Use equal casing. Compare these module identifiers:
* /Users/mypc/monorepo/clients/mobile/node_modules/@nativescript/webpack/dist/loaders/nativescript-worker-loader/index.js!/Users/mypc/monorepo/clients/mobile/node_modules/@nativescript/webpack/dist/loaders/nativescript-hot-loader/index.js??ruleSet[1].rules[2].use[0]!/Users/mypc/monorepo/clients/mobile/app/components/script/Select.js
    Used by 2 module(s), i. e.
    /Users/mypc/monorepo/clients/mobile/node_modules/@nativescript/webpack/dist/loaders/xml-namespace-loader/index.js!/Users/mypc/repo42/clients/mobile/app/views/profile/profile.xml
* /Users/mypc/monorepo/clients/mobile/node_modules/@nativescript/webpack/dist/loaders/nativescript-worker-loader/index.js!/Users/mypc/monorepo/clients/mobile/node_modules/@nativescript/webpack/dist/loaders/nativescript-hot-loader/index.js??ruleSet[1].rules[2].use[0]!/Users/mypc/monorepo/clients/mobile/app/components/script/select.js
    Used by 3 module(s), i. e.
    /Users/mypc/monorepo/clients/mobile/app|sync|/.(xml%7Cjs%7Cs?css)$/
 @ ./app/views/profile/profile.xml 3:7-88 19:7-88
 @ ./app/ sync .(xml%7Cjs%7Cs?css)$ ./views/profile/profile.xml
 @ ./app/__@nativescript_webpack_virtual_entry_javascript__ 3:16-89
```

So I got into this problem because of the following structure:
- components (folder)
   - select.js
   - index.js (this file will export class `Select` from select.js among with other component stuff
   - and more

My xml namespace is "components". XML loader will first attempt to load modules using module path (namespace/elementName.extension), which in my case will be components/Select.js. Since require has case-sensitivity problems in mac OS, loader will attempt to load `components/select.js` despite the difference in casing, resulting in warnings.

Attempting to load namespaces first seems to solve this issue. On top of that, it doesn't seem to break anything.

PR #9875, opened 2022-04-15
fix(ios): Do not call setNeedsLayout() when another layout update request is in process

CatchABus:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/tools/notes/CONTRIBUTING.md#commit-messages.
- [ ] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/tools/notes/DevelopmentWorkflow.md#running-unit-tests-application.
- [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/tools/notes/WritingUnitTests.md.

## What is the current behavior?
There are cases that a layout contains children that often get updated because of bindings.
In my case, it was a list of GridLayouts with nested images, the visibility of which was changing through user interactivity.

Upon 2-3 uses, my app was crashing with an assertion failed message:
```sh
====== Assertion failed ======
Native stack trace:
1          0x103a1e7d8 tns::Assert(bool, v8::Isolate*) + 128
2          0x1039a6848 tns::ArgConverter::Invoke(v8::Local<v8::Context>, objc_class*, v8::Local<v8::Object>, tns::V8Args&, tns::MethodMeta const*, bool) + 100
3          0x1039f9658 tns::MetadataBuilder::InvokeMethod(v8::Local<v8::Context>, tns::MethodMeta const*, v8::Local<v8::Object>, tns::V8Args&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, bool) + 88
4          0x1039f8ec0 tns::MetadataBuilder::MethodCallback(v8::FunctionCallbackInfo<v8::Value> const&) + 220
5          0x103b300e4 v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo) + 544
[...]

Issue #9820, opened 2022-03-10
NS 8.2.0+ iOS Base64 images disappearing after scrolling ListView/RadListView


felixkrautschuk:
### Issue Description

When upgrading @nativescript/core from 8.1.5 to latest 8.2.1 (or 8.2.0), I notice that images inside ListView/RadListView are disappearing after scrolling when using base64 ImageSources on iOS. They appear again when removing the items from ObservableArray and add them again.

```
<ListView items="{{ items }}">
            <ListView.itemTemplate>
                <GridLayout columns="auto,20,*>
                    <Image col="0" src="{{ imageSrc }}" className="image-list"/>
                    <Label col="2" text="{{ title }}" verticalAlignment="center"/>
                </GridLayout>
            </ListView.itemTemplate>
</ListView>
```

```
export const vm = fromObject({
  items: new ObservableArray([
      {title: "Item 1", imageSrc: ImageSource.fromBase64Sync("/9j/4AAQSkZJRgABAQAASABIAAD/4QCMRXhpZg...")},
      {title: "Item 2", imageSrc: ImageSource.fromBase64Sync("/9j/4AAQSkZJRgABAQEBLAEsAAD/7QJoUGhvdG...")},
      {title: "Item 3", imageSrc: null},
      //...
  ])
});
```

The output:
https://user-images.githubusercontent.com/6443021/157667068-8effbe14-f17d-4360-ac4b-f8d1f2157712.mov

With nativescript/core 8.1.5, it was working as expected:

https://user-images.githubusercontent.com/6443021/157667249-c57d8be4-3b16-4e3f-9238-6efe862a557c.mov

### Reproduction

Sample app to reproduce:
[ns-8-2-ios-image.zip](https://github.com/NativeScript/NativeScript/files/8223567/ns-8-2-ios-image.zip)

- run the app on ios
- tap the first button to navigate to the page with RadListView and base64 images
- scroll the list down and up again -> images disappearing

### Relevant log output (if applicable)

```shell
-
```

### Environment

```yaml
OS: macOS 12.2.1
CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
Shell: /bin/zsh
node: 14.18.1
npm: 6.14.15
nativescript: 8.1.5

# android
java: 11.0.11
ndk: Not Found
apis: Not Found
build_tools: Not Found
system_images: Not Found

# ios
xcode: 13.2.1/13C100
cocoapods: 1.11.2
[...]

PR #9671, opened 2021-11-17
fix: ensure we clear ios view.

farfromrefug:
This has been like this for a long time without explanation. And it has side effects like here https://github.com/NativeScript/NativeScript/issues/7469.
It does not seem to have any side effects from my tests.

I open this as draft right now but we should be able to release this

PR #9879, opened 2022-04-23
fix(ios): proper disposal and recreation of iOS native views

CatchABus:
## PR Checklist

Native iOS views are not disposed and cannot be recreated when their view instances are re-added to the view tree.

### Native View Disposal

This PR was initially created for the sole purpose of destroying old native views (until more problems showed up in the process).
This is based on Martin's @farfromrefug  PR #9671 which was reverted because of few problems.

@rigor789 If I'm not mistaken, the most common error case was this one:
```
***** Fatal JavaScript exception - application has been terminated. *****
NativeScript encountered a fatal error: Uncaught TypeError: Cannot set property 'delegate' of null
at
onUnloaded(file: src/webpack:/node_modules/@nativescript/core/ui/text-field/index.ios.js:114:0)
at (file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:321:0)
at callFunctionWithSuper(file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:312:0)
at callUnloaded(file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:321:0)
at unloadView(file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:466:0)
at (file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:249:0)
at eachChildView(file: src/webpack:/node_modules/@nativescript/core/ui/layouts/layout-base-common.js:101:0)
at eachChild(file: src/webpack:/node_modules/@nativescript/core/ui/core/view/view-common.js:772:0)
at onUnloaded(file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:248:0)
at (file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:321:0)
at callFunctionWithSuper(file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:312:0)
at callUnloaded(file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:321:0)
at unloadView(file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:466:0)
at (file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:249:0)
at eachChildView(file: src/webpack:/node_modules/@nativescript/core/ui/layouts/layout-base-common.js:101:0)
at eachChild(file: src/webpack:/node_modules/@nativescript/core/ui/core/view/view-common.js:772:0)
at onUnloaded(file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:248:0)
at (file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:321:0)
at callFunctionWithSuper(file: src/webpack:/node_modules/@nativescript/core/ui/core/view-base/index.js:312:0)
at callUnloaded(file:///app/vend<…>
```

After a bit of research, I concluded the root of this cause is `_tearDownUI` calls being called before the view gets unloaded.
So, what are the cases for a view to get unloaded too late when it's already destroyed?

One case is view controllers which have a `viewDidDisappear` method that is called once owner view is disposed. This method will call `callUnloaded` which will attempt to unload view and all its children in view tree. Let's keep in mind that if a view is disposed by means other than `removeChild`, it's still in view hierarchy.

One primary suspect for this crime is method `destroyNode` in `View` class. I'm not familiar with NS Angular, but its renderer seems to wrap this method inside another `destroyNode`. One of errors mentioned in #9671 has also occured in an Angular app.

I couldn't locate any other error-prone cases and I'm not completely sure all errors have occured because of the flaw in that certain method, but for start I added a missing `callUnloaded` call inside `destroyNode`.
If there are more causes that trigger this issue, I believe we can track them upon further app usage and solve them immediately, since we know what can cause this now. Core has got a dozen of `_tearDownUI` calls that are supposed to get called after view is unloaded but who knows!

Just as @farfromrefug pointed out, perhaps we should make sure those two calls are done in proper order in the future, like adding `callUnloaded` inside `_tearDownUI` for example which needs some refactoring.

So, all these details refer to https://github.com/NativeScript/NativeScript/pull/9879/files#diff-86ce61f77399a980c83c5a14ee2ac54b8c672d375d0280597c217eb17811a076, the change for disposing native views for iOS.

### Re-using iOS view instance (non-reusable native view)

What if one wants to reuse the same NS instance even though its native view is not reusable? Upon re-adding an already removed view instance to the view tree, it should generate a new native view. This leads us to new problems which I'll use an example to describe below.

I'll use a non-reusable animated drawer view of mine embedded in `RootLayout`, in a plain NS app. This is a good example since there is a constant use of `insertChild` and `removeChild` methods, and view will also get transformed it gets hidden to the left side of the screen.

In short, attempting to open drawer will result in calling `insertChild`, while trying to close it will result in calling `removeChild`.

This is how my drawer originally looks like when opening it for first time:
image

If it's removed and re-added to view tree in current NS core version, it will look like this:
image

It looks like this because images undergo a cleanup on disposal since NS 8.2. As view is not currently disposed properly, native view will not be created anew and this results in images not being reloaded.

Now, the following screenshot displays the result of using changes from PR #9671 alone.
image

View layout seems completely messed up and that is so confusing. This problem occured because view instances keep cache of last used native frame.
And here comes the addition of `disposeNativeView` in the iOS version of `View` class: https://github.com/NativeScript/NativeScript/pull/9879/files#diff-025d8125f2c2388468769424324a18b7a896921f7b44f720db8014eb653b5f84R67

This comes with really good results:
image

Yet, it seems few children are not aligned properly, and there's no good explanation.
I concluded that the cause of this is the native view `translateX` transform occuring too early. That transform messes up view position in native window, resulting in padding and inset miscalculations.

That is why we should prevent transforms from happening when layout is still invalid, and apply them once view content layout is ready.
See here: https://github.com/NativeScript/NativeScript/pull/9879/files#diff-025d8125f2c2388468769424324a18b7a896921f7b44f720db8014eb653b5f84R132

The final result:
image

The proper disposal of native views works quite well with the few recent additions of GC inside core.
About reuse, I believe this is a really good step for re-using iOS views even when their native instances are not reusable. This has always been possible in android, but it seems we lacked the proper view cleanup to make this work in iOS and that created a bit of confusion.

This PR is a possible candidate for fixing the iOS part of #9820, and its view cleanup is a good solution for my issue described in #9875

### DIFF
--- a/apps/automated/src/ui/lifecycle/lifecycle-tests.ts
+++ b/apps/automated/src/ui/lifecycle/lifecycle-tests.ts
@@ -1,7 +1,7 @@
 import * as helper from '../../ui-helper';
 import * as btnCounter from './pages/button-counter';
 import * as TKUnit from '../../tk-unit';
-import { isIOS, isAndroid } from '@nativescript/core';
+import { isIOS } from '@nativescript/core';
 
 // Integration tests that asser sertain runtime behavior, lifecycle events atc.
 
@@ -163,7 +163,7 @@ export function test_css_sets_properties() {
 	page.content = stack;
 
 	// TODO: The check counts here should be the same as the counts before removing from the page.
-	const expectedNativeSettersAfterReaddedToPage = isAndroid ? [2, 4, 4, 4] : expectedChangesAfterResettingClasses;
+	const expectedNativeSettersAfterReaddedToPage = [2, 4, 4, 4];
 	for (let i = 0; i < buttons.length; i++) {
 		TKUnit.assertEqual(buttons[i].colorSetNativeCount, expectedNativeSettersAfterReaddedToPage[i], `Expected ${buttons[i].id} native set to not be called when added to page.`);
 		TKUnit.assertEqual(buttons[i].colorPropertyChangeCount, expectedChangesAfterResettingClasses[i], `Expected ${buttons[i].id} change notifications for css properties to not occur when added to page.`);
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@nativescript/core",
-  "version": "8.2.3",
+  "version": "8.2.4-alpha.0",
   "description": "A JavaScript library providing an easy to use api for interacting with iOS and Android platform APIs.",
   "main": "index",
   "types": "index.d.ts",
--- a/packages/core/ui/animation/animation-interfaces.ts
+++ b/packages/core/ui/animation/animation-interfaces.ts
@@ -69,5 +69,5 @@ export interface AnimationDefinitionInternal extends AnimationDefinition {
 export interface IOSView extends View {
 	_suspendPresentationLayerUpdates();
 	_resumePresentationLayerUpdates();
-	_isPresentationLayerUpdateSuspeneded();
+	_isPresentationLayerUpdateSuspended();
 }
--- a/packages/core/ui/core/view-base/index.ts
+++ b/packages/core/ui/core/view-base/index.ts
@@ -909,6 +909,7 @@ export abstract class ViewBase extends Observable implements ViewBaseDefinition 
 
 	public destroyNode(forceDestroyChildren?: boolean): void {
 		this.reusable = false;
+		this.callUnloaded();
 		this._tearDownUI(forceDestroyChildren);
 	}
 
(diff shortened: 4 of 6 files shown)


CatchABus:
@cla-bot check

PR #9881, opened 2022-04-26
fix: Apply proper border radius to box shadow and view sublayers

### LINKED ISSUE #9869: [Box shadow iOS] combining linear-gradient background, border-radius and box-shadow
### Issue Description

When combining box-shadow, border-radius and linear-gradient as a background on iOS, it shows the element as a rectangle with shadow. When the background is a solid color, it works normally.

Box-shadow + border-radius + linear-gradient background:
Simulator Screen Shot - iPhone 13 Pro Max - 2022-04-13 at 11 17 07

Box-shadow + border-radius + solid color background:
Simulator Screen Shot - iPhone 13 Pro Max - 2022-04-13 at 11 17 38

### Reproduction

style an element like this on iOS:
```
element {
    background: linear-gradient(#fafafa, #ccc);
    box-shadow: 0 0 6px rgba(0, 0, 0, 0.3);
    border-radius: 24;
}
```

_No response_

### Environment

```yaml
OS: macOS 12.1
CPU: (8) arm64 Apple M1 Pro
Shell: /bin/zsh
node: 14.18.2
npm: 8.1.4
nativescript: 8.2.2

# android
java: 11.0.14
ndk: Not Found
apis: 28, 29, 30, 31
build_tools: 30.0.2, 30.0.3
system_images: 
  - android-23 | Google APIs Intel x86 Atom
  - android-25 | Google APIs ARM 64 v8a
  - android-29 | Google APIs ARM 64 v8a
  - android-30 | Google Play ARM 64 v8a
  - android-31 | Google APIs ARM 64 v8a

# ios
xcode: 13.2.1/13C100
cocoapods: 1.11.2
python: 2.7.18
python3: 3.8.9
ruby: 2.6.8
platforms: 
  - DriverKit 21.2
  - iOS 15.2
  - macOS 12.1
  - tvOS 15.2
  - watchOS 8.3
```

### Dependencies

```json
"dependencies": {
  "@nativescript/core": "~8.2.0",
  "@nativescript/theme": "~3.0.2",
  "axios": "^0.26.1",
  "lodash": "^4.17.15",
  "moment": "^2.26.0",
  "nativescript-vue": "~2.9.1",
  "vuex": "3.1.2"
},
"devDependencies": {
  "@nativescript/android": "~8.2.0",
  "@nativescript/ios": "~8.2.0",
  "@nativescript/webpack": "~5.0.0",
  "@types/jest": "^26.0.7",
  "@types/node": "^14.0.27",
  "@vue/devtools": "^5.3.3",
  "chalk": "^4.1.0",
  "dotenv": "^8.2.0",
  "husky": "^4.2.5",
  "jest": "^26.1.0",
  "lazy": "1.0.11",
  "nativescript-vue-devtools": "^1.4.0",
  "nativescript-vue-template-compiler": "~2.9.0",
  "sass": "~1.39.0",
  "source-map": "^0.7.3",
  "superagent": "^6.1.0",
  "ts-jest": "^26.1.4",
  "tslib": "^2.1.0",
  "typescript": "~4.3.5",
  "xml-js": "^1.6.11"
}
```

### Please accept these terms

- [X] This is a bug report

CatchABus:
## PR Checklist

In iOS:
- Border radius is miscalculated for drop-shadow
- Gradient background border radius is miscalculated when there is drop-shadow

I found those problems while using an iPad. Box-shadow radius was probably miscalculated because of a scale factor that may be higher than 1.
D1C14C66-80AC-4594-A7ED-AC2BA2CE741E1FC2E34A-D732-4460-A13B-E4B3265F4F83

## What is the new behavior?
Border radius will be properly calculated for view box-shadow and sublayers. It seems that `toDeviceIndependentPixels` works better and returns the same value as the rest of applied view border-radiuses.
This change will also attempt to apply corner radius to all sublayers. Sublayers will usually consist of a gradient layer and a border layer.
There are also cases that border layer has sublayers but there is a need for a whole refactor to make those look good, that is why we apply radius to top-level layers for now.

This is new shadow radius with gradient background:
2DD8A8F8-9BCF-4DB4-AB60-8F1D323198BB_1_101_o

Fixes/Closes #9869 .

### DIFF
--- a/packages/core/ui/styling/background.ios.ts
+++ b/packages/core/ui/styling/background.ios.ts
@@ -726,7 +726,7 @@ function drawBoxShadow(nativeView: NativeScriptUIView, view: View, boxShadow: CS
 	cornerRadius = Math.min(Math.min(renderSize.width / 2, renderSize.height / 2), cornerRadius);
 
 	// Apply corner radius to sub layers as clipToBounds and masksToBounds are now set to false
-	if (nativeView.layer.sublayers?.count) {
+	if (nativeView.layer?.sublayers?.count) {
 		const count = nativeView.layer.sublayers.count;
 		for (let i = 0; i < count; i++) {
 			const subLayer = nativeView.layer.sublayers.objectAtIndex(i);
@@ -763,7 +763,7 @@ function drawBoxShadow(nativeView: NativeScriptUIView, view: View, boxShadow: CS
 
 function clearBoxShadow(nativeView: NativeScriptUIView) {
 	nativeView.clipsToBounds = true;
-	if (nativeView.layer.sublayers?.count) {
+	if (nativeView.layer?.sublayers?.count) {
 		const count = nativeView.layer.sublayers.count;
 		for (let i = 0; i < count; i++) {
 			const subLayer = nativeView.layer.sublayers.objectAtIndex(i);
--- a/packages/core/ui/styling/background.ios.ts
+++ b/packages/core/ui/styling/background.ios.ts
@@ -726,9 +726,11 @@ function drawBoxShadow(nativeView: NativeScriptUIView, view: View, boxShadow: CS
 	cornerRadius = Math.min(Math.min(renderSize.width / 2, renderSize.height / 2), cornerRadius);
 
 	// Apply corner radius to sub layers as clipToBounds and masksToBounds are now set to false
-	if (nativeView.layer.sublayers != null) {
-		for (let childLayer of nativeView.layer.sublayers) {
-			childLayer.cornerRadius = cornerRadius;
+	if (nativeView.layer.sublayers?.count) {
+		const count = nativeView.layer.sublayers.count;
+		for (let i = 0; i < count; i++) {
+			const subLayer = nativeView.layer.sublayers.objectAtIndex(i);
+			subLayer.cornerRadius = cornerRadius;
 		}
 	}
 
@@ -761,9 +763,11 @@ function drawBoxShadow(nativeView: NativeScriptUIView, view: View, boxShadow: CS
 
 function clearBoxShadow(nativeView: NativeScriptUIView) {
 	nativeView.clipsToBounds = true;
-	if (nativeView.layer.sublayers != null) {
-		for (let childLayer of nativeView.layer.sublayers) {
-			childLayer.cornerRadius = 0.0;
+	if (nativeView.layer.sublayers?.count) {
+		const count = nativeView.layer.sublayers.count;
+		for (let i = 0; i < count; i++) {
+			const subLayer = nativeView.layer.sublayers.objectAtIndex(i);
+			subLayer.cornerRadius = 0.0;
 		}
 	}
 
--- a/packages/core/ui/styling/background.ios.ts
+++ b/packages/core/ui/styling/background.ios.ts
@@ -726,11 +726,10 @@ function drawBoxShadow(nativeView: NativeScriptUIView, view: View, boxShadow: CS
 	cornerRadius = Math.min(Math.min(renderSize.width / 2, renderSize.height / 2), cornerRadius);
 
 	// Apply corner radius to sub layers as clipToBounds and masksToBounds are now set to false
-	if (nativeView.layer?.sublayers?.count) {
-		const count = nativeView.layer.sublayers.count;
-		for (let i = 0; i < count; i++) {
-			const subLayer = nativeView.layer.sublayers.objectAtIndex(i);
-			subLayer.cornerRadius = cornerRadius;
+	const sublayers = nativeView.layer?.sublayers;
+	if (sublayers) {
+		for (let i = 0; i < sublayers.count; i++) {
+			sublayers.objectAtIndex(i).cornerRadius = cornerRadius;
 		}
 	}
 
@@ -763,11 +762,10 @@ function drawBoxShadow(nativeView: NativeScriptUIView, view: View, boxShadow: CS
 
 function clearBoxShadow(nativeView: NativeScriptUIView) {
 	nativeView.clipsToBounds = true;
-	if (nativeView.layer?.sublayers?.count) {
-		const count = nativeView.layer.sublayers.count;
-		for (let i = 0; i < count; i++) {
-			const subLayer = nativeView.layer.sublayers.objectAtIndex(i);
-			subLayer.cornerRadius = 0.0;
+	const sublayers = nativeView.layer?.sublayers;
+	if (sublayers) {
+		for (let i = 0; i < sublayers.count; i++) {
+			sublayers.objectAtIndex(i).cornerRadius = 0.0;
 		}
 	}
 
--- a/packages/core/ui/styling/background.ios.ts
+++ b/packages/core/ui/styling/background.ios.ts
@@ -712,6 +712,7 @@ function drawNoRadiusNonUniformBorders(nativeView: NativeScriptUIView, backgroun
 // TODO: use sublayer if its applied to a layout
 function drawBoxShadow(nativeView: NativeScriptUIView, view: View, boxShadow: CSSShadow, background: BackgroundDefinition, useSubLayer: boolean = false) {
 	const layer: CALayer = iOSNativeHelper.getShadowLayer(nativeView, 'ns-box-shadow');
+	const renderSize = view.getActualSize() || { width: 0, height: 0 };
 
 	layer.masksToBounds = false;
 	nativeView.clipsToBounds = false;
@@ -720,10 +721,22 @@ function drawBoxShadow(nativeView: NativeScriptUIView, view: View, boxShadow: CS
 	// nativeView.clipsToBounds doesn't work
 	view.setProperty('clipToBounds', false);
 
+	// This should match the view's border radius (only uniform radius is supported for now)
+	let cornerRadius = layout.toDeviceIndependentPixels(background.borderTopLeftRadius);
+	cornerRadius = Math.min(Math.min(renderSize.width / 2, renderSize.height / 2), cornerRadius);
+
+	// Apply corner radius to sub layers as clipToBounds and masksToBounds are now set to false
+	if (nativeView.layer.sublayers != null) {
+		for (let childLayer of nativeView.layer.sublayers) {
+			childLayer.cornerRadius = cornerRadius;
+		}
+	}
+
 	if (!background.color?.a) {
 		// add white background if view has a transparent background
 		layer.backgroundColor = UIColor.whiteColor.CGColor;
 	}
+
 	// shadow opacity is handled on the shadow's color instance
 	layer.shadowOpacity = boxShadow.color?.a ? boxShadow.color?.a / 255 : 1;
 	layer.shadowRadius = Length.toDevicePixels(boxShadow.blurRadius, 0.0);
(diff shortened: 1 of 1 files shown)


NathanWalker (approved):

packages/core/ui/styling/background.ios.ts
```diff
@@ -720,10 +721,22 @@ function drawBoxShadow(nativeView: NativeScriptUIView, view: View, boxShadow: CS
 	// nativeView.clipsToBounds doesn't work
 	view.setProperty('clipToBounds', false);
 
+	// This should match the view's border radius (only uniform radius is supported for now)
+	let cornerRadius = layout.toDeviceIndependentPixels(background.borderTopLeftRadius);
+	cornerRadius = Math.min(Math.min(renderSize.width / 2, renderSize.height / 2), cornerRadius);
+
+	// Apply corner radius to sub layers as clipToBounds and masksToBounds are now set to false
+	if (nativeView.layer.sublayers != null) {
+		for (let childLayer of nativeView.layer.sublayers) {
```
NathanWalker, line 730: @CatchABus nativeView.layer.sublayers is a NSArray I believe so don't believe that for/of would work there. Likely need (would update condition too):
```
if (nativeView?.layer?.sublayers?.count) {
  for (let i = 0; i < nativeView.layer.sublayers.count; i++) { 
    nativeView.layer.sublayers.objectAtIndex(i).cornerRadius = cornerRadis
  }
}
```

NathanWalker (changes requested):

packages/core/ui/styling/background.ios.ts
CatchABus: @NathanWalker Strange thought it may seem, this loop works and I'm not sure why either.
At first, I used it in one of my plugins some time ago and I thought that runtime was designed to support this kind of iteration.

I can change it to a `for` loop as suggested, but it would also be interesting to know why it works that way too.

NathanWalker: Might check if project testing against is using v8 or JSC runtime? To be safest I would use NSArray iterator as that is for sure what this is.

CatchABus: > Might check if project testing against is using v8 or JSC runtime? To be safest I would use NSArray iterator as that is for sure what this is.

I'm using v8.
I also prefer the safest approach so I updated loops as pointed out above. Please let me know if there is more I can do to improve this.

packages/core/ui/styling/background.ios.ts
```diff
@@ -720,10 +721,24 @@ function drawBoxShadow(nativeView: NativeScriptUIView, view: View, boxShadow: CS
...
+	// This should match the view's border radius (only uniform radius is supported for now)
+	let cornerRadius = layout.toDeviceIndependentPixels(background.borderTopLeftRadius);
+	cornerRadius = Math.min(Math.min(renderSize.width / 2, renderSize.height / 2), cornerRadius);
+
+	// Apply corner radius to sub layers as clipToBounds and masksToBounds are now set to false
+	if (nativeView.layer?.sublayers?.count) {
+		const count = nativeView.layer.sublayers.count;
+		for (let i = 0; i < count; i++) {
+			const subLayer = nativeView.layer.sublayers.objectAtIndex(i);
+			subLayer.cornerRadius = cornerRadius;
```
NathanWalker, line 733: To reduce native calls, one last touch here will help ensure this is as performant as possible.
```
const sublayers = nativeView?.layer?.sublayers;
if (sublayers) {
  for (let i = 0; i < sublayers.count; i++) {
			nativeView.layer.sublayers.objectAtIndex(i).cornerRadius = cornerRadius;
  }
}
```
All throughout core we try to limit unneeded references and native property traversal to absolutely minimum - helps perf in the long run.

NathanWalker (changes requested):

packages/core/ui/styling/background.ios.ts
CatchABus: > To reduce native calls, one last touch here will help ensure this is as performant as possible.
> 
> ```
> const sublayers = nativeView?.layer?.sublayers;
> if (sublayers) {
>   for (let i = 0; i < sublayers.count; i++) {
>     sublayers.objectAtIndex(i).cornerRadius = cornerRadius;
>   }
> }
> ```
> 
> All throughout core we try to limit unneeded references and native property traversal to absolute minimum - helps perf in the long run.

Done! :)

CatchABus:
@NathanWalker Whenever possible, I believe we should revert this one as it's unsafe.
There is the case that corner radius will cause unexpected results if children are added to view "before" view shadow is applied for some reason.
I have thought of a good way to solve this and maintain a good performance at the same time.

farfromrefug:
@CatchABus if of any interest to you I had done this in another cross platform framework Titanium. I can fins the code if you need

CatchABus:
> @CatchABus if of any interest to you I had done this in another cross platform framework Titanium. I can fins the code if you need

Sure! It could be a good addition!

Note: I tend to think that using main layer with masksToBounds and clipsToBounds for shadows will keep causing trouble in the future. Maybe we should also consider the use of a sublayer for shadows as one solution?

farfromrefug:
@CatchABus it s all here https://github.com/Akylas/titanium_mobile/blob/master/iphone/Classes/TiUIView.m#L610= . If you have any question ping me on discord

rigor789:
> Note: I tend to think that using main layer with masksToBounds and clipsToBounds for shadows will keep causing trouble in the future. Maybe we should also consider the use of a sublayer for shadows as one solution?

I agree - I was working on the initial box-shadow implementation and had tinkered with the idea of using proper sublayers for shadows - allowing stacking multiple shadows etc, however there was a major roadblock (fixable I would imagine, but still was major)...

Given an ios view, you can add as many subviews as you want, however you cannot move them behind the parent view, only above it (at least if I remember correctly, been a while...) - which meant we couldn't easily add a shadow layer to existing views and move them behind the view. The approach I was playing around with was adding an extra parent view to hold the original view + the shadow layers, but managing that was kinda tricky as core View handling is set up to handle the root view and not anything above.

Wrapping every view in a extra view seemed like a bad solution, as it would double the number of views rendered even if most of them don't render a shadow.

Ideally the following structure could work:

```
  View
    -> nativeView (UIView or whatever)
```

```
  View with shadow
    -> decorView
       -> nativeView (UIView or whatever)
       -> shadowLayer1
       -> shadowLayer2
       -> etc.
```

CatchABus:
> > Note: I tend to think that using main layer with masksToBounds and clipsToBounds for shadows will keep causing trouble in the future. Maybe we should also consider the use of a sublayer for shadows as one solution?
> 
> I agree - I was working on the initial box-shadow implementation and had tinkered with the idea of using proper sublayers for shadows - allowing stacking multiple shadows etc, however there was a major roadblock (fixable I would imagine, but still was major)...
> 
> Given an ios view, you can add as many subviews as you want, however you cannot move them behind the parent view, only above it (at least if I remember correctly, been a while...) - which meant we couldn't easily add a shadow layer to existing views and move them behind the view. The approach I was playing around with was adding an extra parent view to hold the original view + the shadow layers, but managing that was kinda tricky as core View handling is set up to handle the root view and not anything above.
> 
> Wrapping every view in a extra view seemed like a bad solution, as it would double the number of views rendered even if most of them don't render a shadow.
> 
> Ideally the following structure could work:
> 
> ```
>   View
>     -> nativeView (UIView or whatever)
> ```
> 
> ```
>   View with shadow
>     -> decorView
>        -> nativeView (UIView or whatever)
>        -> shadowLayer1
>        -> shadowLayer2
>        -> etc.
> ```

Perhaps, this issue gets solved using `insertSublayerAtIndex` or setting layer `zPosition` to negative value (default is 0 for all layers).
I'm using those in my neumorphism shadow plugin.

rigor789:
> > > Note: I tend to think that using main layer with masksToBounds and clipsToBounds for shadows will keep causing trouble in the future. Maybe we should also consider the use of a sublayer for shadows as one solution?
> > 
> > 
> > I agree - I was working on the initial box-shadow implementation and had tinkered with the idea of using proper sublayers for shadows - allowing stacking multiple shadows etc, however there was a major roadblock (fixable I would imagine, but still was major)...
> > Given an ios view, you can add as many subviews as you want, however you cannot move them behind the parent view, only above it (at least if I remember correctly, been a while...) - which meant we couldn't easily add a shadow layer to existing views and move them behind the view. The approach I was playing around with was adding an extra parent view to hold the original view + the shadow layers, but managing that was kinda tricky as core View handling is set up to handle the root view and not anything above.
> > Wrapping every view in a extra view seemed like a bad solution, as it would double the number of views rendered even if most of them don't render a shadow.
> > Ideally the following structure could work:
> > ```
> >   View
> >     -> nativeView (UIView or whatever)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> >   View with shadow
> >     -> decorView
> >        -> nativeView (UIView or whatever)
> >        -> shadowLayer1
> >        -> shadowLayer2
> >        -> etc.
> > ```
> 
> Perhaps, this issue gets solved using `insertSublayerAtIndex` or setting layer `zPosition` to negative value (default is 0 for all layers). I'm using those in my neumorphism shadow plugin.

I'm not sure - I'm pretty sure I tried every possible combination, and could not move my shadow layer below the actual view - inserting it at various indices or zPositions, was always above the view. But if that works, that'd be perfect 👍

CatchABus:
@rigor789 I'm not sure if I understood correctly but I have this plugin for neumorphism styling: https://github.com/CatchABus/nativescript-plugins/tree/master/packages/ui-neumorphiclayout
It uses 2 sublayers because support for 2 different shadows is needed. So far, I have nested children inside those views without problems.
Maybe, this concept could serve for adding multiple shadows support to NS.

rigor789:
@CatchABus I just remembered - I never deleted the tests I did here (they are ignored, as we return early): https://github.com/NativeScript/NativeScript/blob/eb4c4a3091a347b9cf8f2fdf5593f20938464ab3/packages/core/utils/native-helper.ios.ts#L126-L203

removing the early return yields the following: 
Screenshot 2022-05-05 at 8 06 58 PM

With some logging - here are the resulting layers:

Screenshot 2022-05-05 at 8 08 38 PM

Haven't had a closer look at your plugin, or what it does differently - but if you want you can try the above:
 - run the toolbox app -> box-shadow -> toggle solid background -> apply shadow
 - comment out the return in the native-helper.ios.ts and let the test logic run
 - make any adjustments in there to see if there's a way to move that shadow layer below the actual view...

CatchABus:
> @CatchABus I just remembered - I never deleted the tests I did here (they are ignored, as we return early):
> 
> https://github.com/NativeScript/NativeScript/blob/eb4c4a3091a347b9cf8f2fdf5593f20938464ab3/packages/core/utils/native-helper.ios.ts#L126-L203
> 
> removing the early return yields the following: Screenshot 2022-05-05 at 8 06 58 PM
> 
> With some logging - here are the resulting layers:
> 
> Screenshot 2022-05-05 at 8 08 38 PM
> 
> Haven't had a closer look at your plugin, or what it does differently - but if you want you can try the above:
> 
> * run the toolbox app -> box-shadow -> toggle solid background -> apply shadow
> * comment out the return in the native-helper.ios.ts and let the test logic run
> * make any adjustments in there to see if there's a way to move that shadow layer below the actual view...
> 
> I guess the difference is that in your plugin you have a `fillColor` property - and not the regular `backgroundColor`/`backgroundImage` I guess, so you have more control of the drawing there?

I see, I get it now! To solve this, we can apply view background color to shadow layer.
After all, it's only the color that's on the bottom.
Screenshot 2022-05-07 at 5 24 52 PM


Seeing where this goes, I think we should stop setting background color to native views directly and have a `backgroundLayer` that replaces shadow and border layers functionality, and background color would also apply to it. It could be used for multiple purposes.

PR #9883, opened 2022-04-27
fix(core): android wrong background state + current value accessors

farfromrefug:
This fixes a bug with my recent addition of `backgroung/foreground` events.
It also adds getters `inBackground` and `suspended` in `application-common` to query the current value for both iOS/Android

### DIFF
--- a/packages/core/application/index.android.ts
+++ b/packages/core/application/index.android.ts
@@ -16,6 +16,7 @@ import { Observable } from '../data/observable';
 import { profile } from '../profiling';
 import { initAccessibilityCssHelper } from '../accessibility/accessibility-css-helper';
 import { initAccessibilityFontScale } from '../accessibility/font-scale';
+import { inBackground, setInBackground, setSuspended, suspended } from './application-common';
 
 const ActivityCreated = 'activityCreated';
 const ActivityDestroyed = 'activityDestroyed';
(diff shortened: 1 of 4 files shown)


NathanWalker (approved):

farfromrefug:
@NathanWalker maybe the getters name should.be different.let me know if you think there would be better names

packages/core/application/index.android.ts
```diff
@@ -48,8 +49,14 @@ export class AndroidApplication extends Observable implements AndroidApplication
...
 	private _systemAppearance: 'light' | 'dark';
-	public paused: boolean;
-	public backgrounded: boolean;
+
+	get paused() {
+		return suspended;
+	}
+	get backgrounded() {
+		return inBackground;
+	}
```
NathanWalker, line 58: @farfromrefug nice changes here, will definitely help. Let's go with this naming, I think will be more clear:
```
get isPaused() {
  return suspended;
}
get isInBackground() {
  return inBackground;
}
```

NathanWalker (changes requested):

packages/core/application/index.android.ts
farfromrefug: @NathanWalker i was more talking about the naming in application-common.ts https://github.com/Akylas/NativeScript/blob/backgroundevent_fix/packages/core/application/application-common.ts#L173=
The change you propose is a breaking change as it was already available. Are we sure we want that to change?
we could actually deprecate the android file one. Now people should use the common one as it works on iOS and android

NathanWalker: Oh! I see my bad. Let’s keep what you have then. Will include in next alpha.

NathanWalker: I agree with deprecating the old one and just make single consistent one from `Application.isPaused` for example - would behave same for all platforms. We can do that separately on the alpha branch itself.

PR #9884, opened 2022-04-27
feat(core): maxLines support for all text components

farfromrefug:
This adds `maxLines` property for all `text-base` components.
It supports number + 'none' as specified. It also works correctly with `whiteSpace`


NathanWalker (approved):

farfromrefug:
@NathanWalker Question why did you merge it this way? You actually dropped `none` which is in the css specs.

farfromrefug:
@NathanWalker OK I will push the fix I guess. i think it is best to stick to CSS specs (though I think it is not yet in all browsers)

PR #9891, opened 2022-05-01
fix: dont include core.aar in production package

farfromrefug:
This prevent the `core.aar` build in dev env to be packaged in npm package.
To do that:
* untrack and ignore it
* add `.npmignore` rules
* remove `files` rule from `package.json` as it conflicts with `.npmignore`
* add nx rule to copy `.npmignore`


packages/core/package.json
```diff
@@ -9,14 +9,6 @@
...
     "./globals"
   ],
-  "files": [
-    "**/*.d.ts",
-    "**/*.js",
-    "**/*.map",
-    "**/platforms/ios/**",
-    "**/platforms/android/**",
-    "**/package.json"
-  ],
```
rigor789, line 19: Let's keep `files` - and drop `.npmignore` instead, files is much more explicit of what is included, and as things change over time, new files won't randomly appear in the package unless we explicitly list them.

farfromrefug: @rigor789
we cant define rule to ignore aar but keep widget lib with `files`
honestly I think there is no point in using files as we copy only the necessary files to dist (except for that core.aar)

NathanWalker:
Branch cleaned up here: https://github.com/NativeScript/NativeScript/pull/9956

PR #9892, opened 2022-05-01
chore: first stage of rewriting animations

farfromrefug:
This is a draft of rewriting the Animation system in N:

* make `Animation` unaware of which properties can be animated
* handle specific cases (rotate, translate, scale)
* make virtually any property animatable on Android. This is done the same way as it is done for `backgroundColor`(which means going back to JS in every single animation step!)
* make width/height properties correctly animate in layout and allowing handling of `auto` size
* iOS: work in progress. Trying to stop handling animations differently if spring or not. Maybe move to `UIViewPropertyAnimator`. My issue right now is how to animate custom properties like text color.

TODO:
* allow `View` subclass to define custom property animation handling
* allow property to define animation handling?
* merge `CSSProperty` and `CSSAnimationProperty` ? have to figure out the real difference between the 2 as there is no inheritance :s Merging would allow any prop to be animated through css


farfromrefug:
Mostly working. Cant go much further within the core:
* refactored Property classes big time to make it clearer and allow interactions with animations
* fixed iOS issues with sping animation, layout updates and originX/Y breaking stuff
* Android can now animation virtually any property
* iOS cant without the use (or impl) of custom tween with update block
* more work to be done to refactor to all custom animation plugin to inherit and use most logic from N Animation

farfromrefug:
I think i am done for now:
* kind of a huge refactor of `Property` classes and `Animation`
* i think all tests are passing though some are not passing locally which i dont understand
* `Animation` on iSOS:
   - uses `UIView.animate` now instead of `CABasicAnimation`. This allows better layout animations
   - `CABasicAnimation` code is still there. This would allow specific properties animations to be done with it when they dont work with `UIView.animate`. An example is `borderRadius` animation
   - also added the code for `UIViewPropertyAnimator` which appeared in ios10. This (if we bump min sdk version to 10) is the new api which allows singular animation cancelling, but also "moving" along an animation time. For example with a slider (or a pager!)
   - refactored to allow subclass for custom animation plugin using 3rd party tween libs
* `Animation` on Android: 
    - it can now virtually animate any property through update black (thus going back to js at every single step). 
    - refactored to allow subclass for custom animation plugin using 3rd party tween libs
    -  there is still a bug that i did not fix (always been there). Using sequential animations wont work. The reason is that from/to of native animations are computed before the first one start. So for example sequential opacity : 0=>0.5=>1 will end up being 0=>0.5 then 0=>1
    
    I think this is ready for review

farfromrefug:
Actually think i can fix some of the iOS tests errors

vallemar:
Hi team @NathanWalker @farfromrefug @rigor789 @triniwiz ! I comment here because one of the reasons why I use the fork is this. Animations work better in the fork than in nativescript core. I think this PR is important, since animations are important today. This PR seems to have been shelved and doesn't seem to be merging anywhere near. Could this be re-evaluated to be able to merge for a version that is not too late?

PR #9897, opened 2022-05-05
fix: expose `inBackground` and `suspended in `Application`

farfromrefug:
and in IOSApplicationt to be like AndroidApplication


packages/core/index.ts
```diff
@@ -6,6 +6,8 @@ export { iOSApplication, AndroidApplication } from './application';
 export type { ApplicationEventData, LaunchEventData, OrientationChangedEventData, UnhandledErrorEventData, DiscardedErrorEventData, CssChangedEventData, LoadAppCSSEventData, AndroidActivityEventData, AndroidActivityBundleEventData, AndroidActivityRequestPermissionsEventData, AndroidActivityResultEventData, AndroidActivityNewIntentEventData, AndroidActivityBackPressedEventData, SystemAppearanceChangedEventData } from './application';
 
 import { fontScaleChangedEvent, launchEvent, displayedEvent, uncaughtErrorEvent, discardedErrorEvent, suspendEvent, resumeEvent, exitEvent, lowMemoryEvent, orientationChangedEvent, systemAppearanceChanged, systemAppearanceChangedEvent, getMainEntry, getRootView, _resetRootView, getResources, setResources, setCssFileName, getCssFileName, loadAppCss, addCss, on, off, notify, hasListeners, run, orientation, getNativeApplication, hasLaunched, android as appAndroid, ios as iosApp, systemAppearance, setAutoSystemAppearanceChanged, ensureNativeApplication, setMaxRefreshRate } from './application';
+import { inBackground, suspended } from 'application/application-common';
```
CatchABus, line 9: @farfromrefug @NathanWalker I know this got merged, but it seems that resolution path is missing `./` here.

farfromrefug: @CatchABus. good catch will fix un a few minutes.

farfromrefug: @NathanWalker it is fixed you can make a new alpha if you want

PR #9915, opened 2022-05-19
fix(android) Connectivity reporting none on resume

jcassidyav:
## PR Checklist

On some android phones, (Google Pixel 5 & Galaxy s20+ running android 12) the final state reported when subscribing to ````Connectivity.startMonitoring```` would report incorrectly report no connectivity.

This happened when the app was in the foreground, and the phone was locked and then unlocked ( i.e. hit power button)
This was because the call to the methods on ConnectivityManager.getActiveNetwork returned null, the documentation seems to suggest that calling any of the syncronus methods on ConnectivityManager is not advised during these callbacks.
https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback

## What is the new behavior?

The new behaviour is to rely on the network capabilities passed to the callback to determine the status, or to report no connectivity when lost is triggered.

### DIFF
--- a/packages/core/connectivity/index.android.ts
+++ b/packages/core/connectivity/index.android.ts
@@ -32,8 +32,12 @@ function getActiveNetworkInfo(): android.net.NetworkInfo {
 function getNetworkCapabilities() {
 	// @ts-ignore
 	const connectivityManager: any = getConnectivityManager();
-	const network = connectivityManager.getActiveNetwork();
-	const capabilities = connectivityManager.getNetworkCapabilities(network);
+	const networkToCheck = connectivityManager.getActiveNetwork();
+	const capabilities = connectivityManager.getNetworkCapabilities(networkToCheck);
+	return parseNetworkCapabilities(capabilities);
+}
+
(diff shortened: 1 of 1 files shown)


packages/core/connectivity/index.android.ts
```diff
@@ -116,24 +120,22 @@ export function startMonitoring(connectionTypeChangedCallback: (newConnectionTyp
...
 				@NativeClass
 				class NetworkCallbackImpl extends ConnectivityManager.NetworkCallback {
-					onAvailable(network: android.net.Network) {
-						if (notifyCallback) {
-							notifyCallback();
-						}
-					}
-
 					onCapabilitiesChanged(network: android.net.Network, networkCapabilities: android.net.NetworkCapabilities) {
+						console.log('NetworkCallbackImpl.onCapabilitiesChanged');
```
farfromrefug, line 136: could you remove the console.log ?

jcassidyav: Don't know how I missed that, gone now.

PR #9922, opened 2022-05-24
fix(android): race condition on offBackgroundColor view change detection

NathanWalker:
## PR Checklist

Various framework integrations, like Angular, may invoke view change detection cycles on property bindings at times when the `nativeView` is not available resulting in errors like:

```
JS: ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'getTrackDrawable' of undefined
JS: TypeError: Cannot read property 'getTrackDrawable' of undefined
JS:     at Switch.setNativeBackgroundColor (file: src/webpack:/app/node_modules/@nativescript/core/ui/switch/index.android.js:48:0)
JS:     at Switch._onCheckedPropertyChanged (file: src/webpack:/app/node_modules/@nativescript/core/ui/switch/index.android.js:61:0)
```

## What is the new behavior?

`Switch` is now resilient to this.

### DIFF
--- a/packages/core/ui/switch/index.android.ts
+++ b/packages/core/ui/switch/index.android.ts
@@ -57,11 +57,13 @@ export class Switch extends SwitchBase {
 	}
 
 	private setNativeBackgroundColor(value: string | number | Color) {
-		if (value instanceof Color) {
-			// todo: use https://developer.android.com/reference/androidx/core/graphics/BlendModeColorFilterCompat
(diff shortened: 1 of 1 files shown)


farfromrefug:
@NathanWalker shouldn't this be handle on the property side ? isn't it already? don't think we want this test everywhere. the property setter handler seems like a better place to me

NathanWalker:
@farfromrefug good question, it seems that because this function is reused several places, it's called outside the `Property` mechanics at times leading to the issue.

PR #9927, opened 2022-06-01
feat(types-minimal): minimal types for only the latest Android and iOS sdks

NathanWalker:
## PR Checklist

Currently Web-based IDEs which auto load *all* type declarations from anything they find in `node_modules` would not perform well due to high memory consumption of loading all types (even all sdks over the years).

## What is the new behavior?

This package `@nativescript/types-minimal` optimizes to only include the latest Android and iOS sdks and allows web-based IDE's to perform extremely well.


farfromrefug:
@NathanWalker love this. Could we split the androidx typing like it is done on iOS? That way there could be a way to get even faster by only using the ones you need. Could make a huge difference.
We could even think about splitting the android typings by separating android.os, android.widgets, java .....

rigor789 (approved):

NathanWalker:
> @NathanWalker love this. Could we split the androidx typing like it is done on iOS? That way there could be a way to get even faster by only using the ones you need. Could make a huge difference. We could even think about splitting the android typings by separating android.os, android.widgets, java .....

We can take this farther as time moves on yes. This will be step 1.

PR #9931, opened 2022-06-06
fix(core): iOS app not responding after rotation

dgmachado:
## PR Checklist

The iOS app isn't responding after rotation. The issue is only reproducible on the iOS phone. Cannot be reproduced on an iOS tablet.

The issue can be reproduced immediately (after the first rotation) on the nested tabs scenario.

Could be reproduced on tabs scenario as well after multiple rotations and switch tabs in between.

When the issue happens, the UI appears to be cut into half and it's not responding at all.

https://user-images.githubusercontent.com/5738416/172222411-58b03653-16cf-4f4f-8c78-9c0b87e28369.mov

## What is the new behavior?

Fix the infinite loop that was occurring after successive rotations of the app so now the app is no longer getting stuck.


CatchABus:
@dgmachado It seems this change causes other issues similar to the one you described.
More specifically, there is a case of infinite loop as described in ticket #9962 
I tried the provided sample with an iPhone 12 simulator and confirmed the validity of this report.

It seems there are cases that `additionalSafeAreaInsets` receive new values over and over resulting in calling `viewSafeAreaInsetsDidChange` and ending up in an infinite loop.

Is it possible that you post an app sample with your case? Perhaps, we can fix this using another approach.

dgmachado:
@CatchABus I found out last week that my change doesn't fix the infinite loop even for my scenario. It's just a lot harder to reproduce the issue.  Let me share with you an app sample with my case [testPopover_tabBar5.zip](https://github.com/NativeScript/NativeScript/files/9241706/testPopover_tabBar5.zip)


Is it possible you also share an app sample with your case?

CatchABus:
> @CatchABus I found out last week that my change doesn't fix the infinite loop even for my scenario. It's just a lot harder to reproduce the issue. Let me share with you an app sample with my case [testPopover_tabBar5.zip](https://github.com/NativeScript/NativeScript/files/9241706/testPopover_tabBar5.zip)
> 
> And about your comment "It seems there are cases that additionalSafeAreaInsets receive new values over and over resulting in calling viewSafeAreaInsetsDidChange and ending up in an infinite loop." that's exactly the same thing I was getting frequently before my change.
> 
> Is it possible you also share an app sample with your case?

Thank you very much for this! I'll also use your sample and see what I can find.
Member who reported the other case has included an app same and a reproduction video in #9962

dgmachado:
@CatchABus I just tested your app using the "@nativescript/core@8.2.4" and "@nativescript/core@8.2.0" to ensure the [#9931](https://github.com/NativeScript/NativeScript/pull/9931) wasn't included and I could reproduce this same issue using simulator iPhone 8 (iOS 15.2). 

@NathanWalker could confirm but I'm almost sure the #9931 was merged into the 8.2.5 version only.

CatchABus:
> @CatchABus I just tested your app using the "@nativescript/core@8.2.4" and "@nativescript/core@8.2.0" to ensure the [#9931](https://github.com/NativeScript/NativeScript/pull/9931) wasn't included and I could reproduce this same issue using simulator iPhone 8 (iOS 15.2).
> 
> [Video showing the infinite loop](https://drive.google.com/file/d/1eA1WrXr1E6PG6bx3U7CR9iK5VSsOMDFg/view?usp=sharing) (Obs: it's an intermittent bug)
> 
> @NathanWalker could confirm but I'm almost sure the #9931 was merged into the 8.2.5 version only.

I believe this is a different case.
In #9962, reporter mentioned everything was good until 8.2.4.
He also had the following output:
```
***** Fatal JavaScript exception - application has been terminated. *****
NativeScript encountered a fatal error: Uncaught RangeError: Maximum call stack size exceeded
at
UILayoutViewController.updateAdditionalSafeAreaInsetsIfNeeded(file: app/webpack:/@nativescript/template-hello-world-ts/node_modules/@nativescript/core/ui/core/view/view-helper/index.ios.js:61:0)
at UILayoutViewController.viewSafeAreaInsetsDidChange(file: app/webpack:/@nativescript/template-hello-world-ts/node_modules/@nativescript/core/ui/core/view/view-helper/index.ios.js:69:0)
at UILayoutViewController.updateAdditionalSafeAreaInsetsIfNeeded(file: app/webpack:/@nativescript/template-hello-world-ts/node_modules/@nativescript/core/ui/core/view/view-helper/index.ios.js:61:0)
at UILayoutViewController.viewSafeAreaInsetsDidChange(file: app/webpack:/@nativescript/template-hello-world-ts/node_modules/@nativescript/core/ui/core/view/view-helper/index.ios.js:69:0)
at UILayoutViewController.updateAdditionalSafeAreaInsetsIfNeeded(file: app/webpack:/@nativescript/template-hello-world-ts/node_modules/@nativescript/core/ui/core/view/view-helper/index.ios.js:61:0)
at UILayoutViewController.viewSafeAreaInsetsDidChange(file: app/webpack:/@nativescript/template-hello-world-ts/node_modules/@nativescript/core/ui/core/view/view-helper/index.ios.js:69:0)
at UILayoutViewController.updateAdditionalSafeAreaInsetsIfNeeded(file: app/webpack:/@nativescript/template-hello-world-ts/node_modules/@nativescript/core/ui/core/view/view-helper/index.ios.js:61:0)
at UILayoutViewController.viewSafeAreaInsetsDidChange(file: app/webpack:/@nativescript/template-hello-world-ts/node_modules/@nativescript/core/ui/core/view/view-helper/index.ios.js:69:0)
at UILayoutViewController.updateAdditionalSafeAreaInsetsIfNeeded(file: app/webpack:/@nativescript/template-hello-world-ts/node_modules/@nativescript/core/ui/core/view/view-helper/index.ios.js:61:0)
at UILayoutViewController.viewSafeAreaInsetsDidChange(file:///app/vendor.js:2524<…>
```

I'll try to reproduce your own case using your sample.

CatchABus:
@dgmachado Thanks for the app, it was really useful.
I also came to realise that your patch made the issue less frequent which makes it easy to get tricked.

After lots of tests, I came up with my first attempt to solve this which I'd like you to try.

Please modify this line: https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/core/view/index.ios.ts#L291

And update it to look like this:
```ts
                // View controllers that need insets calculate it on their own
                if (this.viewController || this.iosIgnoreSafeArea) {
			return frame;
		}
```

This simple change enforces a rule that we should not expand or shrink insets to views with view controllers by default.
This covers the cases of NS `UILayoutViewController` and `UIViewControllerImpl` which modify `additionalSafeAreaInsets`.

I did few tests in the sample app and infinite loop seems to be gone. If possible, please leave some feedback regarding the efficiency of this fix and we may be able to merge this if it looks good.

PR #9945, opened 2022-06-29
fix(ios): label measure correct height when using custom numberOfLines

### LINKED ISSUE #6037: Label nativeElement numberOfLines doesn't resize frame

Yes; I have not found any similar issues.

### Tell us about the problem

When trying to set a max number of lines with the following code: 
`
let label = <Label>ref.nativeElement;
if (platformModule.isIOS) {
    label.ios.numberOfLines = 3;
} else {
    label.android.setMaxLines(3);
}
`
the Label is correctly cut off after three labels. However, it looks like on iOS the frame of the label does not get smaller.

Expected behaviour (this is taken on Android): 

image

Observed behaviour (this is taken on iOS):

image

iOS

- CLI: 4.1.2
- Cross-platform modules: 4.1.0
- Runtime(s): iOS: 4.1.1, android: 4.1.3

I hope this is enough to fix this problem. If there is anything else you need, please contact me.

mukaschultze:
## PR Checklist

Right now, iOS labels include all the lines on the height measurement and do not respect the value specified by the `numberOfLines` property.

## What is the new behavior?

The new behavior respects the `numberOfLines` property and properly measures the label height. When the user doesn't specify a custom value for the property it'll be zero and thus all the lines will be considered in the measurement, just as before.

Fixes #6037

### DIFF
--- a/packages/core/ui/label/index.ios.ts
+++ b/packages/core/ui/label/index.ios.ts
@@ -104,7 +104,7 @@ export class Label extends TextBase implements LabelDefinition {
 	private _measureNativeView(width: number, widthMode: number, height: number, heightMode: number): { width: number; height: number } {
 		const view = <UILabel>this.nativeTextViewProtected;
 
-		const nativeSize = view.textRectForBoundsLimitedToNumberOfLines(CGRectMake(0, 0, widthMode === 0 /* layout.UNSPECIFIED */ ? Number.POSITIVE_INFINITY : layout.toDeviceIndependentPixels(width), heightMode === 0 /* layout.UNSPECIFIED */ ? Number.POSITIVE_INFINITY : layout.toDeviceIndependentPixels(height)), 0).size;
+		const nativeSize = view.textRectForBoundsLimitedToNumberOfLines(CGRectMake(0, 0, widthMode === 0 /* layout.UNSPECIFIED */ ? Number.POSITIVE_INFINITY : layout.toDeviceIndependentPixels(width), heightMode === 0 /* layout.UNSPECIFIED */ ? Number.POSITIVE_INFINITY : layout.toDeviceIndependentPixels(height)), view.numberOfLines).size;
 
 		nativeSize.width = layout.round(layout.toDevicePixels(nativeSize.width));
 		nativeSize.height = layout.round(layout.toDevicePixels(nativeSize.height));


NathanWalker:
Thank you 🙏

NathanWalker (approved):

packages/core/ui/label/index.ios.ts
```diff
@@ -104,7 +104,7 @@ export class Label extends TextBase implements LabelDefinition {
 	private _measureNativeView(width: number, widthMode: number, height: number, heightMode: number): { width: number; height: number } {
 		const view = <UILabel>this.nativeTextViewProtected;
 
-		const nativeSize = view.textRectForBoundsLimitedToNumberOfLines(CGRectMake(0, 0, widthMode === 0 /* layout.UNSPECIFIED */ ? Number.POSITIVE_INFINITY : layout.toDeviceIndependentPixels(width), heightMode === 0 /* layout.UNSPECIFIED */ ? Number.POSITIVE_INFINITY : layout.toDeviceIndependentPixels(height)), 0).size;
+		const nativeSize = view.textRectForBoundsLimitedToNumberOfLines(CGRectMake(0, 0, widthMode === 0 /* layout.UNSPECIFIED */ ? Number.POSITIVE_INFINITY : layout.toDeviceIndependentPixels(width), heightMode === 0 /* layout.UNSPECIFIED */ ? Number.POSITIVE_INFINITY : layout.toDeviceIndependentPixels(height)), view.numberOfLines).size;
```
NathanWalker, line 107: This will be a breaking change and thus targeting 8.3 however we should regression test with this due to fact this would not default to 0 as it was this way anymore.
https://developer.apple.com/documentation/uikit/uilabel/1620539-numberoflines
> The default value for this property is 1. To remove any maximum limit, and use as many lines as needed, set the value of this property to 0.

mukaschultze: @NathanWalker Yes, the default native value is 1, but on {N} setting `textWrap` to true (or the `whiteSpace` style) sets it to 0, and this `_measureNativeView` method is only called when textWrap is true, otherwise it uses `layout.measureNativeView`. 

Following the code path I don't see any way that this is called with the value of 1 unless the user sets the `numberOfLines` prop to 1 explicitly

NathanWalker: Great analysis @mukaschultze I'm going to include this on the next 8.3.x alpha - we are beginning to qualify things around the next minor and would make great candidate for that at any rate. Thanks for your effort here.

PR #9947, opened 2022-07-01
feat(types-minimal): include AVFAudio and ARKit

NathanWalker:
## PR Checklist

With Preview 2.0 you could not access ARKit or AVAudio apis.

## What is the new behavior?

With Preview 2.0 you can now access ARKit or AVAudio apis.

### DIFF
--- a/packages/types-minimal/package.json
+++ b/packages/types-minimal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@nativescript/types-minimal",
-  "version": "8.2.0",
+  "version": "8.2.1",
   "description": "NativeScript 'Minimal' Types for only the latest Android and iOS sdks.",
   "homepage": "https://nativescript.org",
   "repository": {
--- a/packages/types-minimal/package.json
+++ b/packages/types-minimal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@nativescript/types-minimal",
-  "version": "8.2.1",
+  "version": "8.2.2",
   "description": "NativeScript 'Minimal' Types for only the latest Android and iOS sdks.",
   "homepage": "https://nativescript.org",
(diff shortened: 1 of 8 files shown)


farfromrefug:
if I may I don't think arkit nor avaudio have their place in minimal. it breaks the "minimal" term. 
instead you should manually reference them both.
I think minimal should really be minimal to make it as fast as possible. I think most apps don't use nor avaudio nor arkit

rigor789:
@farfromrefug the reason we want to add it is to allow av/ar examples in the new Preview through StackBlitz (which was the motivation behind types-minimal as the regular types crash the browser)

NathanWalker:
@farfromrefug we may add 2-3 more but still carefully evaluating which frameworks are most valuable to have in types-minimal - these 2 in particular are actually common for example showcases and thus good candidates for Preview usage and also did not affect the perf of them at all (which we will evaluate at each inclusion and will continue to do so). It's even possible that subsequent inclusions may contain customized slimmed down versions of frameworks depending on the needs.

farfromrefug:
i understand but then why not singularly reference each module you need? or create a package specifically for the preview. like everyone searching on npm will find this thinking it is minimal when it is not. the issue not only happens on the web, on.most computer you ll have slow completion with full typings

NathanWalker:
If you’d like to submit a PR for a subsequent package we could definitely do that. It’s more matter of simplicity and avoiding so many different types packages on npm. But always open to your pull requests going in alternate directions. Can also discuss at next TSC meeting.

farfromrefug:
@NathanWalker sure I can but I don't have access to the preview code right ? the simplest and I think the best way would be to fix it there without this PR(if I am not wrong in undstanding the issue)

PR #9956, opened 2022-07-09
chore: switch to npmignore over package files

NathanWalker:


NathanWalker:
We'll leave it with package files since exclusions can be marked there as well.

edusperoni:
We can use .nxignore locally and NX shouldn't even copy or hash that file during the build process

PR #9959, opened 2022-07-16
feat(config): logRuntimeDetail to enable verbose platform runtime logs

NathanWalker:
## PR Checklist

No way to view verbose details around platform runtime calls.

## What is the new behavior?

The `nativescript.config` has a new boolean `logRuntimeDetail` which can enable verbose platform runtime calls. This can reveal platform marshaling calls your app is making as well as reveal detail regarding classes and methods being invoked on the target platform runtime.

Relies on this [runtime commit](https://github.com/NativeScript/ns-v8ios-runtime/commit/ac7237c0727685fd166fdd40d7844cd4c5101cee) to engage.


rigor789:
Could we do a string-based setting instead of a boolean?

Something like `runtimeLogVerbosity` and values `silent | verbose`?

NathanWalker:
> Could we do a string-based setting instead of a boolean?
> 
> Something like `runtimeLogVerbosity` and values `silent | verbose`?

Good idea. Perhaps it could be an Array, for example:

`runtimeLogVerbosity: ['interop', 'metadata']`
or...
`runtimeLogVerbosity: ['all']`

That way you could mix/match different detail or just view is `all`.

PR #9995, opened 2022-08-16
feat(core): support font-variation-settings

lekhmanrus:
## PR Checklist

[`font-variation-settings`](https://drafts.csswg.org/css-fonts/#font-variation-settings-def) CSS rule not supported.

## What is the new behavior?
`font-variation-settings` is supported on Android and I'm thinking about the proper implementation for iOS. I guess it should be similar to VFont - using attributes. It would be fine if someone has any suggestions/directions here. 🙂


farfromrefug:
@lekhmanrus warning the font builder was only added in API 29. if I read your changes right it will crash on < 29.

lekhmanrus:
@farfromrefug, [Typeface.Builder](https://developer.android.com/reference/android/graphics/Typeface.Builder#Builder(java.io.File)) added in API level 26. But yeah - I probably need to add a few conditions in the code.

farfromrefug:
@lekhmanrus you are right. it says 26. it is weird as I am sure I saw an android doc saying 29... anyway...

lekhmanrus:
@farfromrefug, I guess it was [Font.Builder](https://developer.android.com/reference/android/graphics/fonts/Font.Builder) - it says 29. But I use [Typeface.Builder](https://developer.android.com/reference/android/graphics/Typeface.Builder#Builder(java.io.File)) here.

packages/core/platforms/ios/typings/objc!NativeScriptUtils.d.ts
```diff
@@ -11,7 +10,7 @@ declare class NativeScriptUtils extends NSObject {
 
 	static getImageDataFormatQuality(image: UIImage, format: string, quality: number): NSData;
 
-	static getSystemFontWeightItalicSymbolicTraits(size: number, weight: number, italic: boolean, symbolicTraits: number): UIFont;
+	static getSystemFont(size: number, weight: UIFontWeight, italic: boolean, symbolicTraits: UIFontDescriptorSymbolicTraits): UIFont;
```
NathanWalker, line 13: This doesn't look right - you can `ns typings ios` inside of `apps/toolbox` to regen the typings and put this typing back to get it correct. Should still be `getSystemFontWeightItalicSymbolicTraits` I believe.

NathanWalker (changes requested):

packages/core/platforms/ios/typings/objc!NativeScriptUtils.d.ts
lekhmanrus: Hi @NathanWalker, Sorry currently do not have Mac in front of me - I'll try that next week.
But why do we need `getSystemFontWeightItalicSymbolicTraits`? We have 0 matches for that in the repo. Where it comes from?

NathanWalker: That is a platform API from iOS therefore it's signature comes from the metadata (`ns typings ios`) - it may or may not be used within scope of core however the typings here are important to match always what the generation determines (should never be manually adjusted).

lekhmanrus: Hi @NathanWalker,
Thanks, just committed typings. 🙂

NathanWalker (approved):

PR #9996, opened 2022-08-17
fix(core): Fix Android navigation crash

dgmachado:
## PR Checklist

Android navigation crash when receiving unexpected value on the Entry struct. The entry.resolvedPage is coming null in a Test App and I suppose the root case is that there's memory leak happening on the navigation.

## What is the new behavior?

Fix the Android navigation crash just ignoring the expected value received. It's a defensive programming and that fixes the this crash issue.

### DIFF
--- a/packages/core/ui/frame/fragment.transitions.android.ts
+++ b/packages/core/ui/frame/fragment.transitions.android.ts
@@ -672,17 +672,16 @@ function transitionOrAnimationCompleted(entry: ExpandedEntry, backEntry: Backsta
 
 	entries.delete(entry);
 	if (entries.size === 0) {
-		if (!entry.resolvedPage) {
-			return;
-		}
-		const frame = entry.resolvedPage.frame;
-
 		// We have 0 or 1 entry per frameId in completedEntries
 		// So there is no need to make it to Set like waitingQueue
 		const previousCompletedAnimationEntry = completedEntries.get(frameId);
 		completedEntries.delete(frameId);
 		waitingQueue.delete(frameId);
 
+		if (!entry.resolvedPage) {
+			return;
+		}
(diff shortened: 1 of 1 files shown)


CatchABus:
@dgmachado Perhaps, this fix needs more thought. Also, return statement may leave unexpected leftovers.
For instance, entry may still reside in waiting queue which may not be the desired result. See https://github.com/NativeScript/NativeScript/blob/90e7e34e9b7cb35cf6c97270dd47b6a9ea5ea5b7/packages/core/ui/frame/fragment.transitions.android.ts#L684

dgmachado:
> @dgmachado Perhaps, this fix needs more thought. Also, return statement may leave unexpected leftovers. For instance, entry may still reside in waiting queue, would this be alright? See
> 
> https://github.com/NativeScript/NativeScript/blob/90e7e34e9b7cb35cf6c97270dd47b6a9ea5ea5b7/packages/core/ui/frame/fragment.transitions.android.ts#L684

dgmachado:
> @dgmachado Perhaps, this fix needs more thought. Also, return statement may leave unexpected leftovers. For instance, entry may still reside in waiting queue, would this be alright? See
> 
> https://github.com/NativeScript/NativeScript/blob/90e7e34e9b7cb35cf6c97270dd47b6a9ea5ea5b7/packages/core/ui/frame/fragment.transitions.android.ts#L684

@CatchABus good catch. Now I'm updating the arrays completedEntries and waitingQueue properly.

edusperoni (approved):
Looks ok to me. I feel like this issue requires some more thought (how did resolvedPage ever become null? And we care about the frame, not necessarily the page), but this solution is good enough for now.

NathanWalker (approved):

PR #10008, opened 2022-08-31
feat(core): support for external XML UI compilers

CatchABus:
## PR Checklist

## What is the new behavior?
This is done in favor of https://github.com/nativescript-community/xml-ui-loader.
These changes are enough for the compiler mentioned above but a cleanup might be needed in the future to support further functionality.

For now, we don't have to update item templates that make use of `Builder.parse`.
Compiler returns templates as callbacks and `Builder.parse` happens to handle callbacks.
This is good for NS plugins too as they need no changes for the time being.

```ts
        static parse(value: string | Template, context?: any): View {
		if (typeof value === 'function') {
			return (<Template>value)();
		} else {
			const exports = context ? getExports(context) : undefined;
			const componentModule = parseInternal(value, exports);

			return componentModule && componentModule.component;
		}
	}
```


farfromrefug:
@CatchABus the point of __UI_USE_XML_PARSER__ was to remove the XML dependencies with webpack. I think ( not 100% sure ) you broke that. can you check that with __UI_USE_XML_PARSER__=false the XML parsing dependencies are removed from vendor?

CatchABus:
> @CatchABus the point of **UI_USE_XML_PARSER** was to remove the XML dependencies with webpack. I think ( not 100% sure ) you broke that. can you check that with **UI_USE_XML_PARSER**=false the XML parsing dependencies are removed from vendor?

I searched but couldn't find such functionality in webpack.
I think we simply set it to true during build time but do nothing more than relying on it for runtime checks.

I still find references of parser modules too:
Screenshot from 2022-09-01 11-43-58

farfromrefug:
@CatchABus It comes from terser. If you have `if (false) {}` then the whole if is removed. 
So using `UI_USE_XML_PARSER=false`  the whole xml + its import where removed from vendor.
You have to build in production to have the code removed

CatchABus:
> @CatchABus It comes from terser. If you have `if (false) {}` then the whole if is removed. So using `UI_USE_XML_PARSER=false` the whole xml + its import where removed from vendor. You have to build in production to have the code removed

Oh I see! Thanks for that note.
I'll add those checks back. Do you think the same behaviour applies for `if else` blocks?

farfromrefug:
@CatchABus yes it does on if/else too.

CatchABus:
@farfromrefug I added back all those `__UI_USE_XML_PARSER__` checks.
Those changes that remain are still serving the purpose of this PR.

CatchABus:
It seems this change alone is good for now. I've been using the NS compiler I'm working on for few days now and there was no need for further compatibility changes in core package.
With this addition, we provide functionality for https://github.com/nativescript-community/xml-ui-loader

NathanWalker (approved):

PR #10014, opened 2022-09-04
feat(ios/web-view): adds allowInlineMediaPlayback property

davecoffin:
New property on WebView (allowInlineMediaPlayback: boolean) will enable inline media playback in a webview on iOS. By default, webview forces iPhone into fullscreen media playback.


CatchABus:
Just a small note. Since it's an ios-specific property, perhaps we should append `ios` suffix like we do with other properties that target a specific platform.

Something like `iosAllowInlineMediaPlayback`.

NathanWalker:
> Just a small note. Since it's an ios-specific property, perhaps we should append platform name prefix like we do with other properties that target a specific platform.
> 
> Something like `iosAllowInlineMediaPlayback`.

Agree good idea, we'll adjust the naming prior to merge.

apps/toolbox/src/pages/webview.xml
```diff
@@ -0,0 +1,17 @@
...
+    <Page.actionBar>
+        <ActionBar title="WebView" class="action-bar">
+        </ActionBar>
+    </Page.actionBar>
+
+    <StackLayout class="p-10">
+      <Label text="iPhone by default plays media content in a webview in full screen, meaning it takes over your whole UI. You can toggle this on and off with a cool new property: allowInlineMediaPlayback" textWrap="true" style="color: gray;" class="m-b-10 v-center text-center" />
+
+      <Label style="color: gray;" textWrap="true" class="m-b-5 text-center" text="This webview plays this youtube content inline." />
+      <WebView allowInlineMediaPlayback="true" src="https://sphere.presonus.com/video/youtube/QnGX0xrv6Dw" height="200" />
```
NathanWalker, line 11: @Ombuweb let's update the property here: `iosAllowInlineMediaPlayback`

Ombuweb: @NathanWalker isn't it supposed to be `< WebView ios:AllowInlineMediaPlayback ...`?

CatchABus: @Ombuweb What you mentioned is a plain NS namespace convention that decides whether to apply properties on instances depending on running platform.
This isn't needed here. We just need an `ios` prefix to property so that we know it's only targeting iOS.
Hence, it should be `iosAllowInlineMediaPlayback`.

Ombuweb: @CatchABus Ah.. ok. Thanks

NathanWalker (dismissed):

Ombuweb (approved):

PR #10028, opened 2022-09-15
fix(ios): Tapping on span does not emit linkTap event

### LINKED ISSUE #9189: Tappable Span on IOS are not working
**Environment**
✔ CLI: 7.1.2
✔ Xcode version 12.4.0 satisfies minimum required version 10.
✔ Getting NativeScript components versions information...
✔ Component nativescript has 7.1.2 version and is up to date.
⚠ Update available for component @nativescript/core. Your current version is 7.0.13 and the latest available version is 7.1.4.
✔ Component @nativescript/ios has 7.1.1 version and is up to date.
✔ Component @nativescript/android has 7.0.1 version and is up to date.

**Describe the bug**
Tappable link are not working correctly on IOS when there is several span.

The "link" doesn't work  on Ios, but work on android. 

```
    <Label>
        <FormattedString>
            <Span
                    text="This is a"></Span>
            <Span (linkTap)="openLink()"
                  color="red"
                  text=" link"></Span>
       </FormattedString>
   </Label>
```

The method openLink is not executed 

```
 openLink() {
        console.log('openlink');
  }
```

Without the first Span, "link" is tappable : 
```
    <Label>
        <FormattedString>
            <Span (linkTap)="openLink()"
                  color="red"
                  text=" link"></Span>
        </FormattedString>
    </Label>
```

CatchABus:
## PR Checklist

In iOS, event `linkTap` is not emitted at all times or happens to emit for wrong tap point.

Issues:
1) Tap point calculations return wrong result if text alignment is not set to center
2) Tap point calculations return wrong result if label is multi-line
3) One can trigger `linkTap` event even if tap occurs outside text bounding area
4) One can trigger `linkTap` event even if tap occurs on a blank point inside text bounding area (method `characterIndexForPointInTextContainerFractionOfDistanceBetweenInsertionPoints` returns first or last character index when no character is found, depending on tap horizontal position)

## What is the new behavior?
An accurate `linkTap` event for iOS that meets the requirements to emulate android native tap for `SpannableStrings`.
It was essential to include paragraph style alignment for tap point calculations to work on multi-line labels.
This PR fixes all issues mentioned above.

It's a good feature when it comes to long, wrapped texts with links inside and there is a good tutorial where Alex introduces its functionality: https://www.youtube.com/watch?v=ACnk_0jOVp0&ab_channel=AlexZiskind

Fixes/Implements/Closes #9189.

### DIFF
--- a/packages/core/platforms/ios/src/UIView+NativeScript.m
+++ b/packages/core/platforms/ios/src/UIView+NativeScript.m
@@ -85,6 +85,7 @@
         } ];
     }
     
+    BOOL isLabel = [self isKindOfClass:[UILabel class]];
     if (lineHeight > 0) {
         NSMutableParagraphStyle *paragraphStyle = [[NSMutableParagraphStyle alloc] init];
         paragraphStyle.lineSpacing = lineHeight;
@@ -96,7 +97,7 @@
             paragraphStyle.alignment = ((UILabel*)self).textAlignment;
         }
         
-        if ([self isKindOfClass:[UILabel class]]) {
+        if (isLabel) {
             // make sure a possible previously set line break mode is not lost when line height is specified
             paragraphStyle.lineBreakMode = ((UILabel*)self).lineBreakMode;
         }
@@ -104,14 +105,14 @@
             0,
             attrText.length
         }];
-    } else {
+    } else if (isLabel || [self isKindOfClass:[UITextView class]]) {
         NSMutableParagraphStyle *paragraphStyle = [[NSMutableParagraphStyle alloc] init];
 
-        if ([self isKindOfClass:[UITextView class]]) {
-            paragraphStyle.alignment = ((UITextView*)self).textAlignment;
-        } else if ([self isKindOfClass:[UILabel class]]) {
+        if (isLabel) {
             // It's important to set paragraph alignment for link tap to work on multi-line spans as NSTextContainer takes it into account
             paragraphStyle.alignment = ((UILabel*)self).textAlignment;
+        } else {
+            paragraphStyle.alignment = ((UITextView*)self).textAlignment;
         }
         [attrText addAttribute:NSParagraphStyleAttributeName value:paragraphStyle range:(NSRange){
             0,
--- a/packages/core/platforms/ios/src/UIView+NativeScript.m
+++ b/packages/core/platforms/ios/src/UIView+NativeScript.m
@@ -92,6 +92,7 @@
         if ([self isKindOfClass:[UIButton class]]) {
             paragraphStyle.alignment = ((UIButton*)self).titleLabel.textAlignment;
         } else {
+            // Paragraph alignment is also important for tappable spans as NSTextContainer takes it into account
             paragraphStyle.alignment = ((UILabel*)self).textAlignment;
         }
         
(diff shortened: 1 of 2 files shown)


farfromrefug:
@CatchABus do you have an example that was failing before?

CatchABus:
> @CatchABus do you have an example that was failing before?

Apart from the open issue, I have confirmed this myself and studied its behaviour until I found the best solution according to apple docs.
Unfortunately, it never worked right for anything apart from single-line, centered texts.
Current code was taken from this like where few of problems were also reported in comments: https://stackoverflow.com/a/35789589
About bad point indexing, there is a similar report for UITextView that also applies to UILabel here: https://stackoverflow.com/questions/49441068/link-methods-get-called-after-the-last-character-in-attributed-text-for-uitextvi

```xml
<!-- Label with text align left (right will also do) -->
<Label textAlignment="left" width="100%">
    <FormattedString>
        <Span text="Click to open link" linkTap="openLink">
        </Span>
    </FormattedString>
</Label>

<!-- Centered label with multiple lines -->
<Label textAlignment="center" textWrap="true">
    <FormattedString>
        <Span text="I am a paragraph with looots of words that wastes your time reading because I'm a humble little sample. If you are brave enough to lose more time, click this">
        </Span>
        <Span text="  link" linkTap="openLink">
        </Span>
    </FormattedString>
</Label>
```

farfromrefug:
@CatchABus thanks ! I want to make sure we don't have the issue either in @nativescript-community/ui-label

CatchABus:
> @CatchABus thanks ! I want to make sure we don't have the issue either in @nativescript-community/ui-label

I recall reading somewhere that UITextView has a built-in support for interaction with links or something like that. I believe that `@nativescript-community/ui-label` has the potential to support this with less hard coding. :D

farfromrefug:
@CatchABus Ok you are right i use UITextView. Now the refactoring which was done in N a while back with `nativeScriptSetFormattedTextDecorationAndTransform` broke `ui-label` and the way it was creating `NSAttributedString` Before it was using its own way but now it cant anymore and is forced to use `nativeScriptSetFormattedTextDecorationAndTransform`.
I hope it would work again and i think i created an issue for that.

Anyway also found out right now that N `nativeScriptSetFormattedTextDecorationAndTransform` seems to be broken for UITextView. Seeing it does not add `NSLinkAttributeName` then `textView:shouldInteractWithURL:inRange:interaction:` cant catch it.

Any chance you can fix it as part of this PR? You can simply use `@nativescript-community/ui-label` to see the issue

CatchABus:
@farfromrefug Then, would it make more sense if we made it flexible for everyone?
I was thinking if we added a callback that gets called before mutable string is assigned to view text property.
Like:
```js
nativeScriptSetFormattedTextDecorationAndTransformLetterSpacingLineHeightBeforeTextChange(details, letterSpacing, lineHeight, function(attributeString) {
    // Assign stuff to attribute string before it gets applied to native view
});
```
What do you think about this?

farfromrefug:
@CatchABus i am not sure the callback would help.  I mean N uses UITextView so you would expect N to support links for it. By that i mean better to handle it natively like it is done for UILabel. (will be faster too).
Does that makes sense to you?

CatchABus:
> @CatchABus i am not sure the callback would help. I mean N uses UITextView so you would expect N to support links for it. By that i mean better to handle it natively like it is done for UILabel. (will be faster too). Does that makes sense to you?

I think N is using UITextView mostly for editable text right now. Even so, adding complete support to N UITextView may be a new feature as I see it.
I believe we can add a minimal support for now so that community plugins can use it easily, then use it for N UITextView in the future if needed.

> Anyway also found out right now that N `nativeScriptSetFormattedTextDecorationAndTransform` seems to be broken for UITextView. Seeing it does not add `NSLinkAttributeName` then `textView:shouldInteractWithURL:inRange:interaction:` cant catch it.

Oh you are talking about UITextView view in N. I think it's only used in `HtmlView` for readonly text, however this view does not extend TextBase and this change may need refactoring.
The rest of N views use it for editable text, so I wonder if we need this there.

It also seems N had no support for tapping links in UITextView even before switching to `nativeScriptSetFormattedTextDecorationAndTransform`.

I think we can merge this one for now and work on the other thing separately.

farfromrefug:
@CatchABus Actually my bad you are right i thought it was working before in N but it wasnt :D 
So 100% agree with you.
Merge this i will find a way to create a PR to fix my plugin

NathanWalker (approved):

PR #10030, opened 2022-09-17
fix(core): Errors get swallowed if thrown inside async event functions

CatchABus:
## PR Checklist

Here is a small scenario. I have this `Page` event callback.
```js
function onNavigatedTo(args) {
    // For a reason, I wish to navigate elsewhere
    Frame.topmost().navigate({moduleName: "I accidentally put an invalid path"});
}
```
There is faulty code inside it that returns a blank page and throws error in console.
```sh
Error: Failed to load component from module: I accidentally put an invalid path
```

So far so good. What if I want to use an asynchronous callback though? Let's say that I would like to use async-await because I like it more than promises.

My callback will look like this:
```js
async function onNavigatedTo(args) {
    var response = await axios.get(...);

    // For a reason, I wish to navigate elsewhere
    Frame.topmost().navigate({moduleName: "I accidentally put an invalid path"});
}
```
Even though the result is the same, error thrown by `Frame.topmost().navigate` will be shallowed because it's thrown inside an asynchronous function.
One would use try-catch to handle this problem in his own custom code, but this problem occurs even with N calls.
As a result, one gets a blank page but has no idea what is happening.

## What is the new behavior?
Depending on whether event callback execution is asynchronous or not, we append a promise `catch` to it and print the error that gets swallowed.

I couldn't think of a better solution to this matter. What I'm sure of is that we have to be informed about throws as error messages usually lead us to the cause of the problem.
I'm open to suggestions on improving this if any.

### DIFF
--- a/packages/core/data/observable/index.ts
+++ b/packages/core/data/observable/index.ts
@@ -303,7 +303,7 @@ export class Observable implements ObservableDefinition {
 			}
 
 			// This ensures errors thrown inside asynchronous functions do not get swallowed
-			if (call instanceof Promise) {
+			if (call && call instanceof Promise) {
 				call.catch((err) => {
 					console.error(err);
 				});
--- a/packages/core/data/observable/index.ts
+++ b/packages/core/data/observable/index.ts
@@ -294,10 +294,19 @@ export class Observable implements ObservableDefinition {
 			if (entry.once) {
 				observers.splice(i, 1);
 			}
+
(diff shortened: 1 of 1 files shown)


NathanWalker:
Pretty good solution, I believe this may help quite a number of conditions I’ve seen over the years. Would be curious others opinions but I like it.

rigor789 (approved):

PR #10040, opened 2022-09-25
fix(core): Color.isValid returned true for null/undefined

CatchABus:
## PR Checklist

For a strange reason, method `Color.isValid` was designed to consider `null` and `undefined` values to be valid colors.

## What is the new behavior?
Method `Color.isValid` will return false if given value is `null` or `undefined`.
Old check was removed completely as `isString` check will handle the possibility of a nullish value.


edusperoni:
This might introduce an issue on equality comparers where a null color can be expected. I don't know where specifically, but I remember there was one case this might be expected. I believe this fix is correct, but we might want to check where this is used and may need to add special conditions for null colors

CatchABus:
> This might introduce an issue on equality comparers where a null color can be expected. I don't know where specifically, but I remember there was one case this might be expected. I believe this fix is correct, but we might want to check where this is used and may need to add special conditions for null colors

Afaik, we use `Color.equals` for equality comparers. I did a quick search inside core package and it seems there is only 1 place we use it right now: https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/animation/animation-common.ts#L160

NathanWalker (approved):

PR #10045, opened 2022-10-01
chore(webpack): update webpack deps for react-nativescript

shirakaba:
## PR Checklist

- [x] There is an issue for the bug/feature this PR is for.

Outdated deps for RNS.

## What is the new behavior?
Up-to-date deps for RNS.

I've only updated dependencies within their originally specified version range, so there should be no breaking changes.


packages/webpack5/package.json
```diff
@@ -17,27 +17,27 @@
 		"prepack": "npm run test && npm run build && npm run copy-stubs && chmod +x dist/bin/index.js"
 	},
 	"dependencies": {
-		"@babel/core": "^7.0.0",
-		"@pmmmwh/react-refresh-webpack-plugin": "^0.5.2",
+		"@babel/core": "^7.19.3",
```
rigor789, line 20:
```suggestion
		"@babel/core": "^7.0.0",
```

Keeping the wider range should still resolve to 7.19.3 but allow deduping dependencies where a different babel/core is installed instead of producing broken builds or duplicate dependencies...

packages/webpack5/package.json
```diff
@@ -17,27 +17,27 @@
...
+		"@pmmmwh/react-refresh-webpack-plugin": "~0.5.7",
 		"acorn": "^8.0.0",
 		"acorn-stage3": "^4.0.0",
-		"babel-loader": "^8.0.0",
+		"babel-loader": "^8.2.5",
```
rigor789, line 24:
```suggestion
		"babel-loader": "^8.0.0",
```

packages/webpack5/package.json
```diff
@@ -17,27 +17,27 @@
...
 		"chalk": "^4.0.0",
 		"cli-highlight": "^2.0.0",
 		"commander": "^8.0.0",
 		"copy-webpack-plugin": "^9.0.0",
 		"css": "^3.0.0",
 		"css-loader": "^6.0.0",
 		"dotenv-webpack": "^7.0.0",
-		"fork-ts-checker-webpack-plugin": "^7.0.0",
+		"fork-ts-checker-webpack-plugin": "^7.2.13",
```
rigor789, line 32:
```suggestion
		"fork-ts-checker-webpack-plugin": "^7.0.0",
```

rigor789 (changes requested):
I'll apply my review suggestions - widening the ranges is important for `@nativescript/webpack` since it has to be able to deal with a wide variety of projects that may also pull in some of these dependencies. If we use a very strict version range, we are making it difficult to run some projects, or would cause duplicate dependencies with different versions (which is also bad, and can cause issues - webpack for instance is very sensitive to that, because it uses `instanceof` checks that fail if the same class was imported from a different version).

rigor789 (approved):

PR #10052, opened 2022-10-11
fix(core): android refactor EditableTextBase

farfromrefug:
to allow sub classes to change/augment delegates.
also unset delegates in disposeNativeView to prevent memory leak

We used that same approach somewhere else but i dont remember where :P 

Now this refactor will allow me to rewrite the input-mask plugin and have it compatible with all N features and will all components inheriting  `EditableTextBase`

It should not have any consequences and i did no see any bugs. But please read the code to ensure we are all ok


NathanWalker (approved):

vallemar:
@farfromrefug  Have you done anything anywhere with the mask plugin? I'm trying to make it work on ios and there is no way. It works fine on android but ios...

PR #10069, opened 2022-10-27
feat: switch core http to okhttp

edusperoni:
WIP. Will fill details later.

Notes:

* needs to use the alpha version of the runtime (fixes to kotlin/jvm methods)
* need to wrap okhttp in an async task so we can process the body in the background
* ideally be more lazy with the http processing
* adapt the devtools agent to recieve okhttp results

other than doing some body decoding in the main thread, this implementation already covers all of the current API + cancellation. Ideeally we should have the web spec AbortController ready for this

### DIFF
--- a/packages/core/http/http-request/index.android.ts
+++ b/packages/core/http/http-request/index.android.ts
@@ -22,7 +22,13 @@ function parseJSON(source: string): any {
 }
 
 let requestIdCounter = 0;
-const pendingRequests = {};
+interface Call {
+	url: string;
+	resolveCallback: (value: httpModule.HttpResponse | PromiseLike<httpModule.HttpResponse>) => void;
+	rejectCallback: (reason?: any) => void;
+	call: okhttp3.Call;
+}
+const pendingRequests: Record<number, Call> = {};
 
 let imageSource: typeof imageSourceModule;
 function ensureImageSource() {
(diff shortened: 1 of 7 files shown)


packages/core/http/http-request/index.android.ts
```diff
@@ -55,43 +61,47 @@ function ensureCompleteCallback() {
...
+	const raw = new java.io.ByteArrayOutputStream(bytes.length);
+	raw.write(bytes, 0, bytes.length);
+	result.body().close();
 
 	callbacks.resolveCallback({
 		content: {
-			raw: result.raw,
-			toArrayBuffer: () => Uint8Array.from(result.raw.toByteArray()).buffer,
+			raw: raw,
+			toArrayBuffer: () => Uint8Array.from(raw.toByteArray()).buffer,
```
mukaschultze, line 98: `Uint8Array.from` iterates over each element of the source array and copies it to the new array.

If you use pass a `java.nio.ByteBuffer` as the argument it'll create the TypedArray from a pointer (O(1) performance and no extra memory allocation).

Creating a buffer from 1MB of data takes about ~700ms, wrapping it in a `java.nio.ByteBuffer` takes only 3ms

```ts
toArrayBuffer: () => {
    const arr = raw.toByteArray();
    const bb = java.nio.ByteBuffer.wrap(arr);
    const buffer = ArrayBuffer.from(bb);
    return buffer;
},
```

PR #10075, opened 2022-11-02
fix(core): iOS Tab page display issue if showing message dialog

dgmachado:
## PR Checklist

If the tabstripitem added with tap event which displays a message dialog, the tab page is not being displayed.

iOS issue detail: "_.... called on <UINavigationControllerImpl 0x14580cc00> while an existing transition or presentation is occurring; the navigation stack will not be updated._"

Reproduce steps:

- Download and run the attached app. (run 'npm i' before running the app)
- Switch to 2nd tab.
- Notice the message dialog is being displayed from the tab item tap event, after closing the dialog, the tab page remains blank.

Before Fix

## What is the new behavior?

Check if an existing transition or presentation is occurring and push the new navigation only after the transition completion.

After Fix

## Example App

[testPopover_tabBar3.zip](https://github.com/NativeScript/NativeScript/files/9922674/testPopover_tabBar3.zip)


farfromrefug:
@dgmachado very interesting fix for something which have been bugging us all for a long time. Will it also apply to showing modal just after a message dialog?

dgmachado:
> @dgmachado very interesting fix for something which have been bugging us all for a long time. Will it also apply to showing modal just after a message dialog?

@farfromrefug this apply also to showing a message dialog just after the modal but the inverse I'm not convinced. If you have a sample app I could take a look.

farfromrefug:
@dgmachado  I can write one tomorrow. But I think simply opening a modal from a alert dialog button callback should trigger the issue. Just had it today (working around it with timeouts)

NathanWalker (approved):

NathanWalker:
@farfromrefug if you end up with a secondary repro on the alternate path you mention we can handle in separate PR.

PR #10080, opened 2022-11-08
fix(webpack): close compiler after run

mukaschultze:
## PR Checklist

Webpack compiler is not closed, this breaks features that rely on the compiler shutdown hook (for example, caching). 

## What is the new behavior?

Now the compiler is closed after running, same behaviour as the webpack CLI.

https://github.com/webpack/webpack/blob/main/lib/webpack.js#L146-L150

Relevant comment about this: https://github.com/webpack/webpack/issues/12345#issuecomment-756099417

### DIFF
--- a/packages/webpack5/src/bin/index.ts
+++ b/packages/webpack5/src/bin/index.ts
@@ -150,7 +150,14 @@ program
 				webpackCompilationCallback
 			);
 		} else {
-			compiler.run(webpackCompilationCallback);
+			compiler.run((err, status) => {
+				compiler.close((err2) =>
+					webpackCompilationCallback(
+						(err || err2) as webpack.WebpackError,
+						status
+					)
+				);
+			});
 		}
 	});
 


rigor789:
Wow! Good catch, I'll give this a try before merging just to make sure, but looks good otherwise! Thank you!

rigor789:
Cool, everything looks good - enabling the FS cache requires adding: 
```ts
webpack.chainWebpack(config => {
  config.cache({
    type: 'filesystem',
  })
}
```

We'll consider that as a default, but should be handled as a separate issue/pr.

In a semi-complex app it saves ~10% (in my test case 400ish ms) but I think it depends on the project.

rigor789 (approved):

mukaschultze:
I'm getting almost 2x speedup by using filesystem caching in a real project. Down from 30 secs to 16 secs. The only issue is that the caching uses a considerable amount of disk space (~500 MB per app).

rigor789:
> 

Nice, I think we should in that case just document how to enable it - rather than setting it by default.

edusperoni:
maybe we should enable as a compiler flag. Currently I have in some projects this function for using the angular cache:

```
function setupCache(config, env) {
  const cacheEnabled = !!env.ngCache;
  if (cacheEnabled) {
    const cacheDirectory = join(webpack.Utils.project.getProjectRootPath(), '.angular', 'cache');
    const oldOpt = config.module.rule('angular-webpack-loader').use('webpack-loader').get('options');
    config.module.rule('angular-webpack-loader').use('webpack-loader').options({
      ...oldOpt,
      cacheDirectory: join(cacheDirectory, 'babel-webpack'),
      optimize: true
    });
    config.cache({
      type: 'filesystem',
      profile: false, //wco.buildOptions.verbose,
      cacheDirectory: join(cacheDirectory, 'angular-webpack'),
      maxMemoryGenerations: 1,
      // We use the versions and build options as the cache name. The Webpack configurations are too
      // dynamic and shared among different build types: test, build and serve.
      // None of which are "named".
      name: createHash('sha1')
        .update(ngVersion)
        // .update(VERSION) // this was from angular-devkit stuff
        .update(webpack.Utils.project.getProjectRootPath())
        .update(JSON.stringify(webpack.Utils.tsconfig.readTsConfig('tsconfig.app.json').options))
        // maybe add tsconfig? not sure
        .update(
          JSON.stringify({
            env,
            // ...wco.buildOptions,
            // Needed because outputPath changes on every build when using i18n extraction
            // https://github.com/angular/angular-cli/blob/736a5f89deaca85f487b78aec9ff66d4118ceb6a/packages/angular_devkit/build_angular/src/utils/i18n-options.ts#L264-L265
            outputPath: undefined,
          })
        )
        .digest('hex'),
    });
    return;
  }
  // if (env.watch) {
  //   config.cache({
  //     type: 'memory',
  //     maxGenerations: 1,
  //   });
  //   return;
  // }

}
```

This was mostly taken from their cli, but oddly enough they use memory cache in watch mode (I commented that out)

PR #10125, opened 2022-12-07
fix(android): box shadow and border radius white border resolution

vallemar:
When you have a dark screen and you apply box-shadow and border radius you see a rather strange white border. Applying this change solves it. I haven't seen any difference on a clear screen, so I think it doesn't break anything

Old:
Screenshot_2022-12-07-13-49-01-283-edit_org nativescript abllapp

Fix:
Screenshot_2022-12-07-13-47-43-779-edit_org nativescript abllapp


rigor789:
Mentioned on Discord already, just adding here for visibility:

The WHITE rect layer is added to mimic the web behavior:
image

So setting it to `TRANSPARENT` is redundant, as we might as well not even add the overlay layer in that case. A proposed solution is to check whether the view already has a background, and only in case it doesn't add the white rect overlay, otherwise skip adding the overlay layer at all.

vallemar:
@rigor789  I had misunderstood you, I thought it was just the passing of the color, now the layer will not be added if it already has a background

vallemar:
can we review this pr?

packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/BoxShadowDrawable.java
```diff
@@ -42,13 +42,15 @@ public BoxShadowDrawable(Drawable wrappedDrawable, String value) {
...
 
-		this.shadowLayer = new ShapeDrawable(new RectShape());
-		this.overlayLayer = this.createOverlayLayer();
 		this.wrappedLayer = wrappedDrawable;
+		this.shadowLayer = new ShapeDrawable(new RectShape());
 
 		// add our layers
 		this.addLayer(shadowLayer);
-		this.addLayer(overlayLayer);
+		if(!((BorderDrawable) this.wrappedLayer).hasBackgroundColor()){
```
rigor789, line 50: This might be unsafe, `this.wrappedLayer` might not always be a BorderDrawable, so casting to it can lead to exceptions. A safer approach would be to check the type of `wrappedLayer`.

rigor789 (changes requested):

vallemar:
@rigor789 @NathanWalker I have updated the PR with the type check

rigor789 (dismissed):
Looks good to me, just left some small nits about the method naming. 👍

rigor789 (approved):

rigor789:
Simplified the logic to hide the overlay layer when the wrapped drawable has a background color, image or gradient. The above issue was caused by changing the order of the `addLayer` calls mainly. Might still try not creating the layer at all - but the order of layers important.

rigor789:
Refactored to not create the layer as originally intended - but preserving the order of layers now correctly.

vallemar:
@rigor789 oh sorry, I thought the order wouldn't matter and put it in an order that I thought was more readable

NathanWalker (approved):

Issue #10092, opened 2022-11-15
IOS iPhone 13 bug, iPhone 14 working, Wrong ActionBar Offset in nested page-router-outlet, likely due to  iosOverflowSafeArea used to compute actionbar offset in nested outlet


cjohn001:
Update: I have just tried the latest ns/core with Angular 15 and the issue still persists. However, I have an additional important info in regards to the issue.

When I use an iPhone 14 simulator with iOS 16.1 the issue is not appearing. However, using other simulators with iOS16.1 the bug persists. I tried with iPhone 13 simulator and iOS 16.1 and iPad Pro 11-inch tablet simulator with iOS 16.1
and on a real iPhone 13 with 16.1.1 and 16.1.2.

The issue seems to be somehow related to  iosOverflowSafeArea being used in nested page-router-outlet to compute the offset of the top element, even though the actionbar is disabled. This is likely the source of the problem.

I tried to find a workaround for the problem. The following code allows to get rid of the extra spacing on the top. However, than the extra spacing appears on the bottom of the page as the layout is not computed for the page without actionbar :(  
Would be great if someone could provide help!

```
public onPageRouterOutletLoaded(args: EventData) {
		const pro = args.object as unknown as Frame;
		pro.iosIgnoreSafeArea = true;
}

<page-router-outlet actionBarVisibility="never" name="outlet-0" (loaded)="onPageRouterOutletLoaded($event)">
		</page-router-outlet>

```

-------------------------------------------------------------------------
### Issue Description

Hello together,
I have a self build tabs control which uses 4 nested page-router-outlets.
I just updated from "@nativescript/core": "8.3.5" to "@nativescript/core": "8.3.6" and run into a problem, 

When I navigate to the page and with it navigate the nested page-router-outlets, they get an actionBar offset which should not be there as the actionBar is disabled for the nested page-router-outlets.

When I only navigate one nested page initially, and than load the other nested pages with some timeout, I can see that the pages navigated with timeout display correctly, and only the initially loaded page has the wrong ActionBar offset.

Seems like the layout for the actionbar is rescaled somehow too late for the first page to render correctly. First image shows the first page loaded when transitioning to the page (wrong ActionBar inset), second one further pages in different page router outlets loaded with some timeout involved:

```
<GridLayout rows="46,2,*" columns="*,*,*,*">
	<!-- white tab bar -->
[...]

Issue #10067, opened 2022-10-25
[iOS 14] ActionBar not respecting safe area, shown under the status bar


fpaaske:
### Issue Description

I have an app using nested page routing outlets, and the ActionBar on the _first tab_ of the _nested page router outlet_ will appear under the status bar. The other tabs works as expected.

This only happens on iOS 14.x.

It happened after moving from @nativescript-community/ui-material-bottom-navigation to @nativescript-community/ui-material-bottomnavigationbar. However, it also happens when I remove the usage of the ui-material-bottomnavigationbar as well, and just use a custom tab bar, so it's probably not really linked to that.

There's a similar issue reported here: https://github.com/nativescript-community/ui-material-components/issues/308

I've tried to set padding-/marginTop to force it down, without luck.

This is how it (correctly) looks on iOS 12, 15 and 16.
![image](https://user-images.githubusercontent.com/1579581/197797356-eafb1ea2-4013-44b7-b31b-d98e2df4415a.png)

![image](https://user-images.githubusercontent.com/1579581/197804169-0477833d-d88a-4185-a2e6-1c5eb2fcb298.png)
---
Here's iOS 14 on iPhone 8
![image](https://user-images.githubusercontent.com/1579581/197804351-796b1804-4ff0-4add-8f37-7293218945ea.png)

![image](https://user-images.githubusercontent.com/1579581/197797543-0093ab4a-b3cd-4fef-84d0-caa4155a609d.png)

iPhone X
![image](https://user-images.githubusercontent.com/1579581/197804425-dfb87cd8-49fb-4c81-850c-b62260727c0f.png)

![image](https://user-images.githubusercontent.com/1579581/197797588-cbf8e05d-d0b5-4990-ac72-ee1e824b15fc.png)

iPhone 12 Pro
![image](https://user-images.githubusercontent.com/1579581/197804615-4209eccc-235b-4f16-9185-bd09b0ca0c86.png)

![image](https://user-images.githubusercontent.com/1579581/197797625-21ca104e-8a16-4d2a-a025-5ba69e1cb9dc.png)

### Reproduction

This is from another, simpler app, but same issue.

app.component

```
<page-router-outlet></page-router-outlet>
```

```
const routes: Routes = [
  {
    path: "",
    redirectTo: "tabs/default",
    pathMatch: "full",
  },
  {
    path: "tabs",
    loadChildren: () =>
      import("./features/tabs/tabs.module").then((m) => m.TabsModule),
  },
];
```

[...]

PR #10044, opened 2022-09-30
fix(ios): ensure view layout when using large titles

edusperoni:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/tools/notes/CONTRIBUTING.md#commit-messages.
- [ ] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/tools/notes/DevelopmentWorkflow.md#running-unit-tests-application.
- [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/tools/notes/WritingUnitTests.md.

## What is the current behavior?
When using large titles with a listview, you'll be unable to scroll up and reveal the large title due to us recalculating the layout on viewDidLayoutSubviews only, constatly glitching while scrolling to top.

This happens because iOS will NOT call `viewDidLayoutSubviews` in those scenarios, only `viewSafeAreaInsetsDidChange`, so there's a size mismatch and the glitch happens

## What is the new behavior?
We now recalculate the layout when safe area insets change. ListView and ScrollView users might still want to use `iosIgnoreSafeArea` to avoid costly layout recalculation as those views' sizes are mostly managed by iOS once large titles are enabled.

PR #10134, opened 2022-12-16
fix(core/ios): reset additional insets if they're zero

mukaschultze:
## PR Checklist

Sometimes `additionalSafeAreaInsets` are correctly set in a layout pass but should be reset back to zero if the size if a sibling view changes. This doesn't happen and causes the safezone to be incorrectly calculated for the view that had the `additionalSafeAreaInsets` set.

## What is the new behavior?

`additionalSafeAreaInsets` is set back to null if the newly calculated additional insets are zero.

### DIFF
--- a/packages/core/ui/core/view/view-helper/index.ios.ts
+++ b/packages/core/ui/core/view/view-helper/index.ios.ts
@@ -80,6 +80,8 @@ class UILayoutViewController extends UIViewController {
 							right: 0,
 						});
 						this.additionalSafeAreaInsets = additionalInsets;
+					} else {
+						this.additionalSafeAreaInsets = null;
 					}
 				}
 			}
--- a/packages/core/ui/page/index.ios.ts
+++ b/packages/core/ui/page/index.ios.ts
@@ -325,6 +325,8 @@ class UIViewControllerImpl extends UIViewController {
 							right: 0,
 						});
 						this.additionalSafeAreaInsets = additionalInsets;
+					} else {
+						this.additionalSafeAreaInsets = null;
 					}
 				}
 			}


rigor789 (approved):
Looks good to me - nice catch 👍

CatchABus:
@mukaschultze @rigor789 Unfortunately, this is not fixing #10092 
It's not that we don't know what's causing this, it's just that fix is not yet released.

After trying to look what causes the problem, @cjohn001 and I concluded the following patches are currently causing troubles:
https://github.com/NativeScript/NativeScript/pull/10044
https://github.com/NativeScript/NativeScript/pull/9773

See https://discord.com/channels/603595811204366337/606457628729671691/1048331850315202654

@edusperoni Removing each one of these separately was fixing the problem so I suspect a conflict between the two of them.
I tend to think we need to re-examine #9773 specifically and see if it's still needed there.

@mukaschultze About the case of `additionalSafeAreaInsets`, I'm not sure if we ever get cases that return zero for top and bottom for view controllers that "already" got positive additional insets (never checked tbh), but I definitely like precautions like this one.

cjohn001:
does not solve https://github.com/NativeScript/NativeScript/issues/10092#issuecomment-1356216580

mukaschultze:
> @mukaschultze About the case of `additionalSafeAreaInsets`, I'm not sure if we ever get cases that return zero for top and bottom for view controllers that "already" got positive additional insets (never checked tbh), but I definitely like precautions like this one.

@CatchABus it's possible, although rare, to get zero for controllers that had the insets already set. For example, if we have something like:

```html
<GridLayout rows="auto, *" #1>
	<GridLayout row="1" *ngIf="someConditionThatChangesDuringRouteChange" #2></GridLayout>
	<GridLayout row="2" #3>
		<page-router-outlet><page-router-outlet>
	</GridLayout>
</StackLayout>
```

When the `someConditionThatChangesDuringRouteChange` changes it will lead to the `Page` (`page-router-outlet`) having an inset even though the inset was set in the GridLayout#2 when it was created.

This is why I thought this PR might fix #10092 too because it's a very similar issue to the one I'm facing. I've tested removing the `layoutIfNeeded` and recalculating the additional insets during the `viewSafeAreaInsetsDidChange` but none of that worked for the example above.

CatchABus:
> > @mukaschultze About the case of `additionalSafeAreaInsets`, I'm not sure if we ever get cases that return zero for top and bottom for view controllers that "already" got positive additional insets (never checked tbh), but I definitely like precautions like this one.
> 
> @CatchABus it's possible, although rare, to get zero for controllers that had the insets already set. For example, if we have something like:
> 
> ```
> <GridLayout rows="auto, *" #1>
> 	<GridLayout row="1" *ngIf="someConditionThatChangesDuringRouteChange" #2></GridLayout>
> 	<GridLayout row="2" #3>
> 		<page-router-outlet><page-router-outlet>
> 	</GridLayout>
> </StackLayout>
> ```
> 
> When the `someConditionThatChangesDuringRouteChange` changes it will lead to the `Page` (`page-router-outlet`) having an inset even though the inset was set in the GridLayout#2 when it was created.
> 
> This is why I thought this PR might fix #10092 too because it's a very similar issue to the one I'm facing. I've tested removing the `layoutIfNeeded` and recalculating the additional insets during the `viewSafeAreaInsetsDidChange` but none of that worked for the example above.

Thank you very much for clarification. Hopefully, this may solve issues like #10067

CatchABus (approved):

PR #10141, opened 2022-12-21
fix(webpack): notify CLI even if there are compilation errors

rigor789:
## PR Checklist

When there's a compilation error, webpack may emit files (depending on how it's configured), however the CLI is never notified of these files, so they are not synced to the device. This is generally the right behavior however there are cases where this isn't ideal, most notably with HMR. When there's a compilation error, a new hash is assigned to the compilation, and files are written but never synced. After fixing the error, yet another hash is generated, however the app will not be able to apply hmr because it must first apply the "errored" hash, and then the fixed one - it works like a sequential chain, and if we don't emit the errored chain links, we break HMR.

## What is the new behavior?

The CLI is notified regardless of errors in the compilation. This may crash the app - however that's expected in some cases. In other cases this should fix HMR breakage after a single error during development.

References:
- https://github.com/nativescript-vue/nativescript-vue/issues/1013


ammarahm-ed:
I hope we can merge this & push a new webpack release. This breaks HMR quite often & i need to restart the CLI time & again.

PR #10142, opened 2022-12-21
fix(core/ios): box shadow value border radius

vallemar:
Without these changes the border with box-shadow was not visible, now the border is clearly visible. The problem when it is a number is that it enters here: https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/styling/style-properties.ts#L78 and the resulting number is much greater than the edge and the view is lost

Without changes
image

image

With changes:
image

image


vallemar:
for this example I am using the class `rounded-3xl`

NathanWalker (dismissed):

NathanWalker (approved):

Issue #10147, opened 2022-12-28
Android - WebView issue - shouldOverrideUrlLoading


prabudevarrajan:
### Issue Description

Android app is unable to handle custom URLs. Due to the following [change](https://github.com/NativeScript/NativeScript/commit/655f3a45c1179088863c8304fd6d72df657ec013#r94456333).

For e.g. if my app has a custom URL bookkeeper://action=close which is used to close the current screen

```
 webview.on(WebView.loadStartedEvent, function (args: LoadEventData) {  
      if ((args.url).indexOf('bookkeeper://action=close') === 0) { // This code is used to override the default webview for loading the URL
                webview.stopLoading();
                this.routerExtensions.back();
      }
}
```

What is happening is shouldOverrideUrlLoading is getting called in android webView and directly passing to Util.openURL API. So, the loadStartedEvent is not getting called. 

### Reproduction

Reproducible in nativescript android core 8.4.0 and above

$ ns run/debug android 

### Relevant log output (if applicable)

```shell
OS: macOS 13.0.1
CPU: (8) x64 Apple M1
Shell: /bin/bash
node: 16.15.0
npm: 8.5.5
nativescript: 8.4.0

# android
java: 18.0.1
ndk: Not Found
apis: Not Found
build_tools: Not Found
system_images: Not Found

# ios
xcode: 14.1/14B47b
cocoapods: 1.11.3
python: 3.9.13
python3: 3.9.13
ruby: 2.6.10
platforms: 
  - DriverKit 22.1
  - iOS 16.1
  - macOS 13.0
  - tvOS 16.1
  - watchOS 9.1


### Dependencies

```json
"dependencies": {
  "@angular/animations": "~14.2.0",
  "@angular/common": "~14.2.0",
  "@angular/compiler": "~14.2.0",
  "@angular/core": "~14.2.0",
  "@angular/forms": "~14.2.0",
  "@angular/platform-browser": "~14.2.0",
  "@angular/platform-browser-dynamic": "~14.2.0",
  "@angular/router": "~14.2.0",
  "@bradmartin/nativescript-urlhandler": "^2.0.1",
  "@nativescript-community/ble": "^3.1.11",
  "@nativescript-community/drawingpad": "^4.1.0",
  "@nativescript-community/perms": "^2.2.21",
  "@nativescript/angular": "^14.2.0",
  "@nativescript/appversion": "^2.0.0",
  "@nativescript/background-http": "^6.0.0",
[...]

PR #10148, opened 2022-12-29
fix(android): Fixes web-view handling url (#10147)

prabudevarrajan:
## PR Checklist

## What is the new behavior?

### DIFF
--- a/packages/core/ui/web-view/index.android.ts
+++ b/packages/core/ui/web-view/index.android.ts
@@ -2,6 +2,7 @@ import { disableZoomProperty, WebViewBase, WebViewClient } from './web-view-comm
 import { Trace } from '../../trace';
 import { knownFolders } from '../../file-system';
 import { openUrl } from '../../utils';
+import * as application from '../../application';
 
 export * from './web-view-common';
 
@@ -27,8 +28,19 @@ function initializeWebViewClient(): void {
 				Trace.write('WebViewClientClass.shouldOverrideUrlLoading(' + url + ')', Trace.categories.Debug);
 			}
 
+			if (android.webkit.URLUtil.isNetworkUrl(url)) {
+				return false;
+			}
+
 			// Handle schemes like mailto, tel, etc
-			if (!android.webkit.URLUtil.isNetworkUrl(url)) {
+			// Make sure to add  <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES"/>  to AndroidManifest.xml
+			// We check if there is any app installed on the device to handle the special URL(mailto, tel, etc)
+			const intent = new android.content.Intent(android.content.Intent.ACTION_VIEW);
+			intent.setData(android.net.Uri.parse(url));
+
+			const packageManager = application.android.context.getPackageManager();
+			if (intent.resolveActivity(packageManager) != null) {
+				// App exists to handle the URL. So, open the app
 				openUrl(url);
 				return true;
 			}
--- a/packages/core/ui/web-view/index.android.ts
+++ b/packages/core/ui/web-view/index.android.ts
@@ -1,8 +1,7 @@
 import { disableZoomProperty, WebViewBase, WebViewClient } from './web-view-common';
 import { Trace } from '../../trace';
 import { knownFolders } from '../../file-system';
-import { openUrl } from '../../utils';
-import * as application from '../../application';
+import { openUrl, ad } from '../../utils';
 
 export * from './web-view-common';
 
@@ -38,7 +37,7 @@ function initializeWebViewClient(): void {
 			const intent = new android.content.Intent(android.content.Intent.ACTION_VIEW);
 			intent.setData(android.net.Uri.parse(url));
 
-			const packageManager = application.android.context.getPackageManager();
+			const packageManager =  ad.getApplicationContext().getPackageManager();
 			if (intent.resolveActivity(packageManager) != null) {
 				// App exists to handle the URL. So, open the app
 				openUrl(url);
--- a/packages/core/ui/web-view/index.android.ts
+++ b/packages/core/ui/web-view/index.android.ts
@@ -1,7 +1,7 @@
 import { disableZoomProperty, WebViewBase, WebViewClient } from './web-view-common';
 import { Trace } from '../../trace';
 import { knownFolders } from '../../file-system';
-import { openUrl, ad } from '../../utils';
+import { openUrl } from '../../utils';
 
 export * from './web-view-common';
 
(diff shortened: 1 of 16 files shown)


packages/core/ui/web-view/index.android.ts
```diff
@@ -2,6 +2,7 @@ import { disableZoomProperty, WebViewBase, WebViewClient } from './web-view-comm
 import { Trace } from '../../trace';
 import { knownFolders } from '../../file-system';
 import { openUrl } from '../../utils';
+import * as application from '../../application';
```
NathanWalker, line 5: Instead of importing `application` here, you can import from utils. We've been trying to pair down some circular imports throughout core over time and `application` contains some which would be better to avoid here for bundling.

Instead, do this:
```
import { openUrl, ad } from '../../utils';
```

## For Clarity

`ad` when importing within core relatively is short for `android` to avoid bumping up against the `android` global. It is exported from the main barrel for consumption as:

```
import { Utils } from '@nativescript/core'

// for example:
Utils.android.getApplicationContext()
```

packages/core/ui/web-view/index.android.ts
```diff
@@ -27,8 +28,19 @@ function initializeWebViewClient(): void {
...
+			}
+
 			// Handle schemes like mailto, tel, etc
-			if (!android.webkit.URLUtil.isNetworkUrl(url)) {
+			// Make sure to add  <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES"/>  to AndroidManifest.xml
+			// We check if there is any app installed on the device to handle the special URL(mailto, tel, etc)
+			const intent = new android.content.Intent(android.content.Intent.ACTION_VIEW);
+			intent.setData(android.net.Uri.parse(url));
+
+			const packageManager = application.android.context.getPackageManager();
```
NathanWalker, line 41: Then you can use it here:
```
const packageManager = ad.getApplicationContext().getPackageManager();
```

This will properly avoid the import of `application` in this file.

NathanWalker (changes requested):
You can also sign the CLA when get a chance, thank you.

prabudevarrajan:
@NathanWalker Updated as per your comments. Please merge accordingly

prabudevarrajan:
@cla-bot check

CatchABus:
I'm quite against this one. I believe we should enhance method `openUrl` to actually check for the existence of an activity that supports schema using package manager. Then, we could return the boolean result of `openUrl` itself in `shouldOverrideUrl`.

@prabudevarrajan I had plans to work on this one but I'm okay with your PR too if you pay more attention to requested changes.

prabudevarrajan:
> I'm quite against merging this one. We duplicate part of code that already resides in `openUrl`. I believe we should enhance method `openUrl` to actually check for the existence of an activity that supports schema using package manager. Then, we could return the boolean result of `openUrl` itself in `shouldOverrideUrl` (which should fix your reported problem).
> 
> @prabudevarrajan I had plans to work on this one but I'm okay with your PR too if you pay more attention to my requested changes. I could even create a PR tonight and see if you agree to my changes.
> 
> Note: Permission `QUERY_ALL_PACKAGES` is quite dangerous. Everything is fine as long as apps contain queries for the schemas they have support for.

@CatchABus  As per your comments, I have updated the [logic](https://github.com/NativeScript/NativeScript/pull/10148/commits/79ce1a521a23fc3f6829fa07337e0ef5a81dc399) accordingly. 

Regarding QUERY_ALL_PACKAGES, it's not recommended unless the user is aware of what he is doing. I have updated the [trace](https://github.com/prabudevarrajan/NativeScript/blob/master/packages/core/utils/index.android.ts#L31) log accordingly.

packages/core/utils/index.android.ts
```diff
@@ -22,8 +22,16 @@ export function openUrl(location: string): boolean {
 	const context = ad.getApplicationContext();
 	try {
 		const intent = new android.content.Intent(android.content.Intent.ACTION_VIEW, android.net.Uri.parse(location.trim()));
+		const packageManager =  context.getPackageManager();
+
 		intent.addFlags(android.content.Intent.FLAG_ACTIVITY_NEW_TASK);
 
+		// Handle schemes like mailto, tel, etc
+		if (intent.resolveActivity(packageManager) == null) {
```
CatchABus, line 30: Since we bother calling `resolveActivity`, we can make use of its return value.
Something like:
```ts
const intentComponent = intent.resolveActivity(packageManager);

// Handle schemes like mailto, tel, etc
if (intentComponent == null) {
    return false;
}

intent.setComponent(intentComponent);
```
This is usually not needed because system will search and find the appropriate class if any, but now that we've already resolved it we can save a bit of work that would be done internally.
See: https://developer.android.com/reference/android/content/Intent#setComponent(android.content.ComponentName)

CatchABus (changes requested):

farfromrefug:
I dont think QUERY_ALL_PACKAGES is necessary here or should even be mentioned. As @CatchABus said it is a dangerous perm in the sense that google is likely to refuse your app. The trace error should remove mention of it. It would wrongly induce users into thinking they need that permission when they dont

prabudevarrajan:
> I dont think QUERY_ALL_PACKAGES is necessary here or should even be mentioned. As @CatchABus said it is a dangerous perm in the sense that google is likely to refuse your app. The trace error should remove mention of it. It would wrongly induce users into thinking they need that permission when they dont

Yup agreed. I have updated the trace error [message](https://github.com/NativeScript/NativeScript/pull/10148/commits/9c6d3e0b9057f125d5a86fa88b49a20186c1d83f) accordingly.

CatchABus:
@prabudevarrajan Can you take a look at this one? https://github.com/NativeScript/NativeScript/pull/10148#discussion_r1059161592

I think we're good if it's done and works.

packages/core/utils/index.android.ts
prabudevarrajan: @CatchABus I believe we should allow the user to pick the appropriate apps for the intent. Setting setComponent() will restrict the app only to the intentComponent. resolveActivity() will always return only one activity compnent even if multiple activities match the intent. 

**resolveActivity()**
If multiple activities are found to satisfy the intent, the one with the highest priority will be used. If there are multiple activities with the same priority, the system will either pick the best activity based on user preference, or resolve to a system class that will allow the user to pick an activity and forward from there.
https://developer.android.com/reference/android/content/Intent#resolveActivity(android.content.pm.PackageManager)

**setComponent()**
If this class is defined, the specified class will always be used regardless of the other fields. You should only set this value when you know you absolutely want a specific class to be used; otherwise it is better to let the system find the appropriate class so that you will respect the installed applications and user preferences.

CatchABus: The whole thing about `startActivity` is a bit of a mystery.
In `resolveActivity` details, you might also read this:
```
This API is called for you as part of starting an activity from an intent. You do not normally need to call it yourself.
```

Still, I'll agree with you that we shouldn't meddle with it more so let's leave it be.

NathanWalker (approved):

NathanWalker:
Nice job @prabudevarrajan !

PR #10160, opened 2023-01-06
Eslint

farfromrefug:
Was getting tired of all the eslint issues

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,7 +24,22 @@
 		{
 			"files": ["*.ts", "*.tsx"],
 			"extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended", "prettier", "plugin:@nrwl/nx/typescript"],
-			"rules": {}
+			"rules": {
+				"no-empty": "off",
+				"no-useless-escape": "off",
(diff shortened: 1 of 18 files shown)


.eslintrc.json
```diff
@@ -24,7 +24,22 @@
...
+				"@typescript-eslint/no-var-requires": "off",
+				"@typescript-eslint/no-empty-function": "off",
+				"@typescript-eslint/ban-ts-comment": "off",
+				"@typescript-eslint/ban-types": "off",
+				"@typescript-eslint/no-unused-vars": "off",
+				"@typescript-eslint/no-explicit-any": "off",
+				"@typescript-eslint/no-this-alias": "warn",
+				"@typescript-eslint/no-namespace": "off",
+				"@typescript-eslint/no-inferrable-types": "off"
+			}
```
wSedlacek, line 42: These seem like reasonable rules to disable given that they appear to not have been being followed anyways.

farfromrefug: I still.kept this alias as warning because I think we should remove those since we use arrow functions. Though there was the risk of regression so I did not do it here

NathanWalker (approved):

PR #10181, opened 2023-01-21
Observable and EventData cleanup

shirakaba:
This PR cherry-picks some of the changes done in the course of #10100. It does *not* implement DOM Events; it just cleans up the existing Observable and EventData source code to reduce the changes needed for DOM Events support (whether implemented into Core or as a separate opt-in package).

## PR Checklist

- [ ] The PR title follows our [guidelines](https://github.com/NativeScript/NativeScript/blob/master/tools/notes/CONTRIBUTING.md#commit-messages) - TODO: I'm keeping them as individual commits for review, but we can rename upon squash.
- [ ] Tests for the changes are [included](https://github.com/NativeScript/NativeScript/blob/master/tools/notes/WritingUnitTests.md) - It's largely a typings change, so no new tests to add.

Inconsistent EventData typings, with lots of `object?: Partial<Observable>`, and `thisArg` not being passed along and/or correctly typed here and there.

We also have conflicts between `observable/index.d.ts` and `observable/index.ts`.

## What is the new behavior?

Consistent, simplified, correct typings. One singular `observable/index.ts`.

No breaking changes anticipated.


shirakaba:
I've built locally and didn't expect to see test failures, so I'll dig into those.

farfromrefug (approved):
Looks awesome to me as it is! 2 things i would love to see added:
* template type for EventData T extending Observable to define "object"
* adding `[k:string]:any` to EventData because most EventData are augmented with data props or others.

shirakaba:
Thanks for the review, @farfromrefug! As for the wishlist:

> * template type for EventData T extending Observable to define "object"

I like it too in theory (have always wanted to just do `data.object` and get a ready-typed object), though not sure how well it will go - would greatly prefer to tackle that as a followup.

But in practice, in most contexts (like defining a click handler in React or Svelte), unless it were inline, you'd need to explicitly specify the generic argument in your callback anyway, so it'd be no less typing saved.

> * adding `[k:string]:any` to EventData because most EventData are augmented with data props or others.

I see the use - though I think the proper solution for this is for the library-maker to properly type their event handlers (e.g. say that the event named "MyEvent" corresponds to the EventData MyEventData). We really need to get away from loose types and enforce strictness. For now, users can assert `as MyEventData` where libraries have mistakenly marked the EventData type as just EventData.

farfromrefug:
About the template type. A simple `<T extends Observable = Observable>` would help a lot not only for you and your callbacks but also for plugins developer creating custom event types. Plus it is the actual correct typing.

As for the other point I understand what you mean. It is just a pain to have tsc always giving errors about EventData properties while everyone know there is always other properties.

shirakaba:
> About the template type. A simple <T extends Observable = Observable> would help a lot not only for you and your callbacks but also for plugins developer creating custom event types. Plus it is the actual correct typing.

I can see that that change would touch a lot of files though, so to reduce the review burden, I'll try preparing a draft PR now that depends on this one that implements just that extra step.

farfromrefug:
@shirakaba it won't change anything! The default T makes it work without a single change !

shirakaba:
> @shirakaba it won't change anything! The default T makes it work without a single change !

It certainly wouldn't break anything, but... as `on`, `off`, `addEventListener`, `removeEventListener`, and every interface extending `EventData` (of which there are many) would need to pass that generic argument around, otherwise there would be no point of use where we'd actually be benefitting from making `EventData` use a generic.

So I could do just that one-line change, but I don't see what it improves as we'd still have to be casting `eventData.object as StackLayout` as usual pretty much everywhere, as all the event handlers would still be typed as non-generic unless we followed this rabbit hole and got them all to pass the generic around. I think there's an extra PR's work in there to get it all hooked up.

farfromrefug:
@shirakaba yes it would be best. But as I said I could already use it in extending EventData or in my callbacks. It is already a free and useful gain.
And no would not have to cast to StackLayout if I use the T in my callback signature

shirakaba:
@farfromrefug I've put together a follow-up commit to implement generic events - as you can see, it's a big change on top of this:

https://github.com/NativeScript/NativeScript/compare/events-cleanup...generic-events

It would most likely be a breaking change for libraries/plugins - any class that were subclassing Observable would now have to add those generic types to the signatures of anything handling event listeners.

It probably wouldn't introduce any breaks for users.

But as it may be a breaking change for libraries, let's keep this commit for generics out of this PR and review it in a follow-up PR.

farfromrefug:
@shirakaba awesome. But I dont think it is a breaking change. You dont need to use generic as all methods / types have default generic.

shirakaba:
The full commit to implement the generic end-to-end *would* be a compile-time breaking change - [playground](https://www.typescriptlang.org/play?ssl=13&ssc=1&pln=1&pc=1#code/JYOwLgpgTgZghgYwgAgKIDcLgCJzHAHgBVkIAPSEAEwGdkB5AIxunTkYBsUBeB519lwB8yAN4AoAJARM4AHJwAthABcyGmCigA5gG4pAe0YArCAjBqi+gL7jxVMxzhQUCJzTpMWUNpxQTkQOQDEGJSCixaPm9fLmReLwE-IQAKGSwLdU0dABpkBDgODkZEAGs1FKo8ODUMDNx8YiEASniRdANgKjywAAtgGgBBKG0AfjU4EABPZrUOrps7NzgPZABlA2UABQ4AV21QcMooxJ9Bf3Eg4JA02UyNLRBtPIKikoRy5Erq2ruGuFa3HanW6yD6A2GYwm01aolstiAA):

image

Any existing plugins extending `Observable` (or `ViewBase` or whatever) and overriding `on` or `addEventListener`, etc., would now find that they need to add the generic to the methods.

Just implementing the one-line change `export interface EventData<T extends Observable = Observable>`, though, would be non-breaking, I agree.

packages/core/abortcontroller/abortsignal.ts
```diff
@@ -1,80 +1,75 @@
...
-        return
-    }
+	if (abortedFlags.get(signal) !== false) {
+		return;
+	}
 
-    abortedFlags.set(signal, true)
-    signal.notify({ eventName: "abort", type: "abort" })
+	abortedFlags.set(signal, true);
+	signal.notify({ eventName: 'abort', type: 'abort', object: this });
```
farfromrefug, line 56: You dont need to pass `object:this` here. It is added by default from `notifiy` (if no `object` set)

packages/core/application/application-common.ts
```diff
@@ -50,10 +49,10 @@ export function setResources(res: any) {
 export const android: AndroidApplication = undefined;
 export const ios: iOSApplication = undefined;
 
-export const on = global.NativeScriptGlobals.events.on.bind(global.NativeScriptGlobals.events);
-export const off = global.NativeScriptGlobals.events.off.bind(global.NativeScriptGlobals.events);
-export const notify = global.NativeScriptGlobals.events.notify.bind(global.NativeScriptGlobals.events);
-export const hasListeners = global.NativeScriptGlobals.events.hasListeners.bind(global.NativeScriptGlobals.events);
+export const on = global.NativeScriptGlobals.events.on.bind(global.NativeScriptGlobals.events) as typeof import('.')['on'];
```
farfromrefug, line 52: wouldn't it be cleaner to do 
```ts
import * as IApplication from '.';
... 
export const on = global.NativeScriptGlobals.events.on.bind(global.NativeScriptGlobals.events) as typeof IApplication.on;
```
I have not "tested" this but i guess it should work

packages/core/global-types.d.ts
```diff
@@ -25,17 +25,7 @@ declare namespace NodeJS {
...
-				off(eventNames: string, callback?: any, thisArg?: any);
-				addEventListener(eventNames: string, callback: (data: any) => void, thisArg?: any);
-				removeEventListener(eventNames: string, callback?: any, thisArg?: any);
-				set(name: string, value: any): void;
-				setProperty(name: string, value: any): void;
-				get(name: string): any;
-				notify<T>(data: any): void;
-				notifyPropertyChange(propertyName: string, value: any, oldValue?: any): void;
-				hasListeners(eventName: string): boolean;
+				[Key in keyof import('data/observable').Observable]: import('data/observable').Observable[Key];
```
farfromrefug, line 28: Same as before. Wouldn't it be cleaner to import `data/observable` as Observable from top and use it here?
maybe it wasn't working?

packages/core/ui/editable-text-base/index.android.ts
```diff
@@ -533,6 +533,7 @@ export abstract class EditableTextBase extends EditableTextBaseCommon {
 
 			this.notify({
 				eventName: EditableTextBase.blurEvent,
+				object: this,
```
farfromrefug, line 536: not needed. Added by default and less JS code

packages/core/global-types.d.ts
shirakaba: I'll have to check. I never have much luck defining imports in `.d.ts` files (have to be very careful not to accidentally turn it into a module, causing it to lose its ambient context). But happy to give that a try.

... Side note, changes to this file are a moot point in the first place as the global types have been broken since `@types/node` was updated to version 16, as they changed the way you have to augment the `global`/`globalThis` context. I have a branch `fix/global-types` that fixes them that I haven't opened a PR for yet.

packages/core/abortcontroller/abortsignal.ts
farfromrefug: @shirakaba I missed that change but that would be a breaking change though. I know i have many plugins depending on that.

shirakaba: How about the middle ground I suggested, though, which preserves the optional member?

shirakaba: image
It seems that the middle ground breaks things... Going to have to assess options.

packages/core/global-types.d.ts
shirakaba: I tried out a few different approaches in a new minimal project using `@types/node@14` (which still uses `NodeJS.Global`).

# Attempt one: importing at top of file

This doesn't work (as expected), as the top-level import changes `global-types.d.ts` into a module, so it stops filling ambient types.

image

# Attempt two: importing within the namespace

This doesn't work either. We'd expect `events` to bear the `eventOne` key, but it doesn't auto-complete.

image

# Attempt three: creating a type alias within the namespace

This does work, but it leaves some pollution - it now means that the users have to put up with `NodeJS.Observable` being in tsc's compilation context. i.e. they can write `const obs = new Observable as NodeJS.Observable`.

image

# Original attempt

The original suggestion gets working Intellisense without leaving any pollution in the user types.

image

So I think we'd best stick with the code in the PR!

packages/core/ui/editable-text-base/index.android.ts
shirakaba: Now reverted.

packages/core/application/application-common.ts
shirakaba: Now done (feel free to resolve this comment).

packages/core/abortcontroller/abortsignal.ts
shirakaba: I applied a solution in the commit `fix: allow `object` to be optional in notify()`. Hope you like it!

vallemar:
a doubt, this will make the type in the event for example: loaded, come defined for the object object?

shirakaba:
@vallemar As of this PR, the behaviour will become:

The callback for the "loaded" event will return an `EventData`:

```ts
export abstract class View extends ViewCommon {
  // ...
  on(event: 'loaded', callback: (args: EventData) => void, thisArg?: any);
  // ...
}
```

`EventData` will look like this (note how `object` is no longer an optional `Partial<Observable>`):

```ts
export interface EventData {
  eventName: string;
  object: Observable;
}
```

Thus, usage will look like this:

```ts
new Button().on("loaded", (eventData: EventData) => {
  const object = eventData.object as Button;
});
```

So, if you want to access any Button-specific APIs, you will still have to cast the Observable to a Button. But if you only need to call Observable APIs, then no type-casting is necessary.

More than anything, the advantage is that the `object` is `Partial`. And in strict-mode TypeScript, you no longer have to check whether it's `undefined`.

This PR paves the way for a follow-up PR (this commit: https://github.com/NativeScript/NativeScript/compare/events-cleanup...generic-events ) where we'd introduce generics. That would improve the API to:

```ts
new Button().on<Button>("loaded", (eventData) => {
  eventData.object; // TypeScript infers this to be a Button
});

// Or (if I can set up the inference correctly by defaulting to the class itself):
new Button().on("loaded", (eventData) => {
  eventData.object; // TypeScript infers this to be a Button
});
```

NathanWalker (approved):

PR #10182, opened 2023-01-21
fix(ios): race conditions with nativeView

NathanWalker:
## PR Checklist

Under various view navigation conditions, anything from background-image css handling to processing of the nativeView can trigger with view bindings at times which lead to race conditions when the nativeView would have been destroyed and crashes such as:

```
  Unhandled Promise rejection: Cannot read property 'frame' of null ; Zone: <root> ; Task: Promise.then ; Value: TypeError: Cannot read property 'frame' of null TypeError: Cannot read property 'frame' of null
  at uiColorFromImage (file: node_modules/@nativescript/core/ui/styling/background.ios.js:317:0)
  at file: node_modules/@nativescript/core/ui/styling/background.ios.js:167:0
```

## What is the new behavior?

iOS stability across broader diverse project/layout binding setups. Under all these types of cases, the existence of `nativeView` does not matter at all and should be ignored.

### DIFF
--- a/packages/core/ui/styling/background.ios.ts
+++ b/packages/core/ui/styling/background.ios.ts
@@ -193,11 +193,13 @@ function setUIColorFromImage(view: View, nativeView: UIView, callback: (uiColor:
 		bitmap = imageSource && imageSource.ios;
 	} else if (imageUri.indexOf('http') !== -1) {
 		style[symbolUrl] = imageUri;
-		ImageSource.fromUrl(imageUri).then((r) => {
-			if (style && style[symbolUrl] === imageUri) {
-				uiColorFromImage(r.ios, view, callback, flip);
-			}
-		});
+		ImageSource.fromUrl(imageUri)
+			.then((r) => {
+				if (style && style[symbolUrl] === imageUri) {
+					uiColorFromImage(r.ios, view, callback, flip);
+				}
+			})
+			.catch(() => {});
 	}
 
 	uiColorFromImage(bitmap, view, callback, flip);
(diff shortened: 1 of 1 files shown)


packages/core/ui/styling/background.ios.ts
```diff
@@ -98,6 +101,9 @@ function onScroll(this: void, args: ScrollEventData): void {
 }
 
 function adjustLayersForScrollView(nativeView: UIScrollView & NativeScriptUIView) {
+	if (!nativeView) {
```
shirakaba, line 104: I think we shouldn't have to second-guess TypeScript - if the method args demand `nativeView` to be populated, we should be able to trust it. Can we handle this upstream of here instead somehow?

Same for the other methods putting a guard around nativeView when the method args specify a non-null, not undefined `nativeView`.

shirakaba (changes requested):

packages/core/ui/styling/background.ios.ts
NathanWalker: The problem in these cases is that it’s tied to a lifecycle which depending on view setup is not guaranteed order - largely due to how different flavors trigger and apply their view bindings.

farfromrefug: @NathanWalker i agree with @shirakaba i think most of those are unecessary.
If i take the example of `adjustLayersForScrollView` and where it is used. It is used in ScrollView on scroll event where there is a test for `nativeView instanceof UIScrollView` so the test in `adjustLayersForScrollView` is uncessary and in a sense makes things slower by adding a if test

NathanWalker: Thanks for feedback @farfromrefug @shirakaba I'll see if can pair it down to resolve upstream for the single issue mentioned `uiColorFromImage` which is the key issue here.

NathanWalker: This is fixed up for the isolated case now and allowed this to improve since this is a case where image request can be aborted/canceled or fail which should not allow this crash to occur.

rigor789 (approved):

PR #10186, opened 2023-01-25
fix(core): android fix for `knownFolders.externalDocuments()` in some…

farfromrefug:
… cases (emulator)

Sometimes (very often actually) it would not work on the emulator. This fixes it


NathanWalker (approved):

NathanWalker:
Don’t worry about unit tests, we’ll check out the jest runner soon - the error is new and likely related to a recent devDep patch.

PR #10193, opened 2023-02-01
Generic events

shirakaba:
With this PR, EventData and methods for adding/removing listeners will now be generic:

```ts
// This event is inherited from View, so eventData.object is
// inferred to be a View.
new Button().on("loaded", (eventData) => {
  eventData.object; // TypeScript infers this to be a View
});

// You can, however, explicitly pass the generic to narrow
// that down to Button:
new Button().on<Button>("loaded", (eventData) => {
  eventData.object; // TypeScript infers this to be a Button
});

// Button-specific events don't require passing the generic:
new Button().on("tap", (eventData) => {
  eventData.object; // TypeScript infers this to be a Button
});
```

Please do give comment on the discussion points I've posted and (once we've reviewed it and prepared a build of it) a try on your projects, as I fully expect it'll result in some TypeScript grumbles.

## PR Checklist

- [ ] The PR title follows our [guidelines](https://github.com/NativeScript/NativeScript/blob/master/tools/notes/CONTRIBUTING.md#commit-messages) - TODO: I'm keeping them as individual commits for review, but we can rename upon squash.
- [ ] Tests for the changes are [included](https://github.com/NativeScript/NativeScript/blob/master/tools/notes/WritingUnitTests.md) - It's largely a typings change, so no new tests to add.

Types for EventData and adding/removing listeners are non-generic, so we always have to write `eventData.object as Button` if we want anything more specific than an Observable.

## What is the new behavior?

Improved inference in most cases. The events in Application have always been *very* messy though, so I've left those as defaulting to Observable even though some could be narrowed to View, as I simply can't get my head round them.

Note that this PR would introduce breaking changes for library authors. As although we do provide default args for each generic, if they've extended `on` anywhere, they'll have to add the generic, as demonstrated in this [playground](https://www.typescriptlang.org/play?ssl=13&ssc=1&pln=1&pc=1#code/JYOwLgpgTgZghgYwgAgKIDcLgCJzHAHgBVkIAPSEAEwGdkB5AIxunTkYBsUBeB519lwB8yAN4AoAJARM4AHJwAthABcyGmCigA5gG4pAe0YArCAjBqi+gL7jxVMxzhQUCJzTpMWUNpxQTkQOQDEGJSCixaPm9fLmReLwE-IQAKGSwLdU0dABpkBDgODkZEAGs1FKo8ODUMDNx8YiEASniRdANgKjywAAtgGgBBKG0AfjU4EABPZrUOrps7NzgPZABlA2UABQ4AV21QcMooxJ9Bf3Eg4JA02UyNLRBtPIKikoRy5Erq2ruGuFa3HanW6yD6A2GYwm01aolstiAA):

image

### What should the generics be?

The generics implemented in this PR are "better than nothing", but still not ideal. See what I ideally wanted to implement vs. what I was able to get working:

```ts
// What I wanted:
export abstract class View extends ViewCommon {
  on<T extends View = this>(event: 'loaded', callback: (args: EventData<T>) => void, thisArg?: any);
}

// The only thing I could get to work:
export abstract class View extends ViewCommon {
  on<T extends Observable = View>(event: 'loaded', callback: (args: EventData<T>) => void, thisArg?: any);
}
```

The former is more strictly correct, but I just couldn't get TypeScript to accept it 🤷

### Should we narrow the default generic for EventData in the first place?

Although I think it *is* correct for the methods for adding/removing event listeners to have more narrow generic defaults than just `Observable`, e.g.:

```ts
export class SegmentedBar extends View implements AddChildFromBuilder, AddArrayFromBuilder {
  on<T extends Observable = SegmentedBar>(event: 'selectedIndexChanged', callback: (args: SelectedIndexChangedEventData<T>) => void, thisArg?: any): void;
}
```

... I am unsure about whether the EventData interfaces should behave the same:

```ts
// Would we prefer a narrow default?
export interface SelectedIndexChangedEventData<T extends Observable = SegmentedBar> extends EventData<T> {
  oldIndex: number;
  newIndex: number;
}

// ... or just stick to Observable?
export interface SelectedIndexChangedEventData<T extends Observable = Observable> extends EventData<T> {
  oldIndex: number;
  newIndex: number;
}
```

Mainly because we don't know whether others might repurpose the same event (SelectedIndexChangedEvent can be used for more than just SegmentedBar, and indeed there are duplicates of this interface across the codebase). Of course consumers can just specify the generic explicitly if they want to override the assumed default of SegmentedBar, so it's not the end of the world whichever choice we go with.

### DIFF
--- a/packages/core/accessibility/accessibility-common.ts
+++ b/packages/core/accessibility/accessibility-common.ts
@@ -28,7 +28,7 @@ export function notifyAccessibilityFocusState(view: View, receivedFocus: boolean
 		eventName: accessibilityFocusChangedEvent,
 		object: view,
 		value: !!receivedFocus,
-	} as EventDataValue);
+	});
 
 	if (receivedFocus) {
 		if (view.page) {
@@ -38,12 +38,12 @@ export function notifyAccessibilityFocusState(view: View, receivedFocus: boolean
 		view.notify({
 			eventName: accessibilityFocusEvent,
 			object: view,
-		} as EventData);
+		});
 	} else if (lostFocus) {
 		view.notify({
 			eventName: accessibilityBlurEvent,
 			object: view,
-		} as EventData);
+		});
 	}
 }
 
--- a/packages/core/accessibility/accessibility-types.ts
+++ b/packages/core/accessibility/accessibility-types.ts
@@ -1,4 +1,4 @@
-import type { EventData } from '../data/observable';
+import type { EventData, Observable } from '../data/observable';
 
 export enum AccessibilityTrait {
 	/**
@@ -264,7 +264,7 @@ export enum AndroidAccessibilityEvent {
 	ALL_MASK = 'all',
 }
 
-export interface AccessibilityEventPerformEscape extends EventData {
+export interface AccessibilityEventPerformEscape<T extends Observable = Observable> extends EventData<T> {
 	cancel?: boolean;
 }
 
--- a/packages/core/application/application-interfaces.ts
+++ b/packages/core/application/application-interfaces.ts
@@ -1,5 +1,5 @@
 // Types
-import { EventData } from '../data/observable';
+import { EventData, Observable } from '../data/observable';
 import { View } from '../ui/core/view';
 
 /**
@@ -12,7 +12,7 @@ export interface NativeScriptError extends Error {
 	nativeError: any;
 }
 
-export interface ApplicationEventData extends EventData {
+export interface ApplicationEventData<T extends Observable = Observable> extends EventData<T> {
 	/**
 	 * UIApplication or undefined, unless otherwise specified. Prefer explicit
 	 * properties where possible.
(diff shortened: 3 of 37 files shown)


ammarahm-ed:
This is ok but I think  it doesn't really solve the problem. All Views "Should" infer to correct types in events automatically without passing some Generic  or anything.

```ts
new Button().on<Button>("loaded", (eventData) => {
  eventData.object; // TypeScript infers this to be a Button
});
```
is almost as much work as doing
```ts
new Button().on("loaded", (eventData) => {
  eventData.object as Button;
});
```
The ideal way would be to not have to pass any generic at all when you are subscribing to an event on Button. For example:

```ts
new Button().on("loaded", (eventData) => {
  eventData.object // is Button automatically
});
```
Also my question, why does this work:

```ts
```

farfromrefug:
@ammarahm-ed be careful some views fire event for which the object is not them self! I know some plugins already do that. And I think some N components also do that. Will try to find them

PR #10207, opened 2023-02-11
fix(ios): Missing support for a11y font scale

CatchABus:
## PR Checklist

In iOS, fonts are not affected by device accessibility font scale settings.

## What is the new behavior?
Fonts will be affected by device accessibility font scale settings.
We had a font scale implementation that actually was an incomplete work so I also performed a cleanup.

@NathanWalker For this to work, I used an API that was introduced at iOS 11. Do you think we should support older devices?
@farfromrefug Do you think this could have a performance impact? We could discuss about adding a property to make it configurable to both platforms perhaps.

Fixes https://discord.com/channels/603595811204366337/1048667697493319851


farfromrefug:
@CatchABus not 100% sure about perf consequences. But I think.it is always best to put that behind a property. Another reason to use a property is that components extending text-base (plugins) will also benefit from it. While if you do it in onCreateView and the plugin has its own onCreateView it won't get it. 
We could try a perf test with collectionview to see If we see any slowdown because of it. But I dont think there will be any.

CatchABus:
> @CatchABus not 100% sure about perf consequences. But I think.it is always best to put that behind a property. Another reason to use a property is that components extending text-base (plugins) will also benefit from it. While if you do it in onCreateView and the plugin has its own onCreateView it won't get it. We could try a perf test with collectionview to see If we see any slowdown because of it. But I dont think there will be any.

I think we could add a `accessibilityAdjustsFontSize` boolean property and register it to text-base. Default value could be `true` (nowadays most frameworks force you to enable most of a11y anyway) and check for native view instance inside since the native API we'll be using has no support for views like UIButton.

dkzeb:
I've tested the changes on iOS and i'm happy to report that it does fix scaling of labels / texts - however - this PR does not fix font-scaling in buttons - its missing an adjustsFontForContentSizeCategory in ui/button/index.ios.ts when creating the native view
I will keep testing font-scaling (as accessibility is a very important for our project), and report back if i find anything else missing

farfromrefug:
@CatchABus on iOS you dont need to find all the necessary views and call requestLayout. Simply relayout all pages
Even if you call requestLayout on a single view, the way it works in N is that it will go up to the parent page and relayout the full page (though not optimized at all)

CatchABus:
> @CatchABus on iOS you dont need to find all the necessary views and call requestLayout. Simply relayout all pages Even if you call requestLayout on a single view, the way it works in N is that it will go up to the parent page and relayout the full page (though not optimized at all)

That was my initial thought but unfortunately it didn't work as expected. You have to call `requestLayout` or `sizeToFit` on those views directly.

We have already gotten an inherited property called `_fontScale` that's doing what we want, and I think it's part of an unfinished work from initial a11y merge to core.

I'm preparing another PR for fixing a problem related to inherited properties, then I think we can complete unfinished work for font scale. :D

CatchABus:
@dkzeb Things took a different turn and I ended up extending current functionality for a number of reasons.
Can you try new changes? They should work along with changes from PR #10225

dkzeb:
@CatchABus - thx - i'll test it out  :)

dkzeb:
@CatchABus, i still doesn't scale text in buttons

CatchABus:
> @CatchABus, i still doesn't scale text in buttons

Strange, I have `Button` texts scale in the app I'm testing this.
Can you try it on a fresh new app or give me an app sample to check?

NathanWalker (approved):

NathanWalker:
If that test fails we’ll swap that, good mention.

CatchABus:
> I'll have to double check, but I believe the added
> 
> ```
> apps/automated/src/ui/styling/style-properties-tests.ios.ts
> ```
> 
> file will prevent running the tests from
> 
> ```
> apps/automated/src/ui/styling/style-properties-tests.ts
> ```
> 
> because it "overrides" it during the webpack build. Checking for `global.isIOS` would likely be a better solution (or whatever is used in other tests that check for the platform).

I think I might have accidentally missed this. The initial plan would be to have a common file and a blank android file that exports tests from common too. If checking for `global.isIOS` is an acceptable option, then sure I can change it to check for platform flag.
@rigor789 Should we display any message for the android plarform?

PR #10209, opened 2023-02-12
fix(android): Change The android api path reference from 17 to 33.

KunaalMangal:
Change The android api path reference from 17 to 33 in types-android/src/lib/android.d.ts for accessing latest android api.

image


rigor789:
Marking this as a draft, as I'd like to bring this up on our next TSC Meeting (Technical Steering Committee).

I agree we should change the default API level, however the default should likely be set to `API30`, and then changed to `API31` around Nov 2023 to match the `targetSdk` requirements (reference: https://apilevels.com/)

packages/types-android/src/lib/android.d.ts
```diff
@@ -1 +1 @@
-/// <reference path="./android-17.d.ts" />
\ No newline at end of file
+/// <reference path="./android-33.d.ts" />
```
NathanWalker, line 1: Per @rigor789 mention, please use 30 as it's the current recommendation from Google.
```suggestion
/// <reference path="./android-30.d.ts" />
```

NathanWalker (changes requested):

KunaalMangal:
@cla-bot check

KunaalMangal:
@cla-bot check

NathanWalker (approved):

PR #10210, opened 2023-02-14
fix(core): new autofillTypes do not work

delaneyb:
## PR Checklist

New autoFill types with uppercase characters do not apply the corresponding field types to the nativeView, due to multiple problems:
* [makeValidator](https://github.com/NativeScript/NativeScript/blob/dcf6a365cd574bd4dfa245ad6c38c22e551ef65c/packages/core/ui/editable-text-base/editable-text-base-common.ts#L54) is not passed the full correct list of valid autofillTypes (missing all except `username` and `password`)
* [makeParser](https://github.com/NativeScript/NativeScript/blob/dcf6a365cd574bd4dfa245ad6c38c22e551ef65c/packages/core/ui/core/properties/index.ts#L1546) converts the specified attribute value to lowercase, while the `isValid` function (returned by makeValidator) that is given to makeParser expects the original camelcase strings from `CoreTypes.AutofillType`.
* The [AutofillType](https://github.com/NativeScript/NativeScript/blob/dcf6a365cd574bd4dfa245ad6c38c22e551ef65c/packages/core/core-types/index.ts#L49) `type` does not actually match with the concrete values of the namespace (why don't we just use `export type AutofillType = keyof typeof AutofillType` to define types for all the CoreTypes in this file to avoid mismatched namespace keys and type members in the first place??)
* The CoreTypes AutofillType namespace is missing member values that the platform implementations support (from ios I can see there should be: `'creditCardNumber', 'email', 'name', 'newUsername', 'phone', 'postalCode'`)

## What is the new behavior?
Do not use makeParser and makeValidator to specify a valueConverter as makeParser converts the attribute value to lowercase, which will not be found in the original set of values passed to makeValidator.

autofillType attribute values other than `username` and `password` work. For example `oneTimeCode`, `email`, `name` and `newPassword` now work where they previously did nothing.

This does not fix all the problems with mismatching definitions of supported values for the attribute.


farfromrefug:
@delaneyb thanks a lot for this. I added those in a rush and actually was about to test those in a realtime app!
Thanks for investigating this!

NathanWalker (approved):

PR #10212, opened 2023-02-14
feat: allow custom image scales on iOS

edusperoni:
## PR Checklist

Images are always loaded with scale 1.0 (if they are files/resources then they're loaded with the appropriate scale when `@2x.png` is provided, for example)

## What is the new behavior?
Allows you to set a hardcoded scale for the image.


edusperoni:
This might be too much of a burden to support in the future (especially if we decide to support orientation as well).

This is a viable alternative that's around 0.02ms slower and doesn't require any code to support:

```
function setImageSourceScale(source: ImageSource, scale: number) {
  const image = source.ios;
  return new ImageSource(UIImage.imageWithCGImageScaleOrientation(image.CGImage, scale, image.imageOrientation));
}
```

PR #10225, opened 2023-02-28
fix(core): Frame to Page property propagation

CatchABus:
## PR Checklist

When I attempt to update some inheritable properties programmatically, they won't propagate from Frame to Page during initialization.
The reason seems to be the fact that we append page to frame using `_addView` before setting new frame entry.
Methods like `eachChildView` and `_childrenCount` rely on current entry resolved page and this is the actual problem.

For example, this is a description of how things currently work on iOS:
- Method `viewWillAppear` in Frame view controller will perform an `_addView` call
- Method `viewDidAppear` will assign new current entry

Property `currentPage` is a getter that returns the resolved page of current entry, and this getter is used by `eachChildView` and `_childrenCount`. Since current entry won't be updated until `viewDidAppear` call, we end up with an `_addView` call that tries to iterate Frame's children in order to update their inheritable properties but `eachChildView` has still got zero iterations as `currentPage` returns null.

## What is the new behavior?
We add a new property called `_childView` in order to separate entry navigation logic from view tree logic. Methods like `eachChildView` are considered a DOM thing and we should not make them rely on navigation entries.

Tests seem good in both platforms but we might want to spend some time and see if this change has any bad impact.

### DIFF
--- a/packages/core/ui/frame/frame-common.ts
+++ b/packages/core/ui/frame/frame-common.ts
@@ -43,6 +43,7 @@ export class FrameBase extends CustomLayoutView {
 
 	public actionBarVisibility: 'auto' | 'never' | 'always';
 	public _currentEntry: BackstackEntry;
+	public _childView: Page;
 	public _animationInProgress = false;
 	public _executingContext: NavigationContext;
 	public _isInFrameStack = false;
@@ -238,6 +239,9 @@ export class FrameBase extends CustomLayoutView {
 		// In case we navigated forward to a page that was in the backstack
 		// with clearHistory: true
 		if (!newPage.frame) {
+			// Used for traversing and manipulating DOM
+			this._childView = newPage;
+
 			this._addView(newPage);
 			newPage._frame = this;
 		}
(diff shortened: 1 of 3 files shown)


farfromrefug:
@CatchABus could we add a comment or rename the variable to express it is exactly the same as currentEntry.resolvedPage but only set sooner?

CatchABus:
> @CatchABus could we add a comment or rename the variable to express it is exactly the same as currentEntry.resolvedPage but only set sooner?

Yes sure, would something like `_resolvedPage` be more descriptive? I originally went for `_currentPage` but its naming is very close to existing `currentPage` getter

farfromrefug:
@CatchABus i like resolvedPage

packages/core/ui/frame/frame-common.ts
```diff
@@ -43,6 +43,14 @@ export class FrameBase extends CustomLayoutView {
 
 	public actionBarVisibility: 'auto' | 'never' | 'always';
 	public _currentEntry: BackstackEntry;
+
+	/**
```
CatchABus, line 47: @farfromrefug What do you think about this? Do we need more details about this change?

farfromrefug: Seems good to.me!

farfromrefug (approved):

PR #10226, opened 2023-03-01
fix(core) export foregroundEvent and backgroundEvent

vallemar:
This does not solve that these two events are not exported, there is some part where I need to put these types but I can't find where,, @NathanWalker  , could you take a quick look?


NathanWalker (approved):

NathanWalker:
Sure thing @vallemar it's fixed up now. They were defined in `application-common` so just needed to be imported from there.

PR #10240, opened 2023-03-17
test: fix safe area tests on iOS16

edusperoni:
no longer do manual calculations on safe area sizing as on iOS 16.

Previous calculation:
status bar + navigation bar height (when present) = safe are top inset. Sounds about right

iOS 16 calculations:
when no navbar:
status bar height + some margin = safe area top inset
when navbar:
status bar height - some negative margin + status bar height = safe area top inset

So essentially:
status bar is of y height
safe area is y + some amount
navbar will overflow the safe area (and status bar a bit), meaning it’ll be y - n
safe area for pages with navbar will be: y - n + navbar height
safe area for pages without navbar will be: window safe area top (y + some amount)

As a result, we just ditch manual calculations and use the values that are available to us (window safe area inset and navbar position and height)


CatchABus:
@edusperoni I seem to get one last failure on safe area tests.
Logs:
```sh
Test: SAFEAREALAYOUT.test_grid_component_cells_layout_in_safe_area_action_bar FAILED: cell height should be equal - cell01<799> - cell11<801>
```

I used iPhone 14 Pro Max emulator with iOS 16.2.

PR #10247, opened 2023-03-23
fix: global declarations fix

farfromrefug:
this changes the way we declare global variables so that it works again as explained here https://stackoverflow.com/questions/59459312/using-globalthis-in-typescript
tsc now picks up the typings and so most fo the any cast are not needed anymore


NathanWalker (approved):
Awesome @farfromrefug ❤️

PR #10269, opened 2023-04-16
fix(core/platform): fixed types for sdkVersion and osVersion

guillemc23:
## PR Checklist

Currently, both `Device.sdkVersion` and `Device.osVersion` return a number in Android. The typings state that this should be a string so expressions like `if (Device.sdkVersion >= 23) {...}` are not possible.

## What is the new behavior?
I added `number | string` to the IDevice interface declaration so these expressions are allowed without the need to add `@ts-ignore`. I also updated the JSDoc to reflect this change.

### DIFF
--- a/packages/core/platform/index.d.ts
+++ b/packages/core/platform/index.d.ts
@@ -46,13 +46,13 @@ export interface IDevice {
 
 	/**
 	 * Gets the OS version.
-	 * For example: 4.4.4(android), 8.1(ios)
+	 * For example: 12(android), 8.1(ios)
 	 */
 	osVersion: number | string;
 
 	/**
 	 * Gets the SDK version.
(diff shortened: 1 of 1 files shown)


NathanWalker (approved):
Excellent thank you @guillemc23

NathanWalker:
@guillemc23 I reverted for the patch release as it does cause a breaking change. The platform types returned do in fact return strings however I think for 8.6 (when a breaking change like this could be introduced) it would likely be better to have it return just a number and let core do the parsing - it's likely debatable given the `string` does often include codenames on release details from the platform that may be desirable to know/get. Possible we may want to introduce a new `number` only getter to use in future.

guillemc23:
Understood! That's a great idea too. Thank you Nathan!

PR #10276, opened 2023-04-24
fix(ios): resilience to nativeView access under edge cases

NathanWalker:
## PR Checklist

Occasionally varied view binding conditions can cause race conditions on property setters which could end up trying to access the `nativeView` when it is not ready or not created yet.

## What is the new behavior?

iOS will be resilient under these conditions.

### DIFF
--- a/packages/core/ui/core/view/index.ios.ts
+++ b/packages/core/ui/core/view/index.ios.ts
@@ -898,13 +898,15 @@ export class View extends ViewCommon implements ViewDefinition {
 			CATransaction.begin();
 		}
 
-		if (value instanceof UIColor) {
-			this.nativeViewProtected.backgroundColor = value;
-		} else {
-			iosBackground.createBackgroundUIColor(this, (color: UIColor) => {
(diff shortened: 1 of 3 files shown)


farfromrefug:
@NathanWalker what about using ? operator now available typescript ? Right now it is not native but it still makes the code cleaner.

NathanWalker:
Yes we can switch to that - we use that all time in userland code today; with core, we had always tried to match existing practices but I think that's one we can incrementally switch to, especially when those moments arise to do so.

NathanWalker:
@farfromrefug we'll do that in cases where we can - turns out in those specific cases we can't without encountering this typescript error: https://bobbyhadz.com/blog/typescript-left-hand-side-of-assignment-not-optional
Optional chaining is useful for property accessors (not really setters):
> The purpose of the optional chaining (?.) operator is [accessing deeply nested properties](https://bobbyhadz.com/blog/typescript-check-if-object-has-key) without erroring out if a value in the chain is equal to null or undefined
> However, the optional chaining operator cannot be used on the left-hand side of an assignment expression.

PR #10291, opened 2023-05-15
fix(dialogs/iOS): do not use view controllers that are being dismissed

mukaschultze:
## PR Checklist

A dialog will not show up on iOS when invoked shortly after a modal has been closed, as the view controller will be playing the dismiss animation and will dismiss the dialog along with it.

## What is the new behavior?
Do not use UIControllers that have the `beingDismissed` flag as the `sourceView` for dialogs.

Alternatively, we could use an async approach like this:
```ts
function getViewController(): Promise<UIViewController> {
	let viewController = ios.rootController as UIViewController;

	while (viewController && viewController.presentedViewController) {
		viewController = viewController.presentedViewController;
	}

	if (viewController.beingDismissed) {
		const dismiss = new Promise<void>((resolve) => viewController.dismissViewControllerAnimatedCompletion(false, () => resolve()));
		return dismiss.then(() => getViewController());
	} else {
		return Promise.resolve(viewController);
	}
}
```
The reason why I didn't use this method is simply that I couldn't find any documentation of the side effects of calling `dismissViewController` in a controller that is already being dismissed.

### DIFF
--- a/packages/core/ui/dialogs/index.ios.ts
+++ b/packages/core/ui/dialogs/index.ios.ts
@@ -45,9 +45,9 @@ function raiseCallback(callback, result) {
 }
 
 function showUIAlertController(alertController: UIAlertController) {
-	let viewController = ios.rootController;
+	let viewController = ios.rootController as UIViewController;
 
-	while (viewController && viewController.presentedViewController) {
+	while (viewController && viewController.presentedViewController && !viewController.presentedViewController.beingDismissed) {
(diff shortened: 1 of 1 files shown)


NathanWalker:
The bool is smart here @mukaschultze thanks. It's likely safer than the async approach callback because wiring that in that case would likely be subject to race conditions.

NathanWalker (approved):

PR #10368, opened 2023-08-25
fix(core): Scroll listener register failure after unregister

CatchABus:
## PR Checklist

There is a case that `ScrollView` scroll event listeners fail to register if a `removeEventListener` call occurs prior.
e.g.
```js
view.off('scroll', onScroll);
view.on('scroll', onScroll); // This won't work because of line above
```
This occurs mostly on iOS because we don't unset property `_delegate` when detaching delegate.
PR will take care of this and improve few other things for both platforms.

## What is the new behavior?
Scroll listeners will register without problems.

### DIFF
--- a/packages/core/ui/scroll-view/index.android.ts
+++ b/packages/core/ui/scroll-view/index.android.ts
@@ -130,10 +130,10 @@ export class ScrollView extends ScrollViewBase {
 
 	protected attachNative() {
 		if (!this.handler) {
-			const that = new WeakRef(this);
+			const viewRef = new WeakRef(this);
 			this.handler = new android.view.ViewTreeObserver.OnScrollChangedListener({
 				onScrollChanged: function () {
-					const owner: ScrollView = that.get();
+					const owner: ScrollView = viewRef.get();
 					if (owner) {
 						owner._onScrollChanged();
 					}
@@ -166,9 +166,13 @@ export class ScrollView extends ScrollViewBase {
 		}
 	}
 
-	protected dettachNative() {
-		this.nativeViewProtected.getViewTreeObserver().removeOnScrollChangedListener(this.handler);
-		this.handler = null;
+	protected detachNative() {
+		if (this.handler) {
+			if (this.nativeViewProtected) {
+				this.nativeViewProtected.getViewTreeObserver().removeOnScrollChangedListener(this.handler);
+			}
+			this.handler = null;
+		}
 	}
 }
 
--- a/packages/core/ui/scroll-view/index.ios.ts
+++ b/packages/core/ui/scroll-view/index.ios.ts
@@ -49,12 +49,6 @@ export class ScrollView extends ScrollViewBase {
 		this._setNativeClipToBounds();
 	}
 
-	disposeNativeView() {
-		this.dettachNative();
-		this._delegate = null;
-		super.disposeNativeView();
-	}
-
 	_setNativeClipToBounds() {
 		if (!this.nativeViewProtected) {
 			return;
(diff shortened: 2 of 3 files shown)


NathanWalker (approved):

JWiseCoder:
This fix seems to have broken the `scroll` event in our NS/Angular code, which is something similar to this:

```
 <ScrollView
        orientation="horizontal"
        scrollBarIndicatorVisible="false"
        (scroll)="onScroll($event)">
```

Basically, when updating to a NativeScript version that includes this fix, the `onScroll` function is never called.  If I go back and copy the 8.5.9 `index.ios.js` in the core/ui/scroll-view folder of the nativescript node_modules, the scroll event fires like normal.  Is this a bug, or is there something we're missing?

CatchABus:
> This fix seems to have broken the `scroll` event in our NS/Angular code, which is something similar to this:
> 
> ```
>  <ScrollView
>         orientation="horizontal"
>         scrollBarIndicatorVisible="false"
>         (scroll)="onScroll($event)"
>         *ngIf="someVariableToShowHideTheScrollView">
> ```
> 
> Basically, when updating to a NativeScript version that includes this fix, the `onScroll` function is never called. If I go back and copy the 8.5.9 `index.ios.js` in the core/ui/scroll-view folder of the nativescript node_modules, the scroll event fires like normal. Is this a bug, or is there something we're missing?

Can you try to get rid of `ngIf` and check if scroll event is triggered?

JWiseCoder:
No, I was already testing that, and getting rid of the ngIf did not change anything.  The component this is a part of is used as a subcomponent, and I even went and removed all the ngIfs up the chain to make sure.  No matter what, the scroll event is not triggering.

CatchABus:
> No, I was already testing that, and getting rid of the ngIf did not change anything. The component this is a part of is used as a subcomponent, and I even went and removed all the ngIfs up the chain to make sure. No matter what, the scroll event is not triggering.

Please create an issue and add a reference to this PR. We'll take care as soon as possible.

JWiseCoder:
I will.  I'm trying to get more info on this issue.  I can create a scroll view elsewhere, even with an ngIf hiding the scroll view initially, and that scroll event fires correctly.  I was hoping it would be something simple, but it looks like there's some complication in here that hasn't been accounted for.

JWiseCoder:
Okay, I've found the culprit.  On our actual code, we were using a class to, among other things, add a background color that will change depending on the light/dark theme.  If I remove the background color, the event fires normally.  I was able to reproduce this using the following test code:

```
<ScrollView
    orientation="horizontal"
    style="background-color: black;"
    (scroll)="onScroll($event)"
>
    <StackLayout
        orientation="horizontal"
        [height]="200">
        <StackLayout width="100" backgroundColor="red"></StackLayout>
        <StackLayout width="100" backgroundColor="blue"></StackLayout>
        <StackLayout width="100" backgroundColor="green"></StackLayout>
        <StackLayout width="100" backgroundColor="yellow"></StackLayout>
    </StackLayout>
</ScrollView>
```

With the background color style in place, the scroll event does not fire.  If I remove it, the scroll event fires normally.

CatchABus:
> Okay, I've found the culprit. On our actual code, we were using a class to, among other things, add a background color that will change depending on the light/dark theme. If I remove the background color, the event fires normally. I was able to reproduce this using the following test code:
> 
> ```
> <ScrollView
>     orientation="horizontal"
>     style="background-color: black;"
>     (scroll)="onScroll($event)"
> >
>     <StackLayout
>         orientation="horizontal"
>         [height]="200">
>         <StackLayout width="100" backgroundColor="red"></StackLayout>
>         <StackLayout width="100" backgroundColor="blue"></StackLayout>
>         <StackLayout width="100" backgroundColor="green"></StackLayout>
>         <StackLayout width="100" backgroundColor="yellow"></StackLayout>
>     </StackLayout>
> </ScrollView>
> ```
> 
> With the background color style in place, the scroll event does not fire. If I remove it, the scroll event fires normally.

I have a brief idea of what's happening, thanks for testing. Could you still create a new issue so I can justify the new PR?

Issue #10373, opened 2023-09-05
`Utils.dismissKeyboard()` does not work in android modals


vallemar:
### Issue Description

When you try to use `Utils.dismissKeyboard()` with focus on a modal view the keyboard does not close

### Reproduction

https://stackblitz.com/edit/nativescript-vue-nativescript-vue-fa2ksg?file=src%2Fcomponents%2FHome.vue,src%2Fcomponents%2FDetails.vue&title=NativeScript%20Starter%20Vue3%20Beta

### Please accept these terms

- [X] I have searched the [existing issues](https://github.com/NativeScript/NativeScript/issues) as well as [StackOverflow](https://stackoverflow.com/questions/tagged/nativescript) and this has not been posted before
- [X] This is a bug report
- [X] I agree to follow this project's [Code of Conduct](https://github.com/NativeScript/NativeScript/blob/master/tools/notes/CONTRIBUTING.md#coc)

farfromrefug:
Might have stumbled upon the same issue just yesterday. Are you trying to dismiss the keyboard while closing the modal ?

vallemar:
@farfromrefug It happens when the modal is open and you try to close the keyboard while keeping the modal open. I have been debugging the code but I don't see anything strange, the code runs fine but the keyboard does not close

CatchABus:
I think support for closing modal soft input is missing.
More specifically, core is missing handling here https://github.com/NativeScript/NativeScript/blob/a959a797df46d47ddf3bfa4514948a97f67758ef/packages/core/utils/android/index.ts#L60

One way to fix this is by retrieving modal window using `_dialogFragment` just like we do for certain properties in `Page` component: https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/page/index.android.ts#L69

farfromrefug:
@vallemar what you can try to figure out the real issue is calling dismissSoftInput with the nativeView argument from your textfield. if it does work it is indeed because we dont get the right  windowToken for the current opened dialog.

PR #10375, opened 2023-09-06
fix(android): `Utils.dismissKeyboard()` not working with modal

vallemar:
Close: https://github.com/NativeScript/NativeScript/issues/10373


packages/core/utils/android/index.ts
```diff
@@ -67,7 +68,7 @@ export function dismissSoftInput(nativeView?: android.view.View): void {
 		}
 		windowToken = nativeView.getWindowToken();
 	} else if (getCurrentActivity() instanceof androidx.appcompat.app.AppCompatActivity) {
-		const decorView = getCurrentActivity().getWindow().getDecorView();
+		const decorView = (topmost().modal?.getClosestWindow() ?? getCurrentActivity().getWindow()).getDecorView();
```
rigor789, line 71:
```suggestion
		const decorView = (topmost()?.modal?.getClosestWindow() ?? getCurrentActivity().getWindow()).getDecorView();
```

rigor789 (changes requested):
Small adjustment, in case `topmost()` is undefined.

vallemar:
@rigor789 fix comment and I have changed the code to ensure removing the focus on modals that the root view is not a Page

CatchABus:
@vallemar We have 2 cases of modals. One rendered with `Frame` and another with `Page` or a layout component I think.
See: https://v7.docs.nativescript.org/ui/components/modal-view

Does this fix cover both cases?

vallemar:
@CatchABus ok, I have it working with Frame, Page and other layout as the root view of the modal.
Additionally, I have also realized that the focus was not being removed when it was a modal. Now it is working in all cases.

Also, I think it would be good to make some function to bring us the android currentView and use it in all these places because it is a bit ugly to do this everywhere:
```ts
    const modalDialog = (topmost()?._modalParent ?? topmost())?.modal?._dialogFragment?.getDialog();
    const view = (modalDialog ?? AndroidUtils.getCurrentActivity())
```

farfromrefug:
@vallemar should mention that it wont work in all cases. For example it wont work with android `PopupWindow` or with `@nativescript-community/ui-material-bottomsheet`. The latter uses `_bottomSheetFragment` and not `_dialogFragment` (nor does it defines modal).
Would be nice to find a implementation where plugins could register for ways to get current windowToken.
I even had a worse case with  `PopupWindow`. I wanted to hide the keyboard when the  `PopupWindow` got hidden. But the `dismiss` event is fired after the  `PopupWindow` view is removed from tree and thus my view `windowToken` is null :s, there is no way around it :s.
My only option was to use a deprecated method to hide keyboard which does not need a windowToken https://github.com/Akylas/NativeScript/commit/825863be9784da7af298a8758a475c9801ca7940

vallemar:
@farfromrefug I understand the problem, and if something common should be done so that the plugins can do this, otherwise we will continue to have this problem with new plugins or views that do not come from NS-core. Maybe you could make a queue of foreground views so you can bring up the last one and grab the current `windowToken`.

@rigor789  While we are thinking about how to do this, I think another PR could be made, since this solves a current problem in the core and I think it should be merged.

farfromrefug:
@vallemar yes it is a question of another PR! 
I LL start thinking about it. But if anyone has an idea dont hesitate ;)

rigor789:
Maybe we need some sort of "fragment stack" to keep track of fragments... but I can see that getting out-of-sync with reality easily. Or maybe we don't need the fragments, just the windowTokens - in which case, a similar `push/pop` stack could be useful as long as it's updated correctly by core/plugins.

NathanWalker (approved):

PR #10392, opened 2023-09-28
feat(visionos): Vision Pro support

NathanWalker:
## PR Checklist

Core did not run directly on visionos.

## What is the new behavior?

Core can now be used on visionos.


migrations.json
rigor789, line 1: Not sure we need to keep these, can't remember...

packages/core/config/config.interface.ts
```diff
@@ -20,11 +20,20 @@ interface IConfigIOS extends IConfigPlatform {
...
+		/**
+		 * Remote package url
+		 */
+		repositoryURL?: string;
+		version?: string;
+		/**
+		 * Local package path relative to project directory
+		 */
+		path?: string;
 	}>;
```
rigor789, line 32: nit: Might be better to copy the same exact types from the trapeze definitions:

```ts
export interface IosRemoteSPMPackageDefinition {
  name: string;
  libs: string[];
  repositoryURL: string;
  version: string;
}

export interface IosLocalSPMPackageDefinition {
  name: string;
  libs: string[];
  path: string;
}

export type IosSPMPackageDefinition = IosRemoteSPMPackageDefinition | IosLocalSPMPackageDefinition;
```

And change this type to `Array<IosSPMPackageDefinition>`. The main difference between the two approaches is that we'd get a warning/error when trying to use `version` or `repositoryURL` in a local SPM package definition (and the other way around, trying to use `path` in a remote definition).

packages/core/platforms/ios/TNSWidgets.xcframework/ios-arm64/dSYMs/TNSWidgets.framework.dSYM/Contents/Resources/Relocations/aarch64/TNSWidgets.yml
```diff
@@ -0,0 +1,82 @@
+---
+triple:          'arm64-apple-darwin'
+binary-path:     '/Users/nstudio/Documents/github/NativeScript/NativeScript/packages/ui-mobile-base/ios/TNSWidgets/build/Release-iphoneos/TNSWidgets.framework/TNSWidgets'
```
rigor789, line 3: This is auto-generated, but I do wonder why it's here now (and if having that path is an issue for consuming these dSYMs).

rigor789 (approved):
LGTM, only added some notes/nits.

migrations.json
NathanWalker: I'll remove it as not needed.

packages/core/config/config.interface.ts
NathanWalker: Thank you, modified to match that now.

packages/core/platforms/ios/TNSWidgets.xcframework/ios-arm64/dSYMs/TNSWidgets.framework.dSYM/Contents/Resources/Relocations/aarch64/TNSWidgets.yml
NathanWalker: Good question, I'm going to remove it from repo and when regenerated prior to release we can check.

PR #10396, opened 2023-10-03
fix(testing): e2e flag no longer needed, testID is now applicable in dev or prod builds

canmertc:
## PR Checklist

- It is not possible to test an accessibility-enabled app by using the same build. Currently, it needs to be passed `--env.e2e` that enables `testID` and blocks accessibility features before building for testing. Passing this argument is needed because `View.setContentDescription` method which is also used by TalkBack to announce items is also used in `testID` property's Android implementation to store the view identifier.
- Maestro (testing framework) is not able to select items by their id.

```js
// Android

<Label text="My Button" testID="my-button" />

// not working
- assertVisible:
     id: "my-button"
     
// not working
- assertVisible:
     text: "My Button"
     
// working (but it shouldn't work this way)
- assertVisible:
     text: "my-button"
 
// working (but it shouldn't work this way)
- assertVisible: "my-button"
```

## What is the new behavior?

- `AccessibilityNodeInfo.setViewIdResourceName` is used to set resource id instead of  `View.setContentDescription` that sets content description on Android.
- No need to pass `--env.e2e` anymore. So the production builds can now be tested.
- `accessibilityIdentifier` will be the source of truth. Setting `testID` property will also set `accessibilityIdentifier`. But setting `accessibilityIdentifier` property will update `testID` only if there is a `testID` already set and has a different value.
- Maestro is now able to select items by their id and text

```js
// Android

<Label text="My Button" testID="my-button" />

// working
- assertVisible:
     id: "my-button"
     
// working
- assertVisible:
     text: "My Button"
     
// working
- assertVisible:
     id: "my-button"
     text: "My Button"
 
// working
- assertVisible: "My Button"
```


NathanWalker (approved):
Brilliant work, thank you @canmertc

NathanWalker (approved):

Issue #3972, opened 2017-04-11
Feature request: text-stroke property and/or text-shadow


ickata:
Hi {N} team,

I wonder if it is technically possible on both platforms and have you got any plans implementing `text-stroke`, and/or `text-shadow` CSS properties? The `text-shadow` is a valid CSS property. `text-stroke` as well, however some vendors support it with prefix (e.g. `-webkit-text-stroke`). 

I need to add outline to text. Semantically `text-stroke` is what I need, but `text-shadow` also does the job when used like this `text-shadow: 0 0 red`.

I'll be happy to know if there is some workaround in tns 2.5 to achieve this.

Thanks!

NordlingArt:
I explained a possible [workaround here](https://github.com/NativeScript/NativeScript/issues/3597#issuecomment-278889890) for iOS. However, it somewhat doesn't render the outline even though it should.

tsonevn:
Hi @ickata,
A similar issue for `text-shadow` support has been logged [here](https://github.com/NativeScript/NativeScript/issues/3597). Please if you have some further suggestions comment on the other issue.

AdityaChoudhary:
+1

NathanWalker:
Reopening - we will be supporting this in the next minor release of core.

PR #10399, opened 2023-10-09
feat(css): text-stroke support

NathanWalker:
## PR Checklist

You could not easily add stroke to text and core did not support `text-stroke`.

## What is the new behavior?

CSS property `text-stroke` is now supported per this spec:
https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-stroke

Example:
```xml
<Label style="text-stroke: 1px black">text-stroke</Label>
```
Screenshot 2023-10-08 at 7 25 12 PM

closes [#3597](https://github.com/NativeScript/NativeScript/issues/3597)
closes [#3972](https://github.com/NativeScript/NativeScript/issues/3972)


farfromrefug:
@NathanWalker I think the StyleableTextView should be native. By creating you actually made every single Label slower because it has o go through JS on every draw.
Seeing label is the most important(used) view, I think it s an issue

packages/core/ui/styling/css-shadow.ts
```diff
@@ -44,67 +40,46 @@ export function parseCSSShadow(value: string): ShadowCSSValues {
 		return null;
 	}
 
-	let colorRaw = 'black';
+	let color = 'black';
```
shirakaba, line 43: Wondering whether the default should be 'transparent' rather than 'black' 🤔 just that 'black' may accidentally look fine in light mode and then surprise people a few hours later in dark mode.

Looks like [-webkit-text-stroke](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-stroke-color) defaults to `currentcolor` (which inherits from `color`). Is there any way to easily get hold of `currentcolor`? I'm guessing there's no precedent, so maybe too hard to implement. I'm not sure to what extent NativeScript supports inheriting color styles in the first place.

packages/core/ui/styling/css-stroke.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { parseCSSStroke } from './css-stroke';
+import { CoreTypes } from '../../core-types';
+import { Length } from './style-properties';
+import { Color } from '../../color';
+
+describe('css-text-stroke', () => {
+	it('empty', () => {
```
shirakaba, line 7: Would be good to have a test for setting to `1px black`, then setting back to `0px black` to see if the style can be set and unset.

shirakaba (commented):
Just some passing thoughts while eating a donut. I haven't looked too closely beyond the bits I've commented on.

packages/core/platforms/ios/src/UIView+NativeScript.m
```diff
@@ -130,4 +130,19 @@ -(void)nativeScriptSetFormattedTextDecorationAndTransform:(NSDictionary*)details
         ((UILabel*)self).attributedText = attrText;
     }
 }
+
+-(void)nativeScriptSetFormattedTextStroke:(CGFloat)width color:(UIColor*)color {
+    if (width > 0) {
```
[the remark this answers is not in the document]
NathanWalker: It’s removed at top of setNativeText before this is called - attrText is always reset when style changes and attributes re-determined.

packages/core/ui/styling/css-shadow.ts
NathanWalker: This maintains same behavior as box shadow - no color setting is black.

NathanWalker:
> @NathanWalker I think the StyleableTextView should be native. By creating you actually made every single Label slower because it has o go through JS on every draw. Seeing label is the most important(used) view, I think it s an issue

Good call @farfromrefug I’ll make it native.

packages/core/ui/styling/css-stroke.spec.ts
NathanWalker: Toolbox demonstrates it working properly now via a binding 👍

Issue #10403, opened 2023-10-12
Error when box-shadow is set no none


asharghi:
### Issue Description

After updating to core 8.6, I get this error. Seems to work with box-shadow: none anyway, but the error should be removed

```
CONSOLE ERROR: Error: Failed to apply property [box-shadow] with value [none] to Button(11). TypeError: Cannot read properties of null (reading 'values')
  JS:     at parseCSSShadow (file: app/webpack:/css-test/node_modules/@nativescript/core/ui/styling/css-shadow.js:58:0)
  JS:     at valueConverter (file: app/webpack:/css-test/node_modules/@nativescript/core/ui/styling/style-properties.js:1146:29)
  JS:     at Style.setCssValue [as css:box-shadow] (file: app/webpack:/css-test/node_modules/@nativescript/core/ui/core/properties/index.js:573:0)
  JS:     at CssState.setPropertyValues (file: app/webpack:/css-test/node_modules/@nativescript/core/ui/styling/style-scope.js:555:47)
  JS:     at file: app/webpack:/css-test/node_modules/@nativescript/core/ui/styling/style-scope.js:431:0
  JS:     at Button._batchUpdate (file: app/webpack:/css-test/node_modules/@nativescript/core/ui/core/view-base/index.js:361:0)
  JS:     at CssState.updateDynamicState (file: app/webpack:/css-test/node_modules/@nativescript/core/ui/styling/style-scope.js:429:0)
  JS:     at CssState.onLoaded (file: app/webpack:/css-test/node_modules/@nativescript/core/ui/styling/style-scope.js:399:0)
  JS:     at Button.onLoaded (file:///data/data/org.nativescript.csst...
```

### Reproduction

Add an element with box-shadow: none in css or inline style

### Environment

```
"dependencies": {
    "@nativescript/core": "~8.6.0",
    "@nativescript/theme": "~3.0.2",
    "nativescript-vue": "~2.9.3"
  },
  "devDependencies": {
    "@nativescript/android": "8.5.4",
    "@nativescript/webpack": "~5.0.18",
    "nativescript-vue-template-compiler": "~2.9.3"
  }
```

### Please accept these terms

- [X] I have searched the [existing issues](https://github.com/NativeScript/NativeScript/issues) as well as [StackOverflow](https://stackoverflow.com/questions/tagged/nativescript) and this has not been posted before
- [X] This is a bug report
[...]

PR #10405, opened 2023-10-13
fix(core): box-shadow 'none' handling 

NathanWalker:
## PR Checklist

box-shadow shorthand 'none' would work but throw a benign error in parsing.

## What is the new behavior?

The value 'none' now applies normally as 'empty'.

closes https://github.com/NativeScript/NativeScript/issues/10403

### DIFF
--- a/packages/core/ui/styling/css-shadow.spec.ts
+++ b/packages/core/ui/styling/css-shadow.spec.ts
@@ -134,4 +134,14 @@ describe('css-shadow', () => {
 		expect(shadow.spreadRadius).toBeUndefined();
 		expect(shadow.color).toEqual(new Color('#333'));
 	});
+
+	it('none', () => {
+		const shadow = parseCSSShadow('none');
+		expect(shadow.inset).toBe(false);
+		expect(shadow.offsetX).toBe(CoreTypes.zeroLength);
+		expect(shadow.offsetY).toBeUndefined();
+		expect(shadow.blurRadius).toBeUndefined();
+		expect(shadow.spreadRadius).toBeUndefined();
+		expect(shadow.color).toEqual(new Color('black'));
+	});
 });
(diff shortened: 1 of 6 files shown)


NathanWalker:
@asharghi I added an extra unit test for this case to prevent in future, good catch.

packages/core/ui/styling/css-shadow.spec.ts
```diff
@@ -134,4 +134,14 @@ describe('css-shadow', () => {
 		expect(shadow.spreadRadius).toBeUndefined();
 		expect(shadow.color).toEqual(new Color('#333'));
 	});
+
+	it('none', () => {
+		const shadow = parseCSSShadow('none');
+		expect(shadow.inset).toBe(false);
+		expect(shadow.offsetX).toBe(CoreTypes.zeroLength);
```
rigor789, line 141: Why is `offsetX` expected to be zeroLength, but `offsetY` `undefined`? Shouldn't both be the same, either zeroLength or `undefined`?

packages/core/ui/styling/css-shadow.ts
```diff
@@ -36,15 +36,13 @@ export function parseCSSShorthand(value: string): {
 	const first = parts[0];
 	const last = parts[parts.length - 1];
 
-	if (first === 'none') {
```
rigor789, line 39: I think short-circuiting and returning in case `first` is `none` is easier/cleaner to read than nesting ifs.

NathanWalker: Refactored to be a general css shorthand parser so could be reused for shadow and stroke, thus returning same structure each time.

NathanWalker: A css-utils.ts may be best as there is quite a bit of overlap in parsing throughout css props.

packages/core/ui/styling/css-shadow.spec.ts
NathanWalker: Ok @rigor789 I cleaned that up to now be properly undefined, even color as well which is much cleaner.

packages/core/ui/styling/css-shadow.ts
NathanWalker: Turns out there was already a css-utils so was able to move there and brought back your early return exit in that case 👍

PR #10439, opened 2023-11-13
fix(android): prevent error while navigation back after using page tr…

farfromrefug:
…ansition

When navigating transition is set on the `currentEntry` but never cleaned up
 which means that on a next navigation forward (without transition) and back
 we will go in `androidFragmentTransactionCallback` with a transition from  `currentEntry` but no state

This is not the perfect fix as the best one would be to clear `transition` on navigation end. But it is working already


packages/core/ui/transition/page-transition.android.ts
```diff
@@ -8,8 +8,9 @@ import { getPageStartDefaultsForType, getRectFromProps, SharedTransition, Shared
 import { ImageSource } from '../../image-source';
 import { ContentView } from '../content-view';
 import { GridLayout } from '../layouts/grid-layout';
-import { ad } from '../../utils';
 import { Screen } from '../../platform';
+import { ExpandedEntry } from '../frame/fragment.transitions.android';
```
CatchABus, line 12: I think we can avoid using android-specific path and import module directly from `fragment.transitions`.

farfromrefug: @CatchABus I LL look but I am not sure ExpandedEntry is exported outside of android

NathanWalker: In these case @farfromrefug it's best to add a stub interface for `ExpandedEntry` in a `fragment.transitions.d.ts` to avoid platform specific imports in bundled code like this.

NathanWalker (approved):

PR #10440, opened 2023-11-15
feat: `defaultVisualState` to override 'normal' default visualState

farfromrefug:
for example to fake disabled state while allowing touch


packages/core/ui/core/view-base/index.ts
```diff
@@ -380,6 +380,11 @@ export abstract class ViewBase extends Observable implements ViewBaseDefinition
 	 */
```
CatchABus, line 380: I think we should move the declarations of view-base into view-common and its types definition file.
Keeping them here might make developers think that views which extend `ViewBase` can support this property.

CatchABus (changes requested):

NathanWalker (approved):

PR #10443, opened 2023-11-16
fix: set modal dialog root view parent

farfromrefug:
now the dialog root view has the `Application` `rootView` as parent to be able to access the `ns-root` class which comply with css specs (for example to read root css variables)


CatchABus:
> now the dialog root view has the `Application` `rootView` as parent to be able to access the `ns-root` class which comply with css specs (for example to read root css variables)

I think it's more like a feature than a fix. Old docs state that you have to use `.ns-modal` for styling modal root view and children.
https://v6.docs.nativescript.org/angular/ui/styling#root-views-css-classes

farfromrefug:
You still can. Ns-modal only apply to modals
But without that PR you can't access root CSS variables from a modal CSS. Which you would expect ( I think). I mean the rootview of application also is the root view for modals, bottom sheet...

CatchABus:
> You still can. Ns-modal only apply to modals But without that PR you can't access root CSS variables from a modal CSS. Which you would expect ( I think). I mean the rootview of application also is the root view for modals, bottom sheet...

I see your point but will this implementation work for the case when root view is a Frame? That class seems to have a special handline for `eachChildView` and it might be a problem: https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/frame/frame-common.ts#L538

vallemar:
@farfromrefug I use this and work fine https://github.com/vallemar/nativescript-reordering/blob/main/src/app.css#L17-L38
https://github.com/vallemar/nativescript-reordering/blob/main/src/app.css#L17-L38

farfromrefug:
@vallemar yes it does. But not when you start modifying them from yes. From what you have root  and modal properties are actually not the same

farfromrefug:
@CatchABus it should not change anything. EachChild and parent are decoupled. Setting the parent does not imply it will be in eachChild. It won't !

CatchABus (dismissed):

PR #10450, opened 2023-11-27
fix(android): crash with material components

farfromrefug:
`setSecureAndKeyboardType` needs to use `nativeTextViewProtected`  As a reminder in text base components there is `nativeViewProtected` and `nativeTextViewProtected`. The reason is that material components (and maybe other) differentiate the layout and the textfield itself inside one view. A while back i made a PR for the whole N to support those two getters. But we need to remain careful on which one we use everytime

## PR Checklist

## What is the new behavior?

### DIFF
--- a/packages/core/ui/text-field/index.android.ts
+++ b/packages/core/ui/text-field/index.android.ts
@@ -40,7 +40,7 @@ export class TextField extends TextFieldBase {
 	setSecureAndKeyboardType(): void {
 		let inputType: number;
 
-		const nativeView = this.nativeViewProtected;
+		const nativeView = this.nativeTextViewProtected;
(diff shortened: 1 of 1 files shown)


CatchABus (changes requested):
I agree we should be more careful and stick to it.
Since we're doing this, I found few more references that could be corrected: https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/text-base/index.android.ts#L508

farfromrefug:
@CatchABus actually this case does not matters a lot as setTextStroke is a method of the custom component used for label. dont think we use it for textfield though. might even crash with N textfield. we should check for method existence

CatchABus (approved):

PR #10452, opened 2023-11-30
feat(wepback): support custom `buildPath` through `nativescript.config`

farfromrefug:
this goes with a cli PR that i will create after

### DIFF
--- a/packages/webpack5/src/configuration/base.ts
+++ b/packages/webpack5/src/configuration/base.ts
@@ -84,7 +84,9 @@ export default function (config: Config, env: IWebpackEnv = _env): Config {
 	// when using hidden-source-map, output source maps to the `platforms/{platformName}-sourceMaps` folder
 	if (env.sourceMap === 'hidden-source-map') {
 		const sourceMapAbsolutePath = getProjectFilePath(
-			`./platforms/${platform}-sourceMaps/[file].map[query]`
+			`./${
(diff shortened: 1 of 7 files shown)


packages/webpack5/src/configuration/base.ts
```diff
@@ -84,7 +84,9 @@ export default function (config: Config, env: IWebpackEnv = _env): Config {
 	// when using hidden-source-map, output source maps to the `platforms/{platformName}-sourceMaps` folder
 	if (env.sourceMap === 'hidden-source-map') {
 		const sourceMapAbsolutePath = getProjectFilePath(
-			`./platforms/${platform}-sourceMaps/[file].map[query]`
+			`./${
+				env.buildPath || 'platforms'
```
rigor789, line 88:
```suggestion
				env.buildPath ?? 'platforms'
```

packages/webpack5/src/helpers/copyRules.ts
```diff
@@ -1,7 +1,7 @@
 import CopyWebpackPlugin from 'copy-webpack-plugin';
 import { basename, relative, resolve } from 'path';
 import Config from 'webpack-chain';
-import { sync as globbySync } from 'globby';
+import { globbySync } from 'globby';
```
rigor789, line 4: Will double check, guessing they are now exporting it as `globbySync` too?

PR #10466, opened 2024-01-02
fix(android): crash on writeText on Android > 29

farfromrefug:
it was failing if you were to pass `content://com.android.externalstorage.documents/document/primary%3ADownload%2Fcom.akylas.documentscanner_settings_2024-01-02.json`.
It was trying to write `com.akylas.documentscanner_settings_2024-01-02.json/com.akylas.documentscanner_settings_2024-01-02.json` instead of `com.akylas.documentscanner_settings_2024-01-02.json`. It seems the `getFile` method is wrong.

This is not the perfect fix but in the meaning it should always work


triniwiz (approved):
lgtm

Issue #10297, opened 2023-05-25
WinterTC Minimum Common Web Platform API Alignment


NathanWalker:
### Is your feature request related to a problem? Please describe.

In addition to what the NativeScript runtimes provide, they should also provide all the common interfaces and global objects expected from any state of the art JavaScript runtime. [WinterCG](https://wintercg.org/) is a community group doing excellent work in this space to help runtime interoperability of APIs across runtimes.

- https://wintercg.org/work
- https://common-min-api.proposal.wintercg.org/

### Describe the solution you'd like

Ensure compliance with at least the [minimum specification](https://common-min-api.proposal.wintercg.org/):

**Interfaces:**

- [ ] [AbortController](https://dom.spec.whatwg.org/#abortcontroller)
- [ ] [AbortSignal](https://dom.spec.whatwg.org/#abortsignal)
- [ ] [Blob](https://w3c.github.io/FileAPI/#dfn-Blob)
- [ ] [ByteLengthQueuingStrategy](https://streams.spec.whatwg.org/#bytelengthqueuingstrategy)
- [ ] [CompressionStream](https://wicg.github.io/compression/#compression-stream)
- [ ] [CountQueuingStrategy](https://streams.spec.whatwg.org/#countqueuingstrategy)
- [x] [Crypto](https://w3c.github.io/webcrypto/#dfn-Crypto) **8.8**
- [ ] [CryptoKey](https://w3c.github.io/webcrypto/#dfn-CryptoKey)
- [ ] [DecompressionStream](https://wicg.github.io/compression/#decompression-stream)
- [ ] [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException)
- [ ] [Event](https://dom.spec.whatwg.org/#event)
- [ ] [EventTarget](https://dom.spec.whatwg.org/#eventtarget)
- [ ] [File](https://w3c.github.io/FileAPI/#dfn-file)
- [ ] [ReadableByteStreamController](https://streams.spec.whatwg.org/#readablebytestreamcontroller)
- [ ] [ReadableStream](https://streams.spec.whatwg.org/#readablestream)
- [ ] [ReadableStreamBYOBReader](https://streams.spec.whatwg.org/#readablestreambyobreader)
- [ ] [ReadableStreamBYOBRequest](https://streams.spec.whatwg.org/#readablestreambyobrequest)
- [ ] [ReadableStreamDefaultController](https://streams.spec.whatwg.org/#readablestreamdefaultcontroller)
[...]

PR #10471, opened 2024-01-08
feat: Initial WinterCG modules

triniwiz:
- TextEncoder
- TextDecoder
- URL
 https://github.com/NativeScript/NativeScript/issues/10297


packages/core/src-native/ios/NativeScript.podspec
```diff
@@ -0,0 +1,55 @@
...
+    s.version      = '1.0.0'
+    s.summary      = "description"
+    s.license      = "MIT"
+
+    s.authors      = "Ammar Ahmed"
+    s.homepage     = "https://github.com/ammarahm-ed/nativescript-v8-module"
+    s.platforms    = { :ios => "12.4" }
+    s.source       = { :git => "https://github.com/ammarahm-ed/nativescript-v8-module.git", :tag => "v1.0.0" }
+    s.cocoapods_version      = ">= 1.10.1"
+    s.vendored_frameworks = "NativeScript.xcframework"
```
shirakaba, line 12: This seems to vend `NativeScript.xcframework`. How were we handling that before, and what's changed now?

It feels like this stuff should be a separate standalone PR. As they're all new files, we can't tell in the git histroy whether anything has changed specifically for WinterCG.

packages/core/url/index.ts
```diff
@@ -0,0 +1,223 @@
+require('../native-helpers');
```
shirakaba, line 1: Is there anything we can do to avoid side-effects? This kind of pattern causes a lot of trouble, making it harder to migrate to new bundlers, tree-shake, opt-out, etc.

packages/core/url/index.ts
```diff
@@ -0,0 +1,223 @@
...
+let sharedPreferences;
+interface BlobItem {
+	blob: Blob;
+	path?: string;
+	type?: string;
+	ext?: string;
+}
+const BLOB_STORE = new Map<string, BlobItem>();
+
+import './url-search';
```
shirakaba, line 27: Again, importing for effects is a troublesome pattern if we can possibly avoid it. Though it is attractive to avoid changing the code of the URLSearchParams polyfill.

packages/core/globals/index.ts
```diff
@@ -311,14 +311,17 @@ export function initGlobal() {
...
+		installPolyfills('text', ['TextDecoder', 'TextEncoder', 'atob', 'btoa']);
 
 		global.registerModule('xhr', () => require('../xhr'));
 		installPolyfills('xhr', ['XMLHttpRequest', 'FormData', 'Blob', 'File', 'FileReader']);
 
 		global.registerModule('fetch', () => require('../fetch'));
 		installPolyfills('fetch', ['fetch', 'Headers', 'Request', 'Response']);
 
+		global.registerModule('url', () => require('../url'));
+		installPolyfills('url', ['URL']);
```
shirakaba, line 323: I'd prefer to separately polyfill each of URL, URLSearchParams, etc. so that we don't have to rely upon side-effects in the URL module itself.

tools/scripts/build-rust-android.sh
```diff
@@ -0,0 +1,32 @@
...
+
+NDK_TARGET=$TARGET
+
+# if [ "$TARGET" = "arm-linux-androideabi" ]; then
+#     NDK_TARGET="armv7a-linux-androideabi"
+# fi
+
+API_VERSION="21"
+NDK_VERSION="23.1.7779620"
+NDK_HOST="darwin-x86_64"
```
shirakaba, line 19: Is this build only gonna work on Intel and Rosetta? Wonder if it can be made more platform-agnostic.

tools/scripts/copy-rust-android.sh
```diff
@@ -0,0 +1,22 @@
+#! /bin/bash
```
shirakaba, line 1: Let's make the script fail upon any command exiting with a non-zero status code.

```suggestion
#! /bin/bash
set -e
```

tools/scripts/build-rust-headers.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
shirakaba, line 1: Let's make the script fail upon any command exiting with a non-zero status code.

```suggestion
#! /bin/bash
set -e
```

tools/scripts/build-rust-android.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
shirakaba, line 1: Let's make the script fail upon any command exiting with a non-zero status code.

```suggestion
#! /bin/bash
set -e
```

packages/ui-mobile-base/android/widgets/src/main/jniLibs/arm64-v8a/libnativescriptcore.so
shirakaba, line 1: I worry about having to maintain native libs. It makes the build process a lot more complicated. I'd prefer to be in a position where we don't need extra expertise to help you maintain these features 🙇‍♂️

packages/ui-mobile-base/android/widgets/src/main/cpp/include/nativescript_core.h
```diff
@@ -0,0 +1,171 @@
+#ifndef CBINDGEN_BINDINGS_H
+#define CBINDGEN_BINDINGS_H
+
+/* Warning, this file is autogenerated by cbindgen. Don't modify this manually. */
```
shirakaba, line 4: If these files are generated by cbindgen, should they be committed to source in the first place? What's the usual approach?

Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[workspace]
```
shirakaba, line 1: Congrats on the first use of Rust in Core 🥳

I would say, however, that I'm worried about extra build tooling as it increases the expertise needed to maintain Core.

shirakaba (changes requested):
Thanks so much for leading the charge on this. It's a crazy amount of work and I'm looking forward to making WinterCG conformance a reality.

I've done a quick review of the areas I understand (so, ignoring Android), but my overall impression is:

* The PR is too big to safely review.
  * Can we break it down into more manageable pieces? GitHub's review interface is crawling 🫠
* It has coupling to V8 and JSI.
  * With the new Node-API-based NativeScript on the way, I worry that this effort would need to be repeated all over again, so I would prefer if we could take an engine-agnostic approach, i.e. no usage of V8 and JSI APIs.
* It looks like a lot of extra complication in build tooling, introducing makefiles and Rust (and a lot more logic in C++-land), which may make it harder to set up on CI and get building on other people's machines, and may slow down builds.
  * Would it be okay to sacrifice some performance in order to simplify the architecture? I'd prefer if we were in a position to help you maintain this 😅

Discussions to be had:
* Is there a reasonable, easy path to opt-in/opt-out of APIs?
  * This is maybe not a deal-breaker, but if it's within reach, that'd be of interest.
* Does this need to live in Core?
  * Core is already complicated to maintain. Could each API perhaps be done as a plugin?
* Would some of these modules be just as good as a Webpack polyfill? Given that JS polyfills exist for these, it'd be a very fair tradeoff for simplicity.

Again, really thanks for the draft to give us something concrete to discuss. Really hope to land this in some way.

packages/core/src-native/ios/NativeScript.podspec
triniwiz: This was needed to make the build happy when doing the cpp builds, the hope is when we moved to pod or spm (gotta test this one) we can drop this pod

packages/core/url/index.ts
triniwiz: I have a few ideas to clean this up

packages/core/url/index.ts
triniwiz: This can move move but this polyfill is only temp until the `URLSearchParams` pr is merged in the rust crate

packages/core/url/index.ts
```diff
@@ -0,0 +1,223 @@
...
+
+export class URL {
```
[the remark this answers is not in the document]
triniwiz: Ok sounds good

tools/scripts/build-rust-android.sh
triniwiz: This works on the m1 afaik they didn't publish a `darwin-aarch64`

packages/ui-mobile-base/android/widgets/src/main/jniLibs/arm64-v8a/libnativescriptcore.so
triniwiz: I can remove these from as they will be generated when calling the build cmd

packages/ui-mobile-base/android/widgets/src/main/cpp/include/nativescript_core.h
triniwiz: This can be ignored as well

triniwiz:
Sure we can sacrifice the performance to simplify the architecture so if needed I can do it as Android and iOS libs.

1. Is there a reasonable, easy path to opt-in/opt-out of APIs?

Turn it into plugins/packages

2. Does this need to live in Core?
No up to the tsc to decide 

3. Core is already complicated to maintain. Could each API perhaps be done as a plugin?
Yeah 🤷🏽

Would some of these modules be just as good as a Webpack polyfill (or even done by deferring to spinning up the system JS runtime, like JSC)?
Yeah we can but will lose some performance here, again up to the tsc.

farfromrefug:
@triniwiz what performance do we loose by using webpack polyfills? I dont see what it will change

To me if those are moved out of this repo we can use rust/c++ as we don't bloat this repo. Now tbh for such libraries is rust really needed? What does it bring to the table instead of simple c++ modules?

If understand correctly where we'll loose in performance is if we become engine agnostic and don't directly use v8 apis. Am i right @triniwiz ? But do we really need that for those APIs? I mean we rarely use those high framerate computing or heavy computation like done in @nativescript/canvas.

Also to point out that this PR is adding around 90mb of so libs which should not be in here

NathanWalker:
Moved here: https://github.com/NativeScript/winter-cg

PR #10474, opened 2024-01-13
Segmented bar

### LINKED ISSUE #10472: [Android] SegmentedBar selectedBackgroundColor not working
### Issue Description

Hello @NathanWalker @farfromrefug,

This for a new PR on The segmentedBar, on android the `selectedBackgroundColor` doesn't work at all, there was some deprecations and some stuff to be done in order to work.

### Reproduction

_No response_

_No response_

### Environment

_No response_

### Please accept these terms

- [X] This is a bug report

kefahB:
## PR Checklist

- [x] All existing tests are passing: I have wort the tests but I couldn't figure out a way to compare the background color, I've manually well tested the changes

On Android, the `selectedBackgroundColor` doesn't work at all (on first load and on select index change), we can not see the selected item.

## What is the new behavior?
The background color does appear at the first load and on selected index changed

Fixes/Implements/Closes #10472 .

I have added a new property for both platforms to control `selectedTextColor` that can be set for more control of the selected state.

### DIFF
--- a/packages/core/ui/segmented-bar/index.android.ts
+++ b/packages/core/ui/segmented-bar/index.android.ts
@@ -51,8 +51,13 @@ function initializeNativeClasses(): void {
 
 		onTabChanged(id: string): void {
 			const owner = this.owner;
-			if (owner.shouldChangeSelectedIndex()) {
-				owner.selectedIndex = parseInt(id);
+			if (owner) {
+				setTimeout(() => {
+					owner.setTabColor(id);
+				});
+				if (owner.shouldChangeSelectedIndex()) {
+					owner.selectedIndex = parseInt(id);
+				}
 			}
 		}
 	}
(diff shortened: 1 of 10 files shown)


packages/core/ui/segmented-bar/index.ios.ts
```diff
@@ -2,7 +2,8 @@ import { Font } from '../styling/font';
 import { SegmentedBarItemBase, SegmentedBarBase, selectedIndexProperty, itemsProperty, selectedBackgroundColorProperty } from './segmented-bar-common';
 import { colorProperty, fontInternalProperty } from '../styling/style-properties';
 import { Color } from '../../color';
-import { iOSNativeHelper } from '../../utils';
+import { Trace } from '../../trace';
+import { Utils } from '../../index';
```
NathanWalker, line 6: Better to use direct deep imports here instead of the "barrels" like `Utils`. So best to use iOSNativeHelper.
This is due to bundling - core needs to stick to direct imports everywhere inside core.

NathanWalker (changes requested):
Nice PR! Small comment.

packages/core/ui/segmented-bar/index.ios.ts
kefahB: Fixed

packages/core/ui/segmented-bar/index.ios.ts
```diff
@@ -3,7 +3,7 @@ import { SegmentedBarItemBase, SegmentedBarBase, selectedIndexProperty, itemsPro
 import { colorProperty, fontInternalProperty } from '../styling/style-properties';
 import { Color } from '../../color';
 import { Trace } from '../../trace';
-import { Utils } from '../../index';
+import { MajorVersion } from 'utils/ios';
```
kefahB, line 6: @NathanWalker this is what you meant ?

NathanWalker: All symbols should be imported via relative imports to their exact location when working within core source. 'utils/ios' is not a relative import. I pushed up the correct usage there for you and will get this merged, thanks again.

NathanWalker (approved):

CatchABus:
@kefahB This PR broke SegmentedBar selection color styling.
Let me note that `selectedBackgroundColor` worked on my device without needing this PR.

This is how it used to be:
Screenshot_20240312-133934 (1)

This is how it is now, after these changes:
Screenshot_20240312-134157 (2)

packages/core/ui/segmented-bar/index.android.ts
```diff
@@ -292,4 +298,42 @@ export class SegmentedBar extends SegmentedBarBase {
...
+				const unselectedTextColor = this.getColorForAndroid(this.color ?? '#6e6e6e');
+				const selectedTextColor = this.getColorForAndroid(this?.selectedTextColor ?? '#000000');
+				const unselectedBackgroundColor = this.getColorForAndroid(this?.backgroundColor ?? '#dbdbdb');
+				const selectedBackgroundColor = this.getColorForAndroid(this?.selectedBackgroundColor ?? this?.backgroundColor ?? 'blue');
+				if (tabWidget) {
+					for (let i = 0; i < tabWidget.getTabCount(); i++) {
+						const view = tabWidget.getChildTabViewAt(i);
+						const item = this.items[i];
+						const textView = item?.nativeViewProtected;
+						view.setBackgroundColor(unselectedBackgroundColor);
```
CatchABus, line 314: @kefahB Still, setting background color here seems wrong as it applies color to entire segment button (it's just as my screenshots indicate). Would it work on your device if we removed those `setBackgroundColor` calls?

CatchABus (commented):
Setting background color this way seems to cause the problem.

packages/core/ui/segmented-bar/index.android.ts
kefahB: I need time to check this because I am. not on the project, but at the time I does not working! Please not that also we can keep it as optional to give same apparence oh both platforms

kefahB:
What do you suggest ?

CatchABus:
> What do you suggest ?

We can leave this be for now but I think we should work on it more and get rid of hackish solutions. If possible, let us know of the phone model and OS version where the problem occured in case we revisit this.

PR #10478, opened 2024-01-20
feat(visionos): ui-mobile-base supporting xros plus improvements to window handling

NathanWalker:
## PR Checklist


farfromrefug:
I dont think you should commit xcframeworks. they should be built locally

NathanWalker:
Completely agree @farfromrefug your PR here will help as well which we need to finish testing with:
https://github.com/NativeScript/NativeScript/pull/10387/files

The hardest debate on that topic is ability to contribute to core quickly/easily. If each contributor has to build an .xcframework just to run core locally to fix something and post a PR? Likely not great (maybe they encounter an issue with the .xcframework build and they become blocked to help fix something?)

farfromrefug:
I dont sée any issue with people having to build something to test. every single android, iOS, c++... requires the user to build and have a dev env. I even think it is good thing .
still I understand what you mean.
I am on my phone right now but that xcframework is the widget one? if so WR could either publish on cocoapod but er could also put the sources directly in platforms/iOS/src. it would make the build a bit slower but I would still tend for this solution :
* it allow anyone to test changing sources directly in their app and then create a PR without even having to clone N
* users could patch sources (advanced usage) for their own need.
* useful for contributors to work directly on widget sources in production apps using XCode. (same thing would be awesome for android).

but if it is not the widgets xcframework my point is useless :p

triniwiz (approved):
LGTM

PR #10506, opened 2024-04-07
feat: enable devtools for android & support network request timings

triniwiz:


farfromrefug:
@triniwiz the code in http should still be hidden behind global variable set through webpack.
with your changes the inspector and everything for debug is included in production builds. It should not and could even be a security issue

NathanWalker:
Good mention @farfromrefug pushed up the fix here:
https://github.com/NativeScript/NativeScript/commit/0b330d8fede7dd7386aab6b10bce13b852032613

NathanWalker (approved):

PR #10522, opened 2024-04-18
fix(android): ListView tap not working after setting children as focusable

CatchABus:
## PR Checklist

We had reports that android ListView `itemTap` event stopped working at core 8.7 and later.
The problem seemed to be that ListView proxy views became focusable due to some a11y calls, thus breaking native item click listener. A thread in StackOverflow explains why ListView does not operate well if list items become focusable: https://stackoverflow.com/questions/11610023/click-is-not-working-on-the-listitem-listview-android

## What is the new behavior?
Views will become focusable or non-focusable only when needed.

This is a continuation of the previous fix related to WebView accessibility focus: https://github.com/NativeScript/NativeScript/commit/92b2ff83a038ece4ab4411cf983ee9e252bba141

@farfromrefug Let me know if there's room for any improvements.

### DIFF
--- a/packages/core/ui/core/view/index.android.ts
+++ b/packages/core/ui/core/view/index.android.ts
@@ -179,8 +179,8 @@ function initializeDialogFragment() {
 		}
 		public onCreate(savedInstanceState: android.os.Bundle) {
 			super.onCreate(savedInstanceState);
-			var ownerId = this.getArguments()?.getInt(DOMID);
-			var options = getModalOptions(ownerId);
+			const ownerId = this.getArguments()?.getInt(DOMID);
+			const options = getModalOptions(ownerId);
 			// The teardown when the activity is destroyed happens after the state is saved, but is not recoverable,
 			// Cancel the native dialog in this case or the app will crash with subsequent errors.
 			if (savedInstanceState != null && options === undefined) {
@@ -325,8 +325,7 @@ export class View extends ViewCommon {
 
 	public _dialogFragment: androidx.fragment.app.DialogFragment;
 	public _manager: androidx.fragment.app.FragmentManager;
-	private _isClickable: boolean;
-	private _isFocusable: boolean;
+	private _isFocusable: boolean = false;
 	private touchListenerIsSet: boolean;
 	private touchListener: android.view.View.OnTouchListener;
 	private layoutChangeListenerIsSet: boolean;
@@ -466,7 +465,6 @@ export class View extends ViewCommon {
 
 	public initNativeView(): void {
 		super.initNativeView();
-		this._isClickable = this.nativeViewProtected.isClickable();
 		this._isFocusable = this.nativeViewProtected.isFocusable();
 		if (this.needsOnLayoutChangeListener()) {
 			this.setOnLayoutChangeListener();
--- a/packages/core/ui/core/view/index.android.ts
+++ b/packages/core/ui/core/view/index.android.ts
@@ -1269,17 +1269,6 @@ export class ContainerView extends View {
 export class CustomLayoutView extends ContainerView implements CustomLayoutViewDefinition {
 	nativeViewProtected: android.view.ViewGroup;
 
-	constructor() {
-		super();
-
-		/**
-		 * mark accessible as false without triggering property change
-		 * equivalent to changing the default
-		 * TODO: Remove this when we have a more flexible API for declaring default property values per type of view
-		 */
-		this.style[accessibilityEnabledProperty.key] = false;
-	}
-
 	public createNativeView() {
 		return new org.nativescript.widgets.ContentLayout(this._context);
 	}
--- a/packages/core/ui/core/view/index.android.ts
+++ b/packages/core/ui/core/view/index.android.ts
@@ -325,7 +325,7 @@ export class View extends ViewCommon {
 
 	public _dialogFragment: androidx.fragment.app.DialogFragment;
 	public _manager: androidx.fragment.app.FragmentManager;
-	private _isFocusable: boolean = false;
+	private _isFocusableByDefault: boolean = false;
 	private touchListenerIsSet: boolean;
 	private touchListener: android.view.View.OnTouchListener;
 	private layoutChangeListenerIsSet: boolean;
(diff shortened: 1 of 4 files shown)


farfromrefug:
@CatchABus only issue I see is that it is not synced with isUserInteractionEnabled.if accessibility is disabled then focusable seems to not be linked to isUserInteractionEnabled

CatchABus:
> @CatchABus only issue I see is that it is not synced with isUserInteractionEnabled.if accessibility is disabled then focusable seems to not be linked to isUserInteractionEnabled

We still have the following scenario. We have this view:
```xml
<stacklayout></stacklayout>
```
Even though `isUserInteraction` is `true` by default, the native view of this layout will still be non-focusable.
So we have a lack of sync regardless of `accessible` value.
What we might need to check this correctly is:
```js
this.isUserInteractionEnabled && this.nativeViewProtected.isFocusable()
```

So if we keep `isUserInteractionEnabled`, it's possible to reduce changes to:
```js
[accessibilityEnabledProperty.setNative](value) {
        // ensure `accessibilityEnabled=false` does not disable focus for view with `isUserInteractionEnabled=true`
        this.nativeViewProtected.setFocusable(!!value || this.isUserInteractionEnabled && this.nativeViewProtected.isFocusable());
        if (value) {
            updateAccessibilityProperties(this);
        }
    }
````

What do you think?

**Comment by farfromrefug** [comment]
@CatchABus You are right. Feels like a lot of things need to be fixed there. 
So what about in the way you did it, store isFocusable default (in createNativeView cause initNativeView is too late as it can be modified by us), then use that default to know if we want to setFocusable when setting isUserInteractionEnabled/accessibilityEnabled to true (or anywhere else really). It could even be cached on a native class bases like Ammar did it for padding?
I still believe we should not go through `accessibilityEnabledProperty.setNative` for the default value. Will check this

**Review by NathanWalker — APPROVED** [review]

File: packages/core/ui/core/view/index.android.ts [hunk 1]
```diff
@@ -465,7 +464,7 @@ export class View extends ViewCommon {
 
 	public initNativeView(): void {
 		super.initNativeView();
-		this._isClickable = this.nativeViewProtected.isClickable();
```
farfromrefug, line 468: @CatchABus why did you remove that line ? I dont see the connection To this PR

CatchABus: @farfromrefug I noticed it's a leftover from an older PR and I cleaned it up considering it's an unneeded JNI call.
https://github.com/NativeScript/NativeScript/commit/c9e29aa9af8b4b82cd259f8402fa11af4640f1b4#diff-6e2aeb3f1b4d4fb9354519de09689dbb6a0a78cc719e1dff37e0d5a4b5d33b5f

I can re-add it if we want to keep it for any reason.

farfromrefug: Good was just wondering

dangrima90:
@CatchABus @farfromrefug @NathanWalker The change done here https://github.com/NativeScript/NativeScript/pull/10522/files#diff-ff996ba7476badc71c6c2ce6924be412385d3ea43a8785248a28159b3f7d203eR19 is breaking Maestro tests on Android.

For example I have a test where a `StackLayout` has a test id, however it is not being applied.

```html
<StackLayout :testID="item.testId" class="mt-10">
   <!-- .... -->
 </StackLayout>
```

I understand that there's a TODO comment to update this eventually, but in the meantime is there a way to workaround this? With this update Android is untestable.

dangrima90:
Update in relation to my above comment:

Adding `:accessible="true"` on the element seems to solve the issue:

```html
<StackLayout :testID="item.testId" :accessible="true" class="mt-10">
   <!-- .... -->
 </StackLayout>
```

Of course this would mean that we'd need to go through all of our code and see where this is needed.

CatchABus:
> Update in relation to my above comment:
> 
> Adding `:accessible="true"` on the element seems to solve the issue:
> 
> ```
> <StackLayout :testID="item.testId" :accessible="true" class="mt-10">
>    
>  </StackLayout>
> ```
> 
> Of course this would mean that we'd need to go through all of our code and see where this is needed.

This PR along with the referenced one are attempts to keep things stable with changes in #10482.
@farfromrefug I wonder whether we should find an appropriate fix or revert the whole update until we find a better way.

farfromrefug:
@CatchABus Here i will let N team decide what to do. In my fork i decided to disable accessibility by default on views so i dont have that commit https://github.com/NativeScript/NativeScript/pull/10522/files#diff-ff996ba7476badc71c6c2ce6924be412385d3ea43a8785248a28159b3f7d203eR19.
You can either revert or try to fix it. Maybe a solution would be to set accessible to true when `testID` is set? It kind of make sense to enforce

CatchABus:
> @NathanWalker @CatchABus, any ideas on a way forward here please?
> 
> As a summary with the above changes I see two main issues:
> 
> 1. Android app cannot be tested via Maestro unless each and every element that is to be accessed during testing has `accessible` set to `true`.
> 2. It's important to also keep in mind that if changes are made the focus shouldn't only be in relation to getting end-to-end tests working, but also keep accessibility in mind. Not sure if you are aware but in the EU soon it will be mandated for software to offer accessibility features and be transparent as to which areas of the application are accessible friendly. (Short article about the impact of the change: https://vaimo.com/blog/the-european-accessibility-act-2025-what-businesses-need-to-do/; Full EU Document: https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX%3A32019L0882)

These points are definitely important but not connected to the current PR as it's a fix addition on top of another 2 PRs and doesn't implement optional a11y focus itself.
Please create an issue so that we can keep track of it.

CatchABus:
> @CatchABus Here i will let N team decide what to do. In my fork i decided to disable accessibility by default on views so i dont have that commit https://github.com/NativeScript/NativeScript/pull/10522/files#diff-ff996ba7476badc71c6c2ce6924be412385d3ea43a8785248a28159b3f7d203eR19. You can either revert or try to fix it. Maybe a solution would be to set accessible to true when `testID` is set? It kind of make sense to enforce

@farfromrefug @dangrima90 @NathanWalker I believe we should proceed with reverting all recent a11y changes related to this until we come with a good solution that's tested thoroughly.

farfromrefug:
@CatchABus i say yes.

CatchABus:
@dangrima90 Core 8.8.6 should take care of it. Let us know.

dangrima90:
Hi @CatchABus I can confirm that I was able to successfully access elements via `testID` when using 8.8.6 on Android.

CatchABus:
> Hi @CatchABus I can confirm that I was able to successfully access elements via `testID` when using 8.8.6 on Android.

Thank you so much!

PR #10531, opened 2024-04-28
fix(core): clean up event-handling in Observable

shirakaba:
This PR cleans up the event-handling in Observable. It's mostly a code-level thing, but see the details below for some minor differences in behaviour (which arguably unbreak things that had been broken to begin with).

## PR Checklist

- It is possible to add duplicates of the same event (i.e. same `callback`, same `thisArg`), which is strange when compared to DOM Events.
- It is possible to add mutiple event listeners at a time (using a comma-separated list of event names) through the instance `addEventListener`, yet not the static one.
- There is duplicated code (`once` is mostly a copy-paste of `addEventListener`).
- There is missed opportunity for code reuse (the static and non-static implementations of `removeEventListener` do the same thing, yet differ).
- We trim whitespace on split-up event names in a loop rather than doing it up-front during the split operation.
- We continue to maintain these very strange static methods despite having discussed that they're probably not in use in practice.

## What is the new behavior?

All of the above points are addressed.

- We now guard against adding duplicate listeners.
- Static and non-static methods now reuse code as much as possible.
- Various code cleanups.
- Deprecation notices added to the static methods.

Potentially breaking changes:

- I add an optional `once?: boolean` parameter to the signature of `addEventListener`. In theory this shouldn't be a breaking change, but in case renderers are relying on passing an arg into that (e.g. as a workaround for supporting `AddEventListenerOptions`), I've added a `addEventListenerSchemaVersion` property to Observable so that they can check at runtime whether the given version of Core follows the expected call signature.
- There is a small chance that users may have been relying on duplicate event listeners. Users can retain the original behaviour by simply passing a new callback each time they call `addEventListener`, rather than reusing an existing one.


shirakaba:
I've not yet added tests, but happy to do so later (bedtime for now) – just opening it up for code review for the time being.

packages/core/data/observable/index.ts
```diff
@@ -191,23 +197,28 @@ export class Observable {
...
-			throw new TypeError('callback must be function.');
+			throw new TypeError('Callback, if provided, must be a function.');
 		}
 
-		const events = eventNames.split(',');
-		for (let i = 0, l = events.length; i < l; i++) {
-			const event = events[i].trim();
-			const list = this._getEventList(event, true);
-			// TODO: Performance optimization - if we do not have the thisArg specified, do not wrap the callback in additional object (ObserveEntry)
+		for (const eventName of eventNames.trim().split(eventNamesRegex)) {
```
CatchABus, line 212: After split, there might be event names with spaces so maybe we should keep old `trim()` call.

packages/core/data/observable/index.ts
```diff
@@ -227,137 +240,160 @@ export class Observable {
...
-
-		const events = _globalEventHandlers[eventClass][eventName];
-		if (thisArg) {
-			for (let i = 0; i < events.length; i++) {
-				if (events[i].callback === callback && events[i].thisArg === thisArg) {
-					events.splice(i, 1);
-					i--;
-				}
+		for (const eventName of eventNames.trim().split(eventNamesRegex)) {
+			const entries = _globalEventHandlers?.[eventClass]?.[eventName];
```
CatchABus, line 326: Same here

CatchABus (commented):
After a quick call, I left a comment about `trim()` calls. We might have to be careful about this in all event name iterations.

packages/core/data/observable/index.ts
shirakaba: I think `\s*,\s*` should preserve spaces within event names and just trim around the commas. Will have to get a REPL up later to sanity-check.

packages/core/data/observable/index.ts
```diff
@@ -100,13 +102,26 @@ export class Observable {
...
+	 * the call signatures of Observable.addEventListener() and
+	 * Observable.prototype.addEventListener() are. This may help us one day
+	 * migrate to an EventTarget-conformant signature (i.e. dropping `thisArg`).
+	 *
+	 * Full context of what each version number means would require referring to
+	 * the source. It's intended that any time a change is made to the call
+	 * signature, we increment the number. Once we've completed the migration, we
+	 * can remove it altogether.
+	 */
+	public static readonly addEventListenerSchemaVersion = 1;
```
shirakaba, line 116: I’m having second thoughts about this. Not sure renderer authors will necessarily find it useful compared to just pinning to a minimum version range of Core.

shirakaba:
Looks like a couple of unit tests are failing. Will check later whether they’re failing for the right reasons or not.

packages/core/data/observable/index.ts
farfromrefug: This is in a sense much slower than the previous code:
* for of is slower than for loop 
* regexp is much slower than split, and personally i feel using a regexp for such a simple case is too heavy

packages/core/data/observable/index.ts
```diff
@@ -227,137 +240,160 @@ export class Observable {
...
+			}
 		}
 	}
 
-	public static addEventListener(eventName: string, callback: (data: EventData) => void, thisArg?: any): void {
-		if (typeof eventName !== 'string') {
-			throw new TypeError('Event must be string.');
+	/**
+	 * Please avoid using the static event-handling APIs as they will be removed
+	 * in future.
```
farfromrefug, line 348: Global event handling will be removed?
Many plugins might break because of that.
At first i was not a fan of global events, but now with the sentry plugin it became a major part of it. Global events allow sentry to keep track of all navigations / modals / bottomsheet ... in the app and log them for easier debugging of issues

packages/core/data/observable/index.ts
```diff
@@ -466,20 +497,9 @@ export class Observable {
...
-				if (entry.callback === callback && entry.thisArg === thisArg) {
-					return i;
-				}
-			} else {
-				if (entry.callback === callback) {
-					return i;
-				}
-			}
-		}
+		thisArg = thisArg || undefined;
```
farfromrefug, line 500: what s the point of this line?

packages/core/data/observable/index.ts
shirakaba: # `for of` vs. `for`

@farfromrefug If you refer to the 10-100 iterations results, the performance is not significantly different. I think that number comfortably contains the general case of how many event listeners people will be using for any given event name. If users are registering 10,000 listeners, I am pretty confident the app will be choking for reasons other than this for loop!

I think readability of our code is important, and `for of` expresses more in fewer lines whilst avoiding out-of-bounds accesses and causing confusion by introducing an iterand (when we are sometimes splicing unrelated arrays using another iterand in these loops).

# Splitting with `RegExp` vs. multiple `split()`

The benchmark seemed unrepresentative, so I [prepared one](https://jsperf.app/hiroco/3) myself. Running on Chrome:

- `"press"`:
  -  old approach is 45.6M ops/sec
  - `RegExp` is 34.5M ops/sec
- `"press,close"`
  - old approach is 33.5M ops/sec
  - `RegExp` is 19.4M ops/sec.
- `" press , close "`
  - old approach is 13.9M ops/sec
  - `RegExp` is 15.9M ops/sec 😄

So it depends which case you want to optimise for. Admittedly, I don't think whitespace is used, so I agree we the old approach looks more attractive performance-wise.

Still, though, these are still running at tens of millions of operations per second, so it likely makes no difference to the user and I'd argue that this is more about readability.

But to be honest, I think this is all moot because multiple event names are probably only used by the gesture-handling code in ViewBase in the first place and we should probably just make _that_ pre-process all its event names so that we don't have to do *any* trimming/splitting in our event handlers. What do you think?

===

Ultimately happy to sacrifice readability in both cases for some small performance here if it's a blocker.

packages/core/data/observable/index.ts
shirakaba: > Global event handling will be removed?

We've been discussing this since 2022, yes – they were only introduced [in the first place](https://github.com/NativeScript/NativeScript/pull/8830) to support the plugins of an ex-community member.

- It follows no standard (I've never seen a framework implement class static event handlers)
- It adds a lot of code duplication
- It's undocumented
- It uniquely has "before" and "after" steps which mean we can't reuse code from the non-static event handler implementation
- It's not used in Core (as far as my VS Code can find - though I wouldn't be surprised if accessibility were using it for some reason)

... so I think it should ideally extracted out of Core as a mixin so that the consumers who still want it can make use of it.

As for Sentry, I'm sure there is another approach that could work (even if it involves writing something new in Core). [React Navigation](https://reactnavigation.org/docs/screen-tracking), for example, doesn't need global event handlers to achieve screen tracking. I think all we'd need to do is dispatch navigation-related events to the Application instance.

The Web-like approach would be to dispatch bubbling events from the navigator itself, but we've had poor reception for propagating events 😄

packages/core/data/observable/index.ts
shirakaba: `thisArg` is ultimately consumed on a truthy/falsy basis:
- if truthy, we bind the callback before calling it.
- if falsy, we don't bind it.

By normalising all falsy values to the same value, we avoid the case whereby someone could add the same listener multiple times with multiple different falsy `thisArg` values (`false`, `null`, `undefined`, `0`, `""`, etc.) and have it be unclear how each case would treated differently downstream.

Another issue is that the call signature is `thisArg?: any`, which suggests that `false` and `null` and `0` and `""` would be treated differently from `undefined`. Users might expect to be able to bind `0` to the this context of their callback, but be surprised when it remains unbound. By normalising, the signature now actually makes sense.

packages/core/data/observable/index.ts
farfromrefug: I agree i was even against in the first place. 
And yes there are alternatives. Just saying it will be a major breaking change in the sense that It could render some plugins ineffective

shirakaba:
I've pushed a commit that I believe will fix the failing CI tests (I'd forgotten to pass along the `once` option after refactoring).

I've spoken with @farfromrefug on Discord and it sounds like neither the `for of` usage nor the regex-based event name trimming are blockers for him, just feedback.

I might as well add some unit tests characterise the things I've mentioned in the description as being fixed by these changes. Welcoming any more input in the meantime.

shirakaba:
Okay, I fixed a unit test, but there's still an iOS E2E test failing (and I haven't looked at Android yet).

I figured it out. In the old behaviour, when you'd call `Observable.prototype.removeEventListener` or `Observable.prototype.off`, passing a falsy `callback`, it'd ignore whatever `thisArg` you'd pass and just remove all events matching your `eventName` arg.

By contrast, in `Observable.removeEventlistener` and `Observable.off`, `thisArg` would be used to narrow down the matches for callbacks.

The test concerns `bindingContextChange`, which relies on the old behaviour for instance methods. Since I brought the two in line, it caused the test to fail. I'll have to think about the best way to solve this.

packages/core/data/observable/index.ts
edusperoni: I'll be honest here and say that we should just put this behind a feature flag. I don't think anyone even uses the `"event1,event2"` format. So maybe we can do:
```
if (debug && has ",") {
  warn user that this is deprecated and to enable an env flag (--env.legacyEvents) to make it work
}
if (legacyEvents) {
  use legacy things
} else {
  let go of regex and trim
}
```

Alternatively, we can think about adding some quick memoization for the event name (meaning we will have a set of all event names that were passed in and if they need to be split or not). It's really better to just remove this bad practice though, as it's another one on the list of  "obscure NativeScript features that shouldn't be used" haha

packages/core/ui/core/bindable/index.ts
```diff
@@ -635,8 +635,9 @@ export class Binding {
 
 		try {
 			if (isEventOrGesture(options.property, <any>optionsInstance) && types.isFunction(value)) {
-				// calling off method with null as handler will remove all handlers for options.property event
-				optionsInstance.off(options.property, null, optionsInstance.bindingContext);
+				// Calling Observable.prototype.off() with just the event name will
+				// remove all handlers under that event name.
+				optionsInstance.off(options.property);
```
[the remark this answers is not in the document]
CatchABus: After taking a better look at Bindable, I think that's safe. We can deal with it if any problems arise.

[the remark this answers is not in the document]
shirakaba: There's an inconsistency between `static removeEventListener` and `removeEventListener`.

`static removeEventListener` narrows by `thisArg`, as you mentioned. By contrast, if `removeEventListener` receives a falsy callback, it actually widens the search to include all event listeners of the given event name:

https://github.com/NativeScript/NativeScript/blob/632a348e9ad8184468cf33a37aa17552d9092ca5/packages/core/data/observable/index.ts#L244-L247

So in Bindable, which uses `removeEventListener`, the old approach actually wasn't narrowing based on `optionsInstance.bindingContext` at all – it was seeing the `null` callback and just removing all event listeners under the given event name.

To make the two consistent, I've chosen to take the approach of `static removeEventListener` as the standard, as it feels more intuitive to me. As far as I can tell, this is the one line in Core that was relying on that quirk, and making this one-line change at the point of usage finally allows us to bring the two implementations in line.

shirakaba:
@CatchABus @farfromrefug @edusperoni is anything holding this back from approval? There's quite a lot more to clean up and I'd like to move onto the next PR!

farfromrefug:
Not on my side.

CatchABus (approved):

NathanWalker (approved):

PR #10534, opened 2024-05-01
fix(core): clean up event-handling in ViewCommon

shirakaba:
Relates to #10531. These two PRs are independent and can be merged in either order.

Cleans up the code related to event-handling in ViewCommon.

It makes it a lot clearer that we're actually splitting event names both here *and again* in Observable! So in a follow-up PR after this and the related PR, we can investigate removing event name splitting altogether.

## PR Checklist

There's some unclear, redundant code.

## What is the new behavior?

I've made the code clear and readable.

### DIFF
--- a/packages/core/ui/core/view/view-common.ts
+++ b/packages/core/ui/core/view/view-common.ts
@@ -33,8 +33,6 @@ import { SharedTransition, SharedTransitionInteractiveOptions } from '../../tran
 // helpers (these are okay re-exported here)
 export * from './view-helper';
 
-const eventNamesRegex = /\s*,\s*/;
-
 let animationModule: typeof am;
 function ensureAnimationModule() {
 	if (!animationModule) {
@@ -311,9 +309,7 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
 			return;
 		}
 
-		for (const eventName of normalizedName.trim().split(eventNamesRegex)) {
-			super.addEventListener(eventName, callback, thisArg);
-		}
+		super.addEventListener(normalizedName, callback, thisArg);
 	}
 
 	public removeEventListener(arg: string | GestureTypes, callback?: (data: EventData) => void, thisArg?: any) {
@@ -335,9 +331,7 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
 			return;
 		}
 
-		for (const eventName of normalizedName.trim().split(eventNamesRegex)) {
-			super.removeEventListener(eventName, callback, thisArg);
-		}
+		super.removeEventListener(normalizedName, callback, thisArg);
 	}
 
 	public onBackPressed(): boolean {
--- a/packages/core/ui/core/view/view-common.ts
+++ b/packages/core/ui/core/view/view-common.ts
@@ -33,6 +33,8 @@ import { SharedTransition, SharedTransitionInteractiveOptions } from '../../tran
 // helpers (these are okay re-exported here)
 export * from './view-helper';
 
+const eventNamesRegex = /\s*,\s*/;
+
 let animationModule: typeof am;
 function ensureAnimationModule() {
 	if (!animationModule) {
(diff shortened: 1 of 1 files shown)


packages/core/ui/core/view/view-common.ts
```diff
@@ -286,61 +288,55 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
...
-			this._observe(<GestureTypes>arg, callback as unknown as (data: GestureEventData) => void, thisArg);
+		// Normalize "ontap" -> "tap"
+		const normalizedName = getEventOrGestureName(arg);
+
+		// Coerce "tap" -> GestureTypes.tap
+		// Coerce "loaded" -> undefined
+		const gesture: GestureTypes | undefined = gestureFromString(normalizedName);
+
+		// If it's a gesture (and this Observable declares e.g. `static tapEvent`)
+		if (gesture && !this._isEvent(normalizedName)) {
```
shirakaba, line 309: A very curious case is when it's a gesture (e.g. "tap") and yet the ViewBase hasn't declared `static tapEvent`. We'll fall through to the next case, which just treats it as a regular event named "tap" rather than setting up a tap gesture observer. This has long/always been the case, so I decided not to change it.

I don't have the appetite to change it during this PR that's just focused on cleanup, but any thoughts about this anyway? Why is it necessary in the first place to declare `static tapEvent` on a ViewBase in order for it to be treated as a gesture event? Is that still an appropriate idea, or would it make sense to remove those across all classes by removing this condition?

packages/core/ui/core/view/view-common.ts
```diff
@@ -286,61 +288,55 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
...
+			this._observe(gesture, callback as unknown as (data: GestureEventData) => void, thisArg);
+			return;
+		}
+
+		for (const eventName of normalizedName.trim().split(eventNamesRegex)) {
```
shirakaba, line 314: Yes, this is redundant because Observable will do the same thing all over again. I've focused just on cleanup here, and in a follow-up PR it would be nice to investigate removing event-name splitting in Observable altogether (in which case we'd actually want to keep this line in place). Hence, I've decided not to remove it just yet.

packages/core/ui/core/view/view-common.ts
```diff
@@ -286,61 +288,55 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
...
+		// Coerce "loaded" -> undefined
+		const gesture: GestureTypes | undefined = gestureFromString(normalizedName);
+
+		// If it's a gesture (and this Observable declares e.g. `static tapEvent`)
+		if (gesture && !this._isEvent(normalizedName)) {
+			this._disconnectGestureObservers(gesture);
+			return;
+		}
+
+		for (const eventName of normalizedName.trim().split(eventNamesRegex)) {
```
shirakaba, line 338: See same comment about this redundant event-name splitting in `addEventListener`.

packages/core/ui/core/view/view-common.ts
```diff
@@ -286,61 +288,55 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
...
-				}
-			}
-		} else if (typeof arg === 'number') {
-			this._observe(<GestureTypes>arg, callback as unknown as (data: GestureEventData) => void, thisArg);
+		// Normalize "ontap" -> "tap"
+		const normalizedName = getEventOrGestureName(arg);
+
+		// Coerce "tap" -> GestureTypes.tap
+		// Coerce "loaded" -> undefined
+		const gesture: GestureTypes | undefined = gestureFromString(normalizedName);
```
shirakaba, line 306: Desperate for TypeScript strict mode :(

packages/core/ui/core/view/view-common.ts
CatchABus: Those static properties are for vanilla to be able to distinguish properties from events because XML format does not help in that matter.

We could get rid of those static properties across all classes possibly by doing a breaking change on vanilla and expect event declarations to have something like `on` prefix in XML.
```xml
<!-- 'on' helps distinguish events from properties -->
<Label text="Hello world" on:tap="dosomething"/>

shirakaba:
For a follow-up PR: would be worth checking that addObservers and removeObservers follow all the same rules as addEventListener and removeEventListener (i.e. checking whether thisArg matches).

farfromrefug (commented):
I think we should already deal with that redundant code for the regexp and split of ,. 
I would even remove support of multiple events directly

shirakaba:
@farfromrefug I've dealt with that redundant code now.

Removing support for multiple events altogether is something I'd prefer to do in a separate followup PR, as it would surely be slowed down by discussion of migration/deprecation path.

farfromrefug (approved):

NathanWalker (approved):

PR #10536, opened 2024-05-02
chore: nx migrate plus keep lock files

NathanWalker:
## PR Checklist


.gitignore
```diff
@@ -7,7 +7,6 @@
 
 # dependencies
 **/node_modules
-**/package-lock.json
```
shirakaba, line 10: Although I think npm is the slowest package manager at installing, it's indeed the one with the lowest barrier to entry for contributors (it's definitely installed), so probably the right one to standardise on 👍

apps/automated/package.json
```diff
@@ -11,9 +11,9 @@
     "nativescript-theme-core": "file:../../node_modules/nativescript-theme-core"
   },
   "devDependencies": {
-    "@nativescript/android": "rc",
-    "@nativescript/ios": "rc",
-    "@nativescript/visionos": "rc",
+    "@nativescript/android": "~8.7.0",
+    "@nativescript/ios": "~8.7.0",
+    "@nativescript/visionos": "~8.7.0",
```
shirakaba, line 16: Is the intention here to pull `~8.7.0` from npm, or to whatever version exists in the codebase?

If it's linked to whatever version exists in the codebase, maybe another version range would be better. Should it be a `file:` reference, or a `*`, or something? In pnpm, we could write `workspace:@nativescript/core`. With [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces), we'd still specify a version range, though. In the NativeScript Node-API monorepo based on npm workspaces, we're [using](https://github.com/DjDeveloperr/runtime/blob/main/packages/ios/package.json#L33C5-L33C53) `"@nativescript/objc-node-api": ">=1.0.0-alpha.5"`, so we don't even need to think about semver.

I'd probably lean towards `file:` to make it totally explicit, or `*` otherwise.

package.json
```diff
@@ -73,16 +70,6 @@
...
-  "overrides": {
-    "@nx/devkit": "18.2.2",
-    "@nx/eslint-plugin": "18.2.2",
-    "@nx/jest": "18.2.2",
-    "@nx/js": "18.2.2",
-    "@nx/node": "18.2.2",
-    "@nx/plugin": "18.2.2",
-    "@nx/workspace": "18.2.2",
-    "nx": "18.2.2"
-  },
```
shirakaba, line 85: Nice that we're able to drop these!

packages/webpack5/package-lock.json
shirakaba, line 1: In nx monorepos, does each package get its own package lock? Just that in npm workspaces and pnpm monorepos, they don't, so checking that it's right for this to be here at all.

apps/automated/package-lock.json
shirakaba, line 1: Looks correct. It's using the specified  local packages (e.g. `@nativescript/core`), it's the [latest lockfileVersion](https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json), and pulls from `registry.npmjs.org`.

shirakaba (approved):
Looks like a great improvement, thanks for taking the time! @NathanWalker

I'm approving, assuming the answer to this is "yes, each package does get a lockfile":

> In nx monorepos, does each package get its own package lock? Just that in npm workspaces and pnpm monorepos, they don't, so checking that it's right for this to be here at all.

If each package is **actually not supposed to have its own lockfile**, then let's get to the bottom of that.

shirakaba:
Actually, that's a point. Either all packages should have their own lockfile, or none of them. I notice that the lockfiles for `devtools` and some other packages are missing, so something's amiss.

It seems I can't withdraw my PR approval for now, so could you please just check on that (or address in a follow-up PR)!

apps/automated/package-lock.json
NathanWalker: We don't need them in apps, I"ll add ignore for this - doesn't hurt either but agreed not needed in `apps` - these are essentially demo apps to support testing/experimentation/verifying of the packages.

NathanWalker:
- apps/* don't need lock files but they don't hurt either.
- packages/* don't need lock files, exception of webpack5 in this workspace for now (can change in future).

shirakaba (approved):
Having discussed on Discord that the package-lock for Webpack5 is present because it's not a properly integrated member of the monorepo (i.e. not a true workspace), I'm happy to approve. Will be happy for Webpack5 to be properly integrated in a follow-up PR in future.

PR #10537, opened 2024-05-03
fix(core): stop accepting GestureTypes enum as an eventName

shirakaba:
This PR is related to https://github.com/NativeScript/NativeScript/pull/10538, but can be merged independently. Whichever is merged first will cause minor conflicts with the other, but it's a simple rebase that I'd be happy to take care of.

## PR Checklist

ViewCommon is a subclass of Observable. When overriding `addEventListener` and `removeEventListener`, it makes the following changes to the signature:

```diff
- addEventListener(eventNames: string, callback: (data: EventData) => void, thisArg?: any, once?: boolean): void;
+ addEventListener(eventNames: string | GestureTypes, callback: (data: EventData) => void, thisArg?: any, once?: boolean): void;

- removeEventListener(eventNames: string, callback?: (data: EventData) => void, thisArg?: any): void;
+ removeEventListener(eventNames: string | GestureTypes, callback: (data: EventData) => void, thisArg?: any);
```

Although this is allowed in TypeScript, even in strict mode, here are my motivations for removing it:

- there's no real need for it. We can just coerce to string upstream of it. Its one advantage is that it supports expressing multiple events at once, but (aside from there being an appetite to remove support for adding/moving multiple events at once) it seems we're not even using that functionality in practice.
- we're not extensively using the feature anywhere within Core in the first place (only four lines across all of Core).
- it complicates the typings and makes them diverge further from EventTarget.

## What is the new behavior?

Accept only strings for event names.

This is not a breaking change within Core (no tests were relying on it, and I've migrated the only four lines relying on it), and any plugins making use of it should see what to do by heeding the updated typings.


apps/automated/src/ui/gestures/gestures-tests.ts
shirakaba, line 1: As the names suggest, these tests are not actually automated tests; they're purely auto-documented code-snippets for use on the old docs site. I thought I'd just remove the now-unsupported ones to save some mental burden when using the search functionality.

packages/core/ui/button/index.android.ts
```diff
@@ -126,9 +126,9 @@ export class Button extends ButtonBase {
 							break;
 					}
 				});
-			this.on(GestureTypes.touch, this._highlightedHandler);
+			this.on('touch', this._highlightedHandler);
```
farfromrefug, line 129: Why do we need to change this? it does not work anymore with `GestureTypes.touch` ? 
It seems preferable to use const we define elsewhere for event names (like we are doing in most places)

shirakaba: > Why do we need to change this? it does not work anymore with `GestureTypes.touch`?

Yes, `Button.on()` calls `Observable.on()` which forwards to `ViewCommon.addEventListener()`, which (as of this PR) accepts only a string (that's the point!).

> It seems preferable to use const we define elsewhere for event names (like we are doing in most places)

Funnily enough, we don't define constants for these 😅 Constants are available for `GestureEvents`, but apparently not for `GestureArgs`. I could add such a set of consts for the first time.

There is of course the `toString()` method in `packages/core/ui/gestures/gestures-common.ts`. However, that method returns a `string` rather than a const, though (because of its odd ability to join multiple event names together), so it's actually less statically analysable than just writing `'touch'` 🫠

Either way, happy to introduce a `GestureArgs` name map for that.

farfromrefug: @shirakaba ok i see now i misplaced `GestureTypes` for an enum of strings when it is an enum of numbers!
Would be nice indeed to define `tapEvent` and others somewhere globally so we can use them everywhere and stop using `'tap'`.
I ll resolve this as it is good as it is

packages/core/ui/gestures/gestures-common.ts
```diff
@@ -61,6 +61,18 @@ export enum GestureTypes {
...
+	tap: 'tap',
+	doubleTap: 'doubleTap',
+	pinch: 'pinch',
+	pan: 'pan',
+	swipe: 'swipe',
+	rotation: 'rotation',
+	longPress: 'longPress',
+	touch: 'touch',
+} as const;
+
```
shirakaba, line 75: @farfromrefug As requested!

packages/core/ui/gestures/gestures-common.ts
```diff
@@ -327,28 +339,8 @@ export function toString(type: GestureTypes, separator?: string): string {
...
-	} else if (t === 'longpress') {
-		return GestureTypes.longPress;
-	} else if (t === 'touch') {
-		return GestureTypes.touch;
-	}
-
-	return undefined;
+export function fromString(type: string): GestureTypes | undefined {
+	return GestureTypeNames[type.trim()];
 }
```
shirakaba, line 344: I've removed the `.toLowerCase()` as:

1. it has mostly been used via the `toString()` convenience function up until now.
2. Now we have a blessed GestureTypeNames map so users should be using that instead.

To be honest, we can likely remove the `.trim()` too, but I'd prefer to do that in the PR where we remove support for multiple event names.

packages/core/ui/gestures/gestures-common.ts
shirakaba: Forgot how enums work. Dropping this object in favour of just using the existing enum as-is.

packages/core/ui/button/index.android.ts
```diff
@@ -126,9 +126,9 @@ export class Button extends ButtonBase {
 							break;
 					}
 				});
-			this.on(GestureTypes.touch, this._highlightedHandler);
+			this.on(GestureTypes[GestureTypes.touch], this._highlightedHandler);
 		} else {
-			this.off(GestureTypes.touch, this._highlightedHandler);
+			this.off(GestureTypes[GestureTypes.touch], this._highlightedHandler);
```
shirakaba, line 131: @farfromrefug As requested, now using the string constants instead (forgot how enums worked).

CatchABus: The thing is we have guided people to use the types for years as explained in old docs: https://old.docs.nativescript.org/ui/gestures

How are we going to make up for that?

shirakaba: Good catch 🤔

Optimistically, I’d imagine they’d see a compiler error, see that it now accepts only a string, and be able to take a good guess at what the string is (as it comes from the same enum).

Arguably, it’s now left unmentioned in the [new docs](https://docs.nativescript.org/guide/gestures), so the typings serve as the up-to-date documentation.

I’d rather not handle it at runtime and just pull the plaster, if possible. It’s unlikely to be the only breaking change regarding event-handling that’s to come, so once the dust has settled, we could provide a migration guide based on what makes it into the release.

shirakaba:
@farfromrefug @edusperoni @CatchABus Anything holding this back from approval?

farfromrefug:
Good with me

NathanWalker (approved):

PR #10539, opened 2024-05-05
fix(core): drop support for plural event/gesture names

shirakaba:
This PR is based onto the following PRs that are in review:

- [x] #10537 (first PR)
- [x] #10538 (second PR)

I'd suggest reviewing this by just selecting the commits from this PR, excluding the commits from those prior two PRs. As per the screenshot, you'd scroll to the bottom of the commits (most recent at the bottom), and select the same commits as I've indicated in blue. In other words, the first included commit should be `chore(core): drop support for plural events`.

image

## PR Checklist

Currently, it is possible to listen for plural events by passing a comma-delimited series of event names (which may optionally be surrounded by whitespace):

```js
observable.addEventListener("propertyChange,loaded");
// Adds listeners for the "propertyChange" and "loaded" events

observable.addEventListener("propertyChange  ,  loaded");
// Adds listeners for the "propertyChange" and "loaded" events
```

It is also possible to observe plural gestures, but due to bugs in the implementation, it is only possible via a plural enum and not a plural string.

```js
observable.addEventListener("tap,doubleTap");
// Fails.
//
// The fromString() method in `gestures-common.ts` doesn't split on commas, so
// the gesture name here coerces to `undefined`, thus
// Observable.prototype.addEventListener() throws a TypeError saying
// "Event name(s) must be a string.".

observable.addEventListener(GestureTypes.tap & GestureTypes.doubleTap);
// Succeeds, though will cease to be allowed if #10537 (which disallows numeric
// event names) is merged.
//
// At the time of writing, adds a single observer that manages two GestureHandlers:
// - one for the "tap" gesture
// - one for the "doubleTap" gesture
```

## What is the new behavior?

We now interpret event names as-is, corresponding to a singular event listener or observer. So we no longer trim event names, nor split on commas. As this PR is based on #10537, we only accept string names, so don't support enums at all (so avoid having to special-case plural enums).

```js
observable.addEventListener("propertyChange,loaded");
// Adds a (probably not very useful) listener for the non-existent
// "propertyChange,loaded" event

observable.addEventListener("  propertyChange  ,  loaded  ");
// Adds a (probably not very useful) listener for the non-existent
// "  propertyChange  ,  loaded  " event

observable.addEventListener("tap,doubleTap");
// Adds a (probably not very useful) listener for the non-existent
// "propertyChange,loaded" event.
//
// Since it doesn't exactly match any gesture name, it is indeed
// an event listener and not a gesture observer.

observable.addEventListener("  propertyChange  ,  loaded  ");
// Adds a (probably not very useful) listener for the non-existent
// "  propertyChange  ,  loaded  " event.
//
// Since it doesn't exactly match any gesture name, it is indeed
// an event listener and not a gesture observer.
```

### Benefits

- Improves performance of event-handling by avoiding unnecessary trimming and splitting of event names.
- Avoids users hitting upon the broken functionality for observing plural gestures when passing a string rather than an enum.
- May make life easier for renderers (reduces divergence from DOM Events)
- Greatly simplifies the codebase, as now we only operate on singular GestureTypes rather than combinatorial ones.

## Breaking changes

This PR drops support for listening for:
- whitespace-padded event names, e.g. `"  propertyChange  "`
- plural events, e.g. `"propertyChange,loaded"`
- plural gestures, e.g. `GestureTypes.tap & GestureTypes.doubleTap` (though support for this form should be removed in #10537, and the string form "tap,doubleTap" was never working to begin with).

### Impact

From TSC discussions, it doesn't seem that this feature is commonly, if ever, used, so impact may be low in practice. Certainly, it wasn't used anywhere within Core except in some test cases (which I've updated accordingly).

### Migration steps

Any plugins or userland code using plural event/gesture names should instead split them into singular names and listen for them individually.

### DIFF
--- a/apps/automated/src/data/observable-tests.ts
+++ b/apps/automated/src/data/observable-tests.ts
@@ -163,7 +163,7 @@ export var test_Observable_addEventListener_MultipleEvents = function () {
 	obj.addEventListener(events, callback);
 	obj.set('testName', 1);
 	obj.test();
-	TKUnit.assert(receivedCount === 2, 'Callbacks not raised properly.');
+	TKUnit.assert(receivedCount === 0, "Expected no event handlers to fire upon the 'propertyChange' event when listening for event name 'propertyChange,tested', as we have dropped support for listening to plural event names.");
 };
 
 export var test_Observable_addEventListener_MultipleEvents_ShouldTrim = function () {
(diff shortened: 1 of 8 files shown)


farfromrefug (changes requested):
@shirakaba shouldn't the regexp for names be removed now ?
The rest looks good

farfromrefug (approved):

shirakaba:
@farfromrefug Thanks for the approval! I pushed a follow-up commit as I can't be sure about `isGesture()`, so will avoid introducing a very hard-to-track-down breakage for now. If we're feeling brave, we can remove the `.trim().toLowerCase()` for extra performance, I just don't know whether it'll break some renderer. Do you think it's worth the risk?

NathanWalker (approved):

PR #10567, opened 2024-07-02
fix(android): reject promise if cancel animation

IlIAntonIlI:
## PR Checklist

When an animation is cancelled, the animation promise ends as if the animation had been played to completion. Thus, it is not possible to determine whether it was cancelled or completed.

## What is the new behavior?
When an animation is cancelled, the animation promise ends as rejected. It helps to separate cancelled animation with animation that had been played to completion.


rigor789:
I believe we had a few discussions about this in the past, and the conclusion was that cancels are not errors - however I do think we should have a way to tell whether or not an animation was cancelled or not, in the `resolve` callback itself.

PR #10577, opened 2024-07-09
feat(winter-cg): crypto, atob, btoa

triniwiz:


farfromrefug:
@triniwiz Why making this part of Nativescript core while you added url into the runtime?
I personally, much prefer for it to be here but i quite dont understand why you switch. I would say everything should be in the same place.

Now it beeing here, wouldnt it be nicer to split that into a separated lib that widgets? while still importing it by default?
It would be cleaner, prevent widgets (and it is not widgets here) from becoming a bag of everything.
It would also easily allow splitting it into a separated "plugin" if need be (yes my fork would do that right away :P)

NathanWalker:
@farfromrefug With the new @nativescript/foundation package being planned out (to eventually supersede core in v10 likely), there will be separate modules.

NathanWalker:
We're splitting up a new @nativescript/winter-cg package on this PR @farfromrefug. It will become a transient to core initially but will become optional with the forthcoming @nativescript/foundation (see Discord).

farfromrefug:
@NathanWalker awesome did not see that on discord. Will look

PR #10581, opened 2024-07-11
fix: re-add lowercase gesture handling and add deprecation notice when using non-string events

edusperoni:
makes gestures case-insensitive again for now


NathanWalker (approved):

SeanKelly369 (commented):
This is needed. Otherwise existing code using GestureTypes.tap, .doubleTap etc will need to be updated.

nick0704350 (approved):

PR #10591, opened 2024-07-25
chore: improve task flow for native core deps

edusperoni:
Creating this as a draft.

Currently we need to manually build ui-mobile-base and winter-cg internal packages before building core since core itself need the artifacts from these packages.

This PR changes it so we just need to run `copy-native-deps` and we can leverage nx caching. We could make `copy-native-deps` a dependency of `build` but it'd require `build` to be ran only from a mac machine, even if not changing any iOS source (currently this works by storing the artifcats themselves in git).

Thoughts on this?

### DIFF
--- a/packages/core/project.json
+++ b/packages/core/project.json
@@ -24,6 +24,17 @@
 				"verbose": true
 			}
 		},
+		"copy-native-deps": {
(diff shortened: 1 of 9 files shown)


NathanWalker:
This looks like great reduction in scripting, I can help bring it up to date with main so we could get this in.

PR #10592, opened 2024-07-25
fix(core): include everything but core.aar in android platforms

edusperoni:
The current configuration excludes winter_cg-release.aar since it's not included in the files. This change now force includes everything *but* the `core.aar` which isn't needed as it's auto-generated by the CLI every time (we should probably remove it and .gitignore it)


triniwiz (approved):
LFG

PR #10602, opened 2024-08-07
perf(core): Avoid setting the same value to view properties

CatchABus:
## PR Checklist

It seems we have 2 portions of dead code when setting view property values.

Sample:
```js
delete oldProperties[property];
if (property in oldProperties && oldProperties[property] === value) {
	// Skip unchanged values
	continue;
}
```

This ends up with a dead if block and core will attempt to re-apply the same values to view properties.
Even though this is getting caught in Observable, we do one more additional check even though it's always going to be `false`.

## What is the new behavior?
This PR corrects property set behaviour and deletes property from list once its existence has been confirmed, enabling core to proceed to value check.
This is trivial, yet it keeps code cleaner and functional.

### DIFF
--- a/packages/core/ui/styling/style-scope.ts
+++ b/packages/core/ui/styling/style-scope.ts
@@ -721,25 +721,40 @@ export class CssState {
 				cssExpsProperties[property] = value;
 				continue;
 			}
-			delete oldProperties[property];
-			if (property in oldProperties && oldProperties[property] === value) {
-				// Skip unchanged values
-				continue;
+
+			if (property in oldProperties) {
+				const oldValue = oldProperties[property];
+
+				delete oldProperties[property];
+
+				if (oldValue === value) {
+					// Skip unchanged values
+					continue;
+				}
 			}
+
 			if (isCssVariable(property)) {
 				view.style.setScopedCssVariable(property, value);
 				delete newPropertyValues[property];
 				continue;
 			}
+
 			valuesToApply[property] = value;
 		}
-		//we need to parse CSS vars first before evaluating css expressions
+
+		// we need to parse CSS vars first before evaluating css expressions
 		for (const property in cssExpsProperties) {
-			delete oldProperties[property];
 			const value = evaluateCssExpressions(view, property, cssExpsProperties[property]);
-			if (property in oldProperties && oldProperties[property] === value) {
-				// Skip unchanged values
-				continue;
+
+			if (property in oldProperties) {
+				const oldValue = oldProperties[property];
+
+				delete oldProperties[property];
+
+				if (oldValue === value) {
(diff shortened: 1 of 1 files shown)


NathanWalker (approved):

rigor789:
This change seems to have introduced an issue. The old code was working by accident, as it was never actually skipping anything because the condition would never be `true`:

```
delete oldProperties[property];
if (property in oldProperties && oldProperties[property] === value) {
```

Essentially because the old property is deleted before the condition, so it's always going to be checking for `undefined === value`...

But now that it actually skips, there's a problem, old and new value are somehow always the same, for example adding some logging, I see:
```
  skipping apply background-color
   - old rgba(241, 241, 255, 0.08)
   - new rgba(241, 241, 255, 0.08)
...
  skipping apply background
   - old linear-gradient(to right, #b21834, #860808)
   - new linear-gradient(to right, #b21834, #860808)
```

But the general issue/side-effect of this is that for example

```html
<Label [ngClass]="{
  'bg-white': !showGradient(),
  'bg-gradient': showGradient()
}"></Label>
```
this should toggle between a white `background-color` and a gradient `background`, since these are separate properties, but it actually goes like this:
```
1. showGradient() === false -> bg-white - OK
2. showGradient() === true -> bg-gradient - OK
3. showGradient() === false -> no background, nor background-color - NOT OK
```

**With Skip**

https://github.com/user-attachments/assets/7bc040f6-fb46-40ee-bff3-55cc8b2602ee

**Without Skip**

https://github.com/user-attachments/assets/3f57f033-fcda-41a9-88ae-6b80df38dc0d

CatchABus:
@rigor789 Is it possible to share that sample? We can definitely revert it but before that, I'd like to see if we can find a solution because old code looked a bit weird.

rigor789:
@CatchABus sure, here's an example of the same setup: https://github.com/rigor789/style-bug-10602/blob/main/src/app/app.component.html

1. clone
2. ns debug ios --no-hmr
3. button is blue ✅ 
4. press button: button has gradient: ✅ 
5. press button again: button loses it's background: ❌ 
6. press button again: button has gradient: ✅

CatchABus:
> @CatchABus sure, here's an example of the same setup: https://github.com/rigor789/style-bug-10602/blob/main/src/app/app.component.html
> 
> 1. clone
> 2. ns debug ios --no-hmr
> 3. button is blue ✅
> 4. press button: button has gradient: ✅
> 5. press button again: button loses it's background: ❌
> 6. press button again: button has gradient: ✅

@rigor789 Thanks, that helped a lot! I believe we should revert my PR for now due to how shorthand properties behave in core.
I have a good idea on how to improve those properties and re-apply these changes at a later time. 😄

PR #10629, opened 2024-09-26
feat(types-ios): latest 18.x types

NathanWalker:
Latest iOS 18.x typings from Xcode 16

We may want to omit some of these new foundational barrels so this needs to be tested with production apps prior to merge.


CatchABus:
@NathanWalker I noticed you bumped typescript eslint dependencies to the new major version.
The problem with this is they use Eslint 9 internally and `eslintrc` configuration files are no longer detectable or usable in that version.

See:
https://eslint.org/docs/latest/use/migrate-to-9.0.0#flat-config
https://eslint.org/docs/latest/use/configure/configuration-files

PR #10648, opened 2024-11-08
feat(core): Typography improvements

CatchABus:
## PR Checklist

Right now, typography is a bit inconsistent between the two platforms and MDN specs.
More specifically, line-height does not act as a height and its visual result is quite different than that of browser.

## What is the new behavior?
- Small android size defaults cleanup
- Added support for line-height dip and percentage units and updated calculation to be closer to that of browser
- Corrected UIButton letter-spacing calculation which used the deprecated font property to read font size
- Corrected iOS Button formatted text default color (it's different than the button with regular text)

BREAKING CHANGES:
- Line height will apply like a height as the name indicates while {N} has long been using it as line-spacing. This might cause strange visual changes in apps.

### DIFF
--- a/packages/core/platforms/ios/src/UIView+NativeScript.m
+++ b/packages/core/platforms/ios/src/UIView+NativeScript.m
@@ -30,16 +30,8 @@
         // make sure a possible previously set text alignment setting is not lost when line height is specified
         if ([self isKindOfClass:[UIButton class]]) {
             paragraphStyle.alignment = ((UIButton*)self).titleLabel.textAlignment;
-
-            if (((UIButton*)self).titleLabel.font) {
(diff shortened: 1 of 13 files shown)


CatchABus:
After a bit of consideration, I removed all changes regarding `line-height`.
The reason is that the css property can also accept a multiplier value which is one of the recommended usages in the web.
NativeScript css dip values are unitless, so if we suddenly treat the `line-height` value as dip then we have no way of setting a multiplier.
In few words, it's best to leave it as is.

CatchABus:
All changes regarding letter-spacing unit have been removed from this PR.
We might revisit this one in the future provided that we don't remove the `em` support since it's very useful when dealing with responsive layouts.

PR #10680, opened 2025-01-28
fix(core): Pseudo-class handlers failing to unsubscribe listeners

CatchABus:
## PR Checklist

Right now, even if a view is unloaded, pseudo-class listeners will not be disposed.
At the same time, once view is loaded again, pseudo-class handlers will add new listeners resulting in a leak.
There's also another case of iOS button which can keep reference of their previous native view if removed and re-added to the view tree. That is also a defect caused by missing pseudo-class handlers cleanup.

## What is the new behavior?
This PR ensures that pseudo-class handlers will follow the subscription cycle of adding and removing listeners.


farfromrefug:
@CatchABus awesome fix!

packages/core/ui/core/control-state-change/index.ios.ts
```diff
@@ -2,18 +2,17 @@
 
 @NativeClass
 class ObserverClass extends NSObject {
-	public callback: WeakRef<ControlStateChangeListenerCallback>;
+	public callback: ControlStateChangeListenerCallback;
```
NathanWalker, line 5: Is this better as not a WeakRef? Just curious if was totally necessary to not use a WeakRef here.

CatchABus: @NathanWalker I moved it elsewhere and focused on turning the {N} view used inside the callback into a weak ref.
https://github.com/NativeScript/NativeScript/pull/10680/files#diff-e51321824b20c30cc9f8cae6badb391594307c53572860272e8177bcd76d1b6dR50

If this one doesn't make much sense, I can revert it.

NathanWalker: Ah that makes total sense thanks for explaining.

NathanWalker (approved):

PR #10690, opened 2025-02-05
fix(core): ignore inserting child if < 0

NathanWalker:
## PR Checklist

A throw could occur during layout introduced from this change which would bubble out to a v8 engine crash:
https://github.com/NativeScript/NativeScript/pull/10679/files#diff-ff996ba7476badc71c6c2ce6924be412385d3ea43a8785248a28159b3f7d203eR54

## What is the new behavior?

Just ignore inserting children if < 0. 

Notes: Usage of CollectionView surfaced this issue.

### DIFF
--- a/packages/core/ui/layouts/layout-base-common.ts
+++ b/packages/core/ui/layouts/layout-base-common.ts
@@ -49,12 +49,14 @@ export class LayoutBaseCommon extends CustomLayoutView implements LayoutBaseDefi
 		this._registerLayoutChild(child);
 	}
 
-	public insertChild(child: View, atIndex: number): void {
+	public insertChild(child: View, atIndex: number): boolean {
 		if (atIndex > -1) {
 			this._subViews.splice(atIndex, 0, child);
 			this._addView(child, atIndex);
 			this._registerLayoutChild(child);
+			return true;
 		}
+		return false;
 	}
 
 	public removeChild(child: View): void {
--- a/packages/core/ui/layouts/layout-base-common.ts
+++ b/packages/core/ui/layouts/layout-base-common.ts
@@ -50,13 +50,11 @@ export class LayoutBaseCommon extends CustomLayoutView implements LayoutBaseDefi
 	}
 
 	public insertChild(child: View, atIndex: number): void {
-		if (atIndex < 0) {
-			throw new Error('Cannot insert a child to a negative index.');
+		if (atIndex > -1) {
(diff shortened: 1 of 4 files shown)


packages/core/ui/layouts/layout-base-common.ts
```diff
@@ -50,13 +50,11 @@ export class LayoutBaseCommon extends CustomLayoutView implements LayoutBaseDefi
 	}
 
 	public insertChild(child: View, atIndex: number): void {
-		if (atIndex < 0) {
-			throw new Error('Cannot insert a child to a negative index.');
+		if (atIndex > -1) {
```
CatchABus, line 53: While this prevents the flow from proceeding, it does not prevent subclass overrides from executing their own logic.

These are the cases that need checking:
https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/layouts/layout-base.ios.ts#L14
https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/layouts/root-layout/index.android.ts#L11

We can either add the `atIndex > -1` check there too or change the return type of the method (let's say `boolean` just like `_addViewToNativeVisualTree`) and be able to perform checks the following instead:

```ts
if (super.insertChild(child, index)) {
    // Proceed
}
```

CatchABus (changes requested):

packages/core/ui/layouts/layout-base-common.ts
NathanWalker: Good idea, pushed up that adjustment. When tests pass here we can merge. 
The only possibility of breaking change this could introduce is likely for a plugin with strict TypeScript enabled which may fail a TS build until an override of insertChild matched signature (instead of void).

CatchABus: Just few things that need tweaking
1) https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/layouts/layout-base.d.ts#L37
2) https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/layouts/root-layout/index.android.ts
3) https://github.com/NativeScript/NativeScript/blob/main/tools/scripts/api-reports/NativeScript.api.md?plain=1#L1347

CatchABus: Seems great, thanks so much!!

CatchABus (approved):

Issue #9046, opened 2020-11-17
iOS 14 navbar large title issue


georgemark0v:
```json
"devDependencies": {
    "@babel/core": "^7.11.6",
    "@babel/preset-env": "^7.11.5",
    "@nativescript/ios": "7.0.5",
    "@nativescript/webpack": "^3.0.4",
    "babel-loader": "^8.1.0",
    "nativescript-vue-template-compiler": "^2.8.1",
    "nativescript-worker-loader": "^0.12.1",
    "sass": "^1.26.10",
    "vue-loader": "^15.9.3"
  }
```

On iOS 14 devices no longer can't use large navbar titles and collapsed searchcontroller. When scrolled up large title doesn't show and listview scrolls back. After some retries large title finally shown. 

```html
<Page @loaded="pageDidLoad" class="page-clean-color">
  <ActionBar title="Контакты"></ActionBar>
  <ListView :items="contacts.listOf" @loaded="listViewDidLoad" height="100%" @itemLoading="itemLoading">
    <v-template></v-template>
  </ListView>
</Page>
```

Video example: https://yadi.sk/i/zbNO5nOwvjhyzQ

rigor789:
Please share your `pageDidLoad` and `listViewDidLoad` methods, maybe related to the issue.

georgemark0v:
> Please share your `pageDidLoad` and `listViewDidLoad` methods, maybe related to the issue.

im sure problem not it

`listViewDidLoad( args ) {
                args.object.ios.separatorInset = UIEdgeInsetsMake( 0, 68, 0, 0 );
                args.object.ios.keyboardDismissMode = UIScrollViewKeyboardDismissModeInteractive;   
            },`

`pageDidLoad(args) {                
                const page = args.object;
                if( this.people.items.getNoOfSections() > 0) {                          
                    this.contacts.prepare();
                    return;
                }
                               
                this.contacts.requests.get().then(() => {});

                this.contacts.listOf.addSection("", [
                    { type: 'location'  },
                    { type: 'invite' },
                    { type: 'requests' }
                ]);

                if( args.object.ios ) {
                    dispatch_async(dispatch_get_current_queue(), () => {
                        this.getContacts( args );
                    });
                }
}`

georgemark0v:
Made some experiments and when listview have content size smaller then screen size its works as expected. https://yadi.sk/i/1FWwsLrHZW_Erg

jerrykeys:
[...]

Issue #5904, opened 2018-06-01
iOS: none translucent navigationBar gets translucent again on resume from background


Cookie8:
### Tell us about the problem
Hi,
For my iOS app I would like to have the navigationBar translucence off. 
It works fine, if the page will initially be loaded, but then, if I bring the app to background and bring it back to foreground again (resume), the navigationBar will be displayed 
translucent again.

### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 4.1.0 (but also occurred with CLI 4.0.2)
- Cross-platform modules: 4.2.0-2018-05-31-01
- Runtime(s): 
tns-android: 4.2.0-2018-05-31-01
tns-ios: 4.1.0
- Plugin(s): "dependencies": {
		"nativescript-iqkeyboardmanager": "^1.3.0",
		"nativescript-theme-core": "~1.0.4",
		"tns-core-modules": "^4.2.0-2018-05-31-01"
	},
	"devDependencies": {
		"babel-traverse": "6.26.0",
		"babel-types": "6.26.0",
		"babylon": "6.18.0",
		"lazy": "1.0.11"
	}

### Please tell us how to recreate the issue in as much detail as possible.
Please use the attached playground project below to reproduce this issue by following these steps:
1. Run the app -> The translucent is off and the NavigationBar looks as expected.
2. Bring the app in background by device home button (it must completely be in background)
3. Bring the app in foreground again -> The NavigationBar background looks to be translucent again, but as I can read in the logs the value of 'page.frame.ios.controller.navigationBar.translucent' will still be 'false'.

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.
Please see my Playground sample app:
[NSPlayground.zip](https://github.com/NativeScript/NativeScript/files/2064068/NSPlayground.zip)

The login page of the Playground sample project contains exactly what I implemented in my project and the issue also occurs in the Playground sample project.

Please just let me know, if you need further infos.
Cheers!

NickIliev:
@Cookie8 there is a property called `flat` which you can use to turn of the translucency property.

For example,
```XML
<ActionBar title="Test" class="action-bar" flat="true">
</ActionBar>
```

And then there is no need to use the native property in the code behind
```JavaScript
const createViewModel = require("./login-view-model").createViewModel;

function loaded(args) {
    console.log("LOGIN-PAGE <-> loaded > start");
    const page = args.object;

[...]

PR #10694, opened 2025-02-10
feat(action-bar): iosLargeTitle and iosShadow attributes

NathanWalker:
Allows setting prefersLargeTitles as well as hiding the shadow/bottom shadow border on the ActionBar. Changes no current defaults, just extra options to customize.

## PR Checklist

Currently you could use loaded event to grab ActionBar reference to set it's prefersLargeTitles property on the UINavigationBar as well as modify it's appearance.

## What is the new behavior?

This introduces 2 new ActionBar attributes:
- `iosLargeTitle`: boolean (defaults to what it is currently which is false).
-  `iosShadow`: boolean (defaults to what it is currently which is true).

Introduces no breaking changes - keeps existing behavior while providing easy declarative options to customize.

closes https://github.com/NativeScript/NativeScript/issues/9046
closes https://github.com/NativeScript/NativeScript/issues/5904

### DIFF
--- a/packages/core/ui/action-bar/index.ios.ts
+++ b/packages/core/ui/action-bar/index.ios.ts
@@ -1,5 +1,5 @@
 import { IOSActionItemSettings, ActionItem as ActionItemDefinition } from '.';
-import { ActionItemBase, ActionBarBase, isVisible, flatProperty, iosIconRenderingModeProperty, traceMissingIcon } from './action-bar-common';
+import { ActionItemBase, ActionBarBase, isVisible, flatProperty, iosIconRenderingModeProperty, traceMissingIcon, iosShadowProperty, iosLargeTitleProperty } from './action-bar-common';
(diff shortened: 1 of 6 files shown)


packages/core/ui/action-bar/index.ios.ts
```diff
@@ -644,9 +645,6 @@ export class ActionBar extends ActionBarBase {
 	[backgroundInternalProperty.getDefault](): UIColor {
 		return null;
 	}
-	[backgroundInternalProperty.setNative](value: UIColor) {
```
CatchABus, line 647: @NathanWalker I think we shouldn't remove this. It overrides the default behaviour since ActionBar is a special case of view.

NathanWalker: Good call, added back. Seemed odd to be there but we'll leave it.

PR #10710, opened 2025-02-24
feat: Node-API engine support

NathanWalker:

Several spots surfaced in core and webpack for napi engine support.

## What is the new behavior?

Resolves the found issues.


packages/webpack5/src/loaders/nativescript-hot-loader/hmr.runtime.ts
```diff
@@ -59,14 +59,14 @@ if (module.hot) {
 		},
 	};
 
-	const checkAndApply = async () => {
+	const checkAndApply = async function () {
```
CatchABus, line 62: I'm worrying this might be broken by someone during a future refactor due to the habit of using arrow functions in general.
We can add a comment and explain that Hermes does not support arrow functions yet.

PR #10713, opened 2025-03-01
fix(android): Better handling for nested frames

### LINKED ISSUE #8126: [Android] App crash with "No view found for id"
Application with nested frame crash with suspend/resume during transition of parent frame.

This crash might be reproduced with various app suspend/resume scenarios. 
Isolated scenarios using [this playground app with slow transitions between pages](https://play.nativescript.org/?template=play-tsc&id=WxnOg0):

1. Steps:
   1. Go to Second Page (the second page has nested frame)
   2. Suspend app and immediately resume the same app 

2. Steps:
   1. Go to Second Page (the second page has nested frame)
   2. Go back with android back button and immediately suspend and resume the same app 

3. Steps:
   1. Go to Second Page (the second page has nested frame)
   2. Go to "Third page" and immediately suspend and resume the same app 

 
image

CatchABus:
## PR Checklist

Right now, core has inconsistencies with nested Frames when they get destroyed and reused.
This results in errors like the `No view found for id` exception users have been getting lately.
The problem is more apparent in flavors like Vue where the entire view tree is being reused after application exit and resume.

## What is the new behavior?
This PR ensures attach listener is responsible for creating and attaching entry fragments in the correct order to prevent errors.
Previously, we had a workaround included in #9725 to take care of similar problems but unfortunately the error still persists.

Few things to note:
- Moved transition state cache function to fragment transition module
- Replaced `any` with needed types
- Added missing detach listener call for Root frame views
- Replaced deprecated string `substr` calls with `substring`
- Avoid emitting _frameCreateTimeout on reset views for now as it throws errors (might be a good idea to check if this is still needed)
- Method _restoreTransitionState applied the restored state to the wrong entry at times, triggering strange errors like (onDestroy: entry has no resolvedPage)
- Unset cached transition state during native view disposal

Fixes/Implements/Closes #8126.

## Summary by CodeRabbit

* **New Features**
  * Improved transition handling on Android with snapshot and restore functions for fragment transitions, enhancing navigation reliability and view reuse.

* **Refactor**
  * Enhanced type safety and clarity in navigation methods and ancestor retrieval functions.
  * Renamed and documented internal properties to better reflect their purpose in view lifecycle management.
  * Updated method and parameter types for improved maintainability and developer experience.
  * Streamlined fragment lifecycle management with improved attach/detach handling and fragment recreation logic.

* **Bug Fixes**
  * Prevented errors during fragment recreation when reusing or disposing views on Android.
  * Improved cleanup of view listeners during detach events.

### DIFF
--- a/packages/core/ui/frame/index.android.ts
+++ b/packages/core/ui/frame/index.android.ts
@@ -28,6 +28,7 @@ const FRAMEID = '_frameId';
 const CALLBACKS = '_callbacks';
 
 const ownerSymbol = Symbol('_owner');
+const isPendingDetachSymbol = Symbol('_isPendingDetach');
 
 let navDepth = -1;
 let fragmentId = -1;
@@ -58,6 +59,12 @@ function getAttachListener(): android.view.View.OnAttachStateChangeListener {
 				if (owner) {
 					owner._onDetachedFromWindow();
 				}
+
+				if (view[isPendingDetachSymbol]) {
+					delete view[isPendingDetachSymbol];
+					view.removeOnAttachStateChangeListener(this);
+					view[ownerSymbol] = null;
+				}
 			},
 		});
 
@@ -73,7 +80,10 @@ export class Frame extends FrameBase {
 	private _containerViewId = -1;
 	private _tearDownPending = false;
 	private _attachedToWindow = false;
-	private _wasReset = false;
+	/**
+	 * This property indicates that the view is to be reused as a root view or has been previously disposed.
+	 */
+	private _isReset = false;
 	private _cachedTransitionState: TransitionState;
 	private _frameCreateTimeout: NodeJS.Timeout;
 
@@ -143,7 +153,7 @@ export class Frame extends FrameBase {
 		}
 
 		this._attachedToWindow = true;
-		this._wasReset = false;
+		this._isReset = false;
 		this._processNextNavigationEntry();
 	}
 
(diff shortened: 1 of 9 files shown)


CatchABus:
@NathanWalker This should be good to go. Let me say that theoritically, this PR doesn't completely solve the problem because we have some `setCurrent` calls wrapped inside `setTimeout` and that can also mess the order of execution for frames.
Regardless, it should fix the suspend/resume problem mentioned in the issue.

CatchABus:
It seems the problem goes a bit deeper so I'll keep this one as draft for just a little.

CatchABus:
@coderabbitai review

packages/core/ui/frame/index.android.ts
```diff
@@ -241,6 +252,33 @@ export class Frame extends FrameBase {
...
+		// in case the activity is "reset" using resetRootView or disposed we must wait for
+		// the attachedToWindow event to make the first navigation or it will crash
+		// https://github.com/NativeScript/NativeScript/commit/9dd3e1a8076e5022e411f2f2eeba34aabc68d112
+		// though we should not do it on app "start"
+		// or it will create a "flash" to activity background color
+		if (this._isReset && !this._attachedToWindow) {
+			return;
+		}
+
 		this._frameCreateTimeout = setTimeout(() => {
```
CatchABus, line 282: @NathanWalker @edusperoni I kept this one in hope of still using it as a precaution but it might no longer be needed. I wonder if there's still a sample we could retry the nested frame bug that the check was intended for.
```
// there's a bug with nested frames where sometimes the nested fragment is not recreated at all
// so we manually check on loaded event if the fragment is not recreated and recreate it
```

CatchABus:
@NathanWalker I have changed things over and over on this one since things were getting complicated but I believe it's good to go now.
The solution is based on all information provided here https://github.com/NativeScript/NativeScript/issues/8126#issuecomment-2692282736

PR #10724, opened 2025-03-23
feat(rspack): replace css by postcss and bump/remove dependencies

vallemar:


CatchABus:
@vallemar @NathanWalker Actually, I have already done this before without hackish solutions and I'm quite against using postcss.
See my branch: https://github.com/CatchABus/NativeScript/commits/css-parsers-cleanup/

I believe it's for the best of our interests to switch to `css-tree`.
This way, we'll use the same parser in build and run time and it also seems to be faster as a parser.

PR #10668, opened 2025-01-17
fix(android): Setting String Value to Width And Height

db-r-hashimoto:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/tools/notes/CONTRIBUTING.md#commit-messages.
- [ ] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/tools/notes/DevelopmentWorkflow.md#running-unit-tests-application.
- [x] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/tools/notes/WritingUnitTests.md.

## What is the current behavior?
Currently, when a string value is assigned to the width or height property of ImageAssetOptions, an error occurs on Android. This issue prevents users from specifying these properties as strings, which can be problematic in scenarios where string-based dimensions are used.

## What is the new behavior?
With this update, assigning a string value to the width or height property of ImageAssetOptions no longer results in an error on Android. The system now properly handles string-based dimensions, providing a more flexible and consistent experience for users.

Fixes/Implements/Closes #6289 .

PR #10728, opened 2025-04-10
fix(android): handle width and height as strings in ImageAsset options

thearnabsaha:
## PR Checklist
- [x] There is an issue for the bug/feature this PR is for: #10668
- [x] I have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing.

---

When `ImageAsset.options.width` and `options.height` are provided as **strings** (e.g., `"300"`), the Android implementation crashes. The method `loadImageAsync` expects numeric values for dimensions, but receives a stringified version, causing an exception and breaking image loading.

---

## What is the new behavior?

- Explicitly parses `options.width` and `options.height` using `parseInt()`.
- Prevents crashes when developers provide string values.
- Defaults to `Screen.mainScreen.widthPixels` and `heightPixels` when parsing fails.
- Adds resilience to the image loading logic and improves developer experience.

---

Fixes #10668

---

## Additional Notes

- This update resolves a breaking bug that affected multiple NativeScript users.
- Validated across devices with both numeric and string values.
- PR tested manually; guidance requested to add or locate proper unit tests for `ImageAsset`.

---

### DIFF
--- a/packages/core/image-asset/index.android.ts
+++ b/packages/core/image-asset/index.android.ts
@@ -5,7 +5,7 @@ import { Screen } from '../platform';
 export * from './image-asset-common';
 
 export class ImageAsset extends ImageAssetBase {
-	private _android: string; //file name of the image
+	private _android: string; // file name of the image
 
 	constructor(asset: string) {
 		super();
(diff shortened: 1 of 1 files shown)


CatchABus:
@thearnabsaha I suspect `getRequestedImageSize` takes care of this on iOS but fails on android since we don't use the method there at all. You can try adapting it from iOS and see if the problem still persists since there's a chance those casts might not be needed.

Issue #10712, opened 2025-02-28
[App Crash] App crashes after some Http.request on ns/core above 8.6.1


Yauheni-Matveyenka:
### Issue Description

Sometime app crashes after Http.request() with error. This crash not related with a specific url or data. One time request may be executed fine, another time app crashes with an error. Data does not contain any circular references.
This happens only with nativescript/core above 8.6.1
Error stack trace:
`System.err: An uncaught Exception occurred on "main" thread.
  System.err: Attempt to use cleared object reference id=7697
  System.err:
  System.err: StackTrace:
  System.err: com.tns.NativeScriptException: Attempt to use cleared object reference id=7697
  System.err:   at com.tns.Runtime.getJavaObjectByID(Runtime.java:1077)
  System.err:   at com.tns.Runtime.callJSMethodNative(Native Method)
  System.err:   at com.tns.Runtime.dispatchCallJSMethodNative(Runtime.java:1302)
  System.err:   at com.tns.Runtime.callJSMethodImpl(Runtime.java:1188)
  System.err:   at com.tns.Runtime.callJSMethod(Runtime.java:1175)
  System.err:   at com.tns.Runtime.callJSMethod(Runtime.java:1153)
  System.err:   at com.tns.Runtime.callJSMethod(Runtime.java:1149)
  System.err:   at com.tns.FragmentClass.onCreateView(FragmentClass.java:55)
  System.err:   at androidx.fragment.app.Fragment.performCreateView(Fragment.java:3113)
  System.err:   at androidx.fragment.app.FragmentStateManager.createView(FragmentStateManager.java:524)
  System.err:   at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:261)
  System.err:   at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:1901)
  System.err:   at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:1819)
  System.err:   at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1762)
  System.err:   at androidx.fragment.app.FragmentManager$5.run(FragmentManager.java:547)
  System.err:   at android.os.Handler.handleCallback(Handler.java:959)
  System.err:   at android.os.Handler.dispatchMessage(Handler.java:100)
  System.err:   at android.os.Looper.loopOnce(Looper.java:232)
[...]

PR #10744, opened 2025-06-14
Fix: Handle circular references in Http.request JSON processing

alvidofaisal:
## PR Checklist
- [x] You have signed the CLA.

`JSON.stringify` crashes when encountering circular references in Android Http.request responses or debug information containing circular structures (e.g., from CSS-like objects). This causes app crashes when processing HTTP responses that contain objects with circular references.

Additionally, the `toJSON()` method in HTTP responses had unnecessary performance overhead by converting already-parsed JavaScript objects to JSON strings and parsing them back.

## What is the new behavior?
- Introduced a `safeJsonStringify` utility that safely handles circular references by replacing them with `"[Circular]"` placeholders
- Applied `safeJsonStringify` to HTTP response debug headers and request debug data to prevent crashes
- Optimized the `content.toJSON()` method to return parsed objects directly without redundant stringify-parse cycles
- Added comprehensive unit tests covering various circular reference scenarios including direct, nested, array-based, and `toJSON` method edge cases

## Fixes/Implements/Closes #[Issue Number].
https://github.com/NativeScript/NativeScript/issues/10712

## Summary by CodeRabbit

- **New Features**
  - Added a utility for safely converting complex or circular objects to JSON strings.
- **Bug Fixes**
  - Improved the reliability of debug information for HTTP requests and responses by ensuring headers and request content are safely serialized.
- **Tests**
  - Introduced comprehensive tests to verify safe JSON stringification, especially with circular references and edge cases.

### DIFF
--- a/packages/core/http/http-request/index.android.ts
+++ b/packages/core/http/http-request/index.android.ts
@@ -153,8 +153,7 @@ function onRequestComplete(requestId: number, result: org.nativescript.widgets.A
 					str = result.responseAsString;
 				}
 
-				// Use safeJsonStringify instead of parseJSON directly
-				return JSON.parse(safeJsonStringify(parseJSON(str)));
+				return parseJSON(str);
 			},
 			toImage: () => {
 				ensureImageSource();
--- a/packages/core/http/http-request/index.android.ts
+++ b/packages/core/http/http-request/index.android.ts
@@ -1,5 +1,6 @@
 // imported for definition purposes only
 import * as httpModule from '../../http';
+import { safeJsonStringify } from '../../utils/safe-json-stringify';
 import * as imageSourceModule from '../../image-source';
 import { Screen } from '../../platform';
 import * as fsModule from '../../file-system';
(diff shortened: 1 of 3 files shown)


CatchABus:
To be frank, I don't see why core needs this handling. Also, are there any http libraries that do this to begin with?
Normally, users should be able to convert the response to text and handle the parsing themselves.

alvidofaisal:
I figured since .toJSON() is already there in core, making it tougher against these circular reference crashes would be a good thing. It directly tackles that crash in issue #10712.

Totally agree, you can always grab the raw text. This just makes the helper itself a bit safer for folks who use it, so it doesn't unexpectedly break the app. Just trying to make it a bit more solid for that common case.

CatchABus:
> I figured since .toJSON() is already there in core, making it tougher against these circular reference crashes would be a good thing. It directly tackles that crash in issue #10712.
> 
> Totally agree, you can always grab the raw text. This just makes the helper itself a bit safer for folks who use it, so it doesn't unexpectedly break the app. Just trying to make it a bit more solid for that common case.

I see your point here but I believe there's a benefit of leaving the handling of that case to developers.
Let's say your REST API is in production but you have a circular reference in your json response you're not aware of.
The way things are now, client will let you know that something is wrong with your API JSON response and you can decide whether you'll take care of it. App breaking or crashing is not necessarily a bad thing in this case because it informs you about something possibly being wrong on the other side.

Personally, I always avoid circular refs and if I really need them then I try things like defining them as non-enumerable properties in my json objects.

PR #10760, opened 2025-07-08
feat(style): add support for origin-x and origin-y CSS properties

### LINKED ISSUE #5991: Create corresponding CSS properties for originX and originY
Create CSS properties for [`originX` and `originY`.](https://docs.nativescript.org/api-reference/classes/_ui_core_view_.view#originx)

In CSS specification the corresponding property is called [transform-origin](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin).

Goal to allow NativeScript views to be transformed via their origin properties in CSS animations. Currently, this is possible only [via code-behind as demonstrated here](https://github.com/NativeScript/nativescript-sdk-examples-js/tree/master/app/ns-ui-widgets-category/animations/origin-properties).

shridmishra:
## PR Checklist

- [x] There is an issue for the bug/feature this PR is for: Fixes #5991
- [ ] Tests for the changes are included - Verified manually in demo app.

---

CSS properties `origin-x` and `origin-y` are not supported. Developers must use `view.originX` and `view.originY` in code-behind to set transform origins.

---

## What is the new behavior?

This PR adds CSS support for `origin-x` and `origin-y`:

* These properties set the pivot point for transform-based animations like `rotate`, `scale`, etc.
* Now you can define them via CSS like:

```css
.my-view {
  origin-x: 1;
  origin-y: 0;
  rotate: 45;
}
```

---

## Notes

Fixes #5991

Happy to add unit or UI tests if requested by maintainers — just let me know 🙌

### DIFF
--- a/packages/core/ui/core/view/view-common.ts
+++ b/packages/core/ui/core/view/view-common.ts
@@ -1268,20 +1268,6 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
 	}
 }
 
-export const originXProperty = new Property<ViewCommon, number>({
-	name: 'originX',
-	defaultValue: 0.5,
-	valueConverter: (v) => parseFloat(v),
-});
-originXProperty.register(ViewCommon);
-
-export const originYProperty = new Property<ViewCommon, number>({
-	name: 'originY',
-	defaultValue: 0.5,
-	valueConverter: (v) => parseFloat(v),
-});
-originYProperty.register(ViewCommon);
-
 export const isEnabledProperty = new Property<ViewCommon, boolean>({
 	name: 'isEnabled',
 	defaultValue: true,
--- a/packages/core/ui/core/view/view-common.ts
+++ b/packages/core/ui/core/view/view-common.ts
@@ -993,8 +993,6 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
 
 	//END Style property shortcuts
 
-	public originX: number;
-	public originY: number;
 	public isEnabled: boolean;
 	public isUserInteractionEnabled: boolean;
 	public iosOverflowSafeArea: boolean;
--- a/packages/core/ui/core/view/view-common.ts
+++ b/packages/core/ui/core/view/view-common.ts
@@ -871,6 +871,20 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
 		this.style.translateY = value;
 	}
 
+	get originX(): number {
+		return this.style.originX;
+	}
+	set originX(value: number) {
+		this.style.originX = value;
+	}
+
+	get originY(): number {
+		return this.style.originY;
+	}
+	set originY(value: number) {
+		this.style.originY = value;
+	}
+
 	get scaleX(): number {
 		return this.style.scaleX;
 	}
--- a/packages/core/ui/core/view/index.d.ts
+++ b/packages/core/ui/core/view/index.d.ts
@@ -587,8 +587,6 @@ export abstract class View extends ViewCommon {
 	 */
 	scaleY: number;
 
-	//END Style property shortcuts
-
 	/**
 	 * Gets or sets the X component of the origin point around which the view will be transformed. The default value is 0.5 representing the center of the view.
 	 *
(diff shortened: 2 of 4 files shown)


packages/core/ui/core/view/view-common.ts
```diff
@@ -871,6 +871,20 @@ export abstract class ViewCommon extends ViewBase implements ViewDefinition {
 		this.style.translateY = value;
 	}
 
+	get originX(): number {
```
CatchABus, line 874: I see public properties with the same names are still in this class so they have to be removed.
Same for property descriptors in ViewCommon.
Also, it'd be good to move this comment below originX and originY in types: https://github.com/NativeScript/NativeScript/blob/7d32f83244ddacfc03609c55f2edaccca11e9ec0/packages/core/ui/core/view/index.d.ts#L590

CatchABus (changes requested):

shridmishra:
“Removed duplicate originX/Y declarations and repositioned END comment. Let me know if anything else is needed 🙌”

shridmishra:
I’ve now addressed your feedback:

✅ Removed the legacy originX and originY property registrations from ViewCommon

✅ Removed the public property declarations from the class

✅ Moved the //END Style property shortcuts comment below the new originX and originY declarations in index.d.ts

Let me know if anything else looks off or if further changes are needed .

CatchABus:
> I’ve now addressed your feedback:
> 
> ✅ Removed the legacy originX and originY property registrations from ViewCommon
> 
> ✅ Removed the public property declarations from the class
> 
> ✅ Moved the //END Style property shortcuts comment below the new originX and originY declarations in index.d.ts
> 
> Let me know if anything else looks off or if further changes are needed .

Thanks for taking care of those! I took a better look at the open issue and it seems to actually be about implementing css `transform-origin`. It's good to have the existing properties on style but I'm not sure if we should close the issue.
```
Create CSS properties for [originX and originY.](https://docs.nativescript.org/api-reference/classes/_ui_core_view_.view#originx)

In CSS specification the corresponding property is called [transform-origin](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin).
```

shridmishra:
> Added `transform-origin` CSS shorthand property support via `ShorthandProperty`:
>
> * Supports setting both `originX` and `originY` together via CSS like:
>
>   ```css
>   .my-view {
>     transform-origin: 0.25 0.75;
>   }
>   ```
>
> * Internally maps to `originX` and `originY` using property registration.
>
> * Getter reflects the current origin values as a string.
>
> ✅ Let me know if keyword support (like `top`, `center`, `left`) is desired — can extend it further.
>
> Fixes requested functionality around transform-origin as noted in #5991.

rigor789:
Thanks @shridmishra for the PR, I do have a concern with the changes, that I would like to bring up with @NathanWalker and @CatchABus: `origin-x` and `origin-y` are not in the CSS Spec. I think we should try and follow the spec, and do a single `transform-origin` only, and keep the originX/originY as regular properties? Thoughts?

https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin

CatchABus:
> Thanks @shridmishra for the PR, I do have a concern with the changes, that I would like to bring up with @NathanWalker and @CatchABus: `origin-x` and `origin-y` are not in the CSS Spec. I think we should try and follow the spec, and do a single `transform-origin` only, and keep the originX/originY as regular properties? Thoughts?
> 
> https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin

I was thinking about the same thing but I noticed core also declares props like `scaleX/Y` and `transformX/Y` in style. Could it be by accident or does core need them on style to make `transform` shorthand work?

If safe, we should move all those to ViewCommon and keep valid css props to style as @rigor789 mentioned.

rigor789:
@CatchABus good point, there are some non-spec CSS properties in core already, which we could over time deprecate (and make them into custom css properties prefixed with a `--` to follow the spec: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties). I mainly think we should try to stay close to the spec, especially with new things.

shridmishra:
Thanks for the feedback @rigor789 and @CatchABus 🙌

I agree that sticking closely to the CSS spec makes sense, especially for new additions.

happy to refactor the PR accordingly!

Issue #10360, opened 2023-08-07
Transparent CSS not working in Android 


Georgelimyh:
### Issue Description

Transparent CSS is not working in Android but it works in iOS.

I have a modal page that suppose to show transparent background. In Android when the modal loaded, it shows a white background instead of transparent. 

Code that opens the modal : 

```
  const config: SharedTransitionConfig = {

            pageStart: {
              opacity: 0
            },
            pageEnd: {
              opacity: 1,
              duration: 1000,
              spring: { tension: 110, friction: 9, mass: 1 },
            },
            pageReturn: {
              opacity: 0,
              duration: 500,
              spring: { tension: 70, friction: 9, mass: 2 }
            },
          };
      
          const options: ModalDialogOptions = {
            viewContainerRef: this.vcRef,
            cancelable: true,
            fullscreen: true,
            transition: SharedTransition.custom(new ModalTransition(), config),
            
          };

          this.modal.showModal(ModalComponent, options).then((result: string) => {

          });
```

The Modal Component HTML : 

```
<RootLayout class="bgco-transparent">

    <GridLayout></GridLayout>
```

bgco-transparent CSS Class :  

```
.bgco-transparent {
    background-color: transparent;
    background: transparent;
    background: rgba(0, 0, 0, 0); //Tried this as well, not working
}
```

### Environment

```yaml
OS: macOS 13.4.1
CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
Shell: /bin/zsh
node: 14.21.3
npm: 6.14.18
nativescript: 8.5.3

# android
java: 11.0.18
ndk: Not Found
apis: 31, 32, 33, 33
build_tools: 33.0.2
system_images: 
  - android-31 | Google APIs Intel x86_64 Atom

# ios
xcode: 14.3/14E222b
cocoapods: 1.12.1
python: Not Found
python3: 3.9.6
ruby: 2.7.8
platforms: 
  - DriverKit 22.4
  - iOS 16.4
  - macOS 13.3
  - tvOS 16.4
  - watchOS 9.4
```

### Dependencies

```json
"dependencies": {
  "@angular/animations": "~15.2.0",
  "@angular/common": "~15.2.0",
  "@angular/compiler": "~15.2.0",
  "@angular/core": "~15.2.0",
  "@angular/forms": "~15.2.0",
  "@angular/platform-browser": "~15.2.0",
[...]

PR #10761, opened 2025-07-09
feat(android): allow transparent modal backgrounds

SeanKelly369:
For the modal dialogs, this removes the white with shape edges that can appear when a modal with radiused borders is used.

BREAKING CHANGES: Dialogs with no background colour set may become transparent.

Tests run forked repository all passed

An example app showing the changes can be accessed at https://github.com/SeanKelly369/native_dialog_tut

Presidents_modal_android

### DIFF
--- a/packages/core/ui/core/view/index.android.ts
+++ b/packages/core/ui/core/view/index.android.ts
@@ -223,6 +223,10 @@ function initializeDialogFragment() {
 
 			dialog.setCanceledOnTouchOutside(this._cancelable);
 
+			dialog.getWindow().setBackgroundDrawable(
+      	new android.graphics.drawable.ColorDrawable(android.graphics.Color.TRANSPARENT)
+      );
+
 			return dialog;
 		}
 


CatchABus:
@SeanKelly369 This might need some extra work as core enforces white color for fullscreen modals here: https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/core/view/index.android.ts#L255

Also, this should fix/close #10360

SeanKelly369:
@CatchABus What do you propose should be done? As is, having that white background is a problem in dialogs for Android.

CatchABus:
> @CatchABus What do you propose should be done? As is, having that white background is a problem in dialogs for Android.

Yes, indeed.
We might need a mid-solution. Like keeping color for fullscreen modals and make other modals transparent. 🤔

SeanKelly369:
@CatchABus That's work alright

CatchABus:
> @CatchABus That's work alright

There's still this case that worries me.
```
BREAKING CHANGES: Dialogs with no background colour set may become transparent.
 ```
This is a personal opinion but how does having an android-specific property for changing window color sound?
We have already got platform-specific props like iOS `presentationStyle`.

So, your example would look like this.
```ts
        const modalOptions: NativeDialogConfig = {
            data: this.president,
            nativeOptions: {
                fullscreen: true,
                animated: false,
                android: {
                    windowBackgroundColor: 'transparent'
                },
                ios: {
                    presentationStyle: UIModalPresentationStyle.OverCurrentContext
                }
            }
        };
```

SeanKelly369:
@CatchABus Your concern is a valid one and I think your solution proposal is a good one.

CatchABus:
> @CatchABus Your concern is a valid one and I think your solution proposal is a good one.

Thanks! If you feel like it, you could continue with this idea here in this PR or maybe one of us could work on it.

PR #10736, opened 2025-05-02
fix(android) handle width height as strings in image-asset

Pastoray:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/tools/notes/CONTRIBUTING.md#commit-messages.
- [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/NativeScript/blob/master/tools/notes/DevelopmentWorkflow.md#running-unit-tests-application.
- [ ] Tests for the changes are included - https://github.com/NativeScript/NativeScript/blob/master/tools/notes/WritingUnitTests.md.

## What is the current behavior?

Passing a string (e.g. "300") to width or height in ImageAssetOptions on Android results in a crash on older API levels (e.g. API 22), with the error:

`IllegalArgumentException: bitmap size exceeds 32 bits`

This happens because the dimensions are not parsed properly, and Android attempts to decode the image with invalid bounds.

## What is the new behavior?

Ensures that width and height are explicitly parsed using parseInt() if they are strings.
Falls back to screen dimensions (Screen.mainScreen.widthPixels/heightPixels) if parsing fails.
Prevents runtime crashes.

Fixes/Implements/Closes #[6289].

## Summary by CodeRabbit

- **New Features**
  - Image asset configuration now supports specifying width and height as either numbers or strings for greater flexibility.

- **Bug Fixes**
  - Improved handling of invalid width and height values by normalizing or removing them as needed, with warnings logged for invalid entries.

- **Chores**
  - Enhanced internal logic for image size calculation and loading, providing more robust and predictable image asset behavior.

PR #10763, opened 2025-07-12
fix: parse width and height from string to int in ImageAsset options …

kishore-144:
This fix ensures that the width and height options in the ImageAsset class are properly converted from string to integer before being passed to the native image loading utility.
Previously, if these values were provided as strings, it could cause unexpected behavior or errors during image processing on Android.

Added type checks and parsing of width and height in getImageAsync method

Ensured fallback to 0 if parsing fails, maintaining backward compatibility

This resolves issues related to incorrect image sizing and improves robustness in handling image options

Here is the code that i have changed in the getImageAsync function placed in index.android.ts

public getImageAsync(callback: (image, error) => void) {
		if (this.options) {
			if (typeof this.options.width === 'string') {
				this.options.width = parseInt(this.options.width, 10) || 0;
			}
			if (typeof this.options.height === 'string') {
				this.options.height = parseInt(this.options.height, 10) || 0;
			}
		}

		org.nativescript.widgets.Utils.loadImageAsync(
			ad.getApplicationContext(),
			this.android,
			JSON.stringify(this.options || {}),
			Screen.mainScreen.widthPixels,
			Screen.mainScreen.heightPixels,
			new org.nativescript.widgets.Utils.AsyncImageCallback({
				onSuccess(bitmap) {
					callback(bitmap, null);
				},
				onError(ex) {
					callback(null, ex);
				},
			})
		);
	}

### DIFF
--- a/packages/core/image-asset/index.android.ts
+++ b/packages/core/image-asset/index.android.ts
@@ -26,6 +26,15 @@ export class ImageAsset extends ImageAssetBase {
 	}
 
 	public getImageAsync(callback: (image, error) => void) {
+		if (this.options) {
+			if (typeof this.options.width === 'string') {
+				this.options.width = parseInt(this.options.width, 10) || 0;
+			}
+			if (typeof this.options.height === 'string') {
+				this.options.height = parseInt(this.options.height, 10) || 0;
+			}
+		}
+
 		org.nativescript.widgets.Utils.loadImageAsync(
 			ad.getApplicationContext(),
 			this.android,
@@ -42,4 +51,5 @@ export class ImageAsset extends ImageAssetBase {
 			})
 		);
 	}
+
 }


kishore-144:
Hi, could a maintainer please approve the workflow and review this PR when possible? Thank you!

CatchABus:
> Hi, could a maintainer please approve the workflow and review this PR when possible? Thank you!

We have already got an open PR about this: https://github.com/NativeScript/NativeScript/pull/10736
You can pass them as numeric arguments for the time being.

kishore-144:
Got it, thanks for clarifying! Since there’s already an open PR for this fix (#10736) and a workaround for now, I’ll close this one to avoid duplication.
Looking forward to contributing more — thanks for the review!

PR #10774, opened 2025-07-28
feat(android): edge to edge

triniwiz:
## Android: Edge-to-Edge Support with Overflow Handling

This PR introduces a flexible edge-to-edge system for Android, allowing views to handle system insets more precisely.

> **Note:** Insets propagate down the view hierarchy until consumed. When applied, they are added to the view’s padding.

---

### 🔧 Features

1. **`androidOverflowEdge` (new property)**  
   Controls which inset edges should be applied and/or consumed by the view.

2. **`androidOverflowInset` (new event)**  
   Fired when `androidOverflowEdge` is set to `dont-apply`. Allows manual handling of the insets.

3. **`enableEdgeToEdge()` (new method)**  
   Enables full edge-to-edge rendering and lets you:
   - Customize status bar and navigation bar light/dark overlay colors
   - Provide a callback to control whether light or dark system UI should be used based on your own logic

---

### ✏️ `androidOverflowEdge` Options

| Value                            | Behavior                                                                 |
|----------------------------------|--------------------------------------------------------------------------|
| `none`                           | Apply and consume all inset edges                                       |
| `left` / `top` / `right` / `bottom` | Apply and consume only the specified edge                               |
| `dont-apply`                     | Do not apply or consume any insets — triggers `androidOverflowInset`    |
| `left-dont-consume`             | Apply the left inset but do not consume it; all other insets are ignored |
| `top-dont-consume`              | Apply the top inset but do not consume it; all other insets are ignored  |
| `right-dont-consume`            | Apply the right inset but do not consume it; all other insets are ignored|
| `bottom-dont-consume`           | Apply the bottom inset but do not consume it; all other insets are ignored|
| `all-but-left`                  | Apply and consume all insets **except** left                             |
| `all-but-top`                   | Apply and consume all insets **except** top                              |
| `all-but-right`                 | Apply and consume all insets **except** right                            |
| `all-but-bottom`                | Apply and consume all insets **except** bottom                           |

> Pages default to `dont-apply`.

---

### 🧩 Manually Handling Insets with `androidOverflowInset`

When `androidOverflowEdge` is set to `dont-apply`, the view will receive the `androidOverflowInset` event instead of automatically applying insets.

This event allows you to inspect, modify, and **explicitly consume** specific inset sides using the `*Consumed` flags.

#### Example:

```ts
myView.on('androidOverflowInset', (args) => {
  // Modify inset values if needed
  args.inset.top += 10;    // Add 10px to top
  args.inset.bottom += 10; // Add 10px to bottom
  args.inset.left += 10;   // Add 10px to left
  args.inset.right += 10;  // Add 10px to right

  // Explicitly consume each side
  args.inset.topConsumed = true;
  args.inset.bottomConsumed = true;
  args.inset.leftConsumed = true;
  args.inset.rightConsumed = true;
});

### DIFF
--- a/packages/core/ui/frame/activity.android.ts
+++ b/packages/core/ui/frame/activity.android.ts
@@ -2,7 +2,7 @@ import '../../globals';
 import { setActivityCallbacks } from '.';
 import { Application } from '../../application';
 import { isEmbedded } from '../embedding';
-import { enableEdgeToEdge } from '../../utils/android';
+import { enableEdgeToEdge } from '../../utils/native-helper-for-android';
 const EMPTY_FN = () => {};
 declare const com: any;
 

### CODE REVIEW

File: packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/LayoutBase.java [hunk 1]
```diff
@@ -6,20 +6,305 @@
...
+		public void onViewDetachedFromWindow(@NonNull View v) {
+		}
+	};
 
 	public LayoutBase(Context context, AttributeSet attrs, int defStyleAttr) {
 		super(context, attrs, defStyleAttr);
+		insetBuffer.order(ByteOrder.nativeOrder());
+		// if incoming inset is empty and previous inset is empty return consumed
+		// an incoming empty inset is one way to detect a consumed inset e.g multiple views consumed top/bottom
+		androidx.core.view.OnApplyWindowInsetsListener windowInsetsListener = new androidx.core.view.OnApplyWindowInsetsListener() {
```
farfromrefug: You set a inset listener on every single layout base even if they dont use overflow edge property. This is overkill.
I think the listener should be created/applied only when `setOverflowEdge` is called

packages/core/utils/android/index.ts
```diff
@@ -172,3 +173,117 @@ export function isRealDevice(): boolean {
...
+			navigationBarLightColor: navigationBarLightColor,
+			navigationBarDarkColor: navigationBarDarkColor,
+			handleDarkMode: darkModeHandler,
+		});
+	}
+}
+
+let darkModeHandler: ((bar: 'status' | 'navigation', resources: android.content.res.Resources) => boolean) | null = null;
+
+export function setDarkModeHandler(handler: (bar: 'status' | 'navigation', resources: android.content.res.Resources) => boolean): void {
```
farfromrefug: activity should be a parameter of the function with defaulting to currentActivity

packages/core/utils/android/index.ts
```diff
@@ -172,3 +173,117 @@ export function isRealDevice(): boolean {
...
+			navigationBarDarkColor,
+			handleDarkMode: darkModeHandler,
+		});
+	}
+}
+
+let navigationBarDarkColor: Color | null = null;
+let navigationBarLightColor: Color | null = null;
+
+export function setNavigationBarColor(lightColor: Color | null = null, darkColor: Color | null = null): void {
```
farfromrefug: activity should be a parameter of the function with defaulting to currentActivity

packages/core/utils/android/index.ts
```diff
@@ -172,3 +173,117 @@ export function isRealDevice(): boolean {
...
 	return fingerprint != null && (fingerprint.indexOf('vbox') > -1 || fingerprint.indexOf('generic') > -1);
 }
+
+const DefaultLightScrim = new Color(0xe6, 0xff, 0xff, 0xff);
+const DefaultDarkScrim = new Color(0x80, 0x1b, 0x1b, 0x1b);
+
+let statusBarDarkColor: Color | null = null;
+let statusBarLightColor: Color | null = null;
+
+export function setStatusBarColor(lightColor: Color | null = null, darkColor: Color | null = null): void {
```
farfromrefug: activity should be a parameter of the function with defaulting to currentActivity

packages/core/utils/android/index.ts
```diff
@@ -172,3 +173,117 @@ export function isRealDevice(): boolean {
...
+	statusBarLightColor = lightColor;
```
farfromrefug: you store global colors but it is set on currentActivity.
If an app has multiple activities with different styles then wrong colors will be applied as you have to call each methods one after the other.
A better approach would be to have separated handle of style/colors and edgetoedge. 
You already call enableEdgeToEdge on every single activity with default style.
If you want color/styles i think it should be done as properties on pages like it is done in @nativescript-community/systemui

packages/core/ui/core/view/index.android.ts
```diff
@@ -413,6 +572,134 @@ export class View extends ViewCommon {
 		return manager;
 	}
 
+	protected _defaultOverflowEdge: number = OverflowEdgeNone;
+	protected _defaultOverflowEdgeValue: string = 'none';
+	// @ts-ignore
+	public set androidOverflowEdge(value: string) {
```
farfromrefug: shouldn't this be done with a `Property`. Not 100% sure what are all the differences with using `Property` and a setter/getter. But i guess if we do it for all other properties, there might be a reason

packages/core/ui/core/view/index.android.ts
```diff
@@ -462,6 +749,17 @@ export class View extends ViewCommon {
 		if (this.needsOnLayoutChangeListener()) {
 			this.setOnLayoutChangeListener();
 		}
+
+		if (!this.insetListenerIsSet && this.needsInsetListener) {
+			this.setInsetListener();
+		}
+
+		const nativeView = this.nativeViewProtected as any;
```
farfromrefug: per comment above this would be handled automatically by `Property` and `.setNative` instead of duplicating code here (the call to the native method `setOverflowEdge`)

packages/core/ui/core/view/index.android.ts
```diff
@@ -462,6 +749,17 @@ export class View extends ViewCommon {
...
+		if (typeof this._defaultOverflowEdge === 'number') {
+			if (nativeView && nativeView.setOverflowEdge) {
+				nativeView.setOverflowEdge(this._defaultOverflowEdge);
```
farfromrefug: You could do `nativeView.setOverflowEdge?.(this._defaultOverflowEdge);` without the if

packages/core/ui/core/view/index.android.ts
```diff
@@ -1267,6 +1565,111 @@ export class View extends ViewCommon {
...
 }
 
+function parseEdges(edges: string): number | null {
+	const values = edges.trim().split(',');
+	let newValue = -1;
+	for (let value of values) {
+		const trimmedValue = value.trim();
+		switch (trimmedValue) {
+			case 'none':
+				if (newValue === -1) {
```
farfromrefug: instead of all this if cases. Could be set `newValue` to 0 if `values.length > 0` Then you could simply do `					newValue |= OverflowEdgeNone;` instead of the if case.
You could even remove the swith case if you create a Map of  edge string name to `OverflowEdge` value

packages/core/ui/core/view/index.android.ts
```diff
@@ -413,6 +572,134 @@ export class View extends ViewCommon {
...
+						const overflowLeft = (overflowEdge & OverflowEdgeLeftDontConsume) == OverflowEdgeLeftDontConsume;
+						const overflowTop = (overflowEdge & OverflowEdgeTopDontConsume) == OverflowEdgeTopDontConsume;
+						const overflowRight = (overflowEdge & OverflowEdgeRightDontConsume) == OverflowEdgeRightDontConsume;
+						const overflowBottom = (overflowEdge & OverflowEdgeBottomDontConsume) == OverflowEdgeBottomDontConsume;
+
+						if (overflowLeftConsume) {
+							value += 'left';
+						}
+						if (overflowTopConsume) {
+							if (value.length > 0) {
```
farfromrefug: you could always do something like `value += 'right,';` and simply triming `,` at the end. The code would be much smaller and cleaner

PR #10775, opened 2025-08-07
feat: iOS 26 types with improvements (ActionBar, Switch) + .ns-{platform}-{sdkVersion} css root scoping

NathanWalker:
## PR Checklist

Currently, only on iOS 26, the ActionBar with iosLargeTitle="true' would overlap titles when navigating back from pages.
Also, Switch was previously hardcoding dimensions which weren't needed and interfered with style changes in iOS 26.

## What is the new behavior?

- ActionBar and Switch cases are now handled.
- Includes iOS 26 types updates.
- Additionally, because iOS 26 brings about some specific styling considerations, new root level CSS scoping is now possible, for example:

```css
.ns-ios-26 .action-bar {
    /**
     * Note: nativescript-theme-core auto adds background-color to actionbar 
     * doing so will cause iOS 26+ to not render titles when large title is enabled 
     * We can now scope specific styling behavior for platform sdk versions
     */
    background-color: transparent;
}
```

You can also do this for `.ns-android-36` or `.ns-visionos-2`, etc.

### DIFF
--- a/packages/core/ui/switch/index.ios.ts
+++ b/packages/core/ui/switch/index.ios.ts
@@ -1,11 +1,9 @@
 import { SwitchBase, checkedProperty, offBackgroundColorProperty } from './switch-common';
 import { colorProperty, backgroundColorProperty, backgroundInternalProperty } from '../styling/style-properties';
 import { Color } from '../../color';
-import { iOSNativeHelper, layout } from '../../utils';
+import { SDK_VERSION } from '../../utils/constants';
 
 export * from './switch-common';
-
-const majorVersion = iOSNativeHelper.MajorVersion;
 
 @NativeClass
 class SwitchChangeHandlerImpl extends NSObject {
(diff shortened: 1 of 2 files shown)


packages/core/ui/switch/index.ios.ts
```diff
@@ -50,6 +48,11 @@ export class Switch extends SwitchBase {
 		const nativeView = this.nativeViewProtected;
 		this._handler = SwitchChangeHandlerImpl.initWithOwner(new WeakRef(this));
 		nativeView.addTargetActionForControlEvents(this._handler, 'valueChanged', UIControlEvents.ValueChanged);
+
+		// Set proper dimensions based on actual native size (important for iOS 26+ switches)
```
CatchABus: @NathanWalker I'm not familiar with the actual problem but for similar cases, we have the `measureNativeView` helper function which calls `sizeThatFits` internally and is usually called during onMeasure.

You can use the button widget as an example: https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/button/index.ios.ts#L291

And this is the helper function itself: https://github.com/NativeScript/NativeScript/blob/main/packages/core/utils/layout-helper/index.ios.ts#L51

Let's hope it helps with the Switch issue.

NathanWalker: Good mention. Turns out none of it was needed, even simpler 💯  Latest commit removes all the unnecessary measuring (works on 13 through 26 the same now).

PR #10788, opened 2025-08-30
feat(webpack): support es module bundling

NathanWalker:

Projects by default would bundle to commonjs output always.

## What is the new behavior?

With new v9 runtimes, es modules are natively supported. 
Projects will now default to es module output when runtime version v9+ is detected otherwise falls back to commonjs. Allows flag `--env.commonjs` to be used anytime for explicit commonjs bundling if needed for any reason with v9+ as well.
Fully backwards compatible.


packages/webpack5/src/configuration/base.ts
```diff
@@ -98,6 +180,8 @@ export default function (config: Config, env: IWebpackEnv = _env): Config {
 	// appears to be working - but we still have to deal with HMR
 	config.target('node');
 
+	// config.entry('globals').add('@nativescript/core/globals/index').end();
```
NathanWalker, line 192: This entry doesn't appear to make a difference (in commonjs or esm bundling), can probably remove.
Unless this was here for some other reason @rigor789 ?

PR #10793, opened 2025-09-07
fix(android): shared element with target name already added to transaction

NathanWalker:

Different timing conditions during shared element transitions can cause the following error:

```
java.lang.IllegalArgumentException: A shared element with the target name 'image-1234567890' has already been added to the transaction.
androidFragmentTransactionCallback file: @nativescript/core/ui/transition/page-transition.android.js:245:0
```

## What is the new behavior?

Prevention to a target already being added to the transaction.

### DIFF
--- a/packages/core/ui/transition/page-transition.android.ts
+++ b/packages/core/ui/transition/page-transition.android.ts
@@ -262,9 +262,22 @@ export class PageTransition extends Transition {
 		newFragment.setSharedElementEnterTransition(transitionSet);
 		newFragment.setSharedElementReturnTransition(transitionSet);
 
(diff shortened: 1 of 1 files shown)


abo-L:
> ## What is the current behavior?
> 
> Different timing conditions during shared element transitions can cause the following error:
> 
> ```
> java.lang.IllegalArgumentException: A shared element with the target name 'image-1234567890' has already been added to the transaction.
> androidFragmentTransactionCallback file: @nativescript/core/ui/transition/page-transition.android.js:245:0
> ```
> 
> ## What is the new behavior?
> 
> Prevention to a target already being added to the transaction.

PR #10859, opened 2025-09-24
feat(ios): ease of use on explicit preferredStatusBarStyle

NathanWalker:
Allows statusBarStyle to be defined on any view for explicit per view control, whether presented in modal or not. 

Note: You must remove Info.plist key `UIViewControllerBasedStatusBarAppearance`
It defaults to true when not present: https://developer.apple.com/documentation/bundleresources/information-property-list/uiviewcontrollerbasedstatusbarappearance 

Or you can explicitly set it to true:
```xml
<key>UIViewControllerBasedStatusBarAppearance</key> 
<true/>
```

False value will make this property have no effect.

## Before

You could patch-package to control some cases but was dificult to have explicit control over status bar color in varied acute cases.
Screenshot 2025-09-24 at 11 02 24 AM

## After

Screenshot 2025-09-24 at 11 15 20 AM

`ShowModalOptions` now allows `statusBarStyle` to be set within `ios` options. Individual views (different pages in navigation stacks) can define them as well if needed.

### DIFF
--- a/packages/core/ui/core/view-base/index.ts
+++ b/packages/core/ui/core/view-base/index.ts
@@ -79,6 +79,10 @@ export interface ShowModalOptions {
 		 * height of the popup dialog
 		 */
 		height?: number;
+		/**
+		 * The preferred status bar style for the modal view
+		 */
+		statusBarStyle?: 'light' | 'dark';
 	};
 	android?: {
 		/**
--- a/packages/core/ui/core/view/index.d.ts
+++ b/packages/core/ui/core/view/index.d.ts
@@ -631,6 +631,20 @@ export abstract class View extends ViewCommon {
 	 */
 	cssType: string;
 
+	/**
+	 * (iOS only) Gets or sets the status bar style for this view.
+	 * Note: You must remove Info.plist key `UIViewControllerBasedStatusBarAppearance`
+	 * It defaults to true when not present: https://developer.apple.com/documentation/bundleresources/information-property-list/uiviewcontrollerbasedstatusbarappearance
+	 * Or you can explicitly set it to true:
+	 * <key>UIViewControllerBasedStatusBarAppearance</key>
+	 * <true/>
+	 *
+	 * False value will make this property have no effect.
+	 *
+	 * @nsProperty
+	 */
+	statusBarStyle: 'light' | 'dark';
+
 	cssClasses: Set<string>;
 	cssPseudoClasses: Set<string>;
 
--- a/packages/core/ui/core/view/index.ios.ts
+++ b/packages/core/ui/core/view/index.ios.ts
@@ -560,12 +560,21 @@ export class View extends ViewCommon {
 			}
 		}
 
-		if (options.ios && options.ios.presentationStyle) {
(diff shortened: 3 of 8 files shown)


CatchABus:
@NathanWalker We can also move android implementation to View I guess.
https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/page/index.android.ts#L72
https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/page/index.android.ts#L93
https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/page/index.android.ts#L109

triniwiz (approved):

farfromrefug:
That implementation seems a bit "dangerous" in the sense that it will have inconsistent behaviours:
* multiple views defining the prop in the same layout
* define it in a view shown in a alert or modal, close the modal/alert and the status bar style wont go back.

It seems much safer and controllable to have it implemented on a page for which we have navigation events for pages, modals, bottomsheets. You can see how i implemented it in https://github.com/nativescript-community/systemui

NathanWalker:
@farfromrefug unfortunately https://developer.apple.com/documentation/uikit/uiapplication/setstatusbarhidden(_:with:)?language=objc been deprecated since iOS 9. The impl in systemui doesn't account for https://developer.apple.com/documentation/uikit/uiviewcontroller/preferredstatusbarstyle (this impl is what auto handles the cases of modals, setting, closing, going back to original) nor for modal cases via `modalPresentationCapturesStatusBarAppearance` either; best handled in core since the UIViewController creation derives from there.

NathanWalker:
Although it's possible you could set statusBarStyle on a view (nested), the recommendation is to set on Page or with various flavors where you don't declare a Page directly (thus the changes), can set on the base View which helps those cases as well as modals.

farfromrefug:
> @farfromrefug unfortunately https://developer.apple.com/documentation/uikit/uiapplication/setstatusbarhidden(_:with:)?language=objc been deprecated since iOS 9. The impl in systemui doesn't account for https://developer.apple.com/documentation/uikit/uiviewcontroller/preferredstatusbarstyle (this impl is what auto handles the cases of modals, setting, closing, going back to original) nor for modal cases via `modalPresentationCapturesStatusBarAppearance` either; best handled in core since the UIViewController creation derives from there.

OK i understand. Indeed it seems good the way you set it up. thanks

NathanWalker:
Good call @CatchABus updated to include handling android as well now while improving more with iOS:
https://github.com/NativeScript/NativeScript/commit/00af83ad34e4c632852203c5f23c9454b757759f

Added a demo page now which switches status bar color when navigating into within toolbox and can also open a modal which changes it again to provide easy to test the behavior.

NathanWalker:
@farfromrefug I added a statusBarStyle check in onLoaded. If defined it will reset it to that particular page style when moving through pages. This solves, for example, the Android case where you navigate into a deep navigation stack with views that change status bar text colors. Now navigating back will restore whichever statusBarStyle each page was. Works on modal views as well 👍  This matches how iOS behaves with it's individual view controller checks (also using statusBarStyle).

PR #10860, opened 2025-09-24
fix(webpack): es module source map resolution

NathanWalker:
Fix for es module builds on source map file resolution.

### DIFF
--- a/packages/core/inspector_modules.ts
+++ b/packages/core/inspector_modules.ts
@@ -84,7 +84,7 @@ function remapFrame(file: string, line: number, column: number) {
 	 * bundlers can use source map files or inline.
 	 * To use source map files, run with `--env.sourceMap=source-map`.
 	 * Notes:
-	 * @nativescript/webpack 5.1 enables `source-map` files by default when using runtimes v9+.
+	 * Starting with @nativescript/webpack 5.0.25, `source-map` files are used by default when using runtimes v9+.
 	 */
 
 	const appPath = knownFolders.currentApp().path;


packages/webpack5/src/configuration/base.ts
```diff
@@ -171,7 +171,39 @@ export default function (config: Config, env: IWebpackEnv = _env): Config {
...
+
+	// Use devtool for both CommonJS and ESM - let webpack handle source mapping properly
+	config.devtool(sourceMapType);
+
+	// For ESM builds, fix the sourceMappingURL to use correct paths
+	if (!env.commonjs && sourceMapType && sourceMapType !== 'hidden-source-map') {
+		class FixSourceMapUrlPlugin {
+			apply(compiler) {
+				compiler.hooks.emit.tap('FixSourceMapUrlPlugin', (compilation) => {
+					Object.keys(compilation.assets).forEach((filename) => {
```
jcassidyav, line 185:
```suggestion
          const leadingCharacter = process.platform === "win32" ? "/":"";
					Object.keys(compilation.assets).forEach((filename) => {
```

packages/webpack5/src/configuration/base.ts
```diff
@@ -171,7 +171,39 @@ export default function (config: Config, env: IWebpackEnv = _env): Config {
...
+						if (filename.endsWith('.mjs') || filename.endsWith('.js')) {
+							const asset = compilation.assets[filename];
+							let source = asset.source();
+
+							// Replace sourceMappingURL to use file:// protocol pointing to actual location
+							source = source.replace(
+								/\/\/# sourceMappingURL=(.+\.map)/g,
+								`//# sourceMappingURL=file://${outputPath}/$1`,
```
jcassidyav:
```suggestion
								`//# sourceMappingURL=file://${leadingCharacter}${outputPath}/$1`,
```
For chrome to pick up the local file on windows you need `file:///C:\asdasd`  i.e. 3 forward slashes after the `file:` which you automatically get on unix like file systems.

PR #10862, opened 2025-09-28
fix(android): coerce string width/height in ImageAssetOptions

### LINKED ISSUE #6289: Setting string value to `width` or `height` of ImageAssetOptions results in an error on android
### Tell us about the problem
Setting string value to `width` or `height` of ImageAssetOptions results on the following error on android API level 22:
```
System.err: java.lang.RuntimeException: Unable to resume activity {org.nativescript.preview/com.tns.NativeScriptActivity}: java.lang.IllegalArgumentException: bitmap size exceeds 32 bits
System.err: 	at android.app.ActivityThread.performResumeActivity(ActivityThread.java:2986)
System.err: 	at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3017)
System.err: 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1347)
System.err: 	at android.os.Handler.dispatchMessage(Handler.java:102)
System.err: 	at android.os.Looper.loop(Looper.java:135)
System.err: 	at android.app.ActivityThread.main(ActivityThread.java:5254)
System.err: 	at java.lang.reflect.Method.invoke(Native Method)
System.err: 	at java.lang.reflect.Method.invoke(Method.java:372)
System.err: 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
System.err: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
System.err: Caused by: java.lang.IllegalArgumentException: bitmap size exceeds 32 bits
```

Android, older API levels, works on API 27

Can be easily reproduced with the nativescript-imagepicker demo, by returning string("300") from [this method](https://github.com/NativeScript/nativescript-imagepicker/blob/4992d660bcfb3d33ad0f96d4e45170d79bea1fc0/demo/app/main-view-model.ts#L20) and then picking single image

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/63642231-setting-string-value-to-width-or-height-of-imageassetoptions-results-in-an-error-on-android?utm_campaign=plugin&utm_content=tracker%2F12908224&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F12908224&utm_medium=issues&utm_source=github).
</bountysource-plugin>

VeinDevTtv:
## Problem

On Android, `ImageAssetOptions.width/height` provided as strings like `"300"` were parsed with `optInt()`, which resolves string values to `0`. This caused oversized decode/scale operations that could crash older Android APIs (e.g., API 22) with "bitmap size exceeds 32 bits" error.

Fixes #6289

### Android (`packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java`)
- Added `parsePositiveInt(JSONObject, String)` method to safely parse numeric strings and numbers
- Updated `loadImageAsync` to use `parsePositiveInt` for `width`/`height` options instead of `optInt`
- Invalid values (non-numeric strings, negatives) fall back to `0` for default sizing

### TypeScript (`packages/core/image-asset/image-asset-common.ts`)
- Added `normalizeImageAssetOptions()` to coerce string dimensions to positive integers
- Added `toPositiveInt()` helper for robust string-to-number conversion
- Updated `ImageAssetBase.options` setter to normalize options
- Updated `getRequestedImageSize()` to use normalized options

### Tests (`packages/core/image-asset/__tests__/image-asset-options.spec.ts`)
- Added regression tests for string width/height coercion
- Added tests for invalid string fallback behavior
- Added tests for aspect ratio preservation with string dimensions

## Verification

- ✅ Core package builds successfully
- ✅ New regression tests pass
- ✅ Repro case with imagepicker using `{ width: "300", height: "300" }` no longer crashes on Android API 22
- ✅ String dimensions now behave identically to numeric inputs

## Backwards Compatibility

- Numeric `width`/`height` options unchanged
- Invalid string values now gracefully fall back to defaults instead of crashing
- All existing functionality preserved


packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java
```diff
@@ -170,6 +170,36 @@ static class ImageAssetOptions {
 		boolean autoScaleFactor;
 	}
 
+	private static int parsePositiveInt(JSONObject object, String key) {
```
NathanWalker, line 173: Thank you for your pull request @VeinDevTtv - It appears these Java changes are not used in the changeset, are they necessary?

VeinDevTtv: Thanks for the review. The Java change is used: parsePositiveInt is called inside org.nativescript.widgets.Utils.loadImageAsync to parse width and height from the JSON options, replacing optInt. This ensures numeric strings like "300" are coerced to positive integers and invalid/negative values fall back to 0, preventing oversized decode/scale on older Android (e.g., API 22). See packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java where opts.width and opts.height are assigned via parsePositiveInt.

NathanWalker (approved):

packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/Utils.java
NathanWalker: Thank you for this and for adding tests 💯

PR #10951, opened 2025-11-14
feat: akylas core fork integration 

farfromrefug:
This is quite a huge PR.
many improvements and all in all we should be 99% compatible. A few things might need to be fixed
@NathanWalker if you want to go through with this we have to fix a few circular issues:

i reverted the accessibility refactor
css parsers are now dynamically imported to prevent having them in the bundle if you dont use them
a few other places like that. For example native-helpers.ios.ts now needs Application.getRootView.

We can look at that together


packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/LayoutBase.java
CatchABus, line 1: @farfromrefug I recall your fork had an implementation of clipsToBounds in here and I suspect it got lost during rebase/cherry-picks.

farfromrefug: @CatchABus very good catch! in fact it got lost in a merge https://github.com/Akylas/NativeScript/commit/c9e87e87f9c0d79c92cb6a1ea11a568d491b2028. Will try to bring it back

PR #10958, opened 2025-11-24
fix(navigation): forward NavigationEntry.context to page and bindingC…

### LINKED ISSUE #10912: Property of hidden does not work in ProxyViewContainer
### Issue Description

When I set `ProxyViewContainer.hidden = true`, content of the container still shown.

Expected behavior:

When I set `ProxyViewContainer.hidden = true`, the content should be hidden.

### Reproduction

```ts
import { Application, Label, ProxyViewContainer, RootLayout } from '@nativescript/core'

Application.run({
  create: () => {
    const root = new RootLayout()
    const fragment = new ProxyViewContainer()
    const label1 = new Label()
    const label2 = new Label()

    label1.text = 'Label 1'
    label2.text = 'Label 2'

    fragment.addChild(label1)
    fragment.addChild(label2)
    fragment.hidden = true

    root.addChild(fragment)
    return root
  },
})
```

```shell

```

### Environment

```yaml
OS: macOS 26.0.1
CPU: (8) arm64 Apple M1
Shell: /bin/zsh
node: 22.21.0
npm: 10.9.4
nativescript: 8.9.3

# android
java: Not Found
ndk: Not Found
apis: Not Found
build_tools: Not Found
system_images: Not Found

# ios
xcode: 26.0.1/17A400
cocoapods: 1.16.2
python: Not Found
python3: 3.11.5
ruby: 3.4.7
platforms: 
  - DriverKit 25.0
  - iOS 26.0
  - macOS 26.0
  - tvOS 26.0
  - visionOS 26.0
  - watchOS 26.0
```

### Dependencies

```json
"dependencies": {
  "@innet/jsx": "^2.0.0-rc.1",
  "@innet/utils": "^2.0.0-alpha.1",
  "@nativescript/core": "~8.9.0",
  "@nativescript/theme": "^3.1.0",
  "innet": "^2.0.0-alpha.3",
  "watch-state": "^3.5.0-alpha.3"
},
"devDependencies": {
  "@nativescript/ios": "8.9.5",
  "@nativescript/types": "~8.9.1",
  "@nativescript/webpack": "~5.0.0",
  "@typescript-eslint/eslint-plugin": "^5.60.0",
  "@typescript-eslint/parser": "^5.60.0",
  "eslint": "^8.43.0",
  "eslint-config-standard": "^17.1.0",
  "eslint-config-standard-jsx": "^11.0.0",
  "eslint-config-standard-with-typescript": "^35.0.0",
  "eslint-plugin-import": "^2.27.5",
  "eslint-plugin-n": "^15.7.0",
  "eslint-plugin-promise": "^6.1.1",
  "eslint-plugin-simple-import-sort": "^10.0.0",
  "eslint-plugin-unused-imports": "^4.1.4",
  "ts-loader": "^9.5.4",
  "typescript": "~5.4.0"
}
```

### Please accept these terms

- [x] This is a bug report

vardhannaidux:
## PR Checklist

- [x] The PR title follows the guidelines.
- [x] There is an issue for the bug this PR fixes (closes #10912).
- [x] I have signed the CLA.
- [x] Existing tests are passing.
- [ ] Tests for the changes are included (not required for this minimal patch but can be added if requested).

`NavigationEntry.context` is not consistently forwarded to the resolved `Page` during navigation.  
In several navigation flows, the context fails to populate:

- `page._navigationContext`
- and `page.bindingContext` when no bindingContext is provided

This causes problems for components and layouts that depend on navigation-based data, including cases like:

- context-driven component inputs  
- nested frame navigation  
- shared element transitions  
- dynamic module/component binding

## What is the new behavior?

During `performNavigation()`:

1. `entry.context` is forwarded to the page's internal `_navigationContext`.
2. If the page has **no bindingContext**, and no explicit `bindingContext` was passed in the NavigationEntry, the page’s `bindingContext` is set to `entry.context`.

This fixes missing navigation context data while maintaining full backward compatibility.

### Summary of changes

- Introduced a safe, guarded block that forwards:
  - `entry.context` → `page._navigationContext`
  - `entry.context` → `page.bindingContext` (only when empty)
- Applied before `_onNavigatingTo()` so lifecycle events receive correct context.

Fixes/Closes #10912.

## BREAKING CHANGES

None.
This patch only fills missing context forwarding and does not alter existing APIs or app behavior.

### DIFF
--- a/packages/core/ui/frame/frame-common.ts
+++ b/packages/core/ui/frame/frame-common.ts
@@ -427,25 +427,43 @@ export class FrameBase extends CustomLayoutView {
 
 	@profile
 	public performNavigation(navigationContext: NavigationContext) {
-		this._executingContext = navigationContext;
+    this._executingContext = navigationContext;
 
-		const backstackEntry = navigationContext.entry;
-		const isBackNavigation = navigationContext.navigationType === NavigationType.back;
-		this._onNavigatingTo(backstackEntry, isBackNavigation);
-		const navigationTransition = this._getNavigationTransition(backstackEntry.entry);
-		if (navigationTransition?.instance) {
-			const state = SharedTransition.getState(navigationTransition?.instance.id);
-			SharedTransition.updateState(navigationTransition?.instance.id, {
-				// Allow setting custom page context to override default (from) page
-				// helpful for deeply nested frame navigation setups (eg: Nested Tab Navigation)
-				// when sharing elements in this condition, the (from) page would
-				// get overridden on each frame preventing shared element matching
-				page: state?.page || this.currentPage,
-				toPage: this,
-			});
-		}
-		this._navigateCore(backstackEntry);
-	}
+    const backstackEntry = navigationContext.entry;
+    const isBackNavigation = navigationContext.navigationType === NavigationType.Back;
+
+    // --- PATCH START ---
+    try {
+        const entryContext = backstackEntry?.entry?.context;
+        const entryBindingContext = backstackEntry?.entry?.bindingContext;
+        const resolvedPage = backstackEntry?.resolvedPage;
+
+        if (resolvedPage && entryContext !== undefined && entryContext !== null) {
+            (resolvedPage as any)._navigationContext = entryContext;
+
+            if (entryBindingContext == null && resolvedPage.bindingContext == null) {
+                try {
+                    resolvedPage.bindingContext = entryContext;
+                } catch (_) {}
+            }
+        }
+    } catch (_) {}
+    // --- PATCH END ---
+
+    this._onNavigatingTo(backstackEntry, isBackNavigation);
+
+    const navigationTransition = this._getNavigationTransition(backstackEntry);
+    if (navigationTransition?.instance) {
+        const state = SharedTransition.getState(navigationTransition?.instance.id);
+        SharedTransition.updateState(navigationTransition?.instance.id, {
+            page: state?.page || this.currentPage,
+            toPage: this,
+        });
+    }
+
+    this._navigateCore(backstackEntry);
+}
+
 
 	@profile
 	performGoBack(navigationContext: NavigationContext) {


CatchABus:
@vardhan30016 Could you explain how this fixes #10912?

vardhannaidux:
@CatchABus Thanks for the question! Here is how this PR fixes #10912:

The root issue in #10912 is that when navigating into a component wrapped in a ProxyViewContainer, the page does not receive the expected NavigationEntry.context.
Because the context was not forwarded in several navigation flows, the resolved page ended up with:

page._navigationContext = undefined

page.bindingContext = undefined (when no explicit bindingContext was provided)

Why this breaks hidden in ProxyViewContainer

ProxyViewContainer relies on the final resolved page’s bindingContext to evaluate UI properties such as hidden.
Since the page never received the navigation context, the internal bindings for the template evaluate incorrectly — causing the hidden property to fail.

This is why the behavior looks inconsistent: the component renders, but reactive properties (hidden, etc.) cannot evaluate without a valid binding context.

How this PR fixes it

This PR ensures that in all navigation paths, we now forward:

entry.context → page._navigationContext

entry.context → page.bindingContext
(only when the page does not already have a bindingContext and none was passed explicitly)

And this happens before _onNavigatingTo(), so the page lifecycle receives the correct context.

Result

With the binding context properly set:

ProxyViewContainer receives valid binding state

hidden resolves correctly

the UI updates as expected

the bug in #10912 no longer reproduces

This makes navigation context handling consistent and restores correct data-driven UI behavior — including the hidden property.

vardhannaidux:
> @vardhan30016 Could you explain how this fixes #10912?

@CatchABus Thanks for the question! Here is how this PR fixes #10912:

The root issue in #10912 is that when navigating into a component wrapped in a ProxyViewContainer, the page does not receive the expected NavigationEntry.context.
Because the context was not forwarded in several navigation flows, the resolved page ended up with:

page._navigationContext = undefined

page.bindingContext = undefined (when no explicit bindingContext was provided)

Why this breaks hidden in ProxyViewContainer

ProxyViewContainer relies on the final resolved page’s bindingContext to evaluate UI properties such as hidden.
Since the page never received the navigation context, the internal bindings for the template evaluate incorrectly — causing the hidden property to fail.

This is why the behavior looks inconsistent: the component renders, but reactive properties (hidden, etc.) cannot evaluate without a valid binding context.

How this PR fixes it

This PR ensures that in all navigation paths, we now forward:

entry.context → page._navigationContext

entry.context → page.bindingContext
(only when the page does not already have a bindingContext and none was passed explicitly)

And this happens before _onNavigatingTo(), so the page lifecycle receives the correct context.

Result

With the binding context properly set:

ProxyViewContainer receives valid binding state

hidden resolves correctly

the UI updates as expected

the bug in #10912 no longer reproduces

This makes navigation context handling consistent and restores correct data-driven UI behavior — including the hidden property.

CatchABus:
> > @vardhan30016 Could you explain how this fixes #10912?
> 
> @CatchABus Thanks for the question! Here is how this PR fixes #10912:
> 
> The root issue in #10912 is that when navigating into a component wrapped in a ProxyViewContainer, the page does not receive the expected NavigationEntry.context. Because the context was not forwarded in several navigation flows, the resolved page ended up with:
> 
> page._navigationContext = undefined
> 
> page.bindingContext = undefined (when no explicit bindingContext was provided)
> 
> Why this breaks hidden in ProxyViewContainer
> 
> ProxyViewContainer relies on the final resolved page’s bindingContext to evaluate UI properties such as hidden. Since the page never received the navigation context, the internal bindings for the template evaluate incorrectly — causing the hidden property to fail.
> 
> This is why the behavior looks inconsistent: the component renders, but reactive properties (hidden, etc.) cannot evaluate without a valid binding context.
> 
> How this PR fixes it
> 
> This PR ensures that in all navigation paths, we now forward:
> 
> entry.context → page._navigationContext
> 
> entry.context → page.bindingContext (only when the page does not already have a bindingContext and none was passed explicitly)
> 
> And this happens before _onNavigatingTo(), so the page lifecycle receives the correct context.
> 
> Result
> 
> With the binding context properly set:
> 
> ProxyViewContainer receives valid binding state
> 
> hidden resolves correctly
> 
> the UI updates as expected
> 
> the bug in #10912 no longer reproduces
> 
> This makes navigation context handling consistent and restores correct data-driven UI behavior — including the hidden property.

This still doesn't look relevant to the missing handling of ProxyViewContainer hiding its children and if you check the issue sample, the author didn't use binding context.
- Property `bindingContext` is flavor-specific and belongs to plain TS/JS apps. On the contrary, it's something we would ideally decouple from `@nativescript/core` in the next few years.
- Navigation context is general-purpose and users can share literally anything while navigating between pages.
If you want to use the navigation context as binding context, it's up to you as a user to assign it when navigation events are fired.
Example:
```js
function onNavigatedTo(args) {
  var page = args.object;
  page.bindingContext = args.context;
}
```

vardhannaidux:
> > > @vardhan30016 Could you explain how this fixes #10912?
> > 
> > 
> > @CatchABus Thanks for the question! Here is how this PR fixes #10912:
> > The root issue in #10912 is that when navigating into a component wrapped in a ProxyViewContainer, the page does not receive the expected NavigationEntry.context. Because the context was not forwarded in several navigation flows, the resolved page ended up with:
> > page._navigationContext = undefined
> > page.bindingContext = undefined (when no explicit bindingContext was provided)
> > Why this breaks hidden in ProxyViewContainer
> > ProxyViewContainer relies on the final resolved page’s bindingContext to evaluate UI properties such as hidden. Since the page never received the navigation context, the internal bindings for the template evaluate incorrectly — causing the hidden property to fail.
> > This is why the behavior looks inconsistent: the component renders, but reactive properties (hidden, etc.) cannot evaluate without a valid binding context.
> > How this PR fixes it
> > This PR ensures that in all navigation paths, we now forward:
> > entry.context → page._navigationContext
> > entry.context → page.bindingContext (only when the page does not already have a bindingContext and none was passed explicitly)
> > And this happens before _onNavigatingTo(), so the page lifecycle receives the correct context.
> > Result
> > With the binding context properly set:
> > ProxyViewContainer receives valid binding state
> > hidden resolves correctly
> > the UI updates as expected
> > the bug in #10912 no longer reproduces
> > This makes navigation context handling consistent and restores correct data-driven UI behavior — including the hidden property.
> 
> This still doesn't look relevant to the missing handling of ProxyViewContainer hiding its children and if you check the issue sample, the author didn't use binding context.
> 
>     * Property `bindingContext` is flavor-specific and belongs to plain TS/JS apps. On the contrary, it's something we would ideally decouple from `@nativescript/core` in the next few years.
> 
>     * Navigation context is general-purpose and users can share literally anything while navigating between pages.
>       If you want to use the navigation context as binding context, it's up to you as a user to assign it when navigation events are fired.
>       Example:
> 
> 
> ```js
> function onNavigatedTo(args) {
>   var page = args.object;
>   page.bindingContext = args.context;
> }
> ```

@CatchABus  Thanks for the clarification — here is the concise reason this fix is still relevant to #10912:

The issue in #10912 happens because the navigation context arrives too late in some navigation flows.
Even when the user does not manually use bindingContext, ProxyViewContainer still evaluates its children based on the state of the resolved Page at that moment.

When NavigationEntry.context isn’t forwarded early:

the Page receives undefined navigation state

ProxyViewContainer evaluates before that state is available

which produces the inconsistent hidden behavior shown in the issue

This PR makes the context consistently available before _onNavigatingTo(), so ProxyViewContainer has the correct state during its evaluation.
The patch doesn’t change any public behavior — it only ensures the navigation context is reliably present, which resolves the timing issue behind #10912.

vardhannaidux:
Hi @CatchABus — just following up.

I've applied all your feedback and clarified the reasoning behind the fix.  
To summarize the key point for #10912:

The issue is not about bindingContext usage by the user — it’s about the
navigation context arriving *too late* for ProxyViewContainer’s internal
evaluation.  
This PR ensures `NavigationEntry.context` is consistently available **before**
`_onNavigatingTo()`, which resolves the timing gap that causes `hidden` to be
evaluated incorrectly.

The change is fully backward-compatible, touches no public API, and only
stabilizes the navigation-state propagation logic.

If everything looks good, could you please review/approve the workflows so it
can be merged?  
I’d really appreciate it — I’m trying to get this fix landed today so I can
continue with related navigation work.

Thanks again for your guidance and time!

PR #10959, opened 2025-11-24
feat(vite): improved workspace and windows compatibility

NathanWalker:
- Improvements to project flavor handling, module resolution, and cross-platform compatibility in the Vite configuration and supporting utilities. 
- Polyfills are now resolved relative to the main entry directory, ensuring correct behavior in both monorepos and standalone projects, and import specifiers are normalized to POSIX style for consistency.


packages/core/ui/repeater/index.ts
```diff
@@ -15,6 +15,8 @@ export interface ItemsSource {
 	getItem(index: number): any;
 }
 
+export type RepeaterItemsLayout = View & Pick<LayoutBase, 'addChild' | 'removeChildren' | 'getChildrenCount' | 'getChildAt'>;
```
CatchABus: @NathanWalker Why not use `LayoutBase` directly here?

NathanWalker: Thank you @CatchABus, part of a refactor that was not needed here, it's now just LayoutBase 💯

packages/core/ui/core/bindable/index.ts
```diff
@@ -576,7 +576,8 @@ export class Binding {
 					indexAsInt--;
 				}
 			} else if (types.isString(index)) {
-				while (result && result.typeName !== index) {
```
CatchABus, line 579: What is the intention of this change? Something to do with Vite having trouble to parse keys of `$parents` references in XML?

NathanWalker: Good question, it makes string-based ancestor lookup compat with Vite runtime naming differences, where relying on typeName alone is brittle. Now, bindings like $parents['StackLayout'] can still resolve even if the runtime name is unstable, because the fallback uses cssType = stacklayout.

NathanWalker:
Closing to begin moving vite to dedicated workspace on it's own which will help split up the framework support into separate packages for better isolated concerns.

PR #11018, opened 2025-12-30
feat: bind overridden value converters to the target

triniwiz:
When a overridden property `valueConverter` is called it will now be bound to the target which can be useful for custom views detecting how to handle the incoming value

```ts
fontSizeProperty.overrideHandlers({
  name: 'fontSize',
  cssName: 'font-size',
  valueConverter: function (value) {
    if (this && this[isACustomView_]) {
      return value as never;
    }
    // fallback to core conversion
    return parseFloat(value as never);
  },
```

### DIFF
--- a/packages/core/ui/core/properties/index.ts
+++ b/packages/core/ui/core/properties/index.ts
@@ -654,7 +654,7 @@ export class CssProperty<T extends Style, U> {
 				delete this[sourceKey];
 			} else {
 				this[sourceKey] = ValueSource.Local;
-				value = valueConverter && typeof newValue === 'string' ? (overrideConverter ? valueConverter.call(view, newValue) : valueConverter(newValue)) : <U>newValue;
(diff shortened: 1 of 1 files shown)


packages/core/ui/core/properties/index.ts
```diff
@@ -200,6 +200,7 @@ export class Property<T extends ViewBase, U> implements TypedPropertyDescriptor<
 		let affectsLayout: boolean = options.affectsLayout;
 		let valueChanged = options.valueChanged;
 		let valueConverter = options.valueConverter;
+		let overrideConverter = false;
```
CatchABus, line 203: @triniwiz I noticed that `this` context is only applied for overridden props.
What about properties that are newly defined in a plugin but not in core? Do you think there could be such cases that also need `this` context?

triniwiz: It will work the same once they exports the property like core does. You never know someone like myself might pop up later on 😄

CatchABus (approved):

PR #11046, opened 2026-01-15
fix(vite): __dirname to work in windows

Tofandel:
a

## PR Checklist

In windows `__dirname` returns `/C:/Users...` which does not resolve properly

## What is the new behavior?

It now returns the correct dirname in windows

### DIFF
--- a/packages/vite/helpers/project.ts
+++ b/packages/vite/helpers/project.ts
@@ -8,7 +8,7 @@ export function getProjectRootPath(): string {
 }
 
 // Get current directory for ES modules (equivalent to __dirname)
-export const __dirname = typeof __dirname !== 'undefined' ? __dirname : import.meta.dirname;
+export const __dirname = import.meta.dirname;
 
 interface IPackageJson {
 	name?: string;


NathanWalker:
Thanks for this @Tofandel, were you able to confirm macOS still ok with dirname in that?

NathanWalker:
Some build errors here:
```
  Compiling TypeScript files for project "vite"...
  packages/vite/helpers/project.ts:11:33 - error TS2448: Block-scoped variable '__dirname' used before its declaration.
  
  11 export const __dirname = typeof __dirname !== 'undefined' ? __dirname : import.meta.dirname;
```
You can `npx nx build vite --skip-nx-cache` to confirm locally.

Tofandel:
Ah yeah typescript doesn't like that both the global and local __dirname are used, since vite doesn't have a CJS build, we can remove it anyways

Tofandel:
I don't have mac OS, but since it just uses an ESM variable instead, I see no reason why it wouldn't work on any platform. This is node 20+ only though; if we need to support less than node 20, then we should use

```js
import { dirname } from 'node:path';
import { fileURLToPath } from 'node:url';
    
export const __dirname = dirname(fileURLToPath(import.meta.url));
```

Tofandel:
Since vite 7 is already node 20.19+, I guess no need for that either

https://github.com/vitejs/vite/blob/main/packages/vite/package.json#L49

NathanWalker:
Nice thanks yes, v9 cli is node 22+ required so 👍

NathanWalker (approved):

PR #11091, opened 2026-02-18
feat: css filter

triniwiz:
Add support for [css filters](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/filter) e.g `drop-shadow(16px 16px 20px red) invert(75%);`

image

### DIFF
--- a/packages/core/platforms/ios/src/CSSFilters.swift
+++ b/packages/core/platforms/ios/src/CSSFilters.swift
@@ -1,0 +1,877 @@
+//
+//  CSSFilters.swift
+//  NativeScript
+//
+//  Created by Osei Fortune on 17/02/2026.
+//
+
+import UIKit
+import CoreImage
+import CoreImage.CIFilterBuiltins
+import MetalKit
+import MetalPerformanceShaders
+
+
+// MARK: - Color Map
+internal let colorMap: [String: UIColor] = [
+  "crimson": UIColor(red: 220/255, green: 20/255, blue: 60/255, alpha: 1),
+  "skyblue": UIColor(red: 135/255, green: 206/255, blue: 235/255, alpha: 1),
+  "black": .black,
+  "silver": UIColor(white: 0.75, alpha: 1),
+  "gray": .gray,
+  "grey": .gray,
+  "white": .white,
+  "maroon": UIColor(red: 0.5, green: 0, blue: 0, alpha: 1),
+  "red": .red,
+  "purple": UIColor(red: 0.5, green: 0, blue: 0.5, alpha: 1),
+  "fuchsia": .magenta,
+  "green": .green,
+  "lime": UIColor(red: 0, green: 1, blue: 0, alpha: 1),
+  "olive": UIColor(red: 0.5, green: 0.5, blue: 0, alpha: 1),
+  "yellow": .yellow,
+  "navy": UIColor(red: 0, green: 0, blue: 0.5, alpha: 1),
+  "blue": .blue,
+  "teal": UIColor(red: 0, green: 0.5, blue: 0.5, alpha: 1),
+  "aqua": .cyan,
+  "orange": .orange,
+  "brown": UIColor(red: 0.65, green: 0.16, blue: 0.16, alpha: 1),
+  "pink": UIColor.systemPink,
+  "transparent": .clear
+]
+
+
+public class CSSFilters: NSObject {
+  
+  enum Filter {
+    case blur(radius: CGFloat)
+    case brightness(value: CGFloat)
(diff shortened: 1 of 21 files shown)


packages/core/ui/text-base/index.android.ts
```diff
@@ -330,7 +330,7 @@ export class TextBase extends TextBaseCommon {
 				break;
 			case 'center':
 				if (supportsRtlTextAlign) {
-					this.nativeTextViewProtected.setTextAlignment(android.view.View.TEXT_ALIGNMENT_CENTER);
+					//	this.nativeTextViewProtected.setTextAlignment(android.view.View.TEXT_ALIGNMENT_CENTER);
```
farfromrefug: why did you comment this?

triniwiz: Nice catch, I was testing something early

PixsaOJ:
Omg! BIG!

packages/core/ui/core/view/index.android.ts
```diff
@@ -1605,6 +1605,18 @@ export class View extends ViewCommon {
 		}
 	}
 
+	[filterProperty.getDefault](): string {
+		return '';
+	}
+	[filterProperty.setNative](value: string) {
+		const nativeView = this.nativeViewProtected;
+		//@ts-ignore
```
CatchABus: @triniwiz Maybe we can do something else other than adding 2 consecutive `ts-ignore` annotations here.
What if we use a condition like instanceof or `typeof nativeView['setFilter'] === 'function' and remove ts-ignores?

triniwiz: I'll switch the check and remove the ignores

packages/core/platforms/ios/src/CSSFilters.swift
```diff
@@ -0,0 +1,877 @@
+//
+//  CSSFilters.swift
+//  NativeScript
+//
+//  Created by Osei Fortune on 17/02/2026.
+//
```
NathanWalker: We could put this in widgets to be included in the .xcframework, that way it's prebuilt and doesn't imply an extra build source on end users.

NathanWalker: I'll reject the PR to keep this reminder and we can work together to do this part.

NathanWalker (changes requested):

packages/core/ui/core/view/index.android.ts
```diff
@@ -1605,6 +1605,19 @@ export class View extends ViewCommon {
 		}
 	}
 
+	[filterProperty.getDefault](): string {
+		return '';
+	}
+	[filterProperty.setNative](value: string) {
+		// LayoutBase, ImageView and StyleableTextView has the setFilter method implemented in the native view, so we need to check before call it.
+		const nativeView: globalAndroid.view.View & {
```
CatchABus, line 1613: @triniwiz Instead of creating a union type here, I have the following thought.
Let's create an interface in ui-mobile-base/widgets (e.g. ICSSFilter.java) with the needed methods and let ImageView, LayoutBase, and StyleableTextView implement it.
This way, we'll have an interface as a type for Java-side and a type in types file to use in TS-side.

packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/CSSFilters.kt
```diff
@@ -0,0 +1,1176 @@
...
+			return ColorMatrix(result)
+		}
+
+		private fun saturate(percent: Float): ColorMatrix {
+			val cm = ColorMatrix()
+			cm.setSaturation(percent)
+			return cm
+		}
+
+		internal fun parseCssColor(value: String): Int {
```
CatchABus, line 993: @triniwiz @farfromrefug I'll just leave my thought on this one for both android and iOS.
This PR adds maps of colors for each platform and adds color parsing methods on native side.
I understand it all stems from the fact that it's cheaper to pass the raw string from core to native but is having duplicate functionality worth over let's say being afraid of passing more complex parameters to JNI calls?
I was thinking core could at least parse the color values alone since we already have the color module that already does that much.
So, filter property could have a `valueConverter` for converting the color values only and pass everything as parameters to native methods afterwards where everything else can be parsed.

triniwiz: Atm this follows the web of taking a string input then building the filter from said input. Nothing wrong with keeping this in swift/objc or kotlin/java

PR #11098, opened 2026-02-20
feat: android widgets

triniwiz:
feat: RemoteView IPC views


farfromrefug:
this should be in the core. the api is very limited and wont fit all use cases.
this should be in a plugin where you can take time to define a shared and nice api for ios/android.
also it seems the best approach to write android widgets is now glance and not remote views which are very limited

triniwiz:
Glance has the same restrictions @farfromrefug .

> Key Point: Glance provides a modern approach to build app widgets using Compose, but is restricted by the limitations of AppWidgets and RemoteViews. Therefore, Glance uses different composables from the Jetpack Compose UI.
[Create an app widget with Glance](https://developer.android.com/develop/ui/compose/glance/create-app-widget)

farfromrefug:
@triniwiz Yes it does but glance APIs are moving really fast and allow much more complex UIs using compose. 
But the main point remains, this should not be in core, it should be a plugin

triniwiz:
You know it's still limited to the same `RemoteViews` API ? I've created a similar reusable wrapper to make building layouts easier also with `Glance` you'll need compose

farfromrefug:
@triniwiz Well not for the UI design part. RemoteViews layout is using xml which makes things very limited, especially when working with theme/size ...
But again the main point is not about glance/RemoteViews. The main point is that it should not be in core (which is bloated enough right now). It should be a plugin which can also allow the API to mature much faster  ( easier to make breaking changes in a beta plugin) to come to something which fits everyone's needs.

NathanWalker:
Agree on plugin for this part (ios didn't need core changes for example to enable). We could introduce @nativescript/widgets which can help bring together all useful widget work.

PR #11129, opened 2026-03-04
fix(ios): secureWithoutAutofill on TextField

NathanWalker:
- improves `secureWithoutAutofill` on`TextField` with iOS to ensure it applies at the right time to prevent the autofill handling

### DIFF
--- a/packages/core/ui/text-field/index.ios.ts
+++ b/packages/core/ui/text-field/index.ios.ts
@@ -13,7 +13,6 @@ class UITextFieldDelegateImpl extends NSObject implements UITextFieldDelegate {
 	public static ObjCProtocols = [UITextFieldDelegate];
 
 	private _owner: WeakRef<TextField>;
-	private firstEdit: boolean;
 
 	public static initWithOwner(owner: WeakRef<TextField>): UITextFieldDelegateImpl {
 		const delegate = <UITextFieldDelegateImpl>UITextFieldDelegateImpl.new();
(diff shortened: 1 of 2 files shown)


packages/core/ui/text-field/index.ios.ts
```diff
@@ -243,6 +237,44 @@ export class TextField extends TextFieldBase {
...
 		this.nativeTextViewProtected.secureTextEntry = value;
+		this._applySecureWithoutAutofillTraits(this.nativeTextViewProtected);
+	}
+
+	private _applySecureWithoutAutofillTraits(textField: UITextField): void {
+		if (!textField || !this.secureWithoutAutofill) {
+			return;
+		}
+
+		const nativeField = textField as any;
```
CatchABus, line 248: Is cast to `any` necessary here? I understand these props are available in UITextField.

packages/core/ui/text-field/index.ios.ts
```diff
@@ -174,14 +175,7 @@ export class TextField extends TextFieldBase {
...
 	public textFieldShouldChangeCharactersInRangeReplacementString(textField: UITextField, range: NSRange, replacementString: string): boolean {
-		if (this.secureWithoutAutofill && !textField.secureTextEntry) {
-			/**
-			 * Helps avoid iOS 12+ autofill strong password suggestion prompt
-			 * Discussed in several circles but for example:
-			 * https://github.com/expo/expo/issues/2571#issuecomment-473347380
-			 */
-			textField.secureTextEntry = true;
-		}
+		this._applySecureWithoutAutofillTraits(textField);
```
CatchABus, line 178: Maybe we need an extra flag to determine if the native props have already been updated and reduce `_applySecureWithoutAutofillTraits` calls since there are cases like `reloadInputViews()` that might be called multiple times while typing.
For instance, the old fix made sure to check whether `secureTextEntry` was already enabled or not before proceeding.

packages/core/ui/text-field/index.ios.ts
NathanWalker: Some properties are sdk level scoped so we can probably rather use SDK_VERSION scope on some.

PR #11131, opened 2026-03-06
perf(TextField): optimize secureWithoutAutofill

NathanWalker:
- ensure input view properties are only updated when necessary, and to reload input views only when changes are made

### DIFF
--- a/packages/core/ui/text-field/index.ios.ts
+++ b/packages/core/ui/text-field/index.ios.ts
@@ -252,49 +252,34 @@ export class TextField extends TextFieldBase {
 			shouldReloadInputViews = true;
 		}
 
-		if (textField.textContentType !== undefined) {
-			const desiredTextContentType = typeof UITextContentTypeOneTimeCode !== 'undefined' ? UITextContentTypeOneTimeCode : '';
-			if (textField.textContentType !== desiredTextContentType) {
-				textField.textContentType = desiredTextContentType;
-				shouldReloadInputViews = true;
-			}
+		if (textField.textContentType !== undefined && textField.textContentType !== UITextContentTypeOneTimeCode) {
+			textField.textContentType = UITextContentTypeOneTimeCode;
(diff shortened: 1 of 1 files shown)


packages/core/ui/text-field/index.ios.ts
```diff
@@ -245,35 +245,60 @@ export class TextField extends TextFieldBase {
...
+		if (!textField.secureTextEntry) {
+			textField.secureTextEntry = true;
+			shouldReloadInputViews = true;
+		}
 
-		if (nativeField.textContentType !== undefined) {
-			nativeField.textContentType = typeof UITextContentTypeOneTimeCode !== 'undefined' ? UITextContentTypeOneTimeCode : '';
+		if (textField.textContentType !== undefined) {
+			const desiredTextContentType = typeof UITextContentTypeOneTimeCode !== 'undefined' ? UITextContentTypeOneTimeCode : '';
+			if (textField.textContentType !== desiredTextContentType) {
```
farfromrefug: all those tests can be simplified:
`typeof UITextContentTypeOneTimeCode !== 'undefined' ? UITextContentTypeOneTimeCode : ''` to `UITextContentTypeOneTimeCode ?? ''`
And everytime both if can be done into one with something like `if (textField.textContentType !== undefined && textField.textContentType !== desiredTextContentType)`
All that making the code more readable

Also i dont understand the `desiredTextContentType` defaulting to `''`. Shouldnt it default to `undefined`. which would also simplify the if test (no need to compare with undefined`

All that would make the code far less verbose and much easier to understand/maintain

CatchABus: @NathanWalker I agree with @farfromrefug, this can be simplified a bit.

NathanWalker: @farfromrefug 💯 , pushed up simplification, much better. the textContentType empty string is not even needed. /cc @CatchABus

PR #11134, opened 2026-03-11
fix(ios): liquid glass flexbox

### LINKED ISSUE #11099: LiquidGlass in FlexboxLayout broken
### Issue Description

Hello together,
I have tried to use a LiquidGlass component in a FlexboutLayout. However, unfortunately this breaks the layout.

### Reproduction

_No response_

```shell

```

### Environment

"@nativescript/core": "9.0.15",

### Please accept these terms

- [x] This is a bug report

VeinDevTtv:
## PR Checklist

* [x] The PR title follows our guidelines: [https://github.com/NativeScript/NativeScript/blob/master/tools/notes/CONTRIBUTING.md#commit-messages](https://github.com/NativeScript/NativeScript/blob/master/tools/notes/CONTRIBUTING.md#commit-messages).
* [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
* [x] You have signed the [CLA](http://www.nativescript.org/cla).
* [x] All existing tests are passing: [https://github.com/NativeScript/NativeScript/blob/master/tools/notes/DevelopmentWorkflow.md#running-unit-tests-application](https://github.com/NativeScript/NativeScript/blob/master/tools/notes/DevelopmentWorkflow.md#running-unit-tests-application).
* [x] Tests for the changes are included - [https://github.com/NativeScript/NativeScript/blob/master/tools/notes/WritingUnitTests.md](https://github.com/NativeScript/NativeScript/blob/master/tools/notes/WritingUnitTests.md).

Currently on iOS, placing a `LiquidGlass` or `LiquidGlassContainer` component inside a `FlexboxLayout` completely breaks the layout (children are measured incorrectly and positioned at the wrong coordinates). This is because LiquidGlass overrides `_addViewToNativeVisualTree` to use a nested `_contentHost` view, which mismatches FlexboxLayout's standard measurement and layout paradigms.

## What is the new behavior?

1. **Measurement fix**: Overrode `onMeasure` in both `LiquidGlass` and `LiquidGlassContainer` to temporarily zero out padding/border properties before delegating to `super.onMeasure()`. This prevents `GridLayout` / `AbsoluteLayout` from double-subtracting padding that `FlexboxLayout` already removed from the child measure spec.

2. **Layout fix**: Overrode `onLayout` in both classes to call `super.onLayout(0, 0, right - left, bottom - top)`. Since `FlexboxLayout` positions the root `UIVisualEffectView` using absolute coordinates, the internal NativeScript layout engine needs normalized `(0, 0, width, height)` bounds to correctly position children within the local `_contentHost` coordinate space.

Fixes #11099.

### DIFF
--- a/apps/toolbox/src/pages/glass-effects.xml
+++ b/apps/toolbox/src/pages/glass-effects.xml
@@ -15,13 +15,13 @@
 
                     <Button row="2" text="Toggle Glass" tap="{{toggleGlassEffect}}" horizontalAlignment="center" verticalAlignment="middle" class="c-white font-weight-bold m-y-20 p-4" fontSize="22" borderRadius="32" width="300" height="100" touchAnimation="{{touchAnimation}}" iosGlassEffect="{{currentEffect}}"/>
 
-                    <LiquidGlass row="3" width="300" height="100" borderRadius="32" iosGlassEffect="{{iosGlassEffectInteractive}}" >
+                    <LiquidGlass row="3" width="300" height="100" borderRadius="32" iosGlassEffect="{{iosGlassEffectInteractive}}">
 
                         <Label text="Glass Interactive" fontSize="22" class="font-weight-bold text-center c-white" />
 
                     </LiquidGlass>
 
-                    
+
                     <LiquidGlassContainer row="4" tap="{{toggleMergeGlass}}" horizontalAlignment="left" verticalAlignment="middle" class="m-t-20" width="300" height="100" columns="*">
                         <LiquidGlass id="glass1" loaded="{{loadedGlass}}" borderRadius="50" width="100" height="100">
                             <Label id="share" text="Share" fontSize="22" class="font-weight-bold text-center" width="100" height="100" loaded="{{loadedGlassLabels}}" />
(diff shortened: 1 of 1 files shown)


packages/core/ui/label/index.ios.ts
```diff
@@ -52,12 +53,14 @@ export class Label extends TextBase implements LabelDefinition {
 	}
 
 	_requestLayoutOnTextChanged(): void {
-		if (this._fixedSize === FixedSize.BOTH) {
-			return;
-		}
-		if (this._fixedSize === FixedSize.WIDTH && !this.textWrap && this.getMeasuredHeight() > 0) {
-			// Single line label with fixed width will skip request layout on text change.
-			return;
+		if (SDK_VERSION < 18) {
```
CatchABus: @VeinDevTtv How does this change affect liquid glass flex?
Because if you did that to make the failing test pass, it's irrelevant and not very desirable.
We should track the real cause of it failing and why the check is there to begin with.

packages/core/utils/constants.ios.ts
```diff
@@ -1,4 +1,9 @@
 export const SDK_VERSION = parseFloat(UIDevice.currentDevice.systemVersion);
+
+function hasRuntimeGlobal(name: 'UIGlassEffect' | 'UIGlassContainerEffect' | 'UIGlassEffectStyle'): boolean {
+	return typeof (globalThis as any)?.[name] !== 'undefined';
```
NathanWalker: Better to just rely on the SDK_VERSION check alone. runtime global checks like this shouldn't be needed when the SDK level is the best defining line. In other words, the `supportsGlass` check should be sufficient alone without this.

NathanWalker (changes requested):

NathanWalker:
Thank you for the adjustments @VeinDevTtv - one last small thing; is it possible to add a quick FlexboxLayout example in https://github.com/NativeScript/NativeScript/blob/main/apps/toolbox/src/pages/glass-effects.xml so it's easily seen in toolbox for this case now and in future?

PR #11136, opened 2026-03-12
fix(transition): prevent circular refs on TransitionListener

edusperoni:
Needs to be tested with the NAPI runtimes


farfromrefug (approved):

NathanWalker:
@edusperoni this looks ready if wanna mark out of draft? Will let next tag receive it for better looks in napi.

PR #11145, opened 2026-03-24
fix(android): reapply tintColor when Image src changes

owenthcarey:
On Android, ColorFilter is applied to the Drawable, not the ImageView itself. When src changes dynamically (e.g. via binding), the new drawable replaces the old one and the tint is lost. This differs from iOS, where tintColor is a view-level property that persists across image changes.

This adds `_reapplyTintColor()` to the Android Image implementation and calls it from both code paths that load a new image:

- `onImageLoaded` -- for images loaded natively via setUri (file/resource paths)
- `[imageSourceProperty.setNative]` -- for images set via `setImageBitmap` (font icons, data URIs, ImageSource/ImageAsset)

### DIFF
--- a/apps/toolbox/src/pages/image-handling.ts
+++ b/apps/toolbox/src/pages/image-handling.ts
@@ -10,10 +10,15 @@ export function navigatingTo(args: EventData) {
 
 export class DemoModel extends Observable {
 	addingPhoto = false;
+	tintTestSrc = 'res://icon';
 	symbolWiggleEffect = ImageSymbolEffects.Wiggle;
 	symbolBounceEffect = ImageSymbolEffects.Bounce;
 	symbolBreathEffect = ImageSymbolEffects.Breathe;
 	symbolRotateEffect = ImageSymbolEffects.Rotate;
+
+	toggleTintTestSrc() {
+		this.set('tintTestSrc', this.tintTestSrc === 'res://icon' ? 'res://add_to_fav' : 'res://icon');
+	}
 
 	pickImage() {
 		const context = create({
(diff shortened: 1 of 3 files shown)


CatchABus:
@owenthcarey Thanks for the PR!

I tried to replicate your problem but I couldn't.
This is a video of switching between 2 bitmaps in 3 seconds using a timer while setting both `src` and `imageSource` properties.
I'm using core v9 and android 16.

The tint-color is applied using css:
```css
#img {
  tint-color: red;
}
```


https://github.com/user-attachments/assets/0892c342-d5c5-4a04-94f9-07d11630e927


Also, regarding the following statement:
```
ColorFilter is applied to the Drawable, not the ImageView itself.
```
The ImageView class handles the color filter internally, so it should normally be persisted between bitmap changes.

Could you please create an issue related to the problem and post an app sample or a recording demonstrating the problem from your device? Maybe it could be a corner case we're unaware of.

CatchABus (commented):
Awaiting response from the author.

NathanWalker:
I believe this case is with dynamic bindings, not css @CatchABus - if the toolbox image-handling is run with the index.android changes commented out the issue shows itself.
<https://github.com/NativeScript/NativeScript/pull/11145/changes#diff-4a3d33d5aaebbf4b40756ca73c95fe2980910999c65e66a33334c40715c460b6R12>

CatchABus:
> I believe this case is with dynamic bindings, not css @CatchABus - if the toolbox image-handling is run with the index.android changes commented out the issue shows itself. https://github.com/NativeScript/NativeScript/pull/11145/changes#diff-4a3d33d5aaebbf4b40756ca73c95fe2980910999c65e66a33334c40715c460b6R12

Thanks for pointing this out @NathanWalker. I'll check it out and continue with the review.

NathanWalker:
This appears to be unique to Angular binding case, not directly represented in core (xml). We'll leave open until more clarity emerges.

PR #11181, opened 2026-04-13
feat: add cross-platform NativeWindow implementation

edusperoni:
## PR Checklist

---

## The concept

1. **Application** — the full context. This is where JS lives. This is the browser "window".
2. **NativeWindow** — a browser tab / OS window. Multiple windows live in the same JS context.

The developer chooses how to handle this. From an Angular background:

1. Every window is a fully new app `bootstrapApplication(...)` — "root" services are per-window, "platform" services are shared.
2. A single application is bootstrapped and every window is a different `createComponent` — "root" services are shared, navigation is similar to modal navigation.

Motivations: iOS 27 makes UIScene mandatory; Android's `singleTask` requirement exists only because the framework binds everything to one root activity; CarPlay/Android Auto need a home; Windows runtimes are coming.

`NativeWindow` existed only as a draft POC. Window content was still owned by the platform application classes, windows died whenever their native surface went away, the window registry was duplicated per platform, and `launch` conflated "the process started" with "build the first window's UI" — which cannot work for a second window or a background launch.

## What is the new behavior?

The POC is now a complete implementation, landed as 14 individually-typechecking commits.

- **`WindowBase` / `NativeWindow` split.** `WindowBase` carries identity, role, state, lifecycle and native accessors; `NativeWindow` adds the view-carrying half. Surfaces with no NativeScript view tree (CarPlay, external displays) can extend `WindowBase` later. Roles: `application | embedded | carplay | externalDisplay`.
- **Session-scoped identity.** A window survives its native surface going away. iOS keys off `UISceneSession.persistentIdentifier` (and handles `didDiscardSceneSessions`), Android mints a UUID persisted in `savedInstanceState`. A scene disconnect or activity recreation raises `detached` — the window stays registered and keeps its listeners — while a real close raises `close` exactly once and then clears the instance's listeners so apps don't leak handlers.
- **Ownership inversion.** All content flows through `window.setContent()`. Every app registers at least one window: scene, non-scene, and embedded alike. App-level root state (`getRootView()`, the global root view, `initRootView`) now mirrors the primary window instead of owning it.
- **`ready` + content resolver.** `ready` fires exactly once per JS context, is never deferred (including background launches), and always precedes the first `windowOpen`. `Application.setWindowContentResolver()` supplies each window's UI on demand. `launch` keeps working as a bridge for the first view-carrying window, including its `root` three-state contract.
- **Per-window configuration.** Orientation, system appearance and layout direction are read live per window, raise per-window changed events, and scope their CSS classes to that window's root view and its modals rather than the process-wide system class list.
- **Shared registry.** `primaryWindow`, `getWindows(role?)`, `getWindowById()`, `openWindow(options?)` and the window events live on `ApplicationCommon`, with protected hooks for iOS-specific bookkeeping.
- **Android `openWindow` is implemented for real** (new activity via intent) and flagged experimental.
- **Public typings** for `native-window` following the `ui/frame` hand-written `index.d.ts` convention, keeping internals and the concrete platform classes out of the published surface.

Test suite: **212 → 287** tests across 19 files.

### Bugs fixed along the way

- **Android `exit` could be lost entirely.** The last-window check initially went where the old code lived, in a callback that only runs for NativeScript activities — but windows are registered for *every* activity, so a third-party SDK activity outliving the main one meant `exit` never fired.
- **`app.css` would never load for resolver-only apps.** `loadAppCss()` is reachable from exactly one place, inside `notifyLaunch()`, which the resolver chain skips.
- **`window.setContent(view)` discarded its argument on Android** and rebuilt from the main entry.
- **`_detach()` disposed native views without `callUnloaded`**, so `unloaded` handlers never ran.
- **`classNameProperty` dropped window-scoped CSS classes** — it clears `cssClasses` and reseeds from the global list, so a `className` change on a root or modal-root view lost `ns-dark`/`ns-landscape`.
- **Rotating one window restyled another window's modals** (`_getRootModalViews()` is process-wide).
- **`Application.ios.on('ready' | 'suspend' | …) did not typecheck`** — both platform classes declared `on()` as method overloads, replacing `ApplicationCommon`'s handler surface. Seven types in `application.d.ts` were referenced but never imported and silently resolved to `any`.
- **A second `launch` dispatch would tear down every `launch` listener** (`events.off('launch', null)`).
- `autoSystemAppearanceChanged` was bypassed by the per-window CSS path; accessibility CSS init was one-shot so windows 2+ never got a11y/font-scale classes; `close` fired twice per teardown; `_getWindows()` leaked the live internal array; the toolbox `Info.plist` was malformed XML (failed `plutil -lint`).

### Known limitations

- A detached Android window that never re-attaches keeps the registry non-empty and suppresses `exit` indefinitely. Suppression is correct for the normal recreation case and there is no clean discriminator.
- `NativeWindow` repeats `WindowBase`'s whole `on()` overload set — TypeScript requires a derived overload set to cover every base overload, and a catch-all cannot stand in for the `SceneEventData`/`AndroidActivity*` shapes.
- `displayed` fires on the primary window only; secondaries deferred.
- `NotificationObserver` is unexported, so it still resolves to `any` in the published typings.

### Still to verify on device

iOS: scene reconnect/restore; `didDiscardSceneSessions` delivered on a later launch; `close()` denial clearing `_closeRequested`; suspend/resume ordering across iOS 15–27 with two scenes; background launch (silent push) content deferral; visionOS; embedded hosts; Vite HMR dev boot.

Android: `openWindow` against `singleTask`/`singleInstance` manifests and OEM recents; `putExtra` overload selection for wrapped primitives (resolved by the runtime, not by tsc); per-activity `ComponentCallbacks2` register/unregister across rotation; window-id round trip after process death.

Both: CSS system-class consumers — modals, dark mode, rotation — since window-scoped classes no longer live in the global list.

---

## BREAKING CHANGES

| Change | Migration |
| --- | --- |
| `NativeWindow.iosWindow` → `.ios`, and its `.window` → `.uiWindow` | `win.ios?.uiWindow`, `win.ios?.scene` |
| `NativeWindow.androidWindow` → `.android` | `win.android?.activity` |
| `SceneEventData.window` is now the `NativeWindow`; the native `UIWindow` moved to `uiWindow` | A `window` payload key always means a `NativeWindow` |
| `getWindows()` is role-filtered, defaulting to `application` + `embedded` | `getWindows('all')` for every registered surface |
| Android `exit` fires only when the **last** window finishes | Previously fired for any finishing activity. iOS `exit` is unchanged (process termination) |
| App `suspend`/`resume` in scene mode reflect **whole-app** state | Use a window's `background`/`foreground` for a single window |
| Scene disconnect / activity recreation raise `detached` instead of destroying the window | The window stays registered and its listeners survive |
| A window clears its listeners after `close` | Do not re-use a closed window instance |
| `CSSUtils.getRootViewCssClasses()` no longer contains orientation/appearance/direction classes | These are per-window; read them from the window |

**Deprecated but still working:** `launch` (use `ready` + `setWindowContentResolver()`), `shouldDelayLaunchEvent` (now a no-op), `Application.orientation()`/`systemAppearance()`/`layoutDirection()` (delegate to `primaryWindow`), and the enumeration getters (`getAllWindows`/`getAllScenes`/`getWindowScenes`/`getPrimaryWindow`/`getPrimaryScene`).

**Un-deprecated:** the `activity*`/`scene*` bridges on `Application`. They are permanent aggregate APIs — they fire for every window, and `args.window` identifies which one.


NathanWalker:
Great start here, it might be helpful to setup a new demo page in toolbox or repurpose `multiple-scenes.ts` to exercise the approach on both platforms?

packages/core/native-window/native-window-common.ts
```diff
@@ -0,0 +1,339 @@
...
+		return undefined;
+	}
+
+	get androidWindow(): { readonly activity: androidx.appcompat.app.AppCompatActivity } | undefined {
+		return undefined;
+	}
+
+	// --- Typed event overloads ---
+
+	on(event: 'activate', callback: (data: NativeWindowEventData) => void, thisArg?: any): void;
```
farfromrefug, line 154: those should go into a native-window.d.ts file (which is missing by the way)

packages/core/application/application.d.ts
```diff
@@ -266,6 +357,68 @@ export class iOSApplication extends ApplicationCommon {
...
+	 *     return config;
+	 *   }
+	 *   return null;
+	 * };
+	 * ```
+	 */
+	onSceneConfiguration: ((application: UIApplication, connectingSceneSession: UISceneSession, options: UISceneConnectionOptions) => UISceneConfiguration | null | undefined) | null;
+
+	/**
+	 * @deprecated Listen on a NativeWindow instance instead.
```
farfromrefug, line 385: i dont think we should mark those as deprecated. There is a lot of power in having Application events.
It is the only place where we can have all windows event from one single listener.
same for orientation and all other events.
I think it is great to have them both in NativeWindow and Application.
plus we ll never be able to remove them. Application events are deeply used in all existing apps and plugins.

packages/core/application/application.d.ts
```diff
@@ -136,17 +171,68 @@ export class AndroidApplication extends ApplicationCommon {
 	 */
 	getRegisteredBroadcastReceivers(intentFilter: string): android.content.BroadcastReceiver[];
 
+	/**
+	 * @deprecated Listen on a NativeWindow instance instead.
```
farfromrefug, line 175: Same thing i would not deprecate those, only explain you can do it through NativeWindw

packages/core/native-window/native-window.android.ts
farfromrefug, line 1: should the android NativeWindow listen to configuration change, just like we do in application?
It should store it s configuration, font scale, orientation... use them for accessors but also notify the NativeWindow about them. Just like we do in Application.
maybe we can even have Application plugged to use accessors, events from the main window?
Not sure about this because in application we use context and not activity i think, not sure what difference it makes for configurations like orientation, font scale ...

packages/core/ui/frame/index.android.ts
```diff
@@ -871,12 +882,24 @@ export class ActivityCallbacksImplementation implements AndroidActivityCallbacks
...
+			// Emit on NativeWindow first
+			const nativeWindow = Application.android._getWindowForActivity(activity);
+			if (nativeWindow) {
+				nativeWindow.notify({
+					...newIntentArgs,
+					object: nativeWindow,
+				} as AndroidActivityNewIntentEventData);
+			}
+
+			// @deprecated - Bridge to Application.android for backward compat
```
farfromrefug, line 901: i wouldnt deprecate here again. It is really nice to have that

CatchABus:
Other than @farfromrefug points that need to be considered, it looks good to me and will definitely bring lots of new possibilities in core.

NathanWalker (approved):

PR #11212, opened 2026-05-07
feat(android): modern Material 3 bottom navigation for TabView

NathanWalker:
- `androidTabsPosition="bottom"` now renders via `com.google.android.material.bottomnavigation.BottomNavigationView` with the M3 active-indicator pill and an 80dp layout, replacing the bottom-pinned TabLayout. 
- androidTabsPosition="top" is unchanged
- This makes working with `TabView` across iOS and Android provide the best of class result on either and consistent experience (with regards to orienting bottom)

https://github.com/user-attachments/assets/b2c48293-889b-4673-a508-cc1565003de0

### DIFF
--- a/apps/toolbox/src/pages/tabview.ts
+++ b/apps/toolbox/src/pages/tabview.ts
@@ -32,6 +32,13 @@ class TabViewDemoModel extends Observable {
 		this.applyFontFamily('ns-playground-font');
 	};
 
+	useResIcons = () => {
+		if (!this.tabView || !this.tabView.items) return;
+		if (!__ANDROID__) return;
+		const resIcons = ['res://outline_home_24', 'res://outline_list_alt_24', 'res://outline_info_24'];
+		this.setIcons(resIcons);
+	};
+
 	clearIcons = () => {
 		if (!this.tabView || !this.tabView.items) return;
 		this.tabView.items.forEach((item) => {
--- a/apps/toolbox/src/pages/tabview.xml
+++ b/apps/toolbox/src/pages/tabview.xml
@@ -4,24 +4,26 @@
     </Page.actionBar>
 
 
-    <TabView id="demoTabView" selectedIndex="0" androidTabsPosition="top" class="m-t-10">
+    <TabView id="demoTabView" selectedIndex="0" androidTabsPosition="bottom" class="m-t-10">
         <!-- First tab -->
         <TabViewItem title="First" iconSource="font://&#x41;">
             <StackLayout padding="20">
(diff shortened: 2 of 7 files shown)


NathanWalker:
This does not supersede or replace community's material components, this just provides the baseline to give best of class on both platforms out of the box since TabView already uses modern iOS API's with glass and on bottom - this now provides consistent expectations to both.

CatchABus:
I think this is a bit sudden and doesn't feel very practical to add a heavy dependency like material just for TabView.
@NathanWalker Core simulated the look 'n feel at a certain point in v6 (BottomNavigation and Tabs?) and there are still some leftover java classes in `ui-mobile-base` we could probably reuse.

https://github.com/NativeScript/NativeScript/blob/main/packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/BottomNavigationBar.java
https://github.com/NativeScript/NativeScript/blob/main/packages/ui-mobile-base/android/widgets/src/main/java/org/nativescript/widgets/TabsBar.java

This link is from BottomNavigation back then:
https://github.com/NativeScript/NativeScript/blob/979130de313e68ced9ad1f60b65ef6050bcfe983/nativescript-core/ui/bottom-navigation/bottom-navigation.android.ts#L284

NathanWalker:
- if can provide alternate way for TabView to provide consistent experience other route, feel free to try in alternate PR
- TabView provides glass tabs (best of class) on iOS 
- TabView on Android does not match that experience (I'd argue right now it provides worst in class experience)
- the dependency case is a concern because it brings it in for anyone using core (whether they use TabView or not)

NathanWalker:
I believe only other alternate view is to just remove all ui components from core and make them all ad hoc pick your own one at a time for each project which there’s a good case for that if clear and easy enough to browse best in class options altogether in official docs. Plan bigger move like that for v10 or v11.

farfromrefug:
I agree 100% with @CatchABus and it was my first  thought when I saw the PR.

* you cant compare IOS design vs material on android. material comes as a dependency and not included In android by default. I think there are multiple reasons for that, one being faster Releasing cycle. However if we follow the basics of core following the platform and other features coming as plugins it should not be in core.  And I think it should NOT be in core. Material.components are here exactly for that. And we can still create templates with them added by default
* that dependency is huge and with the things we keep on adding N is becoming a fat framework which it should not. I always said that (I think the same with the runtimes). And I even think I am the one which had it removed in N 6 or something. and remember N still does not support correct tree shaking and proguard yet without tricky integration like I do. If we did things would be a bit different even though still mean longer build for no reason for everyone not using it.

I do believe this PR should not get in.

NathanWalker:
Let's close this, we'll continue chats in discord on the ui decoupling.

PR #11214, opened 2026-05-12
fix(core): rebuild nested-frame fragment when its view was destroyed on reattach

dangrima90:
## PR Checklist

As described on [discord](https://discord.com/channels/603595811204366337/1343512351529242645/1501909266347462716) I have encountered an issue where a nested frame results in having no content. As stated I have investigated this via claude, not mentioning it to not take any responsibility 😄 but to highlight that code changes and comments in the code were added by claude. I did direct it to simplify certain comments.

As I mentioned on discord I did try replicating this issue in a clean application but I didn't manage. My suspicion was that there's was minor issue in NativeScript that's being triggered by as specific use case in my application.

## What is the new behavior?

Short description via claude of this update:

The update here extends the `_ensureEntryFragment` recovery from #10713 to cover the case where the fragment object survives but its view has been destroyed. Without this, the nested frame renders blank after the parent is briefly covered and dismissed. Also guards entry.fragment = null in onDestroy so the old fragment's teardown doesn't overwrite the new reference assigned by the rebuild path.

Two changes under `packages/core/ui/frame`:

1. **`index.android.ts`** — `_ensureEntryFragment` also rebuilds when `fragment.getView() == null`, gated by a new `_pendingFragmentCheck` flag set by `onLoaded()`. The flag is needed because `_onAttachedToWindow` also calls `_ensureEntryFragment` during outgoing transitions, where a null view is expected and a rebuild would be wrong.

2. **`frame-helper-for-android.ts`** — `onDestroy` only nulls out `entry.fragment` if it still points to the destroying fragment. The rebuild above assigns a new fragment to `entry.fragment`, and without this guard the old fragment's destruction would overwrite that new reference.

### DIFF
--- a/packages/core/ui/frame/frame-helper-for-android.ts
+++ b/packages/core/ui/frame/frame-helper-for-android.ts
@@ -236,8 +236,15 @@ export class FragmentCallbacksImplementation implements AndroidFragmentCallbacks
 		// [nested frames / fragments] see https://github.com/NativeScript/NativeScript/issues/6629
 		// retaining reference to a destroyed fragment here somehow causes a cryptic
 		// "IllegalStateException: Failure saving state: active fragment has cleared index: -1"
-		// in a specific mixed parent / nested frame navigation scenario
-		entry.fragment = null;
+		// in a specific mixed parent / nested frame navigation scenario.
+		//
+		// Only null out entry.fragment if it still points to THIS fragment. _ensureEntryFragment
+		// may have already replaced it with a new fragment (when recovering one whose view was
+		// destroyed) — if THIS old fragment is destroyed afterwards, we must not overwrite the
+		// newer reference.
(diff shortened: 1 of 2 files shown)


CatchABus:
I paste the video indicating the problem here:
https://github.com/user-attachments/assets/d627e4e2-2592-4cef-b657-58523d13a56f

NathanWalker:
@coderabbitai review

farfromrefug (commented):
It seems fine to me. I am not sure about code rabbit proposed changr though. I am having a hard time wrapping my head around the flow

CatchABus:
> It seems fine to me. I am not sure about code rabbit proposed changr though. I am having a hard time wrapping my head around the flow

The author and I are doing a few last tests to see if an old PR has caused the regression.
If we don't find anything else, I'll also leave my approval here.

PR #11216, opened 2026-05-16
perf(core): reduce padding native setter calls

### LINKED ISSUE #8335: [android]padding property should not be a shorthand
Right now the `padding` is a shorthand for all the 4 padding properties.
Not an issue JS wise but it becomes one on Android.
Screen Shot 2020-02-13 at 09 53 14
On android it means 4 native calls which takes 5ms on my phone.
It is not necessary as it could be done in [1 call](https://developer.android.com/reference/android/view/View.html#setPadding(int,%20int,%20int,%20int))

More over it is actually done [twice](https://github.com/NativeScript/NativeScript/blob/e8f5ac8522954b4433fa8932032d9fc03d33819e/nativescript-core/ui/core/view/view.android.ts#L1003) when there is a background set! Why is the padding set with backgroundInternal?

CatchABus:
## PR Checklist

Right now, core calls padding native setter method (e.g. setPadding) for each shorthand property.
In a few words, setting css `padding` will call native setter 4 times.

## What is the new behavior?
The new approach ensures padding keeps its shorthand identity and reduces the native setter calls at the same time.
This is achieved by sticking to core's practices:
- Implement an internal property that is changed by the shorthanded props (see background, font)
- Use Property.isSet to check when property is unset (good for resetting padding to default)

Following a couple of rules helps make use of core's batch update mechanism and its suspendedUpdates bag that stores pending updates per property key.
This is where `paddingInternal` comes to play and as a property it makes sure there's only one single update in the bag.
To trigger updates for the new property we set its value to be the 4 padding props as a concatenated string.
Then, the `setNative` corresponding method will use the 4 `effectivePadding` props to update native view padding.
Additionally, the PR improves the way default paddings are restored for both platforms by replacing effective padding props with getters and at the same time removes repetitive code here and there.

Closes #8335

### DIFF
--- a/packages/core/ui/button/index.android.ts
+++ b/packages/core/ui/button/index.android.ts
@@ -1,7 +1,6 @@
 import { ButtonBase } from './button-common';
 import { PseudoClassHandler } from '../core/view';
-import { paddingLeftProperty, paddingTopProperty, paddingRightProperty, paddingBottomProperty, zIndexProperty, minWidthProperty, minHeightProperty } from '../styling/style-properties';
-import { Length } from '../styling/length-shared';
+import { zIndexProperty, minWidthProperty, minHeightProperty, paddingInternalProperty } from '../styling/style-properties';
 import { textAlignmentProperty } from '../text-base';
 import { CoreTypes } from '../../core-types';
 import { profile } from '../../profiling';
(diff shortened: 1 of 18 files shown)


NathanWalker:
@coderabbitai review

NathanWalker (approved):

PR #11226, opened 2026-05-23
fix(android): Added missing nested frame restoration step

CatchABus:
## PR Checklist

Right now, the nested frame first simulated navigation does not work if already done once.
That usually occurs because the cached transition state is not unset after the simulated navigation.

## What is the new behavior?
This PR is a continuation of #9725 & #10713 and contains the following:
- Added missing transition restoration step for nested frames to fragment `onResume` lifecycle
- Avoid unsetting the entry fragment during `onDestroy` lifecycle of the old fragment if entry doesn't own it anymore
- Removed frame create timeout workaround from view `loaded` lifecycle (this used to help solve the problem before #10713 which broke it)
- Added handling to unload root view during `resetRootView` in case flavors reuse the same root view
- Corrected AndroidFrame types that were broken during NS9 refactoring

Fixes issues mentioned in PR #11214.

Just a few notes about the transition restore change.
In the future, we could keep a reference of the last executed context at all times and set it as executing to simulate that navigation. This way, the simulated navigations will always reach `setCurrent` call.
Right now, this isn't possible as it's breaking current `TabView` functionality and automated tests fail.
Also, there are checks that proceed to actions based on whether a context is executing or not so it will need extended testing.
The first step to leaning to that direction is to take care of the fragment manager mess on `TabView` elements. Then, we can think about it.

Luckily, this PR keeps the simulated navigation workaround as is more or less and solves the apparent problem.

### DIFF
--- a/packages/core/ui/frame/fragment.transitions.android.ts
+++ b/packages/core/ui/frame/fragment.transitions.android.ts
@@ -38,22 +38,22 @@ interface ExpandedAnimator extends android.animation.Animator {
 }
 
 export interface ExpandedEntry extends BackstackEntry {
-	enterTransitionListener: ExpandedTransitionListener;
-	exitTransitionListener: ExpandedTransitionListener;
-	reenterTransitionListener: ExpandedTransitionListener;
-	returnTransitionListener: ExpandedTransitionListener;
+	enterTransitionListener?: ExpandedTransitionListener;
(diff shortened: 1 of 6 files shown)


NathanWalker:
@coderabbitai review

NathanWalker (approved):

PR #11259, opened 2026-06-04
fix(vite): support Vite 8 / Rolldown

JumpLink:
## What

Makes `@nativescript/vite` build under **Vite 8**, which swaps the bundler from Rollup to Rolldown. No behavior change on Vite ≤ 7 / webpack / rspack or the existing flavors.

## Why it breaks on Vite 8 / Rolldown

Rolldown reimplements several Vite-internal plugins in Rust, changing two contracts:

1. **`resolve.alias` function replacements are rejected.** Rolldown's native `ViteAlias` plugin only accepts string replacements and throws: `Failed to convert builtin plugin 'ViteAlias' … function replacement into rust type String`.
2. **The explicit `@rollup/plugin-commonjs` crashes.** The bundled `commonjs({...})` instance throws `Cannot read properties of undefined (reading 'currentLoadingModule')` under Rolldown. Rolldown transforms CommonJS natively, so the plugin is redundant there.

## Fixes

**1. Remove function-replacement `resolve.alias` entries** (`configuration/base.ts`, `helpers/ts-config-paths.ts`):
- `@nativescript/core/(.+)/index` becomes a plain string replacement (`${NS_CORE_ROOT}/$1`) — identical result.
- The redundant `packagePlatformAliases(...)` alias is dropped: its package.json `main` + platform-variant lookup is already performed by the `nativescriptPackageResolver` `resolveId` plugin in the same config. The now-unused helper file is removed.
- The tsconfig **wildcard** `paths` mapping moves out of a function-replacement alias into a new `resolveId` plugin, `tsConfigPathsResolverPlugin`. The fs-based platform-specific (`.android`/`.ios`) and directory-index resolution is preserved byte-for-byte in a shared helper. Exact (non-wildcard) tsconfig path aliases are unchanged.

**2. Skip the explicit `@rollup/plugin-commonjs` on Vite 8 / Rolldown** (`configuration/base.ts`):
- Gated on Vite's reported major version (`>= 8`). Vite ≤ 7 keeps the plugin exactly as before; on Vite 8 / Rolldown it is omitted and Rolldown's native CJS handling takes over.

## Verification

Validated empirically: a real `@nativescript/canvas` app (a three.js teapot) that fails to build on Vite 8.0.16 builds successfully (612 modules → working bundle, `ns run android` launches and renders on the emulator) once these fixes are applied. The change set was developed + proven via an independent implementation of the same two fixes before being applied here; the source changes are minimal and backward-compatible. Full-monorepo `tsc`/build is left to CI.

## Backward compatibility

- Only FIX 2 (the commonjs gate) is version-conditional, keyed on `version` from `vite` (also true for the `rolldown-vite` override). Vite ≤ 7 keeps the exact existing plugin pipeline.
- FIX 1 is unconditional and behavior-equivalent on both Rollup and Rolldown.
- `build.commonjsOptions` and `@rollup/plugin-replace` are untouched.

## Note

`helpers/prelink-angular.ts` still contains a function-replacement alias, but `createAngularPrelinkPlugin` is currently dead code (not imported anywhere — `configuration/angular.ts` uses `angularLinkerVitePlugin`), so it does not affect a Vite 8 build today. Left untouched per minimal scope; flagging in case it is ever wired up.


NathanWalker:
Excellent @JumpLink, I'll work on mapping this into https://github.com/NativeScript/NativeScript/tree/feat/vite-improvements-with-version-8-hmr which has some really nice hmr additions (with new runtime features upcoming with 9.1) so maybe this can release sooner while 9.1 is still being prepared.

NathanWalker:
Thanks again for this @JumpLink, I've brought them into other Vite 8 resolutions and release is targeted by end of month.

PR #11260, opened 2026-06-04
fix: improve view teardown process for reusable views

edusperoni:
## PR Checklist

When destroying a reusable view's node, two issues exist:

1. `destroyNode()` calls `callUnloaded()` unconditionally, even when the view was never loaded. This can fire `unloaded` lifecycle events (and run unload logic) on views that are not in a loaded state.
2. During `_tearDownUI()`, children are torn down by inheriting the parent's `force` flag (`child._tearDownUI(force)`). When the parent is being destroyed but a child is marked as `reusable`, the child's native view is preserved and skipped, leaving orphaned/leaked native views behind.

## What is the new behavior?

1. `destroyNode()` now uses `unloadView(this)`, which guards on `view.isLoaded` before calling `callUnloaded()`, so the unloaded path only runs for views that are actually loaded.
2. When a view is torn down, its children are now always torn down with `force = true` (`child._tearDownUI(true)`), regardless of whether they are marked reusable. Once a parent is being destroyed, its children are destroyed too. Developers who want to reuse a child elsewhere are responsible for detaching it beforehand.

Together these changes make the teardown process for reusable views more correct and prevent native view leaks.

### DIFF
--- a/packages/core/ui/core/view-base/index.ts
+++ b/packages/core/ui/core/view-base/index.ts
@@ -1264,7 +1264,7 @@ export abstract class ViewBase extends Observable {
 	 */
 	public destroyNode(forceDestroyChildren?: boolean): void {
 		this.reusable = false;
-		this.callUnloaded();
+		this.unloadView(this);
 		this._tearDownUI(forceDestroyChildren);
 	}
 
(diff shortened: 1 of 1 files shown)


farfromrefug:
@edusperoni `destroyNode` is only used by angular?

NathanWalker (approved):

PR #11261, opened 2026-06-06
fix(ios): re-attach root view after in-process runtime reload

NathanWalker:
When the runtime is reloaded in-process (NativeScriptRuntime.reloadApplication, used for OTA soft reboots) the app re-runs main in a fresh JS isolate while the process stays alive. iOSApplication.run() sees a live UIApplication and routes to runAsEmbeddedApp()

- The previous UIWindow was deallocated along with the old isolate (it was retained only by the JS-side SceneDelegate._window), and sceneWillConnectToSession, which creates the window, does not re-fire on a reload. getWindow() then returned null and runAsEmbeddedApp() bailed early, leaving a black screen.
- With no embedder delegate (NativeScript owns the UIApplication), the new root was presented modally over the old, now-dead root controller instead of replacing it, leaking the old controller on every reload and breaking when an overlay is mid-presentation.

runAsEmbeddedApp() works same as before but now supports this use case:
- recreates a window bound to the active UIWindowScene when none is found (initWithWindowScene + _setWindowForScene/_setupWindowForScene);
- moves the rootViewController guard into the embedder-only branch, since a freshly recreated window has no controller yet;
- replaces the window root via setWindowRootView() for the NativeScript-owned case

Supports 9.1 runtime features for OTA installed hot reloading of runtime isolates.

### DIFF
--- a/packages/core/application/application.ios.ts
+++ b/packages/core/application/application.ios.ts
@@ -127,6 +127,16 @@ function supportsMultipleScenes(): boolean {
 	return UIApplication.sharedApplication?.supportsMultipleScenes;
 }
 
+/**
+ * Number of times the JS runtime has been soft-rebooted in this process via
+ * NativeScriptRuntime.reloadApplication / restartWithConfig. 0 on first boot.
+ * Provided as a global by the iOS runtime (v9+); older runtimes report 0.
+ */
+function getRuntimeReloadCount(): number {
+	const runtime = (globalThis as any).NativeScriptRuntime;
+	return runtime && typeof runtime.reloadCount === 'number' ? runtime.reloadCount : 0;
+}
+
 @NativeClass
 class Responder extends UIResponder implements UIApplicationDelegate {
 	get window(): UIWindow {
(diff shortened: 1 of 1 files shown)


NathanWalker:
@coderabbitai review

packages/core/application/application.ios.ts
```diff
@@ -360,11 +386,18 @@ export class iOSApplication extends ApplicationCommon implements IiOSApplication
 		});
 
 		if (embedderDelegate) {
+			// Embed into host app.
+			// present over the host's existing root view controller.
+			if (!rootController) {
+				return;
+			}
```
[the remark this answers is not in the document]
NathanWalker: @copilot you can make the suggested fix.

packages/core/application/application.ios.ts
```diff
@@ -331,17 +331,43 @@ export class iOSApplication extends ApplicationCommon implements IiOSApplication
...
 
-		const rootController = window.rootViewController;
-		if (!rootController) {
+		if (!window) {
 			return;
 		}
 
+		// May be null on a freshly recreated window — expected; the replace-root
+		// path below sets it. Only the embedder path needs an existing controller.
+		const rootController = window.rootViewController;
```
NathanWalker, line 395:
```suggestion
		const rootController = window.rootViewController;
        const embedderDelegate = NativeScriptEmbedder.sharedInstance().delegate;

        // Embed into host app requires an existing root view controller
        if (embedderDelegate && !rootController) {
            return;
        }
```

packages/core/application/application.ios.ts
```diff
@@ -331,17 +331,43 @@ export class iOSApplication extends ApplicationCommon implements IiOSApplication
...
+
 		const controller = this.getViewController(rootView);
 		const embedderDelegate = NativeScriptEmbedder.sharedInstance().delegate;
```
NathanWalker:
```suggestion
```

packages/core/application/application.ios.ts
NathanWalker:
```suggestion
```

PR #11271, opened 2026-06-18
feat(android): skip transition graph for non-animated navigation

edusperoni:
## PR Checklist

On Android, every non-animated page navigation sets up a zero-duration `NoTransition`/`CustomTransition` graph. Its only purpose is to receive a transition-end callback that drives navigation completion — there is no actual animation to run. Besides being wasteful, this graph is the source of a memory leak: androidx attaches end listeners to the (shared) `AnimatorSet` that are never removed for a zero-duration run, so each navigation leaves a stale listener retaining the previous fragment and its page.

## What is the new behavior?

Regular (non-nested) non-animated navigation no longer creates any fragment transitions. Navigation completion is already driven by `FragmentCallbacksImplementation.onResume -> frame.setCurrent` whenever nothing is animating, so the zero-duration transition was unnecessary. The nested-first-navigation path is left untouched (it runs once per nested frame, not per navigation, so it does not accumulate). This removes the leak at its source for the common case and avoids building/tearing down a transition + dummy animator on every navigation.

## Related / split PRs

This is the **root-cause** half of the fix and is intentionally split from #11270 because it is **more behaviorally risky**: it makes the existing `onResume` fallback the primary completion path for non-animated navigation, which is worth extra scrutiny across edge cases (`clearHistory`, modal frames, rapid/queued navigations, app suspend/resume mid-navigation).

#11270 is the lower-risk native fix (run each `CustomTransition` on a per-invocation `AnimatorSet` clone so framework listeners can't accumulate) and stands on its own. The two are complementary: #11270 protects all remaining `CustomTransition` users (animated/custom transitions and the nested path); this PR additionally avoids creating the graph at all for the common non-animated case. #11270 can ship independently of this one.


NathanWalker:
Possible a test may need adjustment but looks like the changes do have some legitimate failing automated tests (ran a 2nd time just to make sure wasn't a false first run).

edusperoni:
closing this as it's clearly missing some pieces

Issue #6441, opened 2018-10-22
Max-width & max height featue request


zanina-yassine:
I am requesting to add the support for max-width and max-height features, this will help alot of UI/UX designers, every other problem I manged to work around it with plug-ins except for this...
Thank you

edusperoni:
This is a must. We currently have to do some crazy hacks to achieve something similar to max-width.

For example, we created https://github.com/MukaSchultze/nativescript-windowed-modal just because we wanted a consistent android/iOS experience and the ModalStack was specifically created so we could have our own margins and a max-width (we set the width+margins so it ends up fitting the device).

Same issue with our login box, in which I had to set a fixed width + margins instead of using something like `width:80%; max-width: 500`

We also have some platform specific code to put a LottieView in the actionbar, since Android/iOS have different heights and fixing the rate at 100% basically screwed with our aspect ratio.

aranmohyeddin:
We also need this extensively in our company. For instance we want to centralize the content of our modals vertically and we use gridlayout with rows="* auto *" and use a scrollview for the middle one so that in small devices it won't overflow and it would be scrollable. but without specifying max-height=100% for the scrollview what happens is that the scrollview just overflows the viewport. There are a ton of other usecases to this. We strongly believe this is a must.

alexhribsek:
Is there any progress on this? If not, has anyone found a workaround, which is not JavaScript? 😄

ArthurMJ96:
+1

shirakaba:
One more reason to adopt Facebook’s Yoga library for flexbox layout – it would give us all CSS layout features, including max-width and max-height, not to mention absolute and relative positioning. But it’s easy to dream!

farfromrefug:
@shirakaba agree though it would only work for this layout. You can want that with label, images....
BTW in a sense I think the code is already there as it is the AT_MOST

7ammer:
+1 for this feature

csimpi:
Although it seems NS is dying I'd give a +1, trying our luck

AdamAtri:
why is this not a thing yet????

Issue #1511, opened 2016-02-08
Add CSS max-width support 


mikeres0:
I think it would be beneficial to include max-width as a CSS attribute within Nativescript. As a web developer I am finding myself wanting to use this attribute regularly, so I think it'd be a great addition to Nativescript's CSS subset.

mikeres0:
Sort of related to #412

mnahkies:
@mikeres0 I'm often finding myself wanting a max-width property as well, particularly when trying to make the same layout look nice on a wide range of devices.

It doesn't look like it was implemented in conjunction with support for percentage based sizes, so perhaps this could be reopened?

kazemihabib:
I need this feature too.
please reopen it.

jdnichollsc:
+1

dali-gharbi:
+1

surdu:
This is not related to #412 ...

markedro:
After looking around, I agree that this is different that #412 and there isn't currently any simple way to achieve a max-width. Min-width already is supported so is there something that supporting max-width would do to break a bunch of other stuff? Please re-open this.

skopekreep:
+1

I agree that this would be a valuable addition.

Jarrich:
+1

ibnYusrat:
Still nothing.. after all these years?

surdu:
@ibnYusrat Probably because this is closed an it's not on NativeScript radar ?

@NickIliev could this please be reopened so it would get more visibility ?

NickIliev:
@ibnYusrat @Jarrich @surdu we do have support for setting percentage for width which makes a good solution for max width. (e.g. setting width to be 80% of the parent width while using Label or a layout with transparent background). We also have [a FlexBoxLayout](https://docs.nativescript.org/ui/layouts/layout-containers#flexboxlayout) which support alighItems and other flex properties (which in practice allows mimicking the max width behavior)

For all wondering why it would be so hard to implement max-width (more or less web concept) in the context of NativeScript mobile world take a look [at this solution for native iOS.](https://stackoverflow.com/a/38308811/4936697). 

Still, I am reopening this issue and marking it as a feature request  Anyone interested in [contributing](https://www.nativescript.org/contribute) can create a PR  - a good blog post about NativeScript contributing can be found  [here](https://www.nativescript.org/blog/your-first-contribution-to-nativescript-a-blog-post).

ibnYusrat:
[...]

PR #11300, opened 2026-07-02
feat(core): support max-width and max-height CSS properties

edusperoni:
## PR Checklist

NativeScript supports `width`, `height`, `min-width` and `min-height`, but there is no `max-width` / `max-height`. There is no way to cap a view's size — including the common responsive `max-width: 100%` pattern. The Android flexbox port even contained dead `MAX_SIZE /* lp.maxWidth */` placeholders showing the feature was anticipated but never wired up.

## What is the new behavior?

Adds first-class `max-width` / `max-height` style properties on both iOS and Android, accepting fixed lengths and percentages (e.g. `max-width: 200`, `max-width: 50%`).

Because percentages resolve against the parent size, the properties follow the `width`/`height` pattern rather than `min-*`: the effective pixel value is computed at measure time (`effectiveMaxWidth` / `effectiveMaxHeight`, defaulting to Infinity = unconstrained). The max clamp is applied after the min clamp, so `max` wins when `min > max`, matching CSS.

- **iOS:** measure-time clamps in the base view and every layout container, plus a central cap in `ViewHelper.getMeasureSpec`; wires up the previously-dead `MAX_SIZE` flexbox placeholders.
- **Android:** enforced natively in the `org.nativescript.widgets` library — `maxWidth`/`maxHeight` (+percent) on `CommonLayoutParams`, clamped in `getMeasureSpec` and resolved in `adjustChildrenLayoutParams`, exposed via new `ViewHelper` setters. The now-shadowing `maxWidth`/`maxHeight` fields were removed from `FlexboxLayout.LayoutParams` so flexbox inherits the unified fields and honors the properties (including percent). `widgets-release.aar` was rebuilt.

Adds automated tests for style-property parsing/sync, flexbox measure (enabling four previously-stubbed max tests), StackLayout measure, and Label max-width/height including percent.

Implements #1511
Implements #6441


CatchABus:
I leave this here for future reference since most people think that {N} fully supports minWidth/minHeight.
The minWidth/minHeight are partial CSS implementations and rather complete android implementations with their corresponding layout limitations which do not match with what CSS does: https://developer.android.com/reference/android/view/View#setMinimumWidth(int)

In simple words, {N} minWidth/minHeight will have no effect if users sets width/height to a fixed size due to how android uses these properties along with its measure modes.

NathanWalker (approved):

PR #11306, opened 2026-07-16
fix(core): remove once listeners by identity, not stale index

jkatins:
## PR Checklist

Regression introduced in 1e55f0c (#10946), first released in core 9.0.0.

`Observable._fireEvent` iterates a copy of the observers array but passes the copy's loop index to `_handleListenerEntry`, which splices the live array at that position. If any listener at a lower index is removed during the same notify pass (a handler calling `off()` on itself, or a preceding `once` entry being consumed), the live array shifts left and the splice for a later `once` entry removes the wrong element or nothing at all. Consequences:

1. The `once` listener stays subscribed and fires on every subsequent notify.
2. An unrelated listener sitting after it can be silently unsubscribed.

The `launch` event is affected deterministically in every app: `NativeScriptGlobalState` registers `_setLaunched` as the first launch listener and it removes itself inside its handler, so every `Application.once('launch', ...)` remains permanently armed. On Android this crashes apps that bootstrap from a launch listener and keep their process alive with a foreground service: recreating the activity (e.g. back-button exit, then relaunch) fires `launch` again, the armed `once` runs the bootstrap a second time and `Application.run()` throws `Error: Application is already started.` inside `NativeScriptActivity.onCreate`.

Minimal repro:

```ts
const obj = new Observable();
obj.on('foo', function handler() {
	obj.off('foo', handler);
});
obj.once('foo', () => console.log('once'));

obj.notify({ eventName: 'foo', object: obj }); // logs "once"
obj.notify({ eventName: 'foo', object: obj }); // logs "once" again

### DIFF
--- a/packages/core/data/observable/index.spec.ts
+++ b/packages/core/data/observable/index.spec.ts
@@ -55,5 +55,48 @@ describe('Observable', () => {
 			expect(callCount1).toBe(1);
 			expect(callCount2).toBe(1);
 		});
+
+		it('fires just once when registered after a listener that removes itself', () => {
+			const observable = new Observable();
+			let selfRemovingCallCount = 0;
+			let onceCallCount = 0;
+			let regularCallCount = 0;
+			const selfRemovingHandler = function () {
+				selfRemovingCallCount++;
+				observable.off('test', selfRemovingHandler);
+			};
+			const onceHandler = function () {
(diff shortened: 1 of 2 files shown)


CatchABus:
@jkatins Nice finding and explanation!

CatchABus (approved):

PR #11307, opened 2026-07-17
perf(core): hot-path optimizations across events, properties, CSS matching and iOS layout

NathanWalker:
### Event system (`data/observable`)
- `notify()` now skips the deprecated static/global event-handler machinery entirely (including the `this.constructor.name` resolution and 4 hash lookups per event) when no global handlers are registered, tracked via a simple module-level counter. Global handler semantics are unchanged when they *are* used.
- `_handleListenerEntry` uses `callback.call(thisArg, data)` instead of `callback.apply(thisArg, [data])`, removing a throwaway array allocation on every listener invocation.
- `hasListeners()` uses a direct property read instead of the `in` operator.
- `_indexOfListener` uses an indexed loop instead of `findIndex` + closure (runs on every `addEventListener`).

### Property system (`ui/core/properties`)
- `isResetValue()` / `isCssWideKeyword()` — these run at the top of every property setter. They now bail out with a single `typeof` check for the overwhelmingly common case of non-string values (numbers, booleans, Colors, objects) instead of comparing against every CSS reset keyword.

### CSS selector matching (`ui/styling`)
- `SelectorScope.getSelectorCandidates()` replaced an O(n·buckets) `reduce`+`concat` (which re-copied the accumulator on every bucket) with a single push-based accumulation. This runs for every view on load and on every class/id change.
- `AttributeSelector` with `ignoreCase` normalizes its value once at construction instead of lowercasing (and reassigning) it on every match.
- `matchMediaQueryString()` caches media-query string splits (they come from parsed stylesheets, so the distinct set is small and stable) and avoids per-call closures.
- `SimpleSelectorSequence.match/mayMatch/trackChanges`, `ComplexSelector.match/trackChanges` and `Selector.ChildGroup.getMatchingNode` use indexed loops instead of `every`/`forEach` closures; `getMatchingNode` also drops the string-keyed method dispatch (`sel[funcName]`).
- `CssState.updateDynamicState` / `setPropertyValues` avoid per-call closure allocations (hoisted kebab-case replacement func, indexed filter loop). These run on every pseudo-class change (every tap).

### View base (`ui/core/view-base`)
- `addPseudoClass`/`deletePseudoClass` no longer allocate an aliased-states array on every visual-state transition (every tap toggles `:highlighted`/`:active`/`:pressed`).
- `classNameProperty.valueChanged` uses indexed loops instead of `forEach` closures (hot in list-item recycling).
- Fixed a latent bug in `getViewByDomId()`: the descendants callback compared `view._domId` (the root argument) instead of `child._domId`, so descendant lookups never matched and always walked the entire subtree.

### iOS layout (`utils/layout-helper`)
- `toDevicePixels` / `toDeviceIndependentPixels` / `getDisplayDensity` now cache the screen scale instead of resolving `window → screen → scale` through the ObjC bridge on **every** dp/px conversion (these are called several times per view per measure/layout pass). The device main-screen scale never changes at runtime — `Screen.mainScreen` already relies on the same invariant, and the Android implementation already caches density this way.

### Micro-benchmarks (vitest bench, committed under `packages/core/__tests__/benchmarks/`)

Node v25, darwin arm64, same machine/settings for both runs:

| Benchmark | Before (ops/s) | After (ops/s) | Δ |
|---|---:|---:|---:|
| `StyleSheetSelectorScope.query` — button w/ id + 4 classes | 282,946 | 392,871 | **+38.9%** |
| `StyleSheetSelectorScope.query` — plain label | 434,778 | 518,608 | **+19.3%** |
| `Observable.notify` — 1 listener (thisArg) | 17,503,308 | 23,439,545 | **+33.9%** |
| `Observable.notify` — 3 listeners (thisArg) | 11,750,316 | 14,388,805 | **+22.5%** |
| `Observable.notify` — no listeners | ~21.3M | 27,112,421 | **+27%** |
| `notifyPropertyChange` | ~18.3M | 22,350,543 | **+22%** |
| `AttributeSelector.match` (ignoreCase) | 22,247,452 | 26,514,154 | **+19.2%** |
| `ComplexSelector.match` (child combinator) | 13,961,654 | 14,904,038 | **+6.7%** |
| `hasListeners` / `on/off` cycles | — | — | flat (no regression) |

Reproduce with: `cd packages/core && npx vitest bench --run`

### Automated iOS test suite (`npx nx debug apps-automated ios`, iPhone 17 Pro simulator)

| | Before | After (run 1) | After (run 2) |
|---|---|---|---|
| Tests | 1808 OK, 0 failed | 1808 OK, 0 failed | 1808 OK, 0 failed |
| Duration | 49,891.65 ms | 45,468.59 ms (**-8.9%**) | 43,018.74 ms (**-13.8%**) |

### DIFF
--- a/packages/core/utils/layout-helper/index.ios.ts
+++ b/packages/core/utils/layout-helper/index.ios.ts
@@ -1,6 +1,16 @@
 import * as layoutCommon from './layout-helper-common';
 import { ios as iosUtils } from '../native-helper';
 
+// Cache the screen scale to avoid repeated marshalling calls into UIKit
+// (window resolution + screen + scale) on every dp/px conversion during
+// measure and layout passes. The scale of the device's main screen never
+// changes at runtime, and `Screen.mainScreen` relies on the same invariant.
+let mainScreenScale = 0;
+
+function getMainScreenScale(): number {
+	return mainScreenScale || (mainScreenScale = iosUtils.getMainScreen().scale);
+}
+
 export namespace layout {
 	// cache the MeasureSpec constants here, to prevent extensive marshaling calls to and from Objective C
 	// TODO: While this boosts the performance it is error-prone in case Google changes these constants
@@ -37,15 +47,15 @@ export namespace layout {
 	}
 
 	export function getDisplayDensity(): number {
-		return iosUtils.getMainScreen().scale;
+		return getMainScreenScale();
 	}
 
 	export function toDevicePixels(value: number): number {
-		return value * iosUtils.getMainScreen().scale;
+		return value * getMainScreenScale();
 	}
 
 	export function toDeviceIndependentPixels(value: number): number {
-		return value / iosUtils.getMainScreen().scale;
+		return value / getMainScreenScale();
 	}
 
 	export function round(value: number) {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -39,7 +39,11 @@
 				"@typescript-eslint/no-this-alias": "warn",
 				"@typescript-eslint/no-namespace": "off",
 				"@typescript-eslint/no-inferrable-types": "off",
-				"@typescript-eslint/no-extra-semi": "error",
+				"@typescript-eslint/no-empty-object-type": "off",
+				"@typescript-eslint/no-unsafe-function-type": "off",
+				"@typescript-eslint/no-wrapper-object-types": "off",
+				"@typescript-eslint/no-unsafe-declaration-merging": "off",
+				"@typescript-eslint/no-unused-expressions": ["error", { "allowShortCircuit": true, "allowTernary": true }],
 				"no-extra-semi": "off"
 			}
 		},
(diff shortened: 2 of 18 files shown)


packages/core/utils/layout-helper/index.ios.ts
```diff
@@ -1,6 +1,16 @@
 import * as layoutCommon from './layout-helper-common';
 import { ios as iosUtils } from '../native-helper';
 
+// Cache the screen scale to avoid repeated marshalling calls into UIKit
+// (window resolution + screen + scale) on every dp/px conversion during
+// measure and layout passes. The scale of the device's main screen never
+// changes at runtime, and `Screen.mainScreen` relies on the same invariant.
+let mainScreenScale = 0;
```
CatchABus, line 8: ~~I don't see the android counterpart for screen scale caching.~~
Actually, I just noticed android used to cache density before iOS started doing so but it still seems strange that only one of them has the internal screen scale helper helper.

CatchABus (changes requested):

CatchABus (approved):

PR #11308, opened 2026-07-17
perf(core): skip-unchanged CSS re-application, traversal cleanups

NathanWalker:
- CssState.setPropertyValues: fix the pre-existing ordering bug where delete oldProperties[property] ran before the "skip unchanged" check, making it dead code — unchanged declarations were re-applied through the property system on every dynamic-state change (every :active toggle). Scoped css-variables are still always re-registered because resetScopedCssVariables clears them earlier in the same pass.
- eachLayoutChild allocates 2 closures per measure/layout pass instead of N+1; onLoaded/onUnloaded traversals use shared module-level callbacks (dispatching via child.parent to preserve loadView/unloadView overrides, e.g. TabView).
- query() appends media-query candidates via indexed loop instead of push-spread (avoids arg-spread limits on large candidate sets).
- Benchmarks: theme-scale stylesheet + recycled-rows query bench; new spec tests for repeated-query consistency incl. media-query non-accumulation.

## Considerations

Additional notes to be considered aside (above/beyond) phase 1 perf opts here: https://github.com/NativeScript/NativeScript/pull/11307

Note: selector candidate memoization keyed by (cssType, id, classes) was prototyped and measured to be a net LOSS at realistic scale (cached lookup 273k ops/s vs 294k direct; recycled rows -21%); string key building plus Map hashing costs more than the push-based rebuild from phase 1, which already made candidate sets cheap. 

| Benchmark | Direct rebuild (phase 1) | Memoized | Δ |
|---|---:|---:|---:|
| `query` — button w/ id + 4 classes | 294,492 ops/s | 273,512 ops/s | −7% |
| `query` — 20 recycled list rows (4 classes each) | 15,572 ops/s | 12,324 ops/s | **−21%** |

Building the string cache key (per-class concatenation) plus hashing it for `Map.get` costs more than the push-based candidate rebuild that phase 1 already made cheap; per-node candidate sets are bounded (universal + id + type + k class buckets, realistically < 100 entries), and `accumulateChanges`+`sort` dominate query cost regardless. Mentioning here so it isn't re-attempted without new evidence.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -39,7 +39,11 @@
 				"@typescript-eslint/no-this-alias": "warn",
 				"@typescript-eslint/no-namespace": "off",
 				"@typescript-eslint/no-inferrable-types": "off",
-				"@typescript-eslint/no-extra-semi": "error",
+				"@typescript-eslint/no-empty-object-type": "off",
+				"@typescript-eslint/no-unsafe-function-type": "off",
+				"@typescript-eslint/no-wrapper-object-types": "off",
+				"@typescript-eslint/no-unsafe-declaration-merging": "off",
+				"@typescript-eslint/no-unused-expressions": ["error", { "allowShortCircuit": true, "allowTernary": true }],
 				"no-extra-semi": "off"
 			}
 		},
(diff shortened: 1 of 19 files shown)


NathanWalker:
Closing since this it's more a reference for later if revisited. Other perf enhancements already went in with <https://github.com/NativeScript/NativeScript/pull/11307>

PR #11312, opened 2026-07-21
fix(ios): update swipe back gesture from Page property

tannerlinsley:
## What changed

- bind `enableSwipeBackNavigationProperty` to the native gesture update hook
- stop `actionBarHidden` changes from overwriting swipe-back state
- add an iOS regression test that toggles the Page property and checks the native recognizer

The handler was lost during the property-system migration: `enableSwipeBackNavigation` remained public and was read during page presentation, but changing it on a visible page did not update either iOS gesture recognizer.

## Validation

- `nx run core:build`
- `nx run core:lint` (passes with existing warnings)
- iOS 26.5 simulator XCUITest: disabling the property prevents an interactive edge pop; re-enabling it restores the native transition

### DIFF
--- a/apps/automated/src/ui/page/page-tests.ios.ts
+++ b/apps/automated/src/ui/page/page-tests.ios.ts
@@ -5,6 +5,23 @@ import { addLabelToPage } from './page-tests-common';
 
 export * from './page-tests-common';
 
+export function test_enableSwipeBackNavigation_updates_the_native_gesture() {
+	const page = new Page();
+	addLabelToPage(page);
+	helper.navigateWithHistory(() => page);
+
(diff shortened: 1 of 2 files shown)


NathanWalker (approved):
Good catch 💯

CatchABus (approved):

tannerlinsley:
The first iOS run reached the Page suite, then the NativeScript runtime asserted in the unrelated `test_WhenModalPageShownShowModalEventsRaisedOnRootModalTabView` test and the job timed out. Android passed. I refreshed the single commit SHA to request a clean rerun, but the new iOS/Android workflows are now `action_required` and need maintainer approval. No implementation content changed in the refresh.

CatchABus (approved):

PR #11321, opened 2026-08-01
feat(ios): shared transition interactive morph dismiss with cornerRadius and alpha fixes

NathanWalker:
- adds a`morph` option to interactive dismiss: the destination view scales/translates with the finger, springs back on cancel, and morphs into the matching source element frame on finish (with shared-element snapshot fly-back).
- Gesture phase tracking (idle/pending/active/ignored) so vertical scrolls on inner scroll views no longer engage the dismiss; the inner scroll view is frozen while a dismiss is active.
- Adds `cornerRadius` to SharedProperties, animated across present/dismiss and driven by percent during interactive dismissal (sheet-style transitions).
- Uses native alpha instead of NS opacity when hiding shared elements during transitions, restoring from NS opacity as the source of truth. Tracks all presenting views (including duplicate sharedTransitionTags in sister lists) so cleanup restores every element and avoids double-image artifacts.
- Detaches imageSourceChange listeners on cleanup and seeds image snapshots from the source image to avoid a blank first frame.
- Disables iOS' built-in edge pan while a custom interactive dismiss is wired so it can't race the custom recognizer.
- snapshotView now restores layer opacity synchronously inside a CATransaction instead of on next tick, so cleanup opacity changes aren't clobbered.

### DIFF
--- a/packages/core/ui/page/index.ios.ts
+++ b/packages/core/ui/page/index.ios.ts
@@ -185,7 +185,12 @@ class UIViewControllerImpl extends UIViewController {
 					// Workaround for disabled backswipe on second custom native transition
 					if (frame.canGoBack()) {
 						const transitionState = SharedTransition.getState(owner.transitionId);
-						if (!transitionState?.interactive) {
+						if (transitionState?.interactive) {
+							// A custom interactive dismiss is wired up (shared transition with
+							// pan gesture). Disable iOS' built-in edge pan so it can't race the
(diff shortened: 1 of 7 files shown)


farfromrefug:
That PR should be splitted into small ones with explanations/images/videos of what it fixes.

NathanWalker:
All improvements are related in this PR and support the attached polished effects.

https://github.com/user-attachments/assets/c9651fad-63c2-4020-a7d1-9e6c5b7281d2

PR #11322, opened 2026-08-01
feat(ios): shared transition improvements

NathanWalker:
- refined presentation/dismissal positioning and TabView integration.
- improved property animations for shared transitions; smoother property matching between shared elements during page transitions.
- shadows with shared transitions; shadow support carried through shared element snapshots across present/dismiss

### DIFF
--- a/packages/core/ui/transition/shared-transition.ts
+++ b/packages/core/ui/transition/shared-transition.ts
@@ -121,6 +121,7 @@ export interface SharedTransitionInteractiveOptions {
 	 * (iOS only) Drop shadow rendered behind the destination view during the
 	 * interactive dismiss — gives the view a sense of depth/elevation as it
 	 * morphs back toward the source (Apple Music–style).
+	 * Note: This feature will be brought to Android in a future release.
 	 *
 	 * - `true` applies sensible defaults
 	 *   (color black, opacity 0.3, radius 30, offset (0, 8)).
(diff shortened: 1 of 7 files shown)


packages/core/ui/transition/shared-transition.ts
```diff
@@ -117,6 +117,27 @@ export interface SharedTransitionInteractiveOptions {
...
+	 * (iOS only) Drop shadow rendered behind the destination view during the
+	 * interactive dismiss — gives the view a sense of depth/elevation as it
+	 * morphs back toward the source (Apple Music–style).
+	 *
+	 * - `true` applies sensible defaults
+	 *   (color black, opacity 0.3, radius 30, offset (0, 8)).
+	 * - Pass an object to customize any field.
+	 * - Omit / `false` to skip.
+	 */
+	shadow?: boolean | SharedTransitionInteractiveDismissShadow;
```
farfromrefug, line 131: It should be named with ios if it is ios only

packages/core/ui/transition/shared-transition.ts
```diff
@@ -181,6 +202,11 @@ class SharedTransitionObservable extends Observable {
 }
 let sharedTransitionEvents: SharedTransitionObservable;
 let currentStack: Array<SharedTransitionState>;
+// Detached NS subtrees (e.g. a TabView's iosBottomAccessory) that aren't reachable
+// from a page via the NS view tree, but whose tagged views should still participate
+// in shared transitions. Registered by their owning component; queried alongside
+// the source page in getSharedElements.
+let externalRoots: Array<ViewBase>;
```
farfromrefug: Shouldn't they be weakref?

farfromrefug:
Description on these PRs should be much more detailed explaining what changed. Especially when written by ai.
Also maybe videos could help understand what changed , what if fixed

NathanWalker:
Several detailed edge cases worth keeping split on this PR to strengthen the other (see video there): https://github.com/NativeScript/NativeScript/pull/11321

I'll address the other comments when able.

NathanWalker:
Thanks @farfromrefug, the WeakRef suggestion was a good catch and is applied. 
For `shadow`, it follows the existing convention in this interface (sharedTransitionTags, morph, interactive) which are already marked iOS-only for now but slated for Android parity in future.

PR #11324, opened 2026-08-01
feat(ios): iosContentInsetAdjustmentBehavior for ScrollView with safe-area aware layouts

NathanWalker:
Adds an iOS-only `iosContentInsetAdjustmentBehavior` property to ScrollView (mapping to UIScrollView.contentInsetAdjustmentBehavior, default `never` to preserve existing behavior). When iOS manages insets, ScrollView no longer subtracts safe-area insets manually (avoiding scroll drift with large titles) and nested layouts (Absolute, Dock, Flexbox, Grid, Stack, Wrap) skip their own safe-area subtraction via IOSHelper.hasIOSManagedInsetAncestor to prevent double-counting the nav-bar inset.

### DIFF
--- a/packages/core/ui/scroll-view/index.ios.ts
+++ b/packages/core/ui/scroll-view/index.ios.ts
@@ -179,17 +179,24 @@ export class ScrollView extends ScrollViewBase {
 			return;
 		}
 
-		const insets = this.getSafeAreaInsets();
+		if (SDK_VERSION > 10) {
+			// https://developer.apple.com/documentation/uikit/uiscrollview/contentinsetadjustmentbehavior
+			// 0=automatic, 1=scrollableAxes, 2=never, 3=always
+			const behavior = this.iosContentInsetAdjustmentBehavior;
+			let nativeBehavior = 2; // .never (preserves prior default)
+			if (behavior === 'automatic') nativeBehavior = 0;
+			else if (behavior === 'scrollableAxes') nativeBehavior = 1;
+			else if (behavior === 'always') nativeBehavior = 3;
+			this.nativeViewProtected.contentInsetAdjustmentBehavior = nativeBehavior;
+		}
+
+		// When iOS adjusts content insets itself, don't also subtract safe-area insets here —
+		// doing both makes contentSize track the dynamic navbar height and causes scroll drift with large titles.
+		const useIOSInsetAdjustment = SDK_VERSION > 10 && this.iosContentInsetAdjustmentBehavior !== 'never';
+		const insets = useIOSInsetAdjustment ? { left: 0, top: 0, right: 0, bottom: 0 } : this.getSafeAreaInsets();
 
 		let scrollWidth = right - left - insets.right - insets.left;
 		let scrollHeight = bottom - top - insets.bottom - insets.top;
-
-		if (SDK_VERSION > 10) {
-			// Disable automatic adjustment of scroll view insets
-			// Consider exposing this as property with all 4 modes
-			// https://developer.apple.com/documentation/uikit/uiscrollview/contentinsetadjustmentbehavior
-			this.nativeViewProtected.contentInsetAdjustmentBehavior = 2;
-		}
 
 		let scrollInsetWidth = scrollWidth + insets.left + insets.right;
 		let scrollInsetHeight = scrollHeight + insets.top + insets.bottom;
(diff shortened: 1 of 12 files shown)


packages/core/ui/scroll-view/scroll-view-common.ts
```diff
@@ -113,3 +114,12 @@ export const isScrollEnabledProperty = new Property<ScrollViewBase, boolean>({
 	valueConverter: booleanConverter,
 });
 isScrollEnabledProperty.register(ScrollViewBase);
+
+const insetAdjustmentConverter = makeParser<'never' | 'automatic' | 'scrollableAxes' | 'always'>(makeValidator('never', 'automatic', 'scrollableAxes', 'always'));
+export const iosContentInsetAdjustmentBehaviorProperty = new Property<ScrollViewBase, 'never' | 'automatic' | 'scrollableAxes' | 'always'>({
+	name: 'iosContentInsetAdjustmentBehavior',
+	defaultValue: 'never',
+	affectsLayout: true,
```
CatchABus: @NathanWalker If we bother to create a property that affects layout, we should let users change the value during runtime.
For this, we'll need the corresponding. `iosContentInsetAdjustmentBehaviorProperty.setNative` method.

packages/core/ui/scroll-view/index.ios.ts
```diff
@@ -179,18 +179,25 @@ export class ScrollView extends ScrollViewBase {
...
-
 		if (SDK_VERSION > 10) {
-			// Disable automatic adjustment of scroll view insets
-			// Consider exposing this as property with all 4 modes
 			// https://developer.apple.com/documentation/uikit/uiscrollview/contentinsetadjustmentbehavior
-			this.nativeViewProtected.contentInsetAdjustmentBehavior = 2;
+			// 0=automatic, 1=scrollableAxes, 2=never, 3=always
+			const behavior = this.iosContentInsetAdjustmentBehavior;
+			let nativeBehavior = 2; // .never (preserves prior default)
+			if (behavior === 'automatic') nativeBehavior = 0;
```
CatchABus: This block can be converted in a function to be used in multiple places (e.g. property's setNative).

packages/core/ui/scroll-view/index.ios.ts
```diff
@@ -179,18 +179,25 @@ export class ScrollView extends ScrollViewBase {
...
+			else if (behavior === 'scrollableAxes') nativeBehavior = 1;
+			else if (behavior === 'always') nativeBehavior = 3;
+			this.nativeViewProtected.contentInsetAdjustmentBehavior = nativeBehavior;
 		}
 
+		// When iOS adjusts content insets itself, don't also subtract safe-area insets here —
+		// doing both makes contentSize track the dynamic navbar height and causes scroll drift with large titles.
```
CatchABus, line 206: We are doing this for all layouts but wouldn't it be cleaner to move the check inside method `getSafeAreaInsets` and return the empty insets from there and avoid repetitive code in those classes?

CatchABus (changes requested):

packages/core/ui/scroll-view/scroll-view-common.ts
CatchABus: Let's change `affectsLayout` to `__APPLE__` since the property may trigger a pointless layout request in android.

CatchABus (changes requested):

packages/core/ui/scroll-view/scroll-view-common.ts
NathanWalker: Good call, adjusted now 👍

packages/core/ui/scroll-view/index.ios.ts
NathanWalker: yes, great idea, moved now.

packages/core/ui/scroll-view/index.ios.ts
NathanWalker: another great call, adjusted now 💯

packages/core/ui/scroll-view/scroll-view-common.ts
NathanWalker: thank you, great call, added now 💯

CatchABus (approved):

farfromrefug (approved):

PR #11326, opened 2026-08-01
fix(core): guard undefined navigation queue entry in Frame

NathanWalker:
A goBack() queued while another navigation is still executing (e.g. an interactive dismiss that finished out-of-band) can leave the head of the navigation queue without an entry, since performGoBack hasn't populated it yet. Skip the head safely instead of crashing on reading `.resolvedPage`.

### DIFF
--- a/packages/core/ui/frame/frame-common.ts
+++ b/packages/core/ui/frame/frame-common.ts
@@ -351,9 +351,9 @@ export class FrameBase extends CustomLayoutView implements FrameDefinition {
 			return;
 		}
 
+		// Entry is undefined for a goBack() queued before performGoBack populates it.
 		const entry = this._navigationQueue[0].entry;
-		const currentNavigationPage = entry.resolvedPage;
-		if (page !== currentNavigationPage) {
+		if (entry && page !== entry.resolvedPage) {
 			// If the page is not the one that requested navigation - skip it.
 			return;
 		}


packages/core/ui/frame/frame-common.ts
```diff
@@ -352,6 +352,16 @@ export class FrameBase extends CustomLayoutView implements FrameDefinition {
 		}
 
 		const entry = this._navigationQueue[0].entry;
+		// Entry can be undefined when a `goBack()` was queued while another
+		// navigation was still executing (e.g. an interactive dismiss that
+		// finished out-of-band). performGoBack hasn't run yet to populate it.
+		// Skip the head safely instead of crashing reading `.resolvedPage`.
+		if (!entry) {
+			this._navigationQueue.shift();
```
CatchABus: @NathanWalker Here, we repeat the same 3 lines that will be executed afterwards.
Might be cleaner to just add the guard directly where it's needed.
e.g.
```ts
if (entry && page !== entry.resolvedPage) {
    // If the page is not the one that requested navigation - skip it.
	return;
}
```

CatchABus (changes requested):

packages/core/ui/frame/frame-common.ts
NathanWalker: Good catch @CatchABus pushed up adjustment for that 💯

CatchABus (approved):

farfromrefug (approved):

PR #11349, opened 2026-08-12
FormattedString ignores font scale on Android — withFontScale hardcodes 1, AbsoluteSizeSpan skips scale multiply

tsushanth:
Fix Android font scale: `withFontScale` (and other `with*` mutators) now preserve/forward scale correctly; `AbsoluteSizeSpan` in `setSpanModifiers` now multiplies `fontSize` by `fontScaleInternal`

Found via automated repo scanning, fix written and reviewed before opening.


farfromrefug (approved):

CatchABus:
This shouldn't be enough to scale spans the right way in most android devices and I know through trial.
Can you confirm a plain label with the same scale as a spanned label are identical in size on your screen by switching scale to different values from a11y settings?

PR #11361, opened 2026-08-26
perf(core): overhaul css matching, cascade, and application

edusperoni:
## PR Checklist

The CSS engine does a large amount of redundant work on every dynamic restyle, and has two correctness bugs in the cascade:

- **The "skip unchanged values" diff in `CssState.setPropertyValues` is dead code.** The previously applied entry is deleted *before* it is compared against, so every matching declaration is re-applied (colors and lengths re-parsed, converters re-run) on every dynamic update — tapping a button re-applies the view's entire computed style.
- **Attribute selectors always answer `mayMatch` with `true`** and unconditionally subscribe to `<attribute>Change`. With Angular's emulated encapsulation (`[_ngcontent-cN]` markers), every view keeps N−1 selectors it can never match and holds change subscriptions for events that are never raised — plain expando assignment does not notify. This is the main reason component-level CSS performs badly enough that it is currently discouraged.
- **Every `StyleScope` builds its own index over the merged application rule list**, so registering a stylesheet (as Angular does once per component at first use) re-sorts every application rule into fresh lookup maps for every scope in the app — O(components × rules × scopes) on navigation. Multiple scopes also write `pos` onto the same shared `RuleSet` selectors, and whichever built last wins.
- **Shorthand expansion never runs.** `ShorthandProperty.register` defines the expanding setter on `cls.prototype.PropertyBag` — the bag's constructor rather than its prototype — so shorthands reach the apply step unexpanded and the cascade breaks: `label { margin: 1 } .a { margin-top: 8 } #x { margin: 4 }` leaves `marginTop` at 8, ignoring the most specific rule.
- `CssState.onChange` tears down and rebuilds every dynamic-selector subscription on every re-match, even when the dependencies are identical.
- `CSSSource.parseCSSAst` is `async`, so parse failures escape the `try/catch` in `parse()` as unhandled rejections and leave the stylesheet silently empty.

## What is the new behavior?

Six commits, each self-contained, aligning the engine with how browsers and the CSS specs handle the same problems:

- The unchanged-value diff actually works: values are compared before the old entry is consumed, css expressions diff against their previously evaluated result, and `!important` stripping/scoped-tag lookups move out of the per-view hot path.
- `AttributeSelector.mayMatch` tests whether the view knows the attribute, and change subscriptions are only registered for attributes that can actually raise `<attribute>Change` (i.e. real properties with accessors). Angular's `[_ngcontent-*]` selectors no longer cost anything on views they can't match.
- Shorthands are expanded into their longhands at parse time, as the spec defines the cascade (a shorthand declaration is equivalent to declaring each longhand in its place). This fixes the specificity bug above, runs each converter once per declaration instead of once per view per update, and keeps values shared so the diff can compare by identity.
- `var()`/`calc()` shorthands, which cannot be expanded lexically, get **pending-substitution values** per the CSS Variables spec ([css-variables-1 §3.2](https://drafts.csswg.org/css-variables/#variables-in-shorthands)): each longhand cascades a shared placeholder, resolved once per view at apply time. Property bags are now longhand-only, so shorthand/longhand overlap is structurally impossible.
- The application stylesheets are indexed once, globally, instead of once per `StyleScope` — mirroring Blink's per-`TreeScope` `ScopedStyleResolver` + merged-collection architecture. Selectors sort by `(specificity, tier, position)` so application and scope-local rules cascade correctly from separate indexes, and component registration appends to the shared index instead of rebuilding it.
- `CssState.onChange` re-matches first and only rebuilds subscriptions when the change map actually differs.
- `parseCSSAst` is synchronous, so parse errors are reported through the existing error path again.

**Benchmarks** (included in `packages/core/__tests__/benchmarks/`; measured per-process against `main`, median, best of 3, node 22 — jitless approximates iOS JavaScriptCore):

| Scenario | JIT | jitless |
|---|---|---|
| re-apply 200 views, nothing changed | 2.90x | 3.00x |
| toggle a class on 200 loaded views | 2.12x | 2.74x |
| add 30 stylesheets + restyle 200 views | 1.84x | 2.28x |
| toggle a pseudo class on 200 views | 1.76x | 2.32x |
| style 400 views (angular attribute css) | 1.31x | 1.34x |
| load 30 components | 1.28x | 1.31x |
| style 400 views (plain css) | 1.00x | 1.00x |
| build a style scope | 0.89x | 0.87x |

The style-scope build is ~10% slower by design: `!important` stripping and shorthand expansion moved into parsing, paid once per stylesheet instead of once per view per update.

380 unit tests pass, including new suites for the cascade (`style-scope.spec.ts`, 28 tests) and selector behavior (`css-selector.spec.ts`). The `apps/automated` STYLE suite on a device is the remaining end-to-end check reviewers may want before merging, as it exercises the cascade against real native views.

### DIFF
--- a/packages/core/ui/styling/css-selector.ts
+++ b/packages/core/ui/styling/css-selector.ts
@@ -27,10 +27,7 @@ export interface Node {
 
 export interface Declaration {
 	property: string;
-	/**
-	 * Usually the raw text from the stylesheet, but a shorthand is expanded while
-	 * parsing and its longhands carry the already-converted value.
-	 */
+	/** Raw stylesheet text, except expanded shorthand longhands, which carry the converted value. */
 	value: any;
 }
 
@@ -107,14 +104,9 @@ interface LookupSorter {
 }
 
 /**
- * Which set of stylesheets a selector came from.
- *
- * The cascade breaks a specificity tie on source order, and source order across
- * stylesheets is really (stylesheet, position within it) - so a selector needs to
- * know its stylesheet's rank, not just its own index. Application styles are
- * consulted for every view and are indexed once; a scope's own styles come after
- * them and are indexed per scope, so their positions are only comparable within a
- * tier.
+ * Rank of the stylesheet set a selector was indexed in. Specificity ties break on
+ * (tier, pos): the application and local indexes are built separately, so positions
+ * are only comparable within a tier.
  */
 export const enum SelectorTier {
 	Application = 0,
@@ -164,13 +156,9 @@ function getNodePreviousDirectSibling(node: Node): null | Node {
 }
 
 /**
- * Cache of "does this view type notify for that attribute", keyed by prototype.
- *
- * `<attribute>Change` events are only raised by registered properties, which are
- * defined as accessors on a prototype. An attribute that is a plain value on the
- * instance - Angular's `_ngcontent-*` markers, anything a renderer assigns
- * directly - can never notify, so subscribing to its change event is pure
- * overhead on every view the selector may match.
+ * `<attribute>Change` is only raised by properties defined as prototype accessors;
+ * a plain instance value (e.g. Angular's `_ngcontent-*` markers) never notifies,
+ * so subscribing to its change event would be pure overhead.
  */
 const notifyingAttributes = new WeakMap<object, Map<string, boolean>>();
 
@@ -440,17 +428,12 @@ export class AttributeSelector extends SimpleSelector {
 		return false;
 	}
 	public mayMatch(node: Node): boolean {
-		// The attribute may still be assigned later, but only if the view knows about it:
-		// registered properties live on the prototype and anything a framework has already
-		// set is an own value. An attribute that is on neither can never start matching
-		// without the css state being invalidated anyway, and treating it as a permanent
-		// "maybe" keeps the selector - and a change subscription for it - on every view it
-		// will never match. Angular's per component `[_ngcontent-cN]` scoping makes that
-		// cost grow with the number of components in the app.
+		// Registered properties live on the prototype and anything already assigned is
+		// an own value; an attribute on neither can never start matching without the
+		// css state being invalidated anyway.
 		return this.attribute in node;
 	}
 	public trackChanges(node: Node, map: ChangeAccumulator): void {
-		// Only subscribe when the attribute can actually raise `<attribute>Change`.
 		if (attributeNotifiesChanges(node, this.attribute)) {
 			map.addAttribute(node, this.attribute);
 		}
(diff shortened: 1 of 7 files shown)


packages/core/ui/styling/css-selector.ts
```diff
@@ -376,10 +440,20 @@ export class AttributeSelector extends SimpleSelector {
...
 	}
 	public mayMatch(node: Node): boolean {
-		return true;
+		// The attribute may still be assigned later, but only if the view knows about it:
+		// registered properties live on the prototype and anything a framework has already
+		// set is an own value. An attribute that is on neither can never start matching
+		// without the css state being invalidated anyway, and treating it as a permanent
+		// "maybe" keeps the selector - and a change subscription for it - on every view it
+		// will never match. Angular's per component `[_ngcontent-cN]` scoping makes that
+		// cost grow with the number of components in the app.
```
farfromrefug: those AI comments are always too long. We mostly never need than 3 lines comments. It also helps keeping comments focused on what s important

packages/core/ui/styling/style-scope.ts
```diff
@@ -42,13 +41,32 @@ type KeyframesMap = Map<string, Keyframes[]>;
...
 const applicationAdditionalKeyframes: Keyframes[] = [];
+let mergedApplicationCssKeyframesInvalid = false;
 
 let applicationCssSelectorVersion = 0;
 
+/**
+ * The application stylesheets are the same for every style scope, so they are
+ * indexed once here instead of once per scope. Rules that were registered on
+ * behalf of a particular stylesheet stay in the index and are filtered out at
+ * match time for scopes that never loaded it - see `matchSelectorCandidates`.
```
farfromrefug: same thing seems far too long if even necessary

packages/core/ui/styling/style-scope.ts
```diff
@@ -57,6 +75,34 @@ const kebabCasePattern = /-([a-z])/g;
 const kebabCaseReplacementFunc = (g: string) => g[1].toUpperCase();
 const pattern = /('|")(.*?)\1/;
 
+/**
+ * Resolve a pending-substitution value into the longhand values it stands for.
+ *
+ * The shorthand is only parsed here, once the expression it holds has been
+ * evaluated against the view - which is the point of the placeholder. A shorthand
+ * that does not survive evaluation leaves its longhands unset.
```
farfromrefug: could be shortened

packages/core/ui/styling/style-scope.ts
```diff
@@ -560,9 +724,18 @@ export class CssState {
 	public onChange(): void {
 		const view = this.viewRef.get();
 		if (view && view.isLoaded) {
-			this.unsubscribeFromDynamicUpdates();
+			// Matching does not read the subscriptions, so the new match can be computed
+			// first and its dependencies compared against the ones already subscribed to.
+			// They are usually identical, and tearing every listener down only to add the
+			// same ones back is pure churn - it also toggles the widget-level pseudo class
+			// handlers off and on for nothing.
```
farfromrefug: too long

packages/core/ui/styling/style-scope.ts
```diff
@@ -716,51 +889,141 @@ export class CssState {
 		// Update values for the scope's css-variables
 		view.style.resetScopedCssVariables();
 
-		const valuesToApply = {};
-		const cssExpsProperties = {};
+		// These stay empty in the common case (nothing changed, no css expressions), so
+		// they are only allocated once there is something to put in them.
```
farfromrefug: not needed

packages/core/ui/styling/style-scope.ts
```diff
@@ -716,51 +889,141 @@ export class CssState {
...
+			const oldValue = hadOldValue ? oldProperties[property] : undefined;
+			if (hadOldValue) {
+				delete oldProperties[property];
+			}
+
+			if (value === unsetValue) {
 				delete newPropertyValues[property];
+			} else {
+				// Remember the resolved value so the next update can tell whether the
+				// shorthand still resolves to what is currently applied.
```
farfromrefug: dont think we need this. Feels like there are too many comments in general

packages/core/ui/styling/css-selector.ts
edusperoni: You're right. I did this spike as an experiment on mobile which doesn't have my usual instructions around comments. Should be fixed in a bit

NathanWalker (approved):

PR #11406, opened 2026-09-02
fix(core): insert iOS child views relative to their sibling, not by raw index

NathanWalker:
On iOS, `insertChild(view, index)` can place the child's native view one position lower than asked, beneath the sibling it should sit above. A view inserted after a `ScrollView` ends up underneath it and stops receiving touches while remaining fully visible; a view inserted before an opaque sibling disappears behind it.

Root cause: every iOS `_addViewToNativeVisualTree` implementation used `insertSubview:atIndex:`. UIKit resolves that index against the layer's **sublayers**, not the `subviews` array, and core installs non-view sublayers of its own:

- a CSS `linear-gradient` background is a `CAGradientLayer` at sublayer 0 (`background.ios.ts`, also `RootLayout`)
- each child with `box-shadow` gets an outer shadow layer inserted below its own layer

With any of those ahead of the insertion point, the raw index is off by the number of such layers. Appending (`addChild`) is unaffected, which is why this only shows up when a child is inserted in the middle: keyed list updates, `@for`/`v-for` prepends, conditional views that mount after their siblings, framework HMR remounts.

### Fix

New `IOSHelper.insertSubview(parentNativeView, childNativeView, atIndex?)` appends when the index is absent or past the end and otherwise inserts with `insertSubview:belowSubview:` against the subview currently at that index, which is independent of extra sublayers. All four raw-index sites use it:

- `View._addViewToNativeVisualTree`
- `Page._addViewToNativeVisualTree`
- `LiquidGlass._addViewToNativeVisualTree`
- `LiquidGlassContainer._addViewToNativeVisualTree`

The `atIndex` contract is unchanged (a subview index, as `ProxyViewContainer` and `_childIndexToNativeChildIndex` already assume), so callers are untouched.

### DIFF
--- a/packages/core/ui/core/view/view-helper/index.ios.ts
+++ b/packages/core/ui/core/view/view-helper/index.ios.ts
@@ -312,6 +312,23 @@ export class IOSHelper {
 		return rootView.safeAreaLayoutGuide;
 	}
 
+	/**
+	 * Add `childNativeView` to `parentNativeView` at subview index `atIndex`, or
+	 * append it when the index is absent or past the end. Never done with
+	 * `insertSubview:atIndex:`: UIKit resolves that index against the layer's
+	 * sublayers, which also hold the non-view layers core installs (a gradient
+	 * background at sublayer 0, an outer shadow layer per shadowed child), so the
+	 * view lands below the sibling it should precede.
+	 */
+	static insertSubview(parentNativeView: UIView, childNativeView: UIView, atIndex?: number): void {
+		const subviews = parentNativeView.subviews;
+		if (typeof atIndex !== 'number' || atIndex >= subviews.count) {
+			parentNativeView.addSubview(childNativeView);
+		} else {
+			parentNativeView.insertSubviewBelowSubview(childNativeView, subviews.objectAtIndex(atIndex));
+		}
+	}
+
 	static layoutView(controller: UIViewController, owner: View): void {
 		let layoutGuide = controller.view.safeAreaLayoutGuide;
 		if (!layoutGuide) {
--- a/packages/core/ui/core/view/index.ios.ts
+++ b/packages/core/ui/core/view/index.ios.ts
@@ -1270,11 +1270,7 @@ export class CustomLayoutView extends ContainerView {
 		const childNativeView: NativeScriptUIView = <NativeScriptUIView>child.nativeViewProtected;
 
 		if (parentNativeView && childNativeView) {
-			if (typeof atIndex !== 'number' || atIndex >= parentNativeView.subviews.count) {
-				parentNativeView.addSubview(childNativeView);
-			} else {
-				parentNativeView.insertSubviewAtIndex(childNativeView, atIndex);
-			}
+			IOSHelper.insertSubview(parentNativeView, childNativeView, atIndex);
 
 			// Add outer shadow layer manually as it belongs to parent layer tree (this is needed for reusable views)
 			if (childNativeView.outerShadowContainerLayer && !childNativeView.outerShadowContainerLayer.superlayer) {
--- a/packages/core/ui/core/view/view-helper/index.d.ts
+++ b/packages/core/ui/core/view/view-helper/index.d.ts
@@ -66,6 +66,11 @@ export namespace IOSHelper {
 	export function invalidateStatusBarAppearance(controller?: any /* UIViewController */, reason?: string): void;
 	export function updateAutoAdjustScrollInsets(controller: any /* UIViewController */, owner: View): void;
 	export function updateConstraints(controller: any /* UIViewController */, owner: View): void;
+	/**
+	 * Add `childNativeView` to `parentNativeView` at subview index `atIndex`, or append it when the index is absent or past the end.
+	 * Uses `insertSubview:belowSubview:` — `insertSubview:atIndex:` resolves the index against the layer's sublayers, which also hold non-view layers (gradient backgrounds, shadow layers).
+	 */
+	export function insertSubview(parentNativeView: any /* UIView */, childNativeView: any /* UIView */, atIndex?: number): void;
 	export function layoutView(controller: any /* UIViewController */, owner: View): void;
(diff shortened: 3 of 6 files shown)


CatchABus:
@NathanWalker This generated description seems really strange.
If there's a problem with box-shadow layer you are facing, we can discuss it since I'm already working on taking down this implementation in favor of a better one that doesn't nest the layer in the parent.
Is there any sample around I could try?

NathanWalker:
@CatchABus thanks for the look; this is not a box-shadow problem (the fix doesn't touch the shadow code) sorry the description buried that. The shadow layer is only one of the extra sublayers that trigger it; the case I actually hit is a CSS linear-gradient background, which core installs as a CAGradientLayer at sublayer 0.

The bug is in how children get inserted: every iOS _addViewToNativeVisualTree uses insertSubview:atIndex:, and UIKit resolves that index against the layer's sublayers, not subviews. So with any non-view sublayer ahead of the insertion point, insertChild(view, 1) lands the view at subview 0. Appends are fine, which is why it only shows for inserts in the middle. insertSubview:belowSubview: doesn't have the problem, which is all the PR changes.

Your shadow rework would remove one source of extra sublayers, but not the gradient layer (or anything a plugin adds), so I think the insert fix stands on its own. 

Sample to repro on your end: 
```
// Drop this file into any NativeScript app (plain TypeScript, no framework needed)
// and call `runInsertOrderRepro(page)` from the page's `loaded` handler:
//
//   <Page loaded="onLoaded"> ... </Page>
//   export function onLoaded(args) { runInsertOrderRepro(args.object); }
//
//   Currently:
//   [nativescript] child order:  a, x, b, c
//   [uikit]  subviews (gradient): x, a, b, c     <- wrong, x went below a
//   [uikit]  subviews (plain):    a, x, b, c     <- the same insert with no extra sublayer
//
// With the fix (insertSubview:belowSubview:), the gradient case prints a, x, b, c too.

import { GridLayout, Label, Page } from '@nativescript/core';

export function runInsertOrderRepro(page: Page): void {
	page.addCss('.repro-gradient { background: linear-gradient(to bottom, #ffffff, #f4f4f5); }');

	const grid = new GridLayout();
	grid.className = 'repro-gradient';
	grid.height = 120;
	const label = (text: string) => {
		const l = new Label();
		l.text = text;
		l.height = 20;
		return l;
	};
	grid.addChild(label('a'));
	grid.addChild(label('b'));
	grid.addChild(label('c'));
	page.content = grid;

	setTimeout(() => {
		grid.insertChild(label('x'), 1);

		const children: string[] = [];
		grid.eachChildView((child) => {
			children.push((child as Label).text);
			return true;
		});
		console.log('[nativescript] child order: ', children.join(', '));
		console.log('[uikit] subviews (gradient):  ', subviewTexts(grid.ios as UIView));

		// Pure UIKit, no NativeScript involved: an extra sublayer ahead of the
		// children shifts insertSubview:atIndex: by one.
		const plain = UIView.alloc().initWithFrame(CGRectMake(0, 0, 100, 100));
		const decorated = UIView.alloc().initWithFrame(CGRectMake(0, 0, 100, 100));
		decorated.layer.insertSublayerAtIndex(CAGradientLayer.new(), 0);
		for (const parent of [plain, decorated]) {
			for (const tag of [1, 2, 3]) parent.addSubview(taggedView(tag));
			parent.insertSubviewAtIndex(taggedView(9), 1);
		}
		console.log('[uikit] insertSubview:atIndex:1, plain parent:    ', subviewTags(plain));
		console.log('[uikit] insertSubview:atIndex:1, extra sublayer:  ', subviewTags(decorated));
	}, 100);
}

function taggedView(tag: number): UIView {
	const view = UIView.alloc().initWithFrame(CGRectMake(0, 0, 10, 10));
	view.tag = tag;
	return view;
}

function subviewTexts(view: UIView): string {
	const out: string[] = [];
	for (let i = 0; i < view.subviews.count; i++) {
		out.push((view.subviews.objectAtIndex(i) as UILabel).text);
	}
	return out.join(', ');
}

function subviewTags(view: UIView): string {
	const out: number[] = [];
	for (let i = 0; i < view.subviews.count; i++) {
		out.push(view.subviews.objectAtIndex(i).tag);
	}
	return out.join(', ');
}
```

packages/core/ui/core/view/view-helper/index.ios.ts
```diff
@@ -312,6 +312,23 @@ export class IOSHelper {
...
 
+	/**
+	 * Add `childNativeView` to `parentNativeView` at subview index `atIndex`, or
+	 * append it when the index is absent or past the end. Never done with
+	 * `insertSubview:atIndex:`: UIKit resolves that index against the layer's
+	 * sublayers, which also hold the non-view layers core installs (a gradient
+	 * background at sublayer 0, an outer shadow layer per shadowed child), so the
+	 * view lands below the sibling it should precede.
+	 */
+	static insertSubview(parentNativeView: UIView, childNativeView: UIView, atIndex?: number): void {
```
CatchABus, line 323: Maybe we could rename the function to `insertNativeSubview` to make it distinct for users and plugin maintainers that it doesn't target {N} JS views.

Also, it would be nice to have an android counterpart which could contain the `View` class implementation by default.

packages/core/ui/core/view/view-helper/index.ios.ts
```diff
@@ -312,6 +312,23 @@ export class IOSHelper {
 		return rootView.safeAreaLayoutGuide;
 	}
 
+	/**
+	 * Add `childNativeView` to `parentNativeView` at subview index `atIndex`, or
```
CatchABus, line 316: The comment needs shrinking and correction.

CatchABus (changes requested):

CatchABus:
@NathanWalker Thanks for the detailed response, it makes more sense now!
Thanks for the sample you created too, I made sure to try it.

I requested a couple of changes and suggestions.

PR #11414, opened 2026-09-03
feat(core): native update scheduler and commit hook (phase 0)

edusperoni:
## PR Checklist

A property setter *is* the native call. Each `Property`/`CssProperty`/`CssAnimationProperty`/`InheritedCssProperty` setter carries its own copy of the "suspended → record in `_suspendedUpdates`, else `[setNative](value)`" block (six copies). The first native pass happens at `onLoaded` in the order properties were first *set* on that instance; afterwards every set is one synchronous native call. There is no way for a widget to see "these properties changed together", no previous committed value, no public batching primitive (`_batchUpdate` is private and per node), and no way for a flavor to coalesce writes across nodes.

## What is the new behavior?

Phase 0 of the native-updates redesign: **additive plumbing with no behavior change**. The new API is implemented on top of the existing storage (`_suspendNativeUpdatesCount` as the holds, `_suspendedUpdates` as the dirty set), so every existing code path produces the same native calls, in the same order, with the same `<name>Change` timing. Six commits, each green on its own; review commit by commit:

1. `test(core): add native-update oracle specs` — golden `[setNative]`/`<name>Change` sequences for every setter kind, load/unload/reload, batch, reset, shorthand, inheritance, keyframes. These specs are byte-identical from commit 1 to HEAD and pass at every commit.
2. `refactor(core): unify native property routing` — the six copies collapse into one routing function; `applyPendingNativeSetters`/`applyAllNativeSetters` share the default-capture helper.
3. `feat(core): add commitNativeUpdates hook and NativeUpdateBatch` — `ViewBase.commitNativeUpdates(batch)` is the ordering hook (override, apply what the class cares about, `super` applies the rest). `onResumeNativeUpdates` → `initNativeView` builds the batch and calls it. `NativeUpdateBatch`: `entries/has/apply/skip/previous/isMount/children` (an entry is a property or an aggregate `Invalidation`; the name avoids "target", which means a view elsewhere in core). Previous values are recorded only for nodes that can read them back (a class overriding the hook, or a property declaring `invalidates`), so default nodes allocate nothing extra in a batch or across unload/reload.
4. `feat(core): add NativeUpdates scheduler and flushNativeUpdates` — `NativeUpdates.batch()/begin()/end()/flush()` coalesce writes per node across nodes (implemented as `_batchUpdate` semantics per touched node); `mode` is `'sync'` only. `ViewBase.flushNativeUpdates({ subtree, force })`; `force` pushes to an existing native view while the node is not loaded and leaves the holds in place.
5. `feat(core): add invalidates option to properties` — `PropertyOptions.invalidates?: Invalidation[]` on every property kind; a node handles an aggregate with `[invalidation.apply](batch)`. No core property declares one yet.
6. `perf(core): commit without a batch for nodes that cannot observe one` — a node whose class uses the default hook and has no pending aggregate invalidation cannot observe a `NativeUpdateBatch`, so its commit runs the existing sweeps directly (same order) instead of building one.

**Fast path.** A live set on a class that does not override `commitNativeUpdates`, with no aggregate invalidations and no open batch, still calls `[setNative]` straight from the setter with no allocation.

**Legacy surface kept** (names, types, exports, semantics): `_suspendNativeUpdatesCount`, `_suspendedUpdates`, `SuspendType` (moved to `view-base/suspend-type.ts`, still internal), `_batchUpdate`, `_suspendNativeUpdates`, `_resumeNativeUpdates`, `onResumeNativeUpdates`, `initNativeView`, `applyPendingNativeSetters`, `applyAllNativeSetters`, `affectsLayout` timing.

### Points for reviewers

- **Cost, measured jitless (`VITEST_NO_OPT=1`, interpreter-only like iOS) with a no-op `setNative`, medians.** Live setter: roughly +40–60 ns per set from collapsing six inlined copies into one routing call (JIT: within noise); a real native write is microseconds, so the relative impact in apps is far smaller. Commit path (per view, µs), base → without commit 6 → with commit 6:

  | scenario | base | c5 | c6 |
  |---|---|---|---|
  | load: construct + 8 props + `_setupUI` + `callLoaded` | 7.8 | 10.2 | 9.4 |
  | reload: unload + 4 props + load | 4.3 | 5.7 | 5.0 |
  | `_batchUpdate`, 8 props, live view | 3.44 | 5.20 | 4.12 |

  The residual over base is two helper calls per property in the sweep; inlining the default-capture into those helpers would recover most of it at the cost of duplicating it, deliberately not done in Phase 0. The setter numbers live in `native-updates-hot-path.spec.ts`, the commit-path ones in `native-updates-commit-cost.spec.ts`; both print on every run.
- **`invalidates` is additive**: a property that declares it still runs its own `[setNative]`, then the aggregates. Whether aggregates should *replace* the property's own setter is a Phase 1 decision.
- **Dirty state is cleared before the commit runs**, so a write made from a handler queues against a fresh set (previously the set was replaced after the sweep). Only observable if a `setNative` handler runs `_batchUpdate` on the same node mid-sweep.
- **`Property.set` with an override present** routes through the commit and uses pending semantics (value re-read at commit). Only affects classes that opt in.
- **Device tests** (`apps/automated/src/ui/lifecycle/lifecycle-tests.ts`: `NativeUpdates.batch` equals `_batchUpdate` counts, flush inside a batch, forced flush while unloaded) are added but were not run here. The existing lifecycle suite there asserts exact `setNative` counts through XML inflation, `_batchUpdate`, navigation and CSS class changes and is the production check for "same native calls". To run on iOS and Android before undrafting.

Tests: 485 → 557 passing (`npx vitest run` in `packages/core`), no pre-existing failures. Typecheck: `npx tsc -p packages/core/tsconfig.lib.json --noEmit` clean.

Related: #11412 (`origin` on `PropertyChangeData`, independent), #11413 (`InheritedProperty` reset cascade fix, found on the way).

### DIFF
--- a/packages/core/ui/core/native-updates/batch.ts
+++ b/packages/core/ui/core/native-updates/batch.ts
@@ -1,0 +1,117 @@
+import type { ViewBase } from '../view-base';
+import type { CssAnimationProperty, CssProperty, Property } from '../properties';
+import type { Invalidation } from './invalidation';
+
+/** A change to a node's child list. Phase 0 never records one. */
+export interface ChildMutation {
+	readonly kind: 'insert' | 'remove' | 'move';
+	readonly child: ViewBase;
+	readonly index: number;
+}
+
+export type NativeUpdateProperty = Property<any, any> | CssProperty<any, any> | CssAnimationProperty<any, any>;
+
+/** What a commit applies: a property, through its `[setNative]`, or an aggregate invalidation. */
+export type NativeUpdateEntry = NativeUpdateProperty | Invalidation;
+
+export type NativeUpdateApplier = (batch: NativeUpdateBatch, entry: NativeUpdateEntry) => void;
+
+const NO_CHILDREN: readonly ChildMutation[] = Object.freeze([]);
+
+/**
+ * Everything that is dirty on one node at commit time, in the order the commit will apply it.
+ * Handed to `ViewBase.commitNativeUpdates`, which may reorder, pre-empt or drop entries.
+ */
+export class NativeUpdateBatch {
+	/** Every value the node carries is dirty: it is being applied to a native view for the first time. */
+	public readonly isMount: boolean;
+	public readonly children: readonly ChildMutation[] = NO_CHILDREN;
+
+	private readonly _entries: NativeUpdateEntry[];
+	private readonly _handled: boolean[];
+	private readonly _apply: NativeUpdateApplier;
+	private readonly _previous: Map<NativeUpdateProperty, unknown> | undefined;
+	private _index: Map<NativeUpdateEntry, number> | undefined;
+
+	constructor(
+		public readonly node: ViewBase,
+		isMount: boolean,
+		entries: NativeUpdateEntry[],
+		previous: Map<NativeUpdateProperty, unknown> | undefined,
+		apply: NativeUpdateApplier,
+	) {
+		this.isMount = isMount;
+		this._entries = entries;
+		this._handled = new Array(entries.length).fill(false);
+		this._previous = previous;
+		this._apply = apply;
+	}
+
+	/** Everything this commit applies, in commit order; `has()` tells whether an entry is still pending. */
+	public get entries(): ReadonlyArray<NativeUpdateEntry> {
+		return this._entries;
+	}
+
+	/** Whether the entry is still waiting to be applied by this commit. */
+	public has(entry: NativeUpdateEntry): boolean {
+		const index = this.indexOf(entry);
+
+		return index >= 0 && !this._handled[index];
+	}
+
+	/** The value the property held at the last commit; `undefined` on a mount. */
+	public previous<T>(property: NativeUpdateProperty): T | undefined {
+		return this._previous?.get(property) as T | undefined;
+	}
+
+	/** Runs the entry's handler now. An entry that is not pending, or already handled, is a no-op. */
+	public apply(entry: NativeUpdateEntry): void {
+		const index = this.indexOf(entry);
+		if (index < 0 || this._handled[index]) {
+			return;
+		}
+
+		this._handled[index] = true;
+		this._apply(this, this._entries[index]);
+	}
+
+	/** Marks the entry handled without touching the native view. */
+	public skip(entry: NativeUpdateEntry): void {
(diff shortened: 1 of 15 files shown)


packages/core/ui/core/native-updates/batch.ts
```diff
@@ -0,0 +1,112 @@
...
+export interface ChildMutation {
+	readonly kind: 'insert' | 'remove' | 'move';
+	readonly child: ViewBase;
+	readonly index: number;
+}
+
+export type NativeUpdateProperty = Property<any, any> | CssProperty<any, any> | CssAnimationProperty<any, any>;
+
+/** What a commit applies: a property, through its `[setNative]`, or an aggregate invalidation. */
+export type NativeUpdateTarget = NativeUpdateProperty | Invalidation;
```
farfromrefug: i have doubt about the name. Target often refer to an View in Nativescript. at first i could not find where the "property" array was in NativeUpdateBatch. Could we rename it to something more obvious?

packages/core/ui/core/native-updates/batch.ts
```diff
@@ -0,0 +1,112 @@
...
+ */
+export class NativeUpdateBatch {
+	/** Every value the node carries is dirty: it is being applied to a native view for the first time. */
+	public readonly isMount: boolean;
+	public readonly children: readonly ChildMutation[] = NO_CHILDREN;
+
+	private readonly _targets: NativeUpdateTarget[];
+	private readonly _handled: boolean[];
+	private readonly _apply: NativeUpdateApplier;
+	private readonly _previous: Map<NativeUpdateProperty, unknown> | undefined;
```
farfromrefug, line 33: what actually is the use for previous? i cant see where it is actually used

edusperoni: It's not used at this moment by anything mostly because it's not present, but it allows you to do something like:
```
if(computeMultipleProps(x, y, z) !== computeMultipleProps(oldX, oldY, oldZ)) {
  // apply native only when changed
}
```

or something like:
```
const myTabItems = [x,y,z]
const newTabItems = [y,z]

applyTabItems(new, old) {
  // only need to remove x. y and z remain the same with no changes
}
```

PR #11416, opened 2026-09-04
fix(ios): do not assign the 'spring' curve as a timing function

farfromrefug:
## Summary

Any view animated with `curve: 'spring'` crashes the app on iOS.

`_createBasicAnimation` uses `curve` to pick `CASpringAnimation`, then falls through to `basicAnimation.timingFunction = curve` and assigns the **string** `'spring'` where CoreAnimation expects a `CAMediaTimingFunction`. The assignment is silent. The abort comes later, when the animation is committed:

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException',
reason: '-[NSTaggedPointerString _getPoints:]: unrecognized selector sent to instance'

-[CASpringAnimation _copyRenderAnimationForLayer:]
CA::Layer::commit_animations(...)
_UIApplicationFlushCATransaction
```

Because it throws at commit time rather than at the call site, the crash lands on whatever frame comes next — which makes it look unrelated to the animation that caused it.

Every other curve value is a real `CAMediaTimingFunction`, so only `'spring'` is affected.

## Testing

Reproduced in an app that animates a slider thumb with `curve: 'spring'`: consistent `SIGABRT` at the next CoreAnimation commit, matching the trace above. With this change the spring animation runs and the crash is gone.

`CASpringAnimation` still gets its `damping`, `duration`, `fromValue` and `toValue` as before — only the bogus `timingFunction` assignment is skipped.


farfromrefug:
Opened against the wrong repository by mistake — this targets the Akylas/NativeScript fork, whose `_createBasicAnimation` differs from upstream. Reopening there. Sorry for the noise.

PR #11426, opened 2026-09-09
fix(vite): shim package roots whose export * targets another bundled package

### LINKED ISSUE #11421: [Vite, Vue, HMR] per-module-served plugin subpath that imports `nativescript-vue` bare evaluates the real package root and double-runs `init()` (Element for AbsoluteLayout already registered)
### Issue Description

Runtime error (dev server / HMR only). `@nativescript/vite` **8.0.5** (same on 8.0.4) · `nativescript-vue` 3.0.3 · `@nativescript/core` 9.1.1 · `vite` 8.2.2 · `@nativescript/ios` 9.1.0 · iOS simulator (iPhone 17 Pro Max).

Scope: iOS simulator only; not tested on Android or a physical device.

A blank Vue app boots fine under HMR. It stops booting the moment I import one plugin subpath whose file is **served per-module** (not shimmed into the deps bundle) and contains a bare `import ... from 'nativescript-vue'`. I only tested that one plugin, but the routing below suggests any file with that shape would do the same.

Bare `nativescript-vue` is mapped in the device import map to the **real package root** (`/ns/m/node_modules/nativescript-vue`), unlike `@nativescript/core`, which is mapped to the `/ns/core` bridge. (Every other package root maps the way `nativescript-vue` does — `@nativescript/core` is the special case.) The real root cannot be shimmed (its `export * from '@vue/runtime-core'` is a bare star), so importing it *evaluates* it — and its module-scope `init()` runs a second time, on top of the copy the deps bundle already evaluated:

```
Error: Element for AbsoluteLayout already registered.
If this is intentional set 'overwriteExisting: true' in 'meta'
    at registerElement       (/ns/deps-bundle.mjs)
    at registerCoreElements  (/ns/deps-bundle.mjs)
    at init                  (/ns/deps-bundle.mjs)
    at                       (/ns/m/node_modules/nativescript-vue:41:1)
```

Concrete trigger: `@nativescript-community/ui-collectionview/vue3`. There is a config-only workaround (`NS_VENDOR_INCLUDE`, below), so this is not a blocker — but there may be a similar case to cover for Vue to what `manifest-collect` already does for Solid.

Expected behavior: importing a plugin's Vue entry should share the one `nativescript-vue` instance the deps bundle already evaluated, the way app-code imports (rewritten to `/ns/rt`) do.

A second, separate bridge problem surfaces once this one is worked around: #11422 (`$navigateTo` "app navigator missing" race).

Line numbers below refer to the **published `@nativescript/vite` 8.0.5 package** (compiled `.js` in `node_modules`) unless marked `.ts`, which refers to `packages/vite` on `main`.

### Reproduction

Minimal repro (blank app, two steps):

1. `ns create` a blank Vue TS app (template-blank-vue 9.1.0 — it ships webpack), then `npx nativescript-vite init` (adds `@valor/nativescript-websockets` plus the two config keys). **Boots under HMR, and live HMR works** — a `Home.vue` edit applies on device.
2. `npm i @nativescript-community/ui-collectionview@6.0.25`, and in `app/app.ts`:

```ts
import { createApp } from 'nativescript-vue';
import CollectionView from '@nativescript-community/ui-collectionview/vue3';
import Home from './components/Home.vue';

const app = createApp(Home);
app.use(CollectionView);
app.start();
```

3. `ns debug ios` → **crashes** with the error above.

Controls, same tree:

| Variant | Result |
|---|---|
| plugin installed but **not** imported | boots |
| `pinia` installed **and** `app.use(createPinia())` | boots |
| plugin imported (as above) | crashes |
| `ns debug ios --no-hmr` | boots |

So the trigger is exactly that one import.

<details>
<summary><b>Mechanism — steps 1-3 from my run's manifest, the served file and the import map; steps 4-5 from reading the shipped code</b></summary>

1. **The `vue3/` subpath isn't in the bundle.** In my run's deps-bundle manifest (`node_modules/.ns-vite/deps-bundle-ios-*.json`), the only keys under `@nativescript-community/ui-collectionview` were `index.ios.js` and `index-common.js` — its root entry closure. `vue3/index.js` and `vue3/component.js` are not bundle keys, so no shim is emitted and they are served as real per-module ESM.

2. **Their `nativescript-vue` import is not rewritten.** Served `vue3/component.js` line 2 is a bare `import { NSVRoot, defineComponent, getCurrentInstance, h, ref, render, toRaw, watch } from "nativescript-vue";` — while line 1 of the *same served file* was rewritten to `import __ns_core_ns1 from ".../ns/core"`. App-code imports of `nativescript-vue` are rewritten to `/ns/rt`; this node_modules file's are not.

3. **The import map sends it to the real root.** `/ns/import-map.json` → `importMap.imports` (1987 entries) maps `"nativescript-vue"` → `http://<host>:5173/ns/m/node_modules/nativescript-vue`, but `"@nativescript/core"` → `http://<host>:5173/ns/core`. That asymmetry is what routes the bare import at the real package root.

4. **The real root can't be shimmed.** `nativescript-vue/dist/index.js:16` is `export * from '@vue/runtime-core'` — a bare star. `collectDepsModuleExportInfo` returns `names: null` for a non-relative `export *` (`hmr/server/deps-bundle.js:781-785`; `.ts` ~`:798-813`), and the shim site (`hmr/server/deps-bundle.js:920-938`) emits a shim only when `names !== null` or `opaqueCjs` is set. Neither holds → per-module serving → the URL **evaluates** the file, and `dist/index.js:10` `init()` runs.

5. **The bundle already ran `init()`.** Every vendor file gets a static namespace import in the synthetic entry (`hmr/server/deps-bundle.js:310-322`), tree-shaking is off (`hmr/server/deps-bundle.js:631-633`; `.ts:658`, comment: "dep packages rely on registration side effects"), and the deps bundle is evaluated as part of the dev-client graph, imported before the app entry (`hmr/shared/runtime/session-bootstrap.js:276` vs `:304`). `init()` → `registerCoreElements()` (`nativescript-vue/dist/nativescript/index.js:3-4`) → `registerElement('AbsoluteLayout')` (`dist/nativescript/elements.js:6`) into a module-local `elementMap` (`dist/registry/index.js:4`) that throws on a repeat (`dist/registry/index.js:39-45`).

The served root's `init()` resolves `registerCoreElements` through `dist/nativescript/index.js`, which **is** served as a shim into the bundle — so the second call writes into the already-populated map and throws.

</details>

<details>
<summary><b>What it is <i>not</i> (all tested)</b></summary>

- Not `dependencies` vs `devDependencies` — `nativescript-vue` is in `dependencies`, and the blank-template layout matches the failing app.
- Not whether `vue` is in the vendor set — a variant with `vue` in `vendorSpecToKey` and 17 `@vue/*` keys boots.
- Not `pinia` — installed and used, boots.
- Not duplicate Vue copies — `@vue/*` resolves to a single hoisted copy, no nested duplicates.
- Not Vite's dep optimizer — `optimizeDeps.noDiscovery`/`exclude` changed nothing.
- Not stale caches — `ns clean` + reinstall reproduces identically.

</details>

<details>
<summary><b>Workaround (config only)</b></summary>

```
NS_VENDOR_INCLUDE=@nativescript-community/ui-collectionview/vue3 ns debug ios
```

Exactly that bare subpath. `addCandidate` accepts subpaths (`hmr/shared/vendor/manifest-collect.js:465-471` — adds to `vendor`, and skips only the *dependency* traversal for non-root specifiers; the env hook is at `:629`, `.ts:616`). `resolveSpecToFile` (`hmr/server/deps-bundle.js:184-200`) falls through to the `/index<ext>` candidate fan-out (`:109-114`) and lands on `vue3/index.js`; esbuild pulls `component.js` into the closure. What I observed directly: with that one env var the manifest gains `vue3/index.js` and `vue3/component.js` as keys, the served `component.js` becomes a `__NS_DEPS_MODULES__` shim, and the app boots.

Verified: the repro boots, live HMR works, and on a real app (larger, same stack) the `already registered` error is gone.

</details>

<details>
<summary><b>Suggested direction (there's precedent in the plugin's own code)</b></summary>

`hmr/shared/vendor/manifest-collect.js:614-628` (`.ts:601-611`) vendors `@tanstack/solid-router`, `@tanstack/router-core` and `@tanstack/history` when `@nativescript/tanstack-router` is a dependency, for three reasons — the third of which is this class of problem: *"Import solid-js internally — loading via HTTP creates a separate solid-js instance"*. The Vue flavor has no equivalent, and given the import map above, any per-module-served node_modules file that imports `nativescript-vue` bare would resolve to the real root the same way (only the collectionview subpath was actually exercised).

Options, in rough order of how targeted they seem:

- Route bare `nativescript-vue` in the import map to the bridge/bundle, the way `@nativescript/core` → `/ns/core` already is.
- Auto-vendor node_modules files whose source imports `nativescript-vue`.
- Make the root shim-able despite the bare `export *` (resolve the star target instead of returning `names: null`).

Happy to test a patch on the repro.

</details>

```shell
Error: Element for AbsoluteLayout already registered.
If this is intentional set 'overwriteExisting: true' in 'meta'
    at registerElement       (/ns/deps-bundle.mjs)
    at registerCoreElements  (/ns/deps-bundle.mjs)
    at init                  (/ns/deps-bundle.mjs)
    at                       (/ns/m/node_modules/nativescript-vue:41:1)
```

### Environment

```yaml
OS: macOS 26.6.2
CPU: (16) arm64 Apple M3 Max
Shell: /bin/zsh
node: 24.21.0
npm: 11.19.0
nativescript: 9.1.0 (project-local)

# ios
xcode: 26.6/17F113
cocoapods: 1.14.2
platforms:
  - iOS 26.5
```

```json
"dependencies": {
  "@nativescript/core": "~9.1.0",
  "nativescript-vue": "~3.0.3",
  "@valor/nativescript-websockets": "^2.0.2",
  "@nativescript-community/ui-collectionview": "6.0.25"
},
"devDependencies": {
  "@nativescript/ios": "~9.1.0",
  "@nativescript/tailwind": "^4.0.0",
  "@nativescript/types": "~9.1.0",
  "@nativescript/vite": "~8.0.4",
  "@types/node": "^20.0.0",
  "@vitejs/plugin-vue": "^6.0.8",
  "@vitejs/plugin-vue-jsx": "^5.1.6",
  "nativescript": "~9.1.0",
  "tailwindcss": "^4.0.0",
  "typescript": "~6.0.0",
  "vite": "^8.2.2",
  "vue-tsc": "^3.3.9"
}
```

### Please accept these terms

- [x] This is a bug report

sitefinitysteve:
## PR Checklist

The Vue server strategy pins bare `nativescript-vue` in the device import map to `/ns/m/node_modules/nativescript-vue`, expecting that URL to serve the deps-bundle shim. It doesn't: `collectDepsModuleExportInfo` returns `names: null` for the package root because `nativescript-vue/dist/index.js` contains a bare `export * from '@vue/runtime-core'`, and `getShimForSpec` only emits a shim when the export surface is enumerable. The real file is served per-module instead, so any node_modules file served per-module that imports `nativescript-vue` bare (for example `@nativescript-community/ui-collectionview/vue3`) evaluates the root a second time, and its module-scope `init()` re-registers the core elements:

```
Error: Element for AbsoluteLayout already registered.
```

Repro, controls and the `NS_VENDOR_INCLUDE` workaround are in #11421.

## What is the new behavior?

`generateDepsBundle` now records, from esbuild's discovery metafile, how the bundle actually resolved every bare import edge: importer key → bare specifier → target key (`bareImportEdges`, persisted in the disk cache; schema bumped so existing caches rebuild). `collectDepsModuleExportInfo` takes an optional resolver for bare `export *` targets, and the bundle service backs it with those recorded edges, so the shim enumerates exactly the file esbuild bundled for that importer — no filesystem guessing, no condition lists. If there is no recorded edge, or no resolver, the collector bails to `names: null` as before and per-module serving is unchanged. With this, `nativescript-vue`'s root gets a real shim and the bare specifier lands in the single bundle realm the strategy already assumes.

Specs: an end-to-end `createDepsBundleService` case with a fixture root that does `export * from 'pkg-i'` asserts the served shim re-exports both its own and the target's names through real esbuild; the recorded edges are asserted and round-tripped through the disk cache; collector unit cases cover resolver hit, no edge, and an opaque-CJS target (bails).

Residual risks worth knowing, none new to this change: registry keys collapse nested and hoisted copies of a package to one identity (pre-existing model for every shim); an ambiguous name across two `export *` targets is unioned and reads as `undefined` if absent from the bundled namespace, the same behaviour relative stars already have; and the guard that skips namespaced (`:`) metafile targets has no dedicated spec.

Verified on device with the #11421 repro (blank Vue app importing `@nativescript-community/ui-collectionview/vue3`, no `NS_VENDOR_INCLUDE`): the published 8.0.5 crashes with the error above; this build boots cold and from the rebuilt disk cache, the served `nativescript-vue` URL is the bundle shim, and live edits hot-reload.

Fixes #11421.

### DIFF
--- a/packages/vite/hmr/server/deps-bundle.spec.ts
+++ b/packages/vite/hmr/server/deps-bundle.spec.ts
@@ -1,9 +1,9 @@
-import { existsSync, mkdirSync, mkdtempSync, readdirSync, realpathSync, rmSync, writeFileSync } from 'node:fs';
+import { existsSync, mkdirSync, mkdtempSync, readdirSync, readFileSync, realpathSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import * as path from 'node:path';
 import { afterAll, describe, expect, it } from 'vitest';
 
-import { DEPS_BUNDLE_PATH, buildDepsBundleEntryCode, buildDepsCandidateSpecs, buildDepsShimCode, buildDepsVendorRuntimeModule, collectDepsModuleExportInfo, computeDepsBundleCacheKey, createDepsBundleService, depsRegistryKeyForFile, generateDepsBundle, isDepsPerModuleServingEnabled, resolveDepsEntriesFromRecording, resolveDepsEntriesFromVendorCollection } from './deps-bundle.js';
+import { DEPS_BUNDLE_PATH, buildDepsBundleEntryCode, buildDepsCandidateSpecs, buildDepsShimCode, buildDepsVendorRuntimeModule, collectDepsModuleExportInfo, computeDepsBundleCacheKey, createDepsBundleService, depsRegistryKeyForFile, generateDepsBundle, isDepsPerModuleServingEnabled, resolveDepsEntriesFromRecording, resolveDepsEntriesFromVendorCollection, tryLoadDepsBundleFromDisk } from './deps-bundle.js';
 
 describe('isDepsPerModuleServingEnabled', () => {
 	it('is off by default and on for 1/true', () => {
@@ -120,12 +120,37 @@ describe('collectDepsModuleExportInfo', () => {
 		expect(collectDepsModuleExportInfo(abs, 'android', new Set()).names).toEqual(['AndroidCanvas', 'own']);
 	});
 
-	it('bails to null for bare export * (name set lives in another package) WITHOUT the opaqueCjs flag', () => {
+	it('bails to null for bare export * without a resolver (name set lives in another package) WITHOUT the opaqueCjs flag', () => {
 		// ESM bails must stay per-module: their bundled namespace has real named
 		// exports the server cannot enumerate, so a default-only shim would
 		// break `import { x }` consumers. Only CJS bails get opaqueCjs.
 		const abs = write('bare-star.js', `export * from 'other-pkg';\nexport const own = 1;\n`);
 		const info = collectDepsModuleExportInfo(abs, 'ios');
+		expect(info.names).toBeNull();
+		expect(info.opaqueCjs).toBeUndefined();
+	});
+
+	it('follows bare export * through the resolver and merges the target names', () => {
+		const leaf = write('bare/leaf.esm.js', `const ref = () => {};\nconst h = () => {};\nexport { h, ref };\n`);
+		const abs = write('bare/root.js', `export * from './local.js';\nexport * from 'leaf-pkg';\nexport { own } from './x';\n`);
+		write('bare/local.js', `export const init = () => {};\n`);
+		const resolver = (spec: string, importer: string) => (spec === 'leaf-pkg' && importer === abs ? leaf : null);
+		const info = collectDepsModuleExportInfo(abs, 'ios', new Set(), resolver);
+		expect(info.names).toEqual(['h', 'init', 'own', 'ref']);
+		expect(info.hasDefault).toBe(false);
+	});
+
+	it('bails to null when the resolver has no edge for the importer', () => {
+		const abs = write('bare/no-edge.js', `export * from 'leaf-pkg';\n`);
+		expect(collectDepsModuleExportInfo(abs, 'ios', new Set(), () => null).names).toBeNull();
+	});
+
+	it('bails to null when the bare export * target is opaque CJS', () => {
+		// A default-only namespace cannot satisfy `export *`.
+		const umd = write('bare/umd.js', `module.exports = (function () { return { a: 1 }; })();\n`);
+		expect(collectDepsModuleExportInfo(umd, 'ios').opaqueCjs).toBe(true);
+		const abs = write('bare/umd-root.js', `export * from 'umd-pkg';\nexport const own = 1;\n`);
+		const info = collectDepsModuleExportInfo(abs, 'ios', new Set(), () => umd);
 		expect(info.names).toBeNull();
 		expect(info.opaqueCjs).toBeUndefined();
 	});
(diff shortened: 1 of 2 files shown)


sitefinitysteve:
@NathanWalker  Looks like we crossed lol [NativeScript/NativeScript#11427](https://github.com/NativeScript/NativeScript/pull/11427) happy to close this in favour of it. 

One thought from this branch in case it's useful... instead of re-resolving the bare target from the importer, I recorded the original -> path edges esbuild already emits in the discovery metafile (import-statement kind only) and followed those, so the shim's names are guaranteed to be the file esbuild bundled even when a scanner and esbuild disagree on an exports condition. Also has a real-esbuild spec for a require-first conditional-exports package.

PR #11432, opened 2026-09-14
fix(android): back press navigates nested frames with history again

edusperoni:
## PR Checklist

Since #11181 (cross-platform NativeWindow), pressing the Android back button in an app whose root `Frame` hosts a nested `Frame` with navigation history exits the app instead of navigating the nested frame back.

Both Android back-press handlers in `ui/frame/index.android.ts` were changed from `Frame.goBack()` to `Frame.goBack(view)`, where `view` is the activity's root view. `FrameBase.goBack(frame)` only ever looks at the given frame and its ancestors, so handing it the root `Frame` skips every nested frame below it. The root frame cannot go back and has no ancestor frames, so `goBack` reports `false`, `callSuper` becomes `true`, and the activity finishes.

Before #11181 the handler let `goBack()` default to `Frame.topmost()`, which is the frame highest in the frame stack - the nested frame that last navigated - so its history was popped correctly.

## What is the new behavior?

The back-press handlers now scope to the window instead of the root view:

```ts
callSuper = view instanceof Frame ? !Frame.goBack(Frame.topmost(nativeWindow)) : true;
```

`Frame.topmost(window)` (added in #11181 for exactly this purpose) returns the frame highest in the stack that belongs to the activity's window, so a nested frame with history is found again. This keeps the multi-window guarantee the original change was after - back press never navigates a frame that lives in another activity's window - while restoring the single-window behaviour from before the regression, since there the window-scoped topmost frame is the global topmost frame.

Two regression specs are added to `frame-common.spec.ts`: one shows that `goBack(topmost(window))` navigates the nested frame, and one pins down that `goBack(rootFrame)` does not reach it.

Reported by Dimitris Katsampas on Discord.


CatchABus (approved):
Looks good to me!

PR #11437, opened 2026-09-16
fix(core): prevent goBack from removing topmost frame from stack

### LINKED ISSUE #11436: [Bug] Android frames are missing from frame stack after backpress and app resume
### Issue Description

In android 12 and later, root launcher activities are no longer finished on Back press.
See https://developer.android.com/about/versions/12/behavior-changes-all#activity-lifecycle

However, {N} core overrides backpress and removes frames from stack, thinking that the activity will finish.
Then, user attempts to launch the app and it will simply resume without recreating views and frames will still be missing from the stack.

That is because the frame stack implementation is very old and did not consider the more recent android 12 behaviour.
The issue affects `Frame.topmost()` and `Frame.getFrameById()`.

This can be replicated in apps with multiple frame instances as back press makes sure core keeps at least a single frame in the stack.

### Reproduction

1. Launch an app with a frame that has an `id` in android 12 or later
2. Log result from `Frame.getFrameById(id)` which should print the frame view
3. Press back to close application
4. Launch app again and it will still be in the last navigated page
5. Log result from `Frame.getFrameById(id)` and it will be undefined

```shell

```

### Environment

_No response_

### Please accept these terms

- [x] This is a bug report

CatchABus:
## PR Checklist

In android 12 and later, root launcher activities are no longer finished on Back press.
However, {N} core overrides backpress and removes frames from stack at times, thinking that the activity will finish.
Then, user attempts to launch the app and it will simply resume without recreating views and frames will still be missing from the stack.

That results in Frame `topmost()` and `getFrameById()` returning a wrong frame instance in apps with multiple frames.

## What is the new behavior?
This PR gets rid of a redundant frame stack pop handling that is around since the early days of NativeScript (see https://github.com/NativeScript/NativeScript/commit/cc829e01524ae7d5a2924bd61e50a738000cc075).

Back then, core didn't have support for native view disposal which is what removes frames from stack nowadays.
See: https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/frame/index.android.ts#L546
https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/frame/index.ios.ts#L76

As a result, the frame helper functions will return the proper `Frame` view instance.

The PR also removes an unneeded override of `_popFromFrameStack`.

Fixes/Closes #11436.


packages/core/ui/frame/index.android.ts
```diff
@@ -547,14 +547,6 @@ export class Frame extends FrameBase {
 		super.disposeNativeView();
 	}
 
-	public _popFromFrameStack() {
```
farfromrefug, line 550: not sure i understand why this was removed here?

CatchABus: @farfromrefug This one seemed like redundant.
See the overridden method: https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/frame/frame-common.ts#L566
as well as the wrapped function that already includes the `_isInFrameStack` check: 
https://github.com/NativeScript/NativeScript/blob/main/packages/core/ui/frame/frame-stack.ts#L63

farfromrefug (approved):

PR #11438, opened 2026-09-18
feat(ios): iPhone Duo support

NathanWalker:
## PR Checklist

No issue exists yet; this comes out of building an app against the iPhone Duo simulator (Xcode 27.1 / iOS 27.1) and hitting the problems below. Reference: [Preparing your app for iPhone Duo](https://developer.apple.com/documentation/technologyoverviews/preparing-your-app-for-iphone-duo).

iPhone Duo has two displays. A scene moves between them as the device opens and closes, and `UIScreen.main` is **always the outer display** (there are two `UIScreen`s while open; `window.screen` follows the scene). Core assumes the scene matches the main screen in two places:

**1. A Frame nested in a layout is sized from `UIScreen.main`.** `Frame.layoutNativeView()` has been a no-op since positioning was handed to UIKit, but for a Frame hosted in a NativeScript layout (e.g. `<GridLayout><Frame/></GridLayout>`, which is what the Angular template's `GridLayout > page-router-outlet` produces) core is the container and never sizes the navigation controller's view. It keeps the `UIScreen.main` bounds UIKit creates it with; the autoresizing mask then only tracks later container changes. Measured in `apps/toolbox` with its root Frame wrapped in a GridLayout, cold-launched with the device open:

| step | container | Frame view (before) | Frame view (after) |
| --- | --- | --- | --- |
| launch on inner display | 951x669 | **466x678** | 951x669 |
| close (outer display) | 466x678 | **84x687** | 466x678 |
| reopen | 951x669 | **569x678** | 951x669 |

Launching closed and then opening happens to work, which hides the problem.

**2. `Screen.mainScreen` goes stale.** It caches the first `UIScreen` it resolves, so after the scene moves displays `widthDIPs`/`heightDIPs` (and the pixel sizes media queries match against) keep reporting the previous display: `951x669` while on the `466x678` outer display.

**3. ActionItems can't take part in vertical bars well.** iOS 27.1 presents bars vertically on iPhone Duo, where fewer items fit and the rest move to an overflow menu that shows image + title. Core creates image-only `UIBarButtonItem`s when `icon` is set (the `text` is only used for accessibility), and there is no way to reach the new `visibilityPriority` / `axisBehavior` properties.

## What is the new behavior?

- **`fix(ios)` Frame**: during layout, a Frame hosted directly in a NativeScript view matches its native view to the container's bounds. Frames under a UIKit container (window root, TabView, SplitView, modals) are untouched, and "UIKit positions the Frame" semantics are unchanged — on a regular iPhone the rects are already equal and this is a no-op.
- **`fix(ios)` Screen**: bounds are read from the window's current screen. `scale` still uses the cached screen: it is on hot paths and is identical on both displays, the invariant `layout-helper` already relies on.
- **`feat(action-bar)`**:
  - `text` is kept as the `UIBarButtonItem` title when an `icon` is set. Bars still render only the image (checked on iOS 26.5 and 27.1); the overflow menu now has a title to show.
  - `ios.visibilityPriority`: `'high' | 'standard' | 'low' | number`
  - `ios.axisBehavior`: `'automatic' | 'horizontalOnly' | 'verticalPreferred'`
  - Both are ignored where the API doesn't exist (`respondsToSelector`). The 27.1 symbols aren't in `types-ios` yet, so they are declared locally in `action-bar/index.ios.ts`; those declarations can go once the typings are regenerated against the 27.1 SDK.

```xml
<ActionItem icon="sys://star" text="Favorite" ios.position="right" ios.visibilityPriority="high" />
<ActionItem icon="sys://tag" text="Tag" ios.position="right" ios.visibilityPriority="low" />
<ActionItem icon="sys://printer" text="Print" ios.position="right" ios.axisBehavior="horizontalOnly" />
```

Verified on the iPhone Duo simulator with the new `apps/toolbox` **action-items** page: with a horizontal and a vertical bar both present, the `horizontalOnly` item stays in the horizontal bar while the rest stack vertically; with the closed device in landscape (the tightest bar) both `low` items and the last `standard` one moved to the overflow menu, listed by icon and title, while the `high` item kept its place.

### Testing

- `npx nx run core:test` — passing.
- `apps/automated` on an iPhone 17 Pro simulator (iOS 26.5) — `1818 OK, 0 failed`, including two new ActionItem tests (title kept alongside the icon; placement settings applied where available and harmless where not).
- The Frame and Screen fixes have no automated test: they only show up when the scene's screen differs from `UIScreen.main`, which needs the iPhone Duo simulator. Reproduce by wrapping the root Frame of `apps/toolbox/src/app-root.xml` in a `GridLayout` and cold-launching with the device open.

### Not in this PR

- Media queries are correct when evaluated but are not re-evaluated when the scene moves between displays (no orientation change fires).
- `UILayoutViewController` copies an ancestor's top safe-area inset onto itself assuming its owner is a TabView item; any other hosted view with a parent picks that up too.
- Regenerating `types-ios` for the 27.1 SDK, and any first-class API for hinge state or reserved regions.


packages/core/platform/screen/index.ios.ts
```diff
@@ -23,10 +23,18 @@ class MainScreen {
...
 	get widthDIPs(): number {
-		return this.screen.bounds.size.width;
+		return this.currentScreen.bounds.size.width;
 	}
 	get heightDIPs(): number {
-		return this.screen.bounds.size.height;
+		return this.currentScreen.bounds.size.height;
+	}
+
+	// A window can move to a screen of another size, e.g. between the outer and inner displays of
```
farfromrefug: that comment is unecessary

packages/core/ui/action-bar/index.d.ts
```diff
@@ -264,6 +264,31 @@ export interface IOSActionItemSettings {
 	 * 23: PageCurl
 	 */
 	systemIcon: number;
+
+	/**
```
farfromrefug, line 268: we should try and prevent AI from creating huge comments like this. Here we only need the link developer doc

packages/core/ui/action-bar/index.ios.ts
```diff
@@ -350,6 +381,11 @@ export class ActionBar extends ActionBarBase {
 			if (img) {
 				const image = img.imageWithRenderingMode(this._getIconRenderingMode());
 				barButtonItem = UIBarButtonItem.alloc().initWithImageStyleTargetAction(image, UIBarButtonItemStyle.Plain, tapHandler, 'tap');
+				if (item.text) {
+					// Bars show the image; the title is what the system uses where an image alone won't do,
+					// such as the overflow menu items move to when a bar runs out of room.
+					barButtonItem.title = item.text + '';
```
farfromrefug: unecessary comment. And why the + '' ? item.text is supposed to be a string

packages/core/ui/frame/index.ios.ts
```diff
@@ -362,7 +362,25 @@ export class Frame extends FrameBase {
 	}
 
 	public layoutNativeView(left: number, top: number, right: number, bottom: number): void {
-		//
+		// Positioning is left to UIKit, but a Frame nested in a layout has no container controller sizing
+		// its view, so it keeps the UIScreen.main bounds UIKit created it with. That is wrong whenever the
```
farfromrefug: this comment can be drastically shorten

packages/core/ui/action-bar/index.d.ts
NathanWalker: good mention, it's concise now per the rules in CodeComments.md

